PR #1022, opened 2018-02-16
Append :8081 to Agent backend URLs if no port set

### LINKED ISSUE #985: Change default ws:// and wss:// ports to match agent-port on backend
If a port is not specified in a backend address in the agent configuration, a default of port 80 is used for `ws://` and port 443 for `wss://`. I'd like to see these default ports match the default `agent-port` (8081) of the backend.

amdprophet:
## What is this change?

Append `:8081` any of an agent's backend urls that do not have a port specified.

Fixes #985.

Yes.

Nope!

Nope!

### DIFF
--- a/util/url/url.go
+++ b/util/url/url.go
@@ -1,0 +1,16 @@
+package url
+
+import "net/url"
+
+// AppendPortIfMissing takes a URL and port as strings and appends the provided
+// port to the URL if it is missing
+func AppendPortIfMissing(u string, p string) (string, error) {
+	parsedURL, err := url.Parse(u)
+	if err != nil {
+		return "", err
+	}
+	if parsedURL.Port() == "" {
+		return parsedURL.String() + ":" + p, nil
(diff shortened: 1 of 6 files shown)


util/url/url.go
```diff
@@ -0,0 +1,16 @@
...
+
+// AppendPortIfMissing takes a URL and port as strings and appends the provided
+// port to the URL if it is missing
+func AppendPortIfMissing(u string, p string) (string, error) {
+	parsedURL, err := url.Parse(u)
+	if err != nil {
+		return "", err
+	}
+	if parsedURL.Port() == "" {
+		return parsedURL.String() + ":" + p, nil
```
jamesdphillips: If the parsed URL contained a path it could cause an issue, might be more safe to do:

```go
parsedURL.Host = parsedURL.Hostname() + ":" + p
return parsedURL.String(), nil
```

Issue #1032, opened 2018-02-16
Add compaction to etcd keyspace


mercul3s:
We need to setup compaction to avoid this error:
```
sensuctl event list
Error: failed to request new refresh token; client returned 'The server returned the error: 500 could not remove the access token from the access list: etcdserver: mvcc: database space exceeded'
```
We should be able to add it as a default setting to etcd.
https://github.com/coreos/etcd/blob/master/Documentation/op-guide/maintenance.md#history-compaction

PR #1046, opened 2018-02-20
Add autocompaction by revision to the etcd config

grepory:
## What is this change?

This adds etcd autocompaction to the embedded etcd config. Every 5minutes (the schedule for revision-based autocompaction), we'll compact every revision except the newest.

Without compaction, we'll retain every version of every key until we hit the db size limit.


jamesdphillips:
Closes #1032

Issue #628, opened 2017-11-28
Agent crashes if backend goes away


amdprophet:
If sensu-backend is shut down, connected agents will crash. Sensu agents should go into a reconnect loop if a connection is lost.

```
{"component":"agent","error":"Connection error: websocket: close 1006 (abnormal closure): unexpected EOF","level":"error","msg":"transport receive error","time":"2017-11-28T13:55:33-08:00"}
{"component":"agent","error":"Connection closed: the websocket connection is no longer open","level":"error","msg":"transport receive error","time":"2017-11-28T13:55:33-08:00"}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x149de8b]

goroutine 37 [running]:
github.com/sensu/sensu-go/agent.(*Agent).receivePump(0xc4201505b0, 0x19916c0, 0xc4201e4940)
        /Users/justin/code/go/src/github.com/sensu/sensu-go/agent/agent.go:372 +0x1cb
created by github.com/sensu/sensu-go/agent.(*Agent).Run
        /Users/justin/code/go/src/github.com/sensu/sensu-go/agent/agent.go:508 +0x33c
```

PR #1057, opened 2018-02-21
Agent reconnection

palourde:
## What is this change?

It implements some agent reconnection logic, in case the websocket connection to the backend is lost.

I also made some cleanup in the agent package, sorry about the noise, use the following commit to see the actual logic: https://github.com/sensu/sensu-go/commit/cd415b04e7524805d6e79d0e943307b54468de0a

Closes https://github.com/sensu/sensu-go/issues/628

Spoiler alert, it probably does. Generally speaking, your change needs a changelog. For more information, see [CONTRIBUTING.md](https://github.com/sensu/sensu-go/blob/master/CONTRIBUTING.md#changelog).

1. I wasn't really sure which values to use for backoff (as of now, it will start with a 500ms delay between retries, and exponentially increase up to a 10 seconds delay,  with unlimited attempts): 

https://github.com/sensu/sensu-go/commit/cd415b04e7524805d6e79d0e943307b54468de0a#diff-6ea70e0a4de93ebcb83000a028edc163R170

2. That's maybe the first time I'm really playing with the `transport` package, I hope it makes sense!

Nope!


util/wait/wait.go
```diff
@@ -0,0 +1,52 @@
+package wait
+
+import (
+	"errors"
+	"time"
+)
+
+// Backoff contains the configuration for exponential backoff
+type Backoff struct {
```
jamesdphillips: The asset manager already uses a simple library for exponential backoff `cenkalti/backoff`. Maybe it can be removed and this package used in it's stead.

util/wait/wait.go
```diff
@@ -0,0 +1,52 @@
...
+func (b *Backoff) ExponentialBackoff(fn BackoffFunc) error {
+	sleep := b.InitialDelayInterval
+
+	for i := 0; i < b.MaxRetryAttempts || b.MaxRetryAttempts == 0; i++ {
+		if i != 0 {
+			// Sleep for the determined duration
+			time.Sleep(sleep)
+
+			// Exponentially increase that sleep duration
+			sleep = time.Duration(float64(sleep) * b.Multiplier)
```
echlebek: We should add jitter, to reduce the likelihood of accidentally-synchronized clients making requests at the same time. The formula could look as simple as:

```
jitter := rand.Int63n(int64(time.Second))
sleep = time.Duration(float64(sleep) * b.Multiplier) + jitter
```

util/wait/wait.go
```diff
@@ -0,0 +1,52 @@
...
+		}
+
+		if ok, err := fn(); err != nil || ok {
```
echlebek: Here I would suggest
```
done, err := fn(i)
if done || err == nil {
    return err
}
if i == MaxRetryAttempts {
    return fmt.Errorf("giving up after %d retries: %s", i, err)
}
```

Then, backoff funcs can return `false, err` when they encounter some error that they want to retry on.

agent/sockets.go
```diff
@@ -0,0 +1,246 @@
...
+				}
+				return
+			} else if bytesRead == 0 {
+				if err := c.Close(); err != nil {
+					logger.Debug(err)
+				}
+				return
+			}
+			// If the message is a ping, return without notifying sender.
+			match, _ := regexp.MatchString("\\s+ping\\s+", string(buf[:bytesRead]))
```
echlebek: This is going to trigger the megacheck linter in my pending PR. The regexp should be compiled and placed at package scope with
```
var pingRe = regexp.MustCompile(`\s+ping\s+`)
```

Which avoids compiling the regexp with every iteration, and also avoids double-escaping by using the backtick syntax.

util/wait/wait.go
```diff
@@ -0,0 +1,52 @@
+package wait
```
echlebek: What do you think about calling this package `retry`?

util/wait/wait.go
echlebek: `retry.ExponentialBackoff`?

util/wait/wait.go
```diff
@@ -0,0 +1,52 @@
...
+// aborted
+type BackoffFunc func() (done bool, err error)
+
+// ErrMaxRetryAttempts is returned when the number of maximal retry attempts is
+// reached
+var ErrMaxRetryAttempts = errors.New("maximal number of retry attempts reached")
+
+// ExponentialBackoff retries the provided func with exponential backoff, until
+// the maximal number of retries is reached
+func (b *Backoff) ExponentialBackoff(fn BackoffFunc) error {
```
echlebek: What about calling this method `Retry`?

agent/sockets.go
palourde: I actually simply moved that code into its own file. I think we should fix the issues found by megacheck within a separate PR if that's alright with you!

util/wait/wait.go
palourde: Good idea!

util/wait/wait.go
palourde: I wasn't sure if it was required at this point but yeah, probably a good thing to add now!

util/wait/wait.go
```diff
@@ -0,0 +1,52 @@
...
+	sleep := b.InitialDelayInterval
+
+	for i := 0; i < b.MaxRetryAttempts || b.MaxRetryAttempts == 0; i++ {
+		if i != 0 {
+			// Sleep for the determined duration
+			time.Sleep(sleep)
```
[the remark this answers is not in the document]
palourde: I’m not sure I agree with you on this point; even if `BackoffFunc` duration is long (e.g. 5 seconds), we should probably still wait 500ms between the first and second attempt, otherwise that's basically the equivalent of having no backoff algorithm and if we use an external resource, it would let the chance to other consumers in case there's some rate-limiting in place, for example.

Also, the sleep duration can be easily tweaked in case we know our func is larger than usual so the intervals between retries are smaller.

util/wait/wait.go
palourde: I'm not sure to understand what difference it would make. The goal here was to let `BackoffFunc` return the following combinaison:

* `false, nil` - The attempt failed and we should retry, and the error has been optionally logged (e.g. https://github.com/sensu/sensu-go/pull/1057/files#diff-6ea70e0a4de93ebcb83000a028edc163R179)
* `true, nil` - The attempt succeeded, we can immediately exit (e.g. https://github.com/sensu/sensu-go/pull/1057/files#diff-6ea70e0a4de93ebcb83000a028edc163R185)
* `false, err` - A fatal error was encountered during this attempt, we should immediately exit and return the error so the caller can handle the failure

agent/agent.go
```diff
@@ -157,247 +150,51 @@ func NewAgent(config *Config) *Agent {
...
+					MaxDelayInterval:     10 * time.Second,
+					MaxRetryAttempts:     0, // Unlimited attempts
+					Multiplier:           1.5,
 				}
-				return
-			} else if bytesRead == 0 {
-				if err := c.Close(); err != nil {
-					logger.Debug(err)
+				if err := backoff.Retry(func(retry int) (bool, error) {
+					logger.Debugf("reconnection attempt #%d", retry)
```
echlebek: Probably worth guarding this with 
```
if retry > 0 {
}
```
to avoid producing 'retry 0' debug calls in the common case.

transport/transport.go
```diff
@@ -120,43 +125,29 @@ func NewTransport(conn *websocket.Conn) Transport {
 	}
 }
 
+// Close attempts to send a "going away" message over the websocket connection.
+// This will cause a Write over the websocket transport, which can cause a
+// panic. We rescue potential panics and consider the connection closed,
+// returning nil, because the connection _will_ be closed. Hay!
+func (t *WebSocketTransport) Close() error {
+	t.mutex.Lock()
+	defer func() {
```
echlebek: Why aren't we locking the mutex with every `WriteMessage`?

grepory: Basically the whole transport needs reworking. This is some of the oldest code in 2.0.

grepory: (There's tight coupling that shouldn't exist that makes things "generally work."

agent/agent.go
palourde: Good catch!

PR #1059, opened 2018-02-21
[Web] Search Box

dabria:
## What is this change?
Adds a not-fulling-functioning filter box to the events page. It currently records input, but does nothing with it. We need to revisit this later when we decide how we should deal with the queries.

image

Requirement for web ui.

Added one.

N/A

No, however we will need to revisit this.

### DIFF
--- a/dashboard/src/components/EventsContainer.js
+++ b/dashboard/src/components/EventsContainer.js
@@ -68,6 +68,7 @@ class EventsContainer extends React.Component {
       `${window.location.pathname}?filter=event.Check.Name=='${newValue}'`,
     );
   };
+
   requeryStatus = newValue => {
     this.props.router.push(
       `${window.location.pathname}?filter=event.Check.Status==${newValue}`,
--- a/dashboard/src/components/EventsContainer.js
+++ b/dashboard/src/components/EventsContainer.js
@@ -14,11 +14,9 @@ import EventsContainerMenu from "./EventsContainerMenu";
 
 const styles = theme => ({
   eventsContainer: {
-    border: "1px solid",
-    borderRadius: 2,
-    borderColor: theme.palette.divider,
     marginTop: 16,
     marginBottom: 16,
+    backgroundColor: theme.palette.background.paper,
   },
   tableHeader: {
     padding: "20px 0 16px",
--- a/dashboard/src/components/EventsContainer.js
+++ b/dashboard/src/components/EventsContainer.js
@@ -6,6 +6,7 @@ import map from "lodash/map";
 import get from "lodash/get";
 import { createFragmentContainer, graphql } from "react-relay";
 import { withStyles } from "material-ui/styles";
+import Paper from "material-ui/Paper";
 
 import checkboxIcon from "material-ui/Checkbox";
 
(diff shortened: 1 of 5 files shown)


dashboard/src/components/SearchBox.js
```diff
@@ -0,0 +1,81 @@
...
+
+  updateInput = event => {
+    this.props.onUpdateInput(event.currentTarget.value);
+  };
+
+  render() {
+    const { classes, FilterIcon } = this.props;
+
+    return (
+      <div className={classes.box}>
```
jamesdphillips: We should use the Paper component, that we we get correct background color and box shadow fo' free

dashboard/src/components/SearchBox.js
```diff
@@ -0,0 +1,81 @@
...
+    backgroundColor: "white",
+    boxShadow:
+      "0 0 2px 0 rgba(0,0,0,0.14), 0 2px 2px 0 rgba(0,0,0,0.12), 0 1px 3px 0 rgba(0,0,0,0.20)",
+  },
+  filterIconContainer: {
+    display: "flex",
+    padding: "6px 8px 8px",
+    height: 36,
+    // TODO find proper reference to this in theme, default icon color
+    color: "rgba(0, 0, 0, 0.54)",
```
jamesdphillips: https://github.com/mui-org/material-ui/blob/v1-beta/src/styles/createPalette.js#L32-L37

possibly `theme.palette.action.active` could probably even flip it to `selected` when the input is active.

dashboard/src/components/SearchBox.js
```diff
@@ -0,0 +1,81 @@
...
+    borderRadius: 3,
+    width: 300,
+    height: 36,
+    fontSize: 14,
+    border: "none",
+    "&:focus": { outline: "none" },
+  },
+  save: {
+    alignSelf: "center",
+    marginRight: 8,
```
jamesdphillips: `theme.spacing.unit`

dashboard/src/pages/EventsPage.js
```diff
@@ -19,17 +38,35 @@ class EventsPage extends React.Component {
     }
   `;
 
+  state = { inputValue: "" };
+
+  requerySearchBox = query => {
+    this.setState({ inputValue: query });
+    // TODO return to this and make it actually query
+    console.log(query);
```
jamesdphillips: Kill or 

```
// eslint-disable-next-line no-console
console.info("query", query);
```

dashboard/src/components/EventsContainer.js
```diff
@@ -14,17 +14,14 @@ import EventsContainerMenu from "./EventsContainerMenu";
 
 const styles = theme => ({
   eventsContainer: {
-    border: "1px solid",
-    borderRadius: 2,
-    borderColor: theme.palette.divider,
     marginTop: 16,
     marginBottom: 16,
+    backgroundColor: theme.palette.background.paper,
```
jamesdphillips: Occurred to me that the events container should probably be a Paper element as well. Unrelated to this PR so it can be handled separately though. 🤔

dashboard/src/components/SearchBox.js
```diff
@@ -0,0 +1,79 @@
...
+    display: "flex",
+    padding: "6px 8px 8px",
+    height: 36,
+    color: theme.palette.action.active,
+  },
+  textField: {
+    borderRadius: 3,
+    width: 300,
+    height: 36,
+    fontSize: 14,
```
jamesdphillips: 🤔 we'll have to go through and make sure _all_ our text consistent at some point.

Issue #1034, opened 2018-02-16
Implement last ok in events


amdprophet:
Add a last_ok field to events that will get updated with a timestamp every time an event with an exit status of 0 is received by a backend.

Sensu 1.x implementation: https://github.com/sensu/sensu/blob/master/lib/sensu/server/process.rb#L584-L586

palourde:
Hi @zbintliff,

That's definitively the right place! If you have any question or need assistance with this, don't hesitate to join us in our [Community Slack](http://bit.ly/sensucommunityslack) #sensu2 channel.

Thanks, this could really be an awesome functionality to add!

PR #1062, opened 2018-02-21
Add LastOk to Event (#1034)

zbintliff:
When Event's status is 0 set LastOk to the timestamp of the event

Signed-off-by: Zach Bintliff <zbintliff@gmail.com>

## What is this change?
Add `LastOk` to Event type

Provides feature parity with 1.X sensu.

Closes https://github.com/sensu/sensu-go/issues/1034

Yes

1. First time working with protobuf would appreciate guidance there and if anything looks off
2. Making sure I merged new and old event correctly so that `LastOk` lasts properly persists.
3. The more I look at my testing the more I think it doesn't do anything. So some guidance on how to test would be great. 

Following instructions on updating


types/event.proto
```diff
@@ -35,4 +35,7 @@ message Event {
 
   // Hooks describes the results of multiple hooks; if event is associated to hook execution.
   repeated Hook hooks = 6 [(gogoproto.nullable) = true];
+  
+  // LastOk displays last time this check was ok; if event status is 0 this is set to timestamp
+  int64 last_ok = 7;
```
echlebek: Let's add `[(gogoproto.customname) = "LastOK"]` to this field, so that we can get something that fits the Go coding standards. (https://github.com/golang/go/wiki/CodeReviewComments#initialisms)

grepory:
The test failure in travis is unrelated to this change, I'm comfortable with merging instead until we fix the tests.

types/event.proto
zbintliff: Yeah I wasn't sure if this follows the ruby name standard `last_ok` but went with that since the issue specifies. And the `Ok` vs `OK` got me too. Thanks

zbintliff:
Updated, not too happy about the test I added as I still don't think it _does_ anything.

types/event.proto
```diff
@@ -35,4 +35,7 @@ message Event {
 
   // Hooks describes the results of multiple hooks; if event is associated to hook execution.
   repeated Hook hooks = 6 [(gogoproto.nullable) = true];
+  
+  // LastOK displays last time this check was ok; if event status is 0 this is set to timestamp
+  int64 last_ok = 7 [(gogoproto.customname) = "LastOK"];
```
grepory: In this case, we're looking at the Check's status. We'll want to throw LastOK in [here](https://github.com/sensu/sensu-go/blob/master/types/check.proto#L201).

backend/eventd/eventd.go
```diff
@@ -158,6 +158,7 @@ func (e *Eventd) handleMessage(msg interface{}) error {
 		}
 
 		event.Check.MergeWith(prevEvent.Check)
+		event.LastOK = prevEvent.LastOK
```
grepory: Given that we move LastOK from Event to Check, this will have to change. Also, this isn't the only desired effect we want. If the current event's status is OK, we want to use this event's timestamp as LastOK.

backend/eventd/eventd_test.go
```diff
@@ -69,6 +69,7 @@ func TestEventHandling(t *testing.T) {
 
 	// Make sure the event has been marked with the proper state
 	assert.Equal(t, types.EventPassingState, event.Check.State)
+	assert.Equal(t, badEvent.Timestamp, event.LastOK)
```
grepory: We'll want to assert that event.Check.LastOK is equal to event.Timestamp.

Publishing a bad event seems out of place in this test, but I can look into that another time. Line 63 is where we're publishing our event. We want to make sure that eventd is correctly updating the LastOK field with the current timestamp.

backend/eventd/eventd.go
zbintliff: This part was just the merge from `prevEvent`. The current event's LastOK is set in `flapping.go` [here](https://github.com/zbintliff/sensu-go/blob/21fb439301d747d270d3f579421b0d040b9b7a25/backend/eventd/flapping.go#L28) if status is OK.

This will be the same if we move `LastOK` into the check as well. Except this line will go into the `MergeWith()`

zbintliff:
We can definitely move the `LastOK` into the `Check`, but that will be a departure from where it is in [the 1.X version](https://github.com/sensu/sensu/blob/master/lib/sensu/server/process.rb#L584-L586) and what the #1034 asked for.

I'm also not familiar with the new architecture so not sure is Event in this project is 1:1 with event in the 1.X version. Thanks for any insight.

backend/eventd/eventd_test.go
```diff
@@ -69,6 +70,10 @@ func TestEventHandling(t *testing.T) {
 
 	// Make sure the event has been marked with the proper state
 	assert.Equal(t, types.EventPassingState, event.Check.State)
+	fmt.Printf("Event Timestamp: %d\n", event.Timestamp)
+	fmt.Printf("Check's LastOK: %d\n", event.Check.LastOK)
+	fmt.Printf("badEvent Timestamp: %d\n", badEvent.Timestamp)
```
zbintliff: @grepory its my understanding that `badEvent` should have an older timestamp but it doesn't. IF you pull down this branch and run `$ go test ./... -run=TestEventHandling -v` you'll see they are all the same. Even with modifying `FixtureCheck` to sleep 5 and print `UnixNano()`. Am I missing some mutations somewhere?

This was just extra code I was writing to check if my test was valid or not.

zbintliff: Figured it out!

zbintliff:
This should be ready for review again whenever you have time.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ set when importing legacy settings.
 - Configured and enabled etcd autocompaction.
 - Add event metrics type, implementing the Sensu Metrics Format.
 - Added non-functional selections for resolving and silencing to web ui
+- Add LastOk to event type. This will be updated to reflect the last timestamp of a succesful event
```
portertech: "to check type"

README.md
```diff
@@ -32,7 +32,7 @@ Otherwise, see the **for non-C++ users** [instructions here.](https://github.com
 
 ### Quick Start
 
-Once you make a change to any `*.proto` file within the **types** package, you will need regenerate the associated `*.pb.go` file. To do so, simply run the [genproto.sh](https://github.com/sensu/sensu-go/blob/master/scripts/genproto.sh) script, which will install all required dependencies and launch the code generation.
+Once you make a change to any `*.proto` file within the **types** package, you will need regenerate the associated `*.pb.go` file. To do so, simply run the [genproto.sh](https://github.com/sensu/sensu-go/blob/master/scripts/genproto.sh) script, which will install all required dependencies and launch the code generation (be sure to run the below `./build.sh deps` first though).
```
grepory: While you're in there. Can you make it "you will need *to* regenerate the associated"?

grepory:
Looking into the build failure.

zbintliff:
Great, I'll make that change in the README.md. Do you want me to squash my commits down?

grepory:
@zbintliff Yes, please, if you wouldn't mind. I don't mind doing it for you, but I'd appreciate it. :) We only squash-merge. (I should put that in the contributing doc). If you would be so kind as to reword your initial commit message to be "Add LastOK to Check (#1034)" that would also be preferable.

grepory:
I may have to rebase your branch and re-push, but I'll get this merged for you tonight. I really appreciate your help with this!

PR #1065, opened 2018-02-22
[Web] Silencing and Resolve

dabria:
## What is this change?
Adds a simple, non-functioning disclosure (3 dots) dropdown for silencing and resolving events. The actions are to be added later.

image

Required for the web ui.

I added one.

N/A

Resolving events has yet to be implemented. So this must be revisited when it's available. In addition, the dialog that will appear for silencing requires more thought, so we will revisit that later as well.

### DIFF
--- a/dashboard/src/components/EventsListItem.js
+++ b/dashboard/src/components/EventsListItem.js
@@ -33,9 +33,11 @@ const styles = theme => ({
     verticalAlign: "top",
     padding: "14px 0",
   },
-  silenceButton: {
+  disclosure: {
     marginRight: 4,
     paddingTop: 14,
+    // TODO add this colour to theme
+    color: "rgba(0, 0, 0, 0.54)",
   },
   content: {
     width: "100%",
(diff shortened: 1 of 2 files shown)


dashboard/src/components/EventsListItem.js
```diff
@@ -55,19 +63,48 @@ class EventListItem extends React.Component {
...
+  onClose = () => {
+    this.setState({ anchorEl: null });
+  };
+
+  handleClick = event => {
+    this.setState({ anchorEl: event.currentTarget });
+  };
+
+  silenceEntity = entity => () => {
+    console.log(entity);
```
jamesdphillips: 🤔 linter should not be passing with these present. Regardless- by convention- add `eslint-disable-next-line` and ideally switch to info for the time being.

dashboard/src/components/EventsListItem.js
```diff
@@ -55,19 +63,48 @@ class EventListItem extends React.Component {
     // eslint-disable-next-line react/forbid-prop-types
     classes: PropTypes.object.isRequired,
     Chevron: PropTypes.func.isRequired,
+    Disclosure: PropTypes.func.isRequired,
   };
 
   static defaultProps = {
     Chevron: chevronIcon,
+    Disclosure: disclosureIcon,
```
jamesdphillips: I don't think we need to have the icons passed in as props, unlikely we'll want to change them.

PR #1066, opened 2018-02-22
Use EntityAgentClass constant when assigning an agent entity's class

nikkiattea:

## What is this change?

Uses EntityAgentClass constant when assigning an agent entity's class.

It's good practice to use our entity class constants when assigning entities their class.

Probably not.

Nah.

Nah.


nikkiattea:
Travis failure looks like it was evident on master prior to this PR. Will follow up in slack to investigate, but merging this for now.

PR #1070, opened 2018-02-23
Upgrade Go to 1.10 for Travis and AppVeyor.

echlebek:

## What is this change?

Go 1.10 has been released, so we should use it. As a side effect, it should fix a build error in a dependency of fileb0x.

### DIFF
--- a/Dockerfile.packaging
+++ b/Dockerfile.packaging
@@ -15,8 +15,8 @@ RUN apt-get update && \
   gem install --no-ri --no-rdoc fpm && \
   gem install --no-ri --no-rdoc rake && \
   gem install --no-ri --no-rdoc package_cloud && \
-  wget https://storage.googleapis.com/golang/go1.9.linux-amd64.tar.gz && \
-  tar -C /usr/local -zxvf go1.9.linux-amd64.tar.gz && \
+  wget https://storage.googleapis.com/golang/go1.10.linux-amd64.tar.gz && \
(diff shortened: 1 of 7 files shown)


palourde:
I would suggest to also:

* Update the `sensu-go-build` image so we use Go 1.10 too. See https://github.com/sensu/engineering/blob/master/documentation/Packaging-Instructions.md#optional-docker-build-image and this: https://github.com/sensu/sensu-go/blob/23c072849a6e93f172b48cacd9b3c80b58a98c59/Dockerfile.packaging#L18
* Remove the `-i ` flag to `go build` in `build.sh` & `build.ps1`

echlebek:
@palourde, I implemented the changes you mentioned, plus also discovered that we had not upgrade the Vagrantfile since 1.8.3, so upgraded that too.

Dockerfile.packaging
```diff
@@ -15,7 +15,7 @@ RUN apt-get update && \
   gem install --no-ri --no-rdoc fpm && \
   gem install --no-ri --no-rdoc rake && \
   gem install --no-ri --no-rdoc package_cloud && \
-  wget https://storage.googleapis.com/golang/go1.9.linux-amd64.tar.gz && \
+  wget https://storage.googleapis.com/golang/go1.10.linux-amd64.tar.gz && \
   tar -C /usr/local -zxvf go1.9.linux-amd64.tar.gz && \
```
palourde: We might want to update that too!

echlebek: Ooooooops

PR #1080, opened 2018-02-26
Implement check occurrences & occurrences_watermark

### LINKED ISSUE #1033: Implement event occurrences
Add an occurrence field to events which will represent the number of times an event has been created for a client/check pair with the same state (check status). Occurrences should not be tracked for client/check pairs with an exit status of 0.

Implementation of occurrences in Sensu 1.x can be found here: https://github.com/sensu/sensu/blob/master/lib/sensu/server/process.rb#L571-L575

amdprophet:

## What is this change?

Implements check occurrences and occurrences watermark from Sensu 1.x.

Fixes an issue with `MergeWith()` in `types/check.go` where the
current check's status and execution time were not added to the
check's history.

I also slipped in a rename of the test from my last PR - I had forgotten to name it, whoops!

Fixes #1033

`MergeWith()` was not functioning as intended.

Yes.

Nope!

I discovered that `MergeWith()` does not functioning as intended and came up with a fix.


types/check.go
```diff
@@ -224,18 +224,20 @@ func (b ByExecuted) Less(i, j int) bool { return b[i].Executed < b[j].Executed }
 func (c *Check) MergeWith(chk *Check) {
```
palourde: I know it's not directly related to your PR but could you rename `chk` to something else, like `originalCheck`, `oldCheck` or `sourceCheck`? It took few moments to understand which on these two checks was merged into the other and I feel the bug you fixed at lines 227-228 might be related to that complexity (thank you for fixing that btw 👏 )

amdprophet: Will do!

PR #1094, opened 2018-02-28
Refactor tests to not use testify.Suite.

### LINKED ISSUE #939: Replace testify/suite with Go subtests
It seems that running parallel testing with `testify/suite` is not supported/does not work very well ([1](https://github.com/golang/go/issues/19280))([2](https://github.com/stretchr/testify/pull/369#issuecomment-280623198)).

Since Go 1.7, we can now use [subtests](https://tip.golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks), which should significantly accelerate some of our tests, such as https://github.com/sensu/sensu-go/blob/master/backend/schedulerd/check_scheduler_test.go.

echlebek:

## What is this change?

Removes all uses of testify's suite package. I did not use subtests, instead favouring a light refactoring that involved creating a test struct in most cases.

Closes #939

### DIFF
--- a/agent/assetmanager/asset_test.go
+++ b/agent/assetmanager/asset_test.go
@@ -8,67 +8,76 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+	"time"
 
 	"github.com/sensu/sensu-go/types"
-	"github.com/stretchr/testify/suite"
+	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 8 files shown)


palourde:
@echlebek Looks like the appveyor tests failed on tests modified in this PR!

PR #1117, opened 2018-03-05
Govaluate's modifier tokens can now be optionally forbidden

palourde:
## What is this change?

It adds a second boolean argument to `eval.ValidateStatements` so modifier tokens can be optionally forbidden.

As per 1.x documentation on filters, arithmetic operators (which are considered as modifiers) can be useful in govaluate expressions: https://sensuapp.org/docs/latest/reference/filters.html#example-handling-repeated-events

Added!

Nope!

Nope!

### DIFF
--- a/util/eval/eval.go
+++ b/util/eval/eval.go
@@ -29,18 +29,25 @@ func Evaluate(expression string, parameters map[string]interface{}) (bool, error
 }
 
 // ValidateStatements ensure that the given statements can be parsed
-// successfully and that it does not contain any modifier tokens.
-func ValidateStatements(statements []string) error {
+// successfully and, optionally, that it does not contain any modifier tokens.
+func ValidateStatements(statements []string, forbidModifier bool) error {
 	for _, statement := range statements {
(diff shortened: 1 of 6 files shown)


util/eval/eval.go
```diff
@@ -29,18 +29,25 @@ func Evaluate(expression string, parameters map[string]interface{}) (bool, error
...
 
-		// Do not allow modifier tokens (eg. +, -, /, *, **, &, etc.)
-		for _, token := range exp.Tokens() {
-			if token.Kind == govaluate.MODIFIER {
-				return fmt.Errorf("forbidden modifier tokens in statement '%s'", statement)
+		// We can optionally forbid modifier tokens if we believe an expression has
+		// no reason to use a modifier operator in its context (e.g. assets
+		// filters). By doing so, we can detect expressions that could possibly
+		// evaluate to something else than a boolean value, and return an error
+		// before saving that expression.
```
jamesdphillips: 🙏

PR #1118, opened 2018-03-05
[Web UI] Simplify filtering events

jamesdphillips:

## What is this change?

Simplify filtering events

From user experience point of view it's confusing that I would have to provide `event` in given filter.

`eval` package required minor refactor to allow passing in a record that satisfied `govaluate.Parameters`.


jamesdphillips:
> The PR #1117 might have caused a big conflict, sorry about that!

I intentionally didn't touch that function because I figured you had a PR coming, I _think_ it should be a simple rebase. 🎉

PR #1125, opened 2018-03-06
Refactor the store, queue and ring.

### LINKED ISSUE #987: Liberate etcd client from store
Packages like queue and ring need be created with an etcd client. Currently, that involves getting the client from the Store struct, and creating custom interfaces in places where you would need to use a store and a queue or ring. Instead, we should be able to pass the client to a queue or ring getter function, and pass the result of that function call down where the queue or ring needs to be used. For example:
```
// this would happen in main or backend
client, err := getEtcdClient()
queueGetter := newQueueGetter(client)

// this would be called where you need to instantiate a queue
queue := queueGetter.GetQueue("myqueue")
```

echlebek:
## What is this change?

This is a significant refactoring that aims to accomplish several
things.

* Don't combine the Store interface with Queue and Ring interfaces.

* Ensure that Queue and Ring are decoupled and represented in the
types package.

* Provide mock and fake implementations of Queue and Ring.

* Don't export internal state of various daemon types. Instead,
create Newers that handle setting the daemons' internal state.
This is more idiomatic and also reduces the chance of needing to
break the API, by having fewer symbols exported.

* Use consistent terminology and APIs between queue and ring
packages for getters.

* Implement types.QueueItem, and have Queue implementations make
use of it.

Closes #987 

With the beta milestone coming up, we need to solidify the API so that it does not break in the future. This change pushes us towards that by reducing API surface, and decoupling some interfaces.

Yes


backend/agentd/agentd.go
```diff
@@ -31,22 +33,36 @@ type Store interface {
...
+	tls        *types.TLSOptions
+}
 
-	Store      Store
-	Host       string
-	Port       int
-	MessageBus messaging.MessageBus
-	TLS        *types.TLSOptions
+// New creates a new Agentd.
+func New(client *clientv3.Client, etcdName, host string, port int, bus messaging.MessageBus, tls *types.TLSOptions) *Agentd {
```
palourde: I don't necessarily have all the context behind https://github.com/sensu/sensu-go/issues/987 but I'm a bit concerned about exposing etcd directly within the daemons.

I loved the idea of having a _generic_ store interface which hides all implementation details of the actual datastore. But maybe that's not relevant anymore (e.g. we do not plan on supporting an another store) or there's more cons than pros to that approach!

echlebek: @palourde,

The daemons themselves still use a generic interface for storage, we just create that storage within the `New` function. How would you feel about a `NewWithEtcdStore` name instead?

palourde: @echlebek Sorry, I wasn't really clear. I meant exposing the etcd package into the daemon packages.

Currently in master, if I wanted to introduce a new datastore other than etcd, I could simply replace the store that get created & passed in `backend/backend.go`. With this PR, the implementation details of the datastore get propagated within the daemon packages. But maybe it's not a bad thing, I'm just concerned it removes an abstraction layer of the store.

Am I making sense?

grepory: Let’s be clear. One does not simply replace etcd. ;)

palourde: Alright, let's ignore my comment then :)

echlebek: @palourde @grepory No, I think Simon's comment actually makes sense. Here is what the Newers would look like if we didn't pass in the client:

```
func New(store store.Store, qg types.QueueGetter, rg types.RingGetter, bus messaging.MessageBus, host string, port int) * Agentd {
}
```

My thought was that we could write new Newers if we ever replaced etcd. I agree with Greg that it would be a non-trivial effort.

We could also do something like this:

```
type DaemonConfig struct {
    Store store.Store
    QueueGetter types.QueueGetter
    RingGetter types.RingGetter
    // blah blah blah
}
```

Then we could add fields without breaking backwards compatibility, and pass a single field in to the Newers.

backend/apid/actions/checks.go
```diff
@@ -35,38 +34,32 @@ var (
...
 // NewCheckController returns new CheckController
-func NewCheckController(store CheckStore) CheckController {
+func NewCheckController(store store.CheckConfigStore, getter types.QueueGetter) CheckController {
 	return CheckController{
-		Store:      store,
-		Policy:     authorization.Checks,
-		checkQueue: store.NewQueue(adhocQueueName),
+		store:      store,
+		policy:     authorization.Checks,
+		checkQueue: getter.GetQueue(adhocQueueName),
```
mercul3s: Hooray for getting NewQueue out of the store 🎉

types/queue.go
```diff
@@ -0,0 +1,39 @@
...
+	Enqueue(ctx context.Context, value string) error
+
+	// Dequeue gets an Item from the queue. It returns the Item and any
+	// error encountered, or if the context is cancelled.
+	Dequeue(ctx context.Context) (*QueueItem, error)
+}
+
+// QueueItem contains a Value and embeds AckNack.
+type QueueItem struct {
+	// Value is the item's underlying value.
```
mercul3s: Out of curiosity, what's the rationale for splitting the item and ack/nack functions? I tend to think of them as coupled, and I guess they are. I think reading the implementation as interfaces here is pretty straightforward and makes sense, but is a little confusing within the queue itself.

backend/agentd/agentd.go
palourde: I like this idea of `DaemonConfig`!

types/queue.go
echlebek: I could have done:

```
type QueueItem interface {
    Ack(context.Context) error
    Nack(context.Context) error
    Value() string
}
```

Would that be preferable?

mercul3s: Yeah, I think that would be a little clearer!

Issue #1135, opened 2018-03-07
Add new issue github template


grepory:
New issues in GitHub should default to a template that describes how to file a good bug report.

Things we absolutely need:

Version information for sensuctl, sensu-backend, and/or sensu-agent.

If this was installed by a package.

Operating system and version.

A stack trace, if they have one.

Any relevant logs they can find.

What they expected to happen.

what actually happened.

The Sensu 1.x repo already has one of these, we can probably adapt it with good success.

portertech:
The Sensu 1.x issue template https://github.com/sensu/sensu/blob/master/.github/ISSUE_TEMPLATE.md

PR #1138, opened 2018-03-08
Add issue template for GitHub

palourde:

## What is this change?

It adds an issue template for GitHub.

Closes https://github.com/sensu/sensu-go/issues/1135

Added!

Nope!

Nope!


palourde:
@grepory That's normal in markdown view, it does the same thing on sensu/sensu: https://github.com/sensu/sensu/blob/master/.github/ISSUE_TEMPLATE.md.

But they do appear in the raw view and they will also appear in the markdown editor when creating a new issue!

grepory:
That is very bizarre and somewhat confusing. Can we just make them _underline_ instead?

grepory:
Or is there some advantage to this that I'm missing?

palourde:
I believe it's simply to reduce the noise if someone doesn't remove these instructions, but I don't really have an opinion on that.

Any thoughts @portertech @cwjohnston?

grepory:
Does it go away if they submit without removing?

grepory:
OK. I see how it works now. Sorry. Thanks :D

Issue #1108, opened 2018-03-02
megacheck is running out of memory


palourde:
Looks like megacheck sometimes gets killed, most likely because it's [running out of memory](https://github.com/travis-ci/travis-ci/issues/4192):

```
./build.sh: line 154: 13415 Killed                  megacheck $(go list ./... | grep -v dashboardd | grep -v agent/assetmanager | grep -v scripts)
The command "./build.sh $TEST_SUITE" exited with 137.
```

We should probably try to split this command into multiple smaller commands:
https://github.com/sensu/sensu-go/blob/a39f11a7820d16d16245ced0eaafb90d680062cc/build.sh#L163

portertech:
MEGACHECK DEMANDS MEGA MEMORY!

![groans](https://user-images.githubusercontent.com/149630/36915946-53477a18-1e07-11e8-9f51-0238c6152db1.gif)

echlebek:
Is there some way we can get a higher ulimit?

palourde:
I saw that suggestion, we should probably test that: https://github.com/travis-ci/travis-ci/issues/4192#issuecomment-161800452

PR #1139, opened 2018-03-08
Increase stack size on Travis CI for megacheck

palourde:

## What is this change?

It increases the stack size to ~1GB in order to prevent megacheck from being killed by Travis CI.

Closes https://github.com/sensu/sensu-go/issues/1108

Added

Nope!

Nope!


portertech:
megatron_laughing

PR #1141, opened 2018-03-08
Add cli support for CRUD mutators

### LINKED ISSUE #1119: Add cli management for mutators
Update sensuctl with the ability to CRUD mutators.

mercul3s:
## What is this change?

Adds support for mutators to the cli.

Closes #1119 

Yup!

Nope!

I also noticed an incorrect flag for filters in the handler interactive code, so I fixed it.

### DIFF
--- a/cli/client/mutator.go
+++ b/cli/client/mutator.go
@@ -8,11 +8,11 @@ import (
 	"github.com/sensu/sensu-go/types"
 )
 
-// ListMutators fetches all mutators from configured Sensu instance
-func (client *RestClient) ListMutators() ([]types.Mutator, error) {
+// ListMutators fetches all mutators from the configured Sensu instance
+func (client *RestClient) ListMutators(org string) ([]types.Mutator, error) {
 	var mutators []types.Mutator
 
-	res, err := client.R().Get("/mutators")
+	res, err := client.R().Get("/mutators?org=" + url.QueryEscape(org))
 	if err != nil {
 		return mutators, err
 	}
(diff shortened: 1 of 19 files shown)


cli/commands/handler/interactive.go
```diff
@@ -54,7 +54,7 @@ func (opts *handlerOpts) withHandler(handler *types.Handler) {
 
 func (opts *handlerOpts) withFlags(flags *pflag.FlagSet) {
 	opts.Command, _ = flags.GetString("command")
-	opts.Filters, _ = flags.GetString("handlers")
+	opts.Filters, _ = flags.GetString("filters")
```
jamesdphillips: 😰

cli/commands/mutator/info.go
```diff
@@ -0,0 +1,81 @@
...
+
+			if format == "json" {
+				if err := helpers.PrintJSON(r, cmd.OutOrStdout()); err != nil {
+					return err
+				}
+			} else {
+				printToList(r, cmd.OutOrStdout())
+			}
+
+			return nil
```
jamesdphillips: Maybe we could simplify to

```golang
if format == "json" {
  return helpers.PrintJSON(r, cmd.OutOrStdout())
}
printToList(r, cmd.OutOrStdout())
return nil
```

jamesdphillips: (On a related note.. honestly `printToList` should probably return an error, but, that's an issue with my `list` package and totally out of scope.)

cli/commands/handler/interactive.go
```diff
@@ -54,7 +54,7 @@ func (opts *handlerOpts) withHandler(handler *types.Handler) {
 
 func (opts *handlerOpts) withFlags(flags *pflag.FlagSet) {
 	opts.Command, _ = flags.GetString("command")
-	opts.Filters, _ = flags.GetString("handlers")
```
palourde: giphy

cli/commands/mutator/update.go
```diff
@@ -0,0 +1,55 @@
...
+			}
+
+			// Administer questionnaire
+			opts := newMutatorOpts()
+			opts.withMutator(mutator)
+			if err := opts.administerQuestionnaire(true); err != nil {
+				return err
+			}
+
+			// Apply given arguments to check
```
palourde: Maybe just update this comment 😄

cli/client/mutator.go
```diff
@@ -25,14 +25,71 @@ func (client *RestClient) ListMutators() ([]types.Mutator, error) {
...
+		return err
+	}
+
+	res, err := client.R().SetBody(bytes).Patch("/mutators/" + url.PathEscape(mutator.Name))
+	if err != nil {
+		return err
+	}
+
+	if res.StatusCode() >= 400 {
+		return fmt.Errorf("%v", res.String())
```
jamesdphillips: Unfortunately it's used inconsistently but ideally you could use `unmarshalError`. See `client/check.go` for example https://github.com/sensu/sensu-go/blob/master/cli/client/check.go#L54-L56.

cli/commands/mutator/info.go
mercul3s: oh yeah, that is much cleaner. Thanks for the suggestion, I'll change it!

mercul3s: Should I add an issue for printToList returning an error?

jamesdphillips: No worries, I can do it.

mercul3s:
One of the tests is oddly failing in the linter stage from OOM errors with mega check; everything else passes, and the linter is fine locally so I'm going to go ahead and merge this.

Issue #1027, opened 2018-02-16
Automatically create a "default" environment when creating a new organization


amdprophet:
After some investigation it seems that creating a new organization does not automatically create a new environment. I'd like to see an environment called, "default" created when a new organization is made.

zbintliff:
This is my lack of overall design knowledge of this project comes in.  The ticket makes it seem like `organization` contain `environments` but its seems the exact opposite (especially with the links going only from environments to organizations).  So then I got into a :rabbit: :hole:  trying to figure out if environments were global/unique, if organizations can live in multiple envs, should there be a reverse link from orgs to their envs, etc. 

From my digging it seems the "default" env should already exist and the  ticket is really just asking to register the org to the default env upon the org's creation. Is that a correct reduction?

grepory:
Each organization has one or more environment, and those environments are unique to that organization. The issue we have is that it’s currently possible to create an organization with no environment. Doing this causes us to hit a few edge cases where we built in the assumption that there is always an environment within an org.

PR #1142, opened 2018-03-08
Automatically create a "default" environment when creating a new organization

palourde:
## What is this change?

It automatically adds an environment `default` when creating a new organization. To do so, we now have a distinction in the store between creating and updating an organization.

Closes https://github.com/sensu/sensu-go/issues/1027.

Added!

Nope!

Nope!


backend/store/etcd/organization_store.go
```diff
@@ -20,6 +20,55 @@ func getOrganizationsPath(name string) string {
...
+	envBytes, err := json.Marshal(env)
+	if err != nil {
+		return err
+	}
+
+	orgKey := getOrganizationsPath(org.Name)
+
+	res, err := s.client.Txn(ctx).
+		If(
+			// Ensure the organization does not already exist
```
echlebek: Why is it important to ensure that the organization does not already exist?

palourde: To me, it's simply a question of UX; if I run `sensuctl organization create`, I don't expect it to update an organization. Same goes with `sensuctl organization update`, I wouldn't expect this command to create an organization if it's missing. Makes sense?

echlebek: Since Organizations are resources that are named by users and not Sensu, they are compatible with PUT semantics. If the store is designed this way, PUT semantics are impossible and we have to rely on POST and PATCH, each of them failing when the preconditions are not what the system expects.

What's nice about PUT is that it is idempotent - the operation has the same effect no matter how many times it is applied. This is really important for something like `sensuctl create -f`, where we create multiple resources, because we can repeat the operation without worry if one of the operations fails.

I think that for these types of resources, create and update should have identical semantics in the store. If an organization's description is replaced, nothing of value is lost. We only really need error-if-exists type semantics in the store in cases where creation has side effects, like destroying other bits of information that the system depends on.

We can still achieve the desired UX in sensuctl by using POST and PATCH - we simply have to attempt to load the resource first before calling `UpdateOrganization`. That's not transactional, but I think that's actually OK in these cases.

palourde: You are totally right, it makes total sense. I'll make the required changes!

palourde: Waiting on https://github.com/sensu/sensu-go/pull/1154 to be merged first!

grepory:
@palourde since #1154 has been merged, can we move forward on this one?

palourde:
@grepory Yep!

backend/store/etcd/organization_store.go
palourde: @echlebek Alright so I updated this PR so the new method `CreateOrganization` only creates an org if it does not already exist, while `UpdateOrganization` can create or update an org!

palourde:
@mercul3s @echlebek Ready for review!

mercul3s:
LGTM!

PR #1152, opened 2018-03-09
[Web UI] Upgrade material-ui dependency to latest

jamesdphillips:
## What is this change?

Update Material-UI dependency for fixes to palette. Compare [v1.0.0-beta.30...v1.0.0-beta.35](https://github.com/mui-org/material-ui/compare/v1.0.0-beta.30...v1.0.0-beta.35).

Previously we were receiving unexpected behavior when hover list items.

We were a ways behind and there were a number of breaking changes.


types/event.go
```diff
@@ -93,6 +93,16 @@ func (e *Event) Get(name string) (interface{}, error) {
 		return e.Silenced, nil
 	case "Hook":
 		return e.Hooks, nil
+	case "HasCheck":
```
echlebek: Why is this Get method necessary? It seems that govaluate should be able to query Events without implementing govaluate.Parameters here.

grepory: @jamesdphillips

jamesdphillips: > It seems that govaluate should be able to query Events without implementing govaluate.Parameters here.

@echlebek This is true when we were using `govaluate.MapParameters`, however for ease of use, in some cases I would prefer it if Event implemented `govaluate.Parameters`.

For instance in the web UI, when filtering a collection of events it is more intuitive to me as a user to write the expression `HasCheck && Check.Status > 0` than `event.HasCheck && event.Check.Status > 0`. Not only is it extra key strokes but other experiences on the web (eg. using lucene syntax) will have taught users that the fields of the record being filtered are present in the root.

I would agree that it is sort of annoying to maintain a `Get` method on the Event. If the API continues to shift, we could probably write a gogoproto plugin to generate the method.

echlebek: @jamesdphillips I see, that makes sense to me. I kind of doubt we'll be adding more accessors to the Event in the future so I wouldn't worry about maintenance.

We've moved Hooks and Silenced out of Event now, so those should be removed.

PR #1154, opened 2018-03-12
Apid/put support

### LINKED ISSUE #1153: Enable HTTP PUT where possible
A lot of our resources have URLs that are directly derived from their names. Checks, Assets, and so on. We should have API routes for PUTting these resources, to support #1056.

We don't need to modify the CLI, just add support for PUT where it makes sense to do so, in apid.

echlebek:
## What is this change?

It puts the put in apid!

Closes #1153 

Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 ## Unreleased
 ### Added
 - Added ability to sort events in web UI.
+- Add PUT support to APId for the various resource types.
 
 ### Changed
 - Changed the maximum number of open file descriptors on a system to from 1024
--- a/backend/apid/actions/assets.go
+++ b/backend/apid/actions/assets.go
@@ -136,3 +136,27 @@ func (a AssetController) Update(ctx context.Context, given types.Asset) error {
 
 	return nil
 }
+
+// CreateOrReplace creates or replaces the asset given.
+func (a AssetController) CreateOrReplace(ctx context.Context, asset types.Asset) error {
+	// Adjust context
+	ctx = addOrgEnvToContext(ctx, &asset)
+	abilities := a.Policy.WithContext(ctx)
+
+	// Verify viewer can make change
+	if !(abilities.CanUpdate() && abilities.CanCreate()) {
+		return NewErrorf(PermissionDenied)
+	}
+
+	// Validate
+	if err := asset.Validate(); err != nil {
+		return NewError(InvalidArgument, err)
+	}
+
+	// Persist Changes
+	if serr := a.Store.UpdateAsset(ctx, &asset); serr != nil {
+		return NewError(InternalErr, serr)
+	}
+
+	return nil
+}
--- a/backend/apid/actions/assets_test.go
+++ b/backend/apid/actions/assets_test.go
@@ -180,6 +180,106 @@ func TestAssetFind(t *testing.T) {
 	}
 }
 
+func TestAssetCreateOrReplace(t *testing.T) {
+	defaultCtx := testutil.NewContext(
+		testutil.ContextWithOrgEnv("default", "default"),
+		testutil.ContextWithRules(
+			types.FixtureRuleWithPerms(
+				types.RuleTypeAsset,
+				types.RulePermCreate,
+				types.RulePermUpdate,
+			),
+		),
+	)
+	wrongPermsCtx := testutil.NewContext(
+		testutil.ContextWithOrgEnv("default", "default"),
+		testutil.ContextWithRules(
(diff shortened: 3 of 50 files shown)


palourde:
I think we should take a step back and reevaluate how our API works. This PR introduces a distinction between `PUT` & `PATCH` methods (which is definitively a good thing) but the problem I see here is that sensuctl does not support the PATH method so I feel like we are introducing unnecessary complexity.

I don't think we ever did this exercice but I believe it could be a good idea to define some specs around HTTP methods. I would like to propose the following:

| HTTP Method | CRUD Equivalent | Route | Description | Return codes |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| DELETE  | Delete  |  /resource/:id  |  Deletes a resource identified by a URI  |  200 (OK). 404 (Not Found).  |
| GET  | Read  |  /resource/{:id}  |  Retrieves all resources or a single resource identified by a URI.  |  200 (OK). 404 (Not Found).  |
| PATCH  | Update/Modify  |  /resource/:id  |  Modifies a given resource identified by a URI. The request only contains the changes to the resource, not the complete resource.  |  200 (OK). 404 (Not Found).  |
| POST  | Create  |  /resource  |  Creates a resource with the given request body  |  201 (Created). 404 (Not Found). 409 (Conflict) if resource already exists.  |
| PUT  | Update/Replace  |  /resource/:id  |  Replaces the resource identified by a URI with the request body. If the resource does not exist, it gets created.  | 200 (OK) if replaced. 201 (Created) if created. 404 (Not Found). | 

From this point, I propose that we remove support, in this PR, for `PATCH` requests in the API & the store, since sensuctl does not support that and the methods in the store for these requests actually replace the resource with the given request body instead of updating the resource with the changes in the request body.

Any thoughts?

echlebek:
@palourde, hmm, I was unaware that we weren't using PATCH in the CLI. We definitely _could_ be, since many of our CLI commands could do partial updates.

Since our API could be used by clients other than sensuctl, I actually think it's OK to have methods that the sensuctl tool does not use.

You bring up a good point around return codes - we need to clean those up a bit and make sure we are returning the right HTTP code in all circumstances.

The PUT case is a little difficult because we don't necessarily know if we are creating or updating with our current store. We could amend the store to be a little more informative there.

palourde:
@echlebek I'm suggesting to remove the `PATCH` routes / methods for now not because `sensuctl` doesn't use it, but because it doesn't properly handle `PATCH` request (which should only contains the changes to the resource, not the complete resource). That being said, we should definitively implement that but probably in a new issue/PR.

Regarding the HTTP return code for PUT request, I think it's fair to return 200 for now, whether it was replaced or created, and implement the proper return code on a subsequent issue so we don't scope creep this PR!

echlebek:
@palourde I see, yes, I believe PATCH is not working correctly in all cases. If others are onboard with removing it I agree. I think eventually it would be nice to have.

jamesdphillips:
> I believe PATCH is not working correctly in all cases. If others are onboard with removing it I agree. I think eventually it would be nice to have.

I'm curious, in which cases is it not working correctly?

echlebek:
@jamesdphillips this is one case I could find. https://github.com/sensu/sensu-go/blob/master/backend/apid/actions/mutators.go#L52

The mut is validated before we copy the fields to the resulting data structure, so it's possible we can't apply a partial.

Looking at it further I think PATCH is actually working in the majority of cases. Maybe we should just make sure partial updates are working in all cases and leave it in.

palourde:
As per [RFC 5789](https://tools.ietf.org/html/rfc5789), a PATCH request should only contain the changes to a resource. E.g.I If I want to update the URL of an asset, I should be able to issue a request to /assets/:id with the following body:

```
{
  "url": "https://foo.bar/baz.tar.gz"
}
```

That's currently not the case, as Eric mentioned:

> PATCH basically means PUT-that-returns-an-error-if-not-found

Having proper support for PATCH requests would significantly reduce the complexity in a lot of sensuctl subcommands (e.g. `sensuctl check set-handlers`) since we don't have to pull the resource, update it, validate it and finally submit it directly within sensuctl; that could automatically be done by the API.

jamesdphillips:
@palourde I understand what you mean now. The intended behavior is definitely not correct. Since we unmarshal the payload into the resource we intend to PATCH, even if you didn't provide all the fields you would end up with their defaults.

I can make a task.

echlebek:
@palourde @jamesdphillips oh wow, I had not event considered the defaults aspect. You are correct, we would have to handle PATCH differently.

echlebek:
I'm going to continue working on this, and change the PATCH calls in the CLI to be PUT, so that the incorrect patch routes can be removed in this PR.

palourde:
Awesome, thank you very much! 🙏

PR #1162, opened 2018-03-14
Bugfix/agentd sessions

### LINKED ISSUE #1161: Agent sessions are never terminated
## Expected Behavior
When an agent disconnects, the agent's session's Stop() should be called, terminating the goroutines that are responsible for it.

## Current Behavior
The Session runs forever.

## Possible Solution
When a websocket error is encountered, shut down the Session.

echlebek:
## What is this change?

Allows sessions to shut down cleanly when the connection is disrupted.

Sessions are not shutting down... ever.

Closes #1161 

Yes

### DIFF
--- a/backend/agentd/session.go
+++ b/backend/agentd/session.go
@@ -97,6 +97,11 @@ func (s *Session) recvPump() {
 	defer func() {
 		logger.Info("session disconnected - stopping recvPump")
 		s.wg.Done()
+		select {
+		case <-s.stopping:
+		default:
+			s.Stop()
+		}
 	}()
 
 	for {
@@ -110,7 +115,6 @@ func (s *Session) recvPump() {
 			switch err := err.(type) {
 			case transport.ConnectionError, transport.ClosedError:
 				logger.Warn("stopping session <%s>: recv error: ", s.ID, err.Error())
-				s.Stop()
 				return
 			default:
 				logger.Error("recv error: ", err.Error())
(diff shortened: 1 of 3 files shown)


backend/agentd/session.go
```diff
@@ -206,6 +194,12 @@ func (s *Session) Start() error {
 	org, env := s.cfg.Organization, s.cfg.Environment
 	agentID := s.cfg.AgentID
 
+	defer func() {
+		if err != nil {
```
palourde: I'm probably missing it but where do we assign something to `err`?

echlebek: @palourde it's the named return variable on line 187

backend/agentd/session.go
```diff
@@ -242,6 +236,9 @@ func (s *Session) Stop() {
 			logger.Debug(err)
 			break
 		}
+	}
+	for _, sub := range s.cfg.Subscriptions {
+		topic := messaging.SubscriptionTopic(org, env, sub)
```
palourde: Why not simply remove `break` at line 237?

backend/agentd/session.go
palourde: Oh my bad, I got confused by how the PR presented the changes!

backend/agentd/session.go
echlebek: If the message bus shuts down, we don't need to continue unsubscribing, but that's not relevant to the ring.

palourde: I see, thanks!

PR #1164, opened 2018-03-14
Only use one lease per ring

### LINKED ISSUE #1163: Ring leases are never revoked
Ring leases need to be revoked, or the keepalive will keep the lease around forever. This leads to a goroutine leak even after the key in question is removed.

echlebek:

## What is this change?

Only uses a single lease, and keepalive, per ring. This keeps us from creating a ton of goroutines.

Closes #1163 

Yes


mercul3s:
cool

PR #1168, opened 2018-03-14
Use a package level cache for ring leases.

### LINKED ISSUE #1163: Ring leases are never revoked

echlebek:
Because new rings are made again and again, even for rings that
already exist, checking for a ring-scoped lease ID is not useful.
Instead, keep a cache of the lease IDs at the package scope,
keyed by ring ID.

## What is this change?

Closes #1163 

The previous fix was suboptimal.

### DIFF
--- a/backend/ring/ring.go
+++ b/backend/ring/ring.go
@@ -30,6 +30,9 @@ var (
 
 	backendID   string
 	backendOnce sync.Once
+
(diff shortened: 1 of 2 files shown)


grepory:
@echlebek :shipit:

PR #1170, opened 2018-03-15
Fixed shutting down message to not include %s

grepory:

## What is this change?

Removes an unnecessary format string parameter in backend shutdown messages.

Cosmetics.

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -188,7 +188,7 @@ type stopGroup []daemonStopper
 
 func (s stopGroup) Stop() (err error) {
 	for _, stopper := range s {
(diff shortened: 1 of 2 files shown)


backend/backend.go
```diff
@@ -188,7 +188,7 @@ type stopGroup []daemonStopper
 
 func (s stopGroup) Stop() (err error) {
 	for _, stopper := range s {
-		logger.Info("shutting down %s", stopper.Name)
+		logger.Info("shutting down ", stopper.Name)
```
echlebek: This will cause two spaces to be between the words "down" and stopper.Name ;)

PR #1171, opened 2018-03-15
Comment out descriptions of PR template headings

grepory:

## What is this change?

This updates the PR template to comment out the descriptions of these heads.

We do this in our issue template, so it seems reasonable to be consistent.

As this is a cosmetic change to the PR template and no functional change to the code, it seems reasonable not to have a changelog entry.


mercul3s:
lgtm

PR #1173, opened 2018-03-15
Add cli flag for environment variables

### LINKED ISSUE #1156: Add option to configure environment vars via sensuctl for handlers and mutators
Add flag env-vars to sensuctl mutators and handlers. It should take a comma separated list of key=value pairs, and turn that into a list of strings that sets the EnvVars attribute of handlers and mutators types.

mercul3s:
## What is this change?

Adds support for setting environment variables via sensuctl.

Closes #1156 

Changelog updated

Nope!

Nope!

### DIFF
--- a/cli/commands/handler/interactive.go
+++ b/cli/commands/handler/interactive.go
@@ -12,14 +12,15 @@ import (
 
 type handlerOpts struct {
 	Name       string `survey:"name"`
-	Type       string `survey:"type"`
-	Mutator    string `survey:"mutator"`
 	Command    string `survey:"command"`
-	Timeout    string `survey:"timeout"`
+	EnvVars    string `survey:"env-vars"`
 	Filters    string `survey:"filters"`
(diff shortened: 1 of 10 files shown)


cli/commands/handler/interactive.go
```diff
@@ -218,6 +229,7 @@ func (opts *handlerOpts) Copy(handler *types.Handler) {
 	handler.Organization = opts.Org
 
 	handler.Command = opts.Command
+	handler.EnvVars = strings.Split(opts.EnvVars, ",")
```
jamesdphillips: See `cli/commands/helpers/flags.go` there is a func for handling whitespace in comma separated lists. May help if someone tries `--env-vars 'ENV=production, WOOF=PACK ,     ETC=ETCD'`

echlebek:
Just felt the need to mention - if we made this whitespace separated instead of comma separated, then people could use the output of `env`.

PR #1174, opened 2018-03-15
Handle round-robin scheduling through WizardBus.

### LINKED ISSUE #1165: Move ring-based message passing into WizardBus out of schedulerd/session
Currently, RoundRobin is a boolean attribute on a check configuration. This is desirable, because now round-robin can be check-based instead of subscription-based and we no longer have to worry about prepending `roundrobin:` in front of subscriptions. We can also get truthy round-robin on a check level.

In order for this to happen, though, we have to centralize knowledge of whether or not a particular subscription has a round-robin check somewhere. The agent session cannot know, because it must remain ignorant of which checks belong in which subscriptions. The scheduler _knows_ which checks are there, but not which agents it needs to talk to.

So, as people do when they have a message bus, we'll add more functionality to the message bus.

Refactor the MessageBus to include publishing methods for both Direct messaging and Broadcast messaging--the current Publish method now being the latter. In this case, I think it's best to just add PublishDirect and leave Publish alone. That should be the least amount of rework.

For each subscription in the WizardBus there exists a WizardTopic. The WizardTopic manages a mapping of agent_id->channel. When WizardPub receives a PublishDirect request, the WizardTopic must be told (somehow) to add its agents to a ring for that topic and then use the ring to determine if it should publish or not.

PublishDirect must call a corresponding SendDirect on the WizardTopic so that the topic knows to use the ring.

echlebek:
* Add PublishDirect method to messaging.MessageBus.
* Use ring for WizardBus PublishDirect implementation.
* Refactor codebase to use messaging.NewWizardBus.
* Remove ring-specific code from schedulerd.
* Remove ring-specific code from agentd.

Closes #1165 

Affirmative

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 (default) to 65535.
 - Increased the default etcd size limit from 2GB to 4GB.
 - Move Hooks and Silenced out of Event and into Check.
+- Handle round-robin scheduling in wizardbus.
 
 ### Fixed
 - Shut down sessions properly when agent connections are disrupted.
--- a/backend/agentd/agentd.go
+++ b/backend/agentd/agentd.go
@@ -45,17 +45,15 @@ type Agentd struct {
 	store      Store
 	bus        messaging.MessageBus
 	tls        *types.TLSOptions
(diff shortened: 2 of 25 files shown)


mercul3s:
thumbs up

PR #1187, opened 2018-03-19
Add sensuctl create command

### LINKED ISSUE #1056: Add sensuctl create command to the CLI
`sensuctl` should be able to create resources from files that specify the resource types to be created.

The UI should look something like
```
sensuctl create -f big_list_o_checks.json
```

To accomplish this, we'll need to add a wrapper type that can specify any number of any resources, and a parser for the type to translate it into Sensu resources.

After parsing, the command will make one or more HTTP requests to the backend to create whatever resources are spec'd by the input file.

The command should only make HTTP requests once all resources are validated. If there are errors in the resource definition, one or more _helpful_ errors messages should be output, detailing the problems. In the context of this tool, _helpful_ means that a person could reasonably use the error message to diagnose the problem with their input spec.

If the `-f` flag is omitted, then `sensuctl create` should expect input from stdin.

echlebek:
commit 2a7a9ea4cdc97d8afeb04fd15c44990da8f8f07b (HEAD -> cli/sensuctl-create, origin/cli/sensuctl-create)
Author: Eric Chlebek <eric@sensu.io>
Date:   Sat Mar 17 14:03:50 2018 -0700

    Add sensuctl create command.
    
    The sensuctl create command can create arbitrary Sensu resources
    by parsing one or more Sensu resource definitions.
    
    As a convenience, users can omit organization and environment in
    the resources, and the command-line values for these fields will
    be used instead.
    
    This commit fixes a few issues with the work in the types package
    that was done in the previous commit.
    
commit 7fd33b0ebf067cb819817d74f37ca547d021da8f
Author: Eric Chlebek <eric@sensu.io>
Date:   Fri Mar 16 11:26:08 2018 -0700

    Add dynamic resource resolution.
    
    * Add type Resource, which is an interface that specs URIPath.
    * Add function ResolveResource, which can deliver a zero-valued
    type which implements Resource, given a valid Resource name.
    * Add type Wrapper, which users can supply in json format. The
    wrapper type contains a type name and a spec. When Wrapper is
    unmarshaled, ResolveResource is called to discover its concrete
    type.
    
    ResolveResource is implemented using go generate and a go program
    that uses godoc to discover all of the exported types in the types
    package.

Closes #1056 

Yes


echlebek:
Requesting reviews from portertech and grepory due to the impact of this feature.

portertech:
cheeky_excited

Giving this a whirl.

echlebek:
Adding Simon and Nikki as reviewers since Sean and Greg are quite busy.

types/hook.go
```diff
@@ -3,6 +3,7 @@ package types
 import (
 	"encoding/json"
 	"errors"
+	fmt "fmt"
```
palourde: giphy

Looks like we have a couple of them!

types/wrapper.go
```diff
@@ -0,0 +1,45 @@
...
+type Wrapper struct {
+	// Type is the fully-qualified type name, e.g.
+	// github.com/sensu/sensu-go/types.Check,
+	// OR, a short-hand name that assumes a package path of
+	// github.com/sensu/sensu-go/types.
+	Type string `json:"type"`
+
+	// Value is a valid Resource of concrete type Type.
+	Value Resource `json:"spec"`
+}
```
palourde: I believe we should also open an issue so sensuctl uses that `Wrapper` struct when printing a resource in JSON, so we can easily do something like this:

```bash
$ sensuctl check info check-cpu --format json > check-cpu.json
# modify the command or something else
$ sensuctl create -f check-cpu.json
```

Makes any sense?

cli/commands/create/create_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/stretchr/testify/require"
+
+	mockclient "github.com/sensu/sensu-go/cli/client/testing"
+	cmdtesting "github.com/sensu/sensu-go/cli/commands/testing"
+)
+
+var resourceSpecTmpl = template.Must(template.New("test").Parse(`
+{"type": "Check", "spec": {{ .Check }} }
+{"type": "Asset", "spec": {{ .Asset }} }
+{"type": "Hook", "spec": {{ .Hook }} }
```
palourde: Do you think we could also easily support uncapitalized resource type? I did instinctively used something like `check` and got the following error:

``` bash
$ sensuctl create -f check-cpu.json
error parsing resource: error parsing spec: the named type could not be found
Error: error parsing resources
```

types/hook.go
echlebek: Annoyingly, `goimports` started doing this out of nowhere. I'll look into it.

types/wrapper.go
echlebek: Makes sense to me. Are you thinking that the API would wrap all of its responses?

cli/commands/create/create_test.go
echlebek: Yeah, that would be just a slight modification to the template. We can support both.

types/wrapper.go
palourde: Oh yes, that could be pretty cool! I'll open an issue for that

palourde: Done: https://github.com/sensu/sensu-go/issues/1221

types/handler.go
```diff
@@ -1,6 +1,10 @@
 package types
 
-import "errors"
+import (
+	"errors"
+	fmt "fmt"
```
grepory: fmt fmt

cli/commands/create/create.go
```diff
@@ -0,0 +1,160 @@
...
+
+func execute(cli *cli.SensuCli) func(*cobra.Command, []string) error {
+	return func(cmd *cobra.Command, args []string) error {
+		var in io.Reader
+		if len(args) > 1 {
+			_ = cmd.Help()
+			return errors.New("invalid argument(s) received")
+		}
+		defaultOrg, _ := cmd.Flags().GetString("organization")
+		defaultEnv, _ := cmd.Flags().GetString("environment")
```
grepory: Does this bypass the configured org/env in the sensuctl config?

types/event.go
```diff
@@ -3,6 +3,8 @@ package types
 import (
 	"encoding/json"
 	"errors"
+	fmt "fmt"
```
grepory: fmt fmt

types/environment.go
```diff
@@ -3,6 +3,7 @@ package types
 import (
 	"errors"
 	fmt "fmt"
```
grepory: fmt fmt

cli/commands/create/create_test.go
```diff
@@ -0,0 +1,78 @@
...
+}{
+	Check: mustMarshal(fixtureCheck),
+	Asset: mustMarshal(fixtureAsset),
+	Hook:  mustMarshal(fixtureHook),
+}
+
+func TestCreateCommand(t *testing.T) {
+	cli := cmdtesting.NewMockCLI()
+	client := cli.Client.(*mockclient.MockClient)
+	client.On("PutResource", mock.AnythingOfType("*types.Check")).Return(nil)
```
grepory: Is it possible to validate that we're correctly deserializing and not just getting zero values here?

echlebek: I can add a test case where validation would fail.

PR #1190, opened 2018-03-20
Rework WizardBus Subscribe() method to remove subscriber ID

### LINKED ISSUE #1097: Agents from different orgs/envs overwrite each other's subscriptions in the message bus
When an agent connects and its subscriptions are started, the org and env of the agent aren't taken into account when providing a subscriber ID to the message bus:

https://github.com/sensu/sensu-go/blob/master/backend/agentd/session.go#L207-L212

We should either include the org+env+agentid in the subscriber ID or we should have the message bus provide an ID for us.

grepory:

## What is this change?

Removes the concept of subscriber IDs from the message bus, and instead Subscribe takes a Subscriber and returns a Subscription (which can be later cancelled).

This fixes #1097. Rather than simply prolonging the life of user-supplied subscriber IDs, we decided to tackle this bit of tech debt now.

Added.

### DIFF
--- a/backend/agentd/session.go
+++ b/backend/agentd/session.go
@@ -232,7 +232,9 @@ func (s *Session) Stop() {
 	s.wg.Wait()
 
 	for sub := range s.subscriptions {
-		sub.Cancel()
+		if err := sub.Cancel(); err != nil {
+			logger.WithError(err).Error("unable to unsubscribe from message bus")
+		}
 	}
 	close(s.checkChannel)
 }
--- a/backend/agentd/session.go
+++ b/backend/agentd/session.go
@@ -38,6 +38,8 @@ type Session struct {
 	sendq        chan *transport.Message
 	checkChannel chan interface{}
 	bus          messaging.MessageBus
+
+	subscriptions chan messaging.Subscription
 }
 
 func newSessionHandler(s *Session) *handler.MessageHandler {
@@ -77,18 +79,24 @@ func NewSession(cfg SessionConfig, conn transport.Transport, bus messaging.Messa
 	logger.Infof("agent connected: id=%s subscriptions=%s", cfg.AgentID, cfg.Subscriptions)
 
 	s := &Session{
-		ID:           id.String(),
-		conn:         conn,
-		cfg:          cfg,
-		stopping:     make(chan struct{}, 1),
-		wg:           &sync.WaitGroup{},
-		sendq:        make(chan *transport.Message, 10),
-		checkChannel: make(chan interface{}, 100),
-		store:        store,
-		bus:          bus,
+		ID:            id.String(),
+		conn:          conn,
+		cfg:           cfg,
+		stopping:      make(chan struct{}, 1),
+		wg:            &sync.WaitGroup{},
+		sendq:         make(chan *transport.Message, 10),
+		checkChannel:  make(chan interface{}, 100),
+		store:         store,
+		bus:           bus,
+		subscriptions: make(chan messaging.Subscription, len(cfg.Subscriptions)),
 	}
 	s.handler = newSessionHandler(s)
 	return s, nil
+}
+
+// Receiver returns the check channel for the session.
+func (s *Session) Receiver() chan<- interface{} {
+	return s.checkChannel
 }
 
 func (s *Session) recvPump() {
@@ -194,7 +202,7 @@ func (s *Session) Start() (err error) {
 	go s.subPump()
 
 	org, env := s.cfg.Organization, s.cfg.Environment
-	agentID := s.cfg.AgentID
+	agentID := fmt.Sprintf("%s:%s:%s", org, env, s.cfg.AgentID)
 
 	defer func() {
 		if err != nil {
@@ -205,11 +213,14 @@ func (s *Session) Start() (err error) {
 	for _, sub := range s.cfg.Subscriptions {
 		topic := messaging.SubscriptionTopic(org, env, sub)
 		logger.Debugf("Subscribing to topic %q", topic)
-		if err := s.bus.Subscribe(topic, agentID, s.checkChannel); err != nil {
+		subscription, err := s.bus.Subscribe(topic, agentID, s)
+		if err != nil {
 			logger.WithError(err).Error("error starting subscription")
 			return err
 		}
+		s.subscriptions <- subscription
 	}
+	close(s.subscriptions)
 
 	return nil
 }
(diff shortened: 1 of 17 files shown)


backend/messaging/message_bus.go
```diff
@@ -22,6 +24,24 @@ const (
 	TopicSubscriptions = "sensu:check"
 )
 
+// A Subscriber receives messages via a channel.
+type Subscriber interface {
+
+	// Receiver returns the channel a subscriber uses to receive messages.
+	Receiver() chan<- interface{}
```
grepory: I'm about to kick myself in the face for asking this, but...

What if we did:
```
Receive(msg interface{})
```

And hid the receiving channel from the messagebus entirely.

echlebek: The problem with this is there's no way to prevent Receive from blocking; we'd have to start up goroutines.

backend/messaging/wizard_topic.go
```diff
@@ -1,41 +1,92 @@
...
+	"strconv"
+	"sync"
+
+	"github.com/sensu/sensu-go/types"
+)
 
 // WizardTopic encapsulates state around a WizardBus topic and its
 // consumer channel bindings.
 type WizardTopic struct {
+	ID       string
```
echlebek: ID, Bindings and Ring are all implementation details and shouldn't be exported.

backend/messaging/wizard_topic.go
```diff
@@ -1,41 +1,92 @@
...
+
+// Unsubscribe a consumer from this topic.
+func (wTopic *WizardTopic) unsubscribe(id int64) {
+	wTopic.Lock()
+	delete(wTopic.Bindings, id)
+	if wTopic.Ring != nil {
+		// TODO: What do we do in the case of this failing?
+		_ = wTopic.Ring.Remove(context.Background(), strconv.FormatInt(id, 10))
+	}
 	wTopic.Unlock()
```
echlebek: defer wTopic.Unlock()

backend/messaging/wizard_topic.go
```diff
@@ -1,41 +1,92 @@
...
+		cancel: wTopic.unsubscribe,
+	}, nil
+}
+
+// Unsubscribe a consumer from this topic.
+func (wTopic *WizardTopic) unsubscribe(id int64) {
+	wTopic.Lock()
+	delete(wTopic.Bindings, id)
+	if wTopic.Ring != nil {
+		// TODO: What do we do in the case of this failing?
```
echlebek: unsubscribe should return an error, and the error should be threaded through to Cancel.

backend/messaging/message_bus.go
```diff
@@ -22,6 +24,24 @@ const (
...
+}
+
+// A Subscription is a cancellable subscription to a WizardTopic.
+type Subscription struct {
+	id     int64
+	cancel func(int64)
+}
+
+// Cancel a WizardSubscription.
+func (t Subscription) Cancel() {
```
echlebek: Cancel should return an error.

backend/messaging/message_bus.go
grepory: Ah right you are.

backend/messaging/wizard_topic.go
grepory: Yeah, good catch.

backend/messaging/wizard_topic.go
grepory: Yeah, they don't need to be, but for that matter, neither does WizardTopic.

backend/messaging/message_bus.go
```diff
@@ -22,6 +24,24 @@ const (
...
+	Receiver() chan<- interface{}
+}
+
+// A Subscription is a cancellable subscription to a WizardTopic.
+type Subscription struct {
+	id     int64
+	cancel func(int64)
+}
+
+// Cancel a WizardSubscription.
```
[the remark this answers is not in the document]
grepory: I don't understand that model at all. That's why I went this route.

backend/agentd/session.go
```diff
@@ -194,21 +202,22 @@ func (s *Session) Start() (err error) {
...
+	for i, sub := range s.cfg.Subscriptions {
 		topic := messaging.SubscriptionTopic(org, env, sub)
 		logger.Debugf("Subscribing to topic %q", topic)
-		if err := s.bus.Subscribe(topic, agentID, s.checkChannel); err != nil {
+		subscription, err := s.bus.Subscribe(topic, s)
+		if err != nil {
 			logger.WithError(err).Error("error starting subscription")
 			return err
 		}
+		s.subscriptions[i] = subscription
```
echlebek: Heh, looks like the CI found a race condition here... seemingly with the Stop method? :man_shrugging:

backend/messaging/message_bus.go
```diff
@@ -22,6 +24,24 @@ const (
...
 
+// A Subscriber receives messages via a channel.
+type Subscriber interface {
+
+	// Receiver returns the channel a subscriber uses to receive messages.
+	Receiver() chan<- interface{}
+}
+
+// A Subscription is a cancellable subscription to a WizardTopic.
+type Subscription struct {
```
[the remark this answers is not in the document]
grepory: I still don't really think this should be an interface. Like I don't really understand your justification for it at all.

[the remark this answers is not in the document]
grepory: TBH, you haven't given any, so that would probably help. :)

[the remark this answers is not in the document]
echlebek: Well if it's not an interface, there's no way to implement messaging.MessageBus outside of the package, making the messaging.MessageBus interface useless.

backend/agentd/session.go
grepory: But only on 386? I didn't even think the race detector worked on anything but amd64

backend/messaging/message_bus.go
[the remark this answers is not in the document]
grepory: It basically _is_ useless. It's a holdover from when there were different types of message bus, but I'm _definitely_ not doing that refactor in this PR.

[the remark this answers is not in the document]
grepory: It can't simply be removed because it's referenced everywhere and mockBus is used somewhere for some reason.

backend/messaging/wizard_topic.go
echlebek: It's definitely worth reducing the API surface as much as possible, so if WizardTopic can be unexported, I think we should do that.

backend/agentd/session.go
grepory: Oh this is annoying. We have the exact same problem anywhere we have Start() and Stop(). Sessions and Daemons aren't safe for concurrent use, but neither of them is used concurrently.

PR #1193, opened 2018-03-20
Updated etcd to 3.3.2

### LINKED ISSUE #1069: Update autocompaction retention setting
Issue [9337](https://github.com/coreos/etcd/issues/9337) has been resolved, and that fix has been backported to 3.3. We should update the config to be 1 instead of 1ns: https://github.com/sensu/sensu-go/blob/master/backend/etcd/etcd.go#L200

mercul3s:
## What is this change?

Updates etcd to 3.3.2 so we can set autocompaction revision to an actual int instead of a time-based value.

Closes #1069 

Added a note under Changed section.

Nope!

Nope!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 - Move Hooks and Silenced out of Event and into Check.
 - Handle round-robin scheduling in wizardbus.
 - Added informational logging for failed entity keepalives.
+- Updated etcd to 3.3.2 from 3.3.1 to fix an issue with autocompaction settings.
 
 ### Fixed
 - Shut down sessions properly when agent connections are disrupted.
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -114,8 +114,8 @@
     "wal",
     "wal/walpb"
   ]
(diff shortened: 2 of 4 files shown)


amdprophet:
@echlebek it's likely due to differing versions of dep on each of our systems. I haven't found a great way to ensure we all have the same version of system-level go tools.

jamesdphillips:
If you're on a mac try running `brew upgrade dep` to update to the latest stable version. Also, double check that you haven't installed it through `go get`.

```
$ dep version
dep:
 version     : v0.4.1
 build date  : 2018-01-27
 git hash    : 37d9ea0
 go version  : go1.9.3
 go compiler : gc
 platform    : darwin/amd64
```

> I haven't found a great way to ensure we all have the same version of system-level go tools.

As far as I'm concerned this is this is something `dep` should be responsible for checking...

echlebek:
I'm not on a mac but I think Mercedes is. I probably won't be installing dep with brew ;)

mercul3s:
Just updated dep - I was running a really old development version! Let me see if that has any effect on the lockfile.

echlebek:
Turns out I was running a development version too. I followed the installation instructions and upgraded to 0.4.1.

PR #1197, opened 2018-03-20
Generate protobufs with go generate.

### LINKED ISSUE #1178: Use 'go generate' instead of scripts/genproto.sh
We shouldn't need to rely on bash scripts for generating protobufs, we can already do it with `go generate`.

Convert the genproto.sh script into a `types_gen.go`. The approach should be able to pick up new proto files as they are created - don't just call protoc for each file.

This will allow developers to work with protobuf generation on Windows or other places bash isn't available.

echlebek:
Closes #1178

## What is this change?

To generate protobufs, issue `go generate` from inside the types directory, or 
```
go generate github.com/sensu/sensu-go/types
```
From anywhere on the filesystem.

It eliminates a bash dependency and a git dependency - now Windows users can generate the protobufs without needing a bash shell, or git. (gogo/protobuf is vendored)


nikkiattea:
I know you updated the readme.. but maybe a changelog entry too.. just for those of us that may forget 😉

echlebek:
Added CL entry.

PR #1198, opened 2018-03-21
[Web UI] Replace fileb0x

### LINKED ISSUE #910: Investigate alternatives to fileb0x
Fileb0x generates some gnarly code, and we're not super thrilled with how it works wrt our build process. We'd like a saner approach to asset management. So we're going to investigate a couple of options:

- https://github.com/tv42/becky
  - example use: https://github.com/echlebek/erickson/tree/master/assets

and

- https://godoc.org/bitbucket.org/mjl/httpasset
  - an implementation of: https://godoc.org/golang.org/x/tools/godoc/vfs
  - which is used in godoc: https://github.com/golang/tools/blob/master/cmd/godoc/main.go

### LINKED ISSUE #1177: fileb0x broken during package builds
## Expected Behavior
Packaging the dashboard succeeds without errors.

## Current Behavior
When running the packaging dockerfile, the following fileb0x error occurs:

```
/go/src/github.com/sensu/sensu-go
panic: bufio.Scanner: token too long

goroutine 1 [running]:
main.main()
        /go/src/github.com/UnnoTed/fileb0x/main.go:127 +0x24ea
Makefile:159: recipe for target 'build_backend' failed
make[2]: *** [build_backend] Error 2
make[2]: Leaving directory '/go/src/github.com/sensu/sensu-go'
Makefile:121: recipe for target 'sensu' failed
make[1]: *** [sensu] Error 2
make[1]: Leaving directory '/go/src/github.com/sensu/sensu-go'
Makefile:103: recipe for target 'linux' failed
make: *** [linux] Error 2
```

## Possible Solution
This could be due to a newer version of fileb0x being pulled in. If we can find a way to pin system-level Golang deps we may be able to solve this by using an older version of fileb0x.

1. `docker build -t sensuapp/sensu-go-build -f Dockerfile.packaging .`
2. `docker run -it -v `pwd`:/go/src/github.com/sensu/sensu-go sensuapp/sensu-go-build`

## Context
Trying to test new packaging work, I've had to disable dashboard builds for now.

## Your Environment
N/A

jamesdphillips:
## What is this change?

Replace fileb0x with working solution.

Fileb0x has been unstable and the code it generates [has lint](https://github.com/UnnoTed/fileb0x/issues/17). This new approach uses `go generate` and `vfsgen`; the immediate benefit being that we can pin the version in our project's dependencies.

Closes #1177 
Closes #910


dashboard/assets_release.go
```diff
@@ -0,0 +1,10 @@
+//+build release
```
jamesdphillips: I'm not convinced this is the most elegant approach but here is the reasoning behind it.

1. Ideally we don't have to run an optimized UI build every time we make a change to the dashboard. At present the build takes around 45 seconds, which while not bad, will likely increase to 5-10 minutes as the codebase grows.
As such I don't think it is viable to have the generated assets committed to the repo. It would sort of be like packaging sensu for each platform each time we committed.
1. Ideally a user should still be able to compile the backend without having to install nodejs and yarn.
1. Ideally a user should be able to `go get` sensu backend.

Of course these ideals conflict to some degree. Eg. if we go with the good developer experience and don't commit the build artifacts to the repo then the build cannot succeed until `go generate` has run; conversely if we do commit it, we are damning ourselves to frequently re-building and commit the dashboard builds.

To avoid this, the default behavior of this package is to expose an "empty" implementation of `http.FileSystem`. In this case the dashboard will be missing, however, the code will still compile.

When the developer / user is ready to build the dashboard they can simply run:

```
go generate -tags release
go build -tags release
```

Or simply run the build script.

jamesdphillips:
@amdprophet I've looped back around and tweaked this clarity. A summary follows.

* Removed the idea of a `release` tag.
* Dashboard is now included when the `dev` tag is not present. From asking around this seems more conventional.
* If `go generate` was not run before the build (eg. user attempted to `go get` the backend) it will still
  compile. However, they will be presented with the following message when they attempt to use the
  dashboard.
 screen shot 2018-03-21 at 9 24 12 pm

* For speedy development if backend is built with the `dev` tag the dashboard isn't built and the
  developer is shown the following message when they hit the dashboard.
  screen shot 2018-03-21 at 9 24 49 pm
* I've remove source maps from release builds. They were ~15mbs, bloated the binary, extended the build time and weren't overly useful. 
  At some point we will want to upload the sourcemaps somewhere when we package releases.
  Having the source maps on hand will make it easier to debug issues if / when end-users encounter them.

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 - Move Hooks and Silenced out of Event and into Check.
 - Handle round-robin scheduling in wizardbus.
 - Added informational logging for failed entity keepalives.
+- Replaced fileb0x with vfsgen for bundling static assets into binary.
```
palourde: Do you mind adding an entry to indicate that the dashboard is now automatically built when building the backend?

amdprophet:
I love this, looks great!

CHANGELOG.md
jamesdphillips: Good call. I should probably update the README as well.

PR #1201, opened 2018-03-21
Support adding rules to role on creation via cli.

### LINKED ISSUE #1196: Sensuctl - create roles with permissions in one call
Currently, sensuctl role create does have flags to allow a user to pass a ruleset on creation:
To get started, you must first create a new read-only role.
```shell 
$ sensuctl role create read-only --organization default --environment default
```
Permissions must be added in a second step:
```shell
$ sensuctl role add-rule read-only --type * --read
```
To fix this, we need to add flags for `--read`, `--write`, `--create`, and `--delete` and allow the user to pass those on create and with add-rule:
```shell
$ sensuctl role create read-write --read --write
```
Alternatively, we could allow a comma separated list on create:
```shell
$ sensuctl role create read-write --abilities read,write

### PR DESCRIPTION
**Pull request by mercul3s**
Signed-off-by: Mercedes Coyle <mercedes@sensu.io>

## What is this change?

Adds type, create, read, update, delete flags to role creation in
sensuctl.

Closes #1196 

Added a note to the **Changed** section.

Nope.

This command is laid out slightly differently than our other commands, in add_rule.go takes the place of interactive.go in our other commands. I'm thinking that add_rule should really be a subcommand, and we create an interactive.go for create/update etc. Thoughts? This is also something we can refactor later if we choose to.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ and yarn are now dependencies for building the backend.
 - Updated etcd to 3.3.2 from 3.3.1 to fix an issue with autocompaction settings.
 - Updated and corrected logging style for variable fields.
 - Build protobufs with go generate.
+- Creating roles via sensuctl now supports passing flags for setting permissions
+  rules.
 
 ### Fixed
 - Shut down sessions properly when agent connections are disrupted.
--- a/cli/commands/role/create.go
+++ b/cli/commands/role/create.go
@@ -21,19 +21,40 @@ func CreateCommand(cli *cli.SensuCli) *cobra.Command {
 				return errors.New("invalid argument(s) received")
 			}
 
+			rule := types.Rule{}
+
 			role := &types.Role{Name: args[0]}
 			if err := role.Validate(); err != nil {
 				return err
 			}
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by mercul3s** [comment]
@palourde ah, I thought I had guarded against that - the intent is to throw an error if the type isn't configured. Let me revisit the code. Thinking about this more - is there any particular reason to create a role without any configured permissions? I tried to preserve that behavior, but I'm not sure that it makes sense to do so.

**Comment by palourde** [comment]
Also, I'm not sure what I'm doing wrong but I have the following error, which is weird:

```
$ sensuctl role create ops -t * -crud
create new roles

Usage:	sensuctl role create [NAME] [flags]

create new roles

Options:
[...]
Error: invalid argument(s) received
```

mercul3s:
@palourde OK asterisks need to be escaped on the cli - "*" or \* should work.

mercul3s:
@palourde updated the role creation so that a type and permission have to be passed as it's a little easier to code and I think makes more sense from a user perspective. If we need the ability to create roles without permissions, lmk and I can figure out a way to do that.

PR #1202, opened 2018-03-21
backend/eventd: don't double write events

grepory:

## What is this change?

Only write an event once when handling it in eventd.

We were erroneously double-writing all events.


backend/eventd/eventd.go
```diff
@@ -185,11 +185,6 @@ func (e *Eventd) handleMessage(msg interface{}) error {
 		return err
 	}
 
-	err = e.store.UpdateEvent(ctx, event)
```
echlebek: I think you actually need to remove the other one, since the if condition returns if true.

PR #1206, opened 2018-03-22
Fixup panic when resolving events

jamesdphillips:

## What is this change?

Message bus was not being passed down to the events router, causing a panic when the controller attempted to publish the event.

Unable to resolve events.

I believe the bug was introduced prior to last release.

### DIFF
--- a/backend/apid/graphql/mutations.go
+++ b/backend/apid/graphql/mutations.go
@@ -136,7 +136,7 @@ func (r *mutationsImpl) ResolveEvent(p schema.MutationResolveEventFieldResolverP
 
 	if event.Check != nil && event.Check.Status > 0 {
 		event.Check.Status = 0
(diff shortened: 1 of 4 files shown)


backend/apid/graphql/mutations.go
```diff
@@ -139,7 +139,7 @@ func (r *mutationsImpl) ResolveEvent(p schema.MutationResolveEventFieldResolverP
 		event.Check.Output = "Manually resolved manually with " + p.Args.Input.Source
```
nikkiattea: Sorry, I know this isn't related, but "manually resolved manually" is bothering me..

jamesdphillips: 🤦‍♂️ thanks. I'll fix that.

PR #1243, opened 2018-03-27
Remove -c check flag

mercul3s:
Signed-off-by: Mercedes Coyle <mercedes@sensu.io>

## What is this change?

Removes the -c (check) flag from the execute command.

Check name is already taken as the first arg to the command, so this
additional flag is redundant and a little confusing as it is marked
required. Removing it in favor of passing check name as the first arg
fits with the conventions of our other commands.

Added note under changed.

Nope!

Nope!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ and yarn are now dependencies for building the backend.
 - Build protobufs with go generate.
 - Creating roles via sensuctl now supports passing flags for setting permissions
   rules.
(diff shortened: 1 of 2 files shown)


mercul3s:
@nikkiattea good catch - I removed the prerun flag option, but kept the logic the same as that's how the interactive prompt grabs the check name from the survey instead of the arg. So a user can do:
```sh
$ sensuctl check execute [NAME]
```
or:
```
$ sensuctl check execute --interactive
```

Issue #167, opened 2017-05-16
Delete a user


palourde:
**As an operator, I can delete users**

- [x] A deleted user can no longer authenticate
- [x] The user entry is in fact not deleted from etcd, but simply disabled to prevent authentication, for consistency.
- [x] Deleted (and therefore disabled) users are not listed
- [x] Both the API & CLI supports deletion a user

PR #1244, opened 2018-03-28
Add statsd server to sensu-agent and log metrics

### LINKED ISSUE #1123: [Agent statsd] Embed the Atlassian gostatsd statsd server within the Sensu Agent
###  Background

See the [proposal](https://github.com/sensu/proposal/blob/master/design/85-agent-statsd.md).

We need to embed the [Atlassian gostatsd statsd server](https://github.com/atlassian/gostatsd#using-the-library) within the Sensu agent.

The Atlassian gostatsd server has the concept of "Backends", a plugin system for sending the collected statsd metrics to a backend service (e.g. a time-series database like Graphite). For now, we should only need to configure the [the built-in "stdout" backend](https://github.com/atlassian/gostatsd/blob/master/pkg/backends/stdout/stdout.go). A backend for Sensu will be created in a [subsequent issue](https://github.com/sensu/sensu-go/issues/1124).

### Acceptance Criteria

- [x] The gostatsd statsd server automatically starts with the agent
- [x] The gostatsd statsd server automatically stops with the agent
- [x] You can write statsd metrics to the listener
- [x] The metrics should be sent to stdout

### Definition of Done

- [ ] Documentation
- [x] Unit Tests
- [x] Integration Test
- [ ] E2E Test

### LINKED ISSUE #1124: [Agent statsd] Implement a Sensu Metric gostatsd backend
###  Background

See the [proposal](https://github.com/sensu/proposal/blob/master/design/85-agent-statsd.md).

The Atlassian gostatsd server has the concept of "Backends", a plugin system for sending the collected statsd metrics to a backend service (e.g. a time-series database like Graphite).

We need to implement a  gostatsd backend that transforms the gostatsd produced metrics into the [Sensu Metric Format](7-sensu-metric-format.md) ([metrics type](https://github.com/sensu/sensu-go/blob/master/types/metrics.proto)).

### Acceptance Criteria

- [x] It can take one or more [gostatsd metric](https://github.com/atlassian/gostatsd/blob/0ff1907c7cc65351a5417c12c555e5b7c692f6bd/metrics.go#L40-L50) and transform it into the Sensu Metric Format
- [x] The backend must satisfies this [interface](https://github.com/atlassian/gostatsd/blob/master/backend.go)
- [x] This backend replaces the `stdout` backend implemented in https://github.com/sensu/sensu-go/issues/1123.
- [x] The transformed metric is simply logged for now.  It will be properly handled in https://github.com/sensu/sensu-go/issues/1126.
- [x] Extra gostatsd fields are converted into Sensu Metric Format tags (e.g. `SourceIP`)

### Definition of Done

- [ ] Documentation
- [x] Unit Tests
- [x] Integration Test
- [ ] E2E Test

### LINKED ISSUE #1126: [Agent statsd] Update the Sensu Metric gostatsd backend to transmit Events
###  Background

See the [proposal](https://github.com/sensu/proposal/blob/master/design/85-agent-statsd.md).

The Atlassian gostatsd server has the concept of "Backends", a plugin system for sending the collected statsd metrics to a backend service (e.g. a time-series database like Graphite).

Update the Sensu Metric gostatsd backend (created in https://github.com/sensu/sensu-go/issues/1124) to transmit Events over the Agent/Backend connection, containing ONLY the produced metrics (in the Sensu Metric Format, e.g. `Event.Metrics`).

### Acceptance Criteria

- [x] Metric events are sent to the backend. (The backend will probably discard that event since it doesn't contain a check, which is fine for now.)

### Definition of Done

- [ ] Documentation
- [x] Unit Tests
- [x] Integration Test
- [ ] E2E Test

nikkiattea:

## What is this change?

Adds a statsd server to the sensu-agent which runs statsd at a configurable flush interval and converts gostatsd metrics to Sensu Metric Format. These metrics are logged for the time being.

Closes #1123 
Closes #1124 
Closes #1126 

Yas.

Nah.

Nah.


agent/cmd/start.go
```diff
@@ -40,8 +40,11 @@ const (
 	flagDeregistrationHandler = "deregistration-handler"
 	flagEnvironment           = "environment"
 	flagExtendedAttributes    = "custom-attributes"
+	flagFlushInterval         = "flush-interval"
```
palourde: Should we also prefix this flag with `metrics`? Or maybe use `statsd` for all related flags? So it's easier to see that it's related to statsd using the flag name only!

nikkiattea:
@palourde They won't be logged for long. The next pr will transmit them to the event pipeline. I can do that now if you'd prefer, it's actually a small change.

agent/cmd/start.go
nikkiattea: Yeah, I'm honestly thinking we should prefix all things statsd related with `statsd`! I noticed the naming convention was somewhat inconsistent.

palourde:
@nikkiattea I'm actually concerned with the simple fact that these metrics are collected; I feel like we don't need those, or at least not by default, but I'm not really familiar with statsd. Maybe Sean has an idea?

nikkiattea:
@palourde I've just used the gostatsd default settings for the server. Perhaps there are fields I haven't discovered yet that would mute these metrics. However, I was under the impression that these were the metrics we _wanted_ from statsd.

palourde:
As discussed, we should find a way to disable these internal metrics!

nikkiattea:
Filed: https://github.com/atlassian/gostatsd/issues/167 to address the inability to disable the internal flush metrics.
Filed: https://github.com/sensu/sensu-go/issues/1259 to disable the internal flush metrics once [#167](https://github.com/atlassian/gostatsd/issues/167) is submitted.

Gopkg.toml
```diff
@@ -163,3 +163,7 @@ required = ["github.com/shirou/w32"]
 [[constraint]]
   branch = "master"
   name = "github.com/NYTimes/gziphandler"
+
+[[constraint]]
+  name = "github.com/atlassian/gostatsd"
+  version = "2.3.0"
```
palourde: I just realized it looks like we are using an older version of gostatsd (latest is `5.4.0`). Any reason?

nikkiattea: I blame govendor! Haha yeah let me update that version 😉

nikkiattea: @palourde So this is why dep installed 2.3.0 and not most recent 🤔
```
Solving failure: No versions of github.com/atlassian/gostatsd met constraints:
	5.4.0: Could not introduce github.com/atlassian/gostatsd@5.4.0 due to a case-only variation: it depends on "github.com/sirupsen/logrus", but "github.com/Sirupsen/logrus" was already established as the case variant for that project root by the following other dependers:
	(root)
	github.com/docker/docker@v1.13.1
	github.com/tmc/grpc-websocket-proxy@timestamp_stdtime_example
```

PR #1248, opened 2018-03-28
Separate Travis CI deploy job into multiple jobs

amdprophet:

## What is this change?

Splits the single Travis CI deploy job into multiple jobs for our various deployments.

If a deploy job script fails, any subsequent scripts will not run. This can lead to scenarios where we won't deploy docker images after package building has failed.

Nope!

Nope!

Nope!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,24 +29,20 @@ jobs:
   include:
   - stage: deploy
     env: GOARCH=amd64 TEST_SUITE=none
-    script: skip
-    deploy:
-    - provider: script
-      skip_cleanup: true
-      script: "./build.sh docker push master"
-      on:
-        branch: master
-    - provider: script
-      skip_cleanup: true
-      script: "./build.sh deploy versioned"
-      on:
-        tags: true
-        branch: master
-    - provider: script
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -27,26 +27,22 @@ install:
...
-        tags: true
-        branch: master
-    - provider: script
-      skip_cleanup: true
-      script: "./build.sh deploy nightly"
-      on:
-        condition: $TRAVIS_EVENT_TYPE = api
+    script: "./build.sh deploy_binaries"
+    if: tag IS present
+  - stage: deploy packages
```
palourde: I believe the stage name should be the same for all deployment if you want multiple jobs running in parallel, otherwise it creates a stage for each type of deployment and run each stage sequentially, which significantly increase the total amount time spent for a build:

screen shot 2018-03-28 at 4 11 21 pm

amdprophet: Fixed in the latest commit. Thanks!

amdprophet:
This is going to add too much time to builds. We will have to try to simulate a clean room environment instead.

PR #1254, opened 2018-04-03
[Web UI] Mobile Push

### LINKED ISSUE #815: [Sketch] Entities page
## 📝 Intent 

- Table w/ most important details surfaced.
  - Name, last keep-alive, status, ???
- Paginated? Infinite loading? Paginate by letter?
- Filters? Suggest filters?

## ✨ Inspiration

**Uchiwa** http://demo.sensuapp.com:3000/#/clients
screen shot 2018-01-08 at 12 15 50 am

jamesdphillips:
## What is this change?

Introduce a number of tweaks to improve experience on mobile.

Closes #815.

## Screenshots

springboardlogindashboard404 page
bulk actions


dashboard/src/components/AppFrame.js
```diff
@@ -4,55 +4,77 @@ import { withStyles } from "material-ui/styles";
 import { createFragmentContainer, graphql } from "react-relay";
 
 import AppRoot from "./AppRoot";
-import AppBar from "./Toolbar";
+import AppBar from "./AppBar";
 import Drawer from "./Drawer";
 import QuickNav from "./QuickNav";
 
-const styles = theme => ({
```
dabria: Unsure why styles were moved into the class declaration, but make sure you keep in consistent in all files and tell me why the change was made (so I understand) if we keep it this way. Obviously we sometimes need certain styles in the class if the styles need to be generated a certain way depending on props, but it didn't seem like it in this case.

dashboard/src/components/Drawer.js
```diff
@@ -83,79 +82,72 @@ class Drawer extends React.Component {
...
+                <Wordmark
+                  alt="sensu"
+                  className={classes.logo}
+                  color="secondary"
+                />
+              </div>
+              <div className={classes.row}>
+                {/* TODO update with global variables or whatever when we get them */}
+                <div className={classes.namespaceIcon}>
+                  <EnvironmentIcon environment={environment} size={36} />
```
dabria: I don't think the comment on like 117 is needed any more because it looks like we have a variable for environment

dashboard/src/components/AppFrame.js
jamesdphillips: In a separate task I'll back around and make the usage of withStyle, compose, createContainer and others consistent.

dabria: Maybe make a ticket now so we don't forget

PR #1255, opened 2018-04-03
Log dropped messages on each topic.

echlebek:
If topic subscribers are not ready to receive messages, then the
messages are dropped. As we are encountering strange failure modes
under load testing, this change should introduce some visibility
around this component.

Each topic will now emit a logger warning if there are any dropped
messages within the last second, with a log field of
'dropped_messages_per_second'.

The counts are inexact.

This commit also introduces a more idiomatic single-letter receiver
name for wizardTopic.


backend/messaging/wizard_topic.go
```diff
@@ -19,75 +21,91 @@ type wizardTopic struct {
...
+}
+
+func (t *wizardTopic) logDroppedMessages() {
+	dropped := atomic.LoadInt64(&t.droppedMessages)
+	if dropped == 0 {
+		return
+	}
+	// Inexact but avoids contention; good enough for our purposes.
+	atomic.StoreInt64(&t.droppedMessages, 0)
+	logger.WithFields(logrus.Fields{
```
grepory: logger is undefined

backend/messaging/wizard_topic.go
```diff
@@ -19,75 +21,98 @@ type wizardTopic struct {
...
+	names := make([]string, 0, len(t.bindings))
+	subscribers := make([]Subscriber, 0, len(t.bindings))
+	for name, subscriber := range t.bindings {
+		names = append(names, name)
+		subscribers = append(subscribers, subscriber)
+	}
+	t.RUnlock()
+	for i, subscriber := range subscribers {
 		select {
 		case subscriber.Receiver() <- msg:
```
palourde: So it look like this bit now causes a race condition, see https://travis-ci.org/sensu/sensu-go/jobs/361886987.

PR #1265, opened 2018-04-05
Fix a deadlock in the monitor.

### LINKED ISSUE #1264: Deadlock in the monitor
After extensive load testing, I have discovered a deadlock in the monitor code.

In this method, `reset`, https://github.com/sensu/sensu-go/blob/master/backend/monitor/monitor.go#L154, a channel send is performed while holding a mutex. However, the consumer of that channel can only consume if the other cases in its select are not trying to acquire the same mutex.

The issue can be trivially fixed by adding a buffer to the channel.

echlebek:
The deadlock only occurs under load; in testing, it took around
5500 agent sessions to expose the problem.

On reset, the monitor acquires a mutex and then performs a channel
send. However, if another select case in the monitor is trying to
acquire the mutex, then the channel send will never occur.

Adding a buffer to the resetChan fixes the problem.

Closes #1264

### DIFF
--- a/backend/monitor/monitor.go
+++ b/backend/monitor/monitor.go
@@ -95,7 +95,9 @@ func (m *Monitor) start() {
 
 	timerDuration := m.Timeout
 	m.timer = time.NewTimer(timerDuration)
-	m.resetChan = make(chan time.Duration)
+	// resetChan needs a buffer, otherwise we could end up holding onto m's
+	// mutex indefinitely.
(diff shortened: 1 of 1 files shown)


grepory:
Do we even need the mutex?

echlebek:
Yes, we need the mutex to guard against a send to a closed channel.

echlebek:
This is strictly a correctness bug in the monitor, and does not pertain to how the message bus or keepalives work.

PR #1266, opened 2018-04-05
Terminate processes gracefully in e2e tests

### LINKED ISSUE #1228: TestAgentReconnection intermittently fails
The TestAgentReconnection test is intermittently failing in CI.

## Expected Behavior
Test should pass.

## Current Behavior
Seeing the following:

```
--- FAIL: TestAgentReconnection (22.14s)
	Error Trace:	agent_reconnection_test.go:57
	Error:      	Received unexpected error:
	            	exit status 1
	Messages:   	Error: Get http://127.0.0.1:42638/events/TestAgentReconnection/keepalive?env=default&org=default: dial tcp 127.0.0.1:42638: connect: connection refused
	Error Trace:	agent_reconnection_test.go:60
	Error:      	Received unexpected error:
	            	invalid character 'E' looking for beginning of value
```

https://travis-ci.org/sensu/sensu-go/jobs/358521947#L1868

## Possible Solution
N/A

Failure is intermittent so it may be tough to reproduce.

## Context
N/A

## Your Environment
Travis CI

cyphus:
## What is this change?

Terminate processes gracefully in e2e tests, allowing ports to be reused.

Fixes #1228 

Yes:

### Fixes
- Terminate processes gracefully in e2e tests, allowing ports to be reused.

I need clarification on the hardest part: naming things. If there is a name or place for `terminate_helper` that would fit project conventions better, I'd love to hear about it.

I spent some time trying to find a way to gracefully terminate CLI processes on Windows -- sadly there is no supported way to do this in Windows.

### DIFF
--- a/testing/e2e/process_test.go
+++ b/testing/e2e/process_test.go
@@ -163,7 +163,7 @@ func (b *backendProcess) Start() error {
 }
 
 func (b *backendProcess) Terminate() error {
-	return terminate_helper(b.cmd.Process)
+	return terminateProcess(b.cmd.Process)
 }
 
 type agentProcess struct {
@@ -299,7 +299,7 @@ func (a *agentProcess) Start(t *testing.T) error {
 }
 
 func (a *agentProcess) Terminate() error {
-	return terminate_helper(a.cmd.Process)
+	return terminateProcess(a.cmd.Process)
 }
 
 type sensuCtl struct {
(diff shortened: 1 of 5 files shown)


testing/e2e/process_test.go
```diff
@@ -355,3 +351,21 @@ func (s *sensuCtl) run(args ...string) ([]byte, error) {
 func (s *sensuCtl) SetStdin(r io.Reader) {
 	s.stdin = r
 }
+
+// Terminate a Process by sending SIGTERM and waiting for it to exit cleanly.
+// On Windows, there is no way to ask a program to exit cleanly. Instead, it
+// must be killed with SIGKILL.
+func terminate_helper(p *os.Process) error {
```
mercul3s: I'd suggest using came case here per Go convention: `terminateHelper`; could also be named `terminateProcess`. Otherwise, I think this is the right place for this function to go 👍

echlebek: +1 to `terminateProcess`

cyphus: Ah of course! I was so focused on the `helper` part of the name that I completely missed the fact that I was using snake case 🤦‍♂️ `terminateProcess` sounds good, I'll update.

PR #1268, opened 2018-04-05
Make wizardbus synchronous

### LINKED ISSUE #1253: Track dropped messages and monitor under load
As an intentional design feature, WizardBus drops fanout messages to subscribers that are not prepared to receive them. I suspect this is causing issues when the number of subscribers becomes large: it becomes increasingly unlikely that another goroutine will be waiting to read the channel send as the number of goroutines in the process increases.

Add a package-global atomic counter and a periodic logger to track the number of messages lost. Once that is in place, repeat Greg's load testing to 6000 agents and observe how many messages we are losing.

### LINKED ISSUE #1267: Synchronous message bus
WizardBus currently drops messages when its subscribers are not ready to receive. However, empirical testing has shown this to be unnecessary and sub-optimal.

Originally, it was thought that the bus should not block when a consumer is not ready to handle a message. However, as load increases, this becomes more and more likely, leading to dropped messages.

Sensu uses the bus for keepalives, and cannot lose many of them before agents begin to drop out. Testing has shown that the backend can actually handle more load with a synchronous backend, so we should change how wizard bus works in this regard.

echlebek:
## What is this change?

This changeset makes WizardBus synchronous, and also makes a few other small improvements.

* topics with zero subscribers are now closed
* spurious pipelined logging is now silenced
* Keepalived now uses a time.Ticker instead of time.Timer

Closes #1253 
Closes #1267


backend/messaging/wizard_bus_test.go
```diff
@@ -87,8 +86,8 @@ func TestWizardBus(t *testing.T) {
 		received++
 		messages = append(messages, m.(string))
 	}
-	assert.Equal(t, 7, len(messages))
-	assert.Equal(t, 7, received)
+	assert.Equal(t, 9, len(messages))
```
nikkiattea: Out of curiosity, could you explain the change in the expected message count here? Were we actually dropping messages in the test too?

echlebek: Yes we were... one of the channels didn't have enough buffer and the messages were being dropped as a result. This was being explicitly tested for.

Issue #1270, opened 2018-04-07
Remove os.Exit(0) from pipelined unit tests


echlebek:
Some of the pipelined unit tests have `os.Exit(0)` in them. This is masking failures in some of the other tests in the package, as it is causing the test runner to terminate early.

Remove the `os.Exit(0)` lines, and sort out why the other tests are failing.

echlebek:
Update: I read through this again and realized that the use of os.Exit(0) is merely clever. However, it is quite confusing to encounter and should probably be removed in favour of another strategy.

Since it's not a critical issue, I'm demoting it to be merely "technical-debt".

echlebek:
@amdprophet could you explain the rationale behind designing these tests this way? If we don't replace them we should at least document them better so I don't get confused again. ;)

echlebek:
Oh! This all makes a lot of sense to me now that you put it in the context of cross-platform behaviour. I see now that I was wrong about all of this.

I'll add some explanatory comments in my current working branch. Thanks!

PR #1272, opened 2018-04-07
Add extensions to pipelined

### LINKED ISSUE #1115: Integrate extensions into pipelined
After #1113 and #1114 are finished, integrate support for extensions into pipelined.

Pipelined should consult the extension registry for handlers and mutators in addition to the internal handlers and mutators that are mapped.

Determine if we need to specify a timeout for RPC so that the pipeline doesn't stall.

Come up with a strategy for how we can communicate failure on RPC errors.

Develop a comprehensive end-to-end test, using the simple handler and mutator extension created in #1114.

echlebek:
## What is this change?

Adds grpc extensions support to pipelined.

Closes #1115 

Filed #1270

### DIFF
--- a/backend/pipelined/handle.go
+++ b/backend/pipelined/handle.go
@@ -8,9 +8,9 @@ import (
 	"net"
 	"time"
 
-	"github.com/sirupsen/logrus"
 	"github.com/sensu/sensu-go/command"
 	"github.com/sensu/sensu-go/types"
+	"github.com/sirupsen/logrus"
 )
 
 const (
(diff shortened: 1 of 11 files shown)


echlebek:
Note that this branch may fail to test cleanly due to #1270

backend/pipelined/handle.go
```diff
@@ -19,6 +19,11 @@ const (
 	DefaultSocketTimeout uint32 = 60
 )
 
+type handlerExtensionUnion struct {
```
cyphus: In the C/C++ universe, the term 'union' means 'two separate types that share the same memory space', I assume it means something different for go? I'm still in learning mode here.

echlebek: Go doesn't have unions, but you can simulate them through anonymous embedding.

PR #1274, opened 2018-04-09
Use lowercase sensitive "github.com/sirupsen/logrus"

### LINKED ISSUE #1261: Use lowercase sensitive "github.com/sirupsen/logrus"
Our vendor dependency tree references "github.com/Sirupsen/logrus" which was changed to "github.com/sirupsen/logrus" in https://github.com/sirupsen/logrus/pull/384. This causes an issue with updating other vendor packages which are dependent on logrus. We should find a way to clean this up in order to update certain packages to more recent versions (ex. github.com/atlassian/gostatsd).

```
Solving failure: No versions of github.com/atlassian/gostatsd met constraints:
	5.4.0: Could not introduce github.com/atlassian/gostatsd@5.4.0 due to a case-only variation: it depends on "github.com/sirupsen/logrus", but "github.com/Sirupsen/logrus" was already established as the case variant for that project root by the following other dependers:
	(root)
	github.com/docker/docker@v1.13.1
	github.com/tmc/grpc-websocket-proxy@timestamp_stdtime_example
```

nikkiattea:
## What is this change?

Updates dependency "Sirupsen/logrus" to "sirupsen/logrus." Updates other applicable dependencies to versions which use the latter. 

Closes #1261.

Yas.

Nah.

~~Updating gostatsd to 5.4.1 removes a deprecated function `NewServer`. I had to rewrite that function for our purposes.~~

### DIFF
--- a/agent/statsd_server.go
+++ b/agent/statsd_server.go
@@ -15,13 +15,12 @@ import (
 	"github.com/atlassian/gostatsd/pkg/statsd"
 	"github.com/sensu/sensu-go/types"
 	"github.com/spf13/viper"
-	"golang.org/x/time/rate"
 )
 
 // NewStatsdServer provides a new statsd server for the sensu-agent.
 func NewStatsdServer(a *Agent) *statsd.Server {
 	c := a.config.StatsdServer
-	s := NewServer()
+	s := statsd.NewServer()
 	backend, err := NewClientFromViper(s.Viper, a)
 	if err != nil {
 		logger.WithError(err).Error("failed to create sensu-statsd backend")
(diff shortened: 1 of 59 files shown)


cyphus:
Changes look good from here, going to make sure I can build locally before giving the official thumbs up.

agent/statsd_server.go
```diff
@@ -31,6 +32,38 @@ func NewStatsdServer(a *Agent) *statsd.Server {
 	return s
 }
 
+// NewServer will create a new statsd Server with the default configuration.
+func NewServer() *statsd.Server {
```
echlebek: This change seems unrelated to the PR topic, a mistake?

nikkiattea: It seems so. NewServer is deprecated in gostatsd version 5.4.1 and we had to update gostatsd for sirupsen/logrus compatibility. I just used the most recent release as that’s what we originally attempted, which prompted this logrus issue.

nikkiattea: Yes! For some reason I assumed the function was deprecated _before_ the logrus fix. But it wasn't, so we can totally roll this back to 2.4.0.

Issue #1247, opened 2018-03-28
Event validation surrounding the presence of metrics and/or checks


nikkiattea:
Solve the issue presented in #1126 
>Metric events are sent to the backend. (The backend will probably discard that event since it doesn't contain a check, which is fine for now.)

## Expected Behavior
Event containing metrics (but no check) is properly handled.

## Current Behavior
Error handling event:
https://github.com/sensu/sensu-go/blob/a27ab4b9a5b3f6c4edcb625ec8f69fef50efa626/backend/agentd/session.go#L137-L139
Caused by event validation:
https://github.com/sensu/sensu-go/blob/a27ab4b9a5b3f6c4edcb625ec8f69fef50efa626/types/event.go#L74-L76
Log:
```
{"component":"agentd","error":"malformed event","level":"error","message":{"Type":"event","Payload":"eyJ0aW1lc..."},"msg":"error handling message","time":"2018-03-28T11:21:36-05:00"}
```

## Possible Solution
1. Change event validation so that it requires either a check or metric `(entity && (check || metric))`. We'll also have to check for mutual exclusivity whenever invoking `event.check` and `event.metric`.
2. Spoof a check to be added to a metrics event - perhaps we could use it to provide additional context to the metrics.

echlebek:
I think the right thing to do here is to make `Event` an interface. We'll split the current Event type into multiple types, `CheckEvent` and `MetricsEvent`. We already have an outstanding issue to make breaking changes for the Event type (by removing Hooks and Silenced), so we may as well go all-in.

(n.b. Hooks and Silenced were moved into Check a while ago and are deprecated for Event now.)

```
type Event interface {
    // Time returns the time the event was emitted at.
    Time() time.Time

    // GetName returns the name of the event
    GetName() string

    // GetEntity() gets the entity associated with the event.
    GetEntity() *types.Entity

    // GetHandlers gets the handlers the event is associated with
    GetHandlers() []string

    Validator
    Resource
}
```

In places where we need to handle check-specific or metrics-specific information, we can use type assertions.

nikkiattea:
@echlebek I am on board for this for a few reasons. Firstly, I already had an unsuccessful attempt at decoupling checks and events without breaking changes. At the time, I considered a proposal to store metrics in a different scope than an event, but interfacing the event did not even cross my mind. Secondly, I think this method will be a much cleaner approach than having to manage `check || metrics` in any given  event.

[...]

PR #1283, opened 2018-04-10
[Agent statsd] Additional unit and integration tests

nikkiattea:

## What is this change?

Adds a unit test and integration test surrounding the sensu-agent's statsd server and udp port. Refactors some reusable code into "fixture" functions.

Testing is 🔑 

Yas.

Nah.

Part of the test is blocked by #1247 but a `TODO` has been created.

### DIFF
--- a/agent/statsd_server_test.go
+++ b/agent/statsd_server_test.go
@@ -1,12 +1,16 @@
 package agent
 
 import (
+	"context"
+	"encoding/json"
+	"net"
 	"testing"
 	"time"
 
 	"github.com/atlassian/gostatsd"
 	"github.com/sensu/sensu-go/types"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestNewStatsdServer(t *testing.T) {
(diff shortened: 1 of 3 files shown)


agent/statsd_server_test.go
```diff
@@ -160,3 +133,150 @@ func TestComposeSetPoints(t *testing.T) {
...
+
+	cfg := FixtureConfig()
+	ta := NewAgent(cfg)
+
+	ctx, cancel := context.WithCancel(context.TODO())
+	defer cancel()
+
+	go ta.statsdServer.Run(ctx)
+
+	udpClient, err := net.Dial("udp", ta.statsdServer.MetricsAddr)
```
echlebek: I wonder what happens if the server doesn't start before we try to open a UDP conn here? Seems like it could lead to a flaky test.

I'm a little uncertain about the intent of the test as well. It seems that all it does is fire UDP packets at statsd, which presumably works. Maybe we don't need it.

nikkiattea: What kind of solution might you suggest? I think we should establish some kind of reasonable time window for the udp server to start. If it does not start within that window, thus failing the test, it would indicate we should file an issue with the dependency.

So right now, the intent of the test is to ensure we can write a metric map to the statsd server with no error. Since the feature is not complete (blocked by #1247) we cannot yet verify the metrics received are written to an event (the `TODO` portion of the test). I still think its important to validate that we're able to establish a udp connection with the statsd server and write metrics to it, otherwise there is something wrong with the semi-complete portion of the feature.

nikkiattea: We do something similar here to test the udp and tcp sockets: https://github.com/sensu/sensu-go/blob/master/agent/sockets_test.go#L29-L32. But yeah, I agree it's definitely not a unit test, and if kept as is, should definitely be tagged as integration.

I'd like to keep the existing test, and explore implementing a mock as well. However, I'm curious with how we would want to go about the mock. It sounds like we wouldn't be mocking the actual server, correct (`server.Run()`)? Are you seeking a test that would just validate different kinds of messages that could be sent to the server?

Issue #1208, opened 2018-03-22
[Build stability] intermittent failure in TestSilencing e2e


nikkiki:
Investigate intermittent failure in e2e TestSilencing:

```
--- FAIL: TestSilencing (18.19s)
	assertions.go:239: 
                          
	Error Trace:	silencing_test.go:135
		
	Error:      	Not equal: 
		
	            	expected: 5
		
	            	actual: 6
```

palourde:
Just I had a similar [failure](https://travis-ci.org/sensu/sensu-go/jobs/358623295):

```
--- FAIL: TestSilencing (16.21s)
	Error Trace:	silencing_test.go:135
	Error:      	Not equal: 
	            	expected: 3
	            	actual: 4
```

palourde:
Happened again [here](https://travis-ci.org/sensu/sensu-go/jobs/364950366)

```
{"component":"schedulerd","env":"default","level":"info","msg":"stopping scheduler","name":"check_silencing","org":"default","time":"2018-04-11T05:47:06Z"}
{"component":"backend","level":"info","msg":"shutting down eventd","time":"2018-04-11T05:47:06Z"}
{"component":"eventd","level":"info","msg":"shutting down eventd","time":"2018-04-11T05:47:06Z"}
{"component":"backend","level":"info","msg":"shutting down pipelined","time":"2018-04-11T05:47:06Z"}
{"component":"backend","level":"info","msg":"shutting down message bus","time":"2018-04-11T05:47:06Z"}
{"component":"etcd","level":"info","msg":"skipped leadership transfer for single member cluster","pkg":"etcdserver","time":"2018-04-11T05:47:06Z"}
WARNING: 2018/04/11 05:47:06 grpc: addrConn.resetTransport failed to create client transport: connection error: desc = "transport: Error while dialing dial tcp 127.0.0.1:33463: connect: connection refused"; Reconnecting to {127.0.0.1:33463 0  <nil>}
WARNING: 2018/04/11 05:47:06 grpc: addrConn.resetTransport failed to create client transport: connection error: desc = "transport: Error while dialing dial tcp 127.0.0.1:33463: connect: connection refused"; Reconnecting to {127.0.0.1:33463 0  <nil>}
[...]

PR #1284, opened 2018-04-11
Add an end-to-end test for extensions

### LINKED ISSUE #1271: Create e2e test for extensions
We need an end-to-end test for extensions.

Build a gRPC service that implements the rpc/extension interface, and write an end-to-end test that makes use of its filter, handler, and mutator.

echlebek:
## What is this change?

This PR adds an end-to-end test for extensions.

Closes #1271 

Yes

Yes.

The good: extensions are working as intended.
The bad: the end-to-end test is a little ugly, but should be relatively robust.
The ugly: We need to implement gRPC connection caching at a later date. This is likely going to be quite difficult.


palourde:
Travis e2e test failed because of the TestSilencing, added a note there: https://github.com/sensu/sensu-go/issues/1208

backend/pipelined/handle.go
```diff
@@ -110,12 +111,15 @@ func (p *Pipelined) expandHandlers(ctx context.Context, handlers []string, level
 
 		if handler == nil {
 			if err != nil {
-				logger.WithError(err).Error("pipelined failed to retrieve a handler")
+				logger.WithError(err).Error("pipelined failed to retrieve a handler:", handlerName)
```
nikkiattea: Might I suggest `WithField("handler", handlerName)`?

backend/pipelined/handle.go
```diff
@@ -110,12 +111,15 @@ func (p *Pipelined) expandHandlers(ctx context.Context, handlers []string, level
...
 				continue
 			}
 			extension, err = p.store.GetExtension(ctx, handlerName)
+			if err == store.ErrNoExtension {
+				continue
+			}
 			if err != nil {
-				logger.WithError(err).Error("pipelined failed to retrieve a handler")
+				logger.WithError(err).Error("pipelined failed to retrieve a handler:", handlerName)
```
nikkiattea: Same here, `WithField` func.

testing/e2e/extensions_test.go
```diff
@@ -0,0 +1,277 @@
...
+					assert.Equal(t, test.Env["SENSU_E2E_FILTER_ERROR"], out)
+				}
+			case "mutate":
+				if err == nil {
+					assert.Equal(t, test.Env["SENSU_E2E_MUTATE_RESPONSE"], out)
+				} else {
+					assert.Equal(t, test.Env["SENSU_E2E_MUTATE_ERROR"], out)
+				}
+			default:
+				t.Fatalf("invalid test type: %q", test.Type)
```
nikkiattea: If our tests are written properly, this code block is unreachable. Instead of establishing `Type` as an arbitrary string value in the test struct, we could store the expected constant(s) (ex. `SENSU_E2E_HANDLE_RESPONSE`) there to minimize the # of lines of code.

backend/pipelined/handle.go
echlebek: Capital suggestion.

testing/e2e/extensions_test.go
echlebek: Can you show me what this might look like? How will we know what to compare against `out`?

echlebek:
Looks like my brilliant strategy of `go install`ing the extension and then trying to run it doesn't work on Windows. I might have to adopt the crazy test runner approach.

testing/e2e/extensions_test.go
nikkiattea: I believe this would eliminate the need for the switch cases and unreachable code. Note: I did not attempt to run this example haha.

```
 	tests := []struct {
  		Name		string
  		Response	string
  		Error		string
  		Env		map[string]string
  		Event		*types.Event
  		ExpErr		bool
  	}{
  		{
  			Name:  "no error",
  			Response: SENSU_E2E_HANDLE_RESPONSE,
  			Error: SENSU_E2E_HANDLE_ERROR,
  			Event: handleEvt,
  			Env: map[string]string{
  				"SENSU_E2E_HANDLE_RESPONSE": "{}",
  			},
  		}, // ... other test cases here
  	}
  
  	for _, test := range tests {
  		t.Run(fmt.Sprintf("%s %s", test.Type, test.Name), func(t *testing.T) {
  			cleanup := newExtension(t, ports[0], test.Env)
  
  			body, _ := json.Marshal(test.Event)
  			resp, err := client.R().SetBody(body).Put(
  				fmt.Sprintf("/events/%s/%s", test.Event.Entity.ID, test.Event.Check.Name))
  			if err != nil {
  				t.Fatal(err)
  			}
  			if status := resp.StatusCode(); status >= 400 {
  				t.Fatalf("bad status: %d", status)
  			}
  
  			out, err := cleanup()
  			if err != nil && !test.ExpErr {
  				fmt.Println(out)
  				t.Fatal(err)
  			}

  			if err == nil {
  					assert.Equal(t, test.Env[test.Response], out)
  				} else {
  					assert.Equal(t, test.Env[test.Error], out)
  				}
  			}
  	}
```

echlebek:
I replaced the problematic extension program with a same-process server. Works much better :P

backend/pipelined/handle.go
```diff
@@ -110,12 +111,21 @@ func (p *Pipelined) expandHandlers(ctx context.Context, handlers []string, level
 
 		if handler == nil {
 			if err != nil {
-				logger.WithError(err).Error("pipelined failed to retrieve a handler")
+				(logger.
+					WithFields(logrus.Fields{"handler": handlerName}).
+					WithError(err).
+					Error("pipelined failed to retrieve a handler", handlerName))
```
nikkiattea: Thank you for adding the `handlerName` as a logrus field! I hate to be logger nanny here, but we can probably remove it from the message in `Error()` since it's already a field.

backend/pipelined/handle.go
```diff
@@ -110,12 +111,21 @@ func (p *Pipelined) expandHandlers(ctx context.Context, handlers []string, level
...
 			extension, err = p.store.GetExtension(ctx, handlerName)
+			if err == store.ErrNoExtension {
+				continue
+			}
 			if err != nil {
-				logger.WithError(err).Error("pipelined failed to retrieve a handler")
+				(logger.
+					WithFields(logrus.Fields{"handler": handlerName}).
+					WithError(err).
+					Error("pipelined failed to retrieve a handler", handlerName))
```
nikkiattea: RT logger nanny.

testing/e2e/extensions_test.go
```diff
@@ -0,0 +1,261 @@
...
+				t.Fatalf("bad status: %d", status)
+			}
+
+			ext.Wait()
+			if ext.err != "" && !test.ExpErr {
+				t.Fatal(ext.err)
+			}
+
+			assert.Equal(t, test.Error, ext.err)
+			assert.Equal(t, test.Type, ext.reqType)
```
nikkiattea: 💯

echlebek:
Oh, just for the sake of posterity, I realized that neither a separate executable nor a "crazy test runner" were necessary solutions for this testing problem. It just took a while.

:man_facepalming:

Issue #1285, opened 2018-04-11
Use constants instead of hardcoded strings for all rules, all orgs, all envs


cyphus:
Inspired by this TODO https://github.com/sensu/sensu-go/blob/master/backend/authorization/authorization.go#L19-L23

Let's make the concept of `"*"` more explicit and manageable by turning all hardcoded string values for all rules, all organizations, and all environments into constants, and updating all references in the codebase/tests.

For example:
`rule := *FixtureRule("*", "*")` would become `rule := *FixtureRule(types.OranizationAll, types.EnvironmentAll)`

annaplotkin:
This is done.

PR #1286, opened 2018-04-11
Create default read-only user

cyphus:
## What is this change?

Seed the sensu RBAC store with a user by the name "sensu" with global read-only permissions.

It's useful for customers to have a non-admin user when setting up sensu. It's a common enough use case to warrant creating one by default.

Yes.

I did need some clarification on the use of constants in RBAC related code. Ended up creating #1285 to track it.

Because the RBAC seed functionality was already in place, adding a second user (after the admin user) was straightforward.

### DIFF
--- a/backend/authorization/authorization.go
+++ b/backend/authorization/authorization.go
@@ -3,8 +3,8 @@ package authorization
 import (
 	"net/http"
 
-	"github.com/sirupsen/logrus"
 	"github.com/sensu/sensu-go/types"
+	"github.com/sirupsen/logrus"
 )
 
 func hasPermission(rule types.Rule, action string) bool {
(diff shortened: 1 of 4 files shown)


cyphus:
I amended my commit that was missing DCO, but it looks like the DCO bot hasn't picked up the changes -- do I need to change the commit message too? I'll try that.

backend/authorization/authorization.go
```diff
@@ -16,10 +16,8 @@ func hasPermission(rule types.Rule, action string) bool {
 	return false
 }
 
-// TODO (JK): this function may end up becoming more complex if
-// we decide to use "*" as more than a way of saying "all resources"
 func matchesRuleType(rule types.Rule, resource string) bool {
-	return rule.Type == resource || rule.Type == "*"
+	return rule.Type == resource || rule.Type == types.RuleTypeAll
```
mercul3s: 👍

PR #1288, opened 2018-04-11
Provide a means of configuring event handlers for metrics

### LINKED ISSUE #1127: [Agent statsd] Provide a means of configuring Event Handlers
###  Background

See the [proposal](https://github.com/sensu/proposal/blob/master/design/85-agent-statsd.md).

Provide a means of configuring which Event Handlers the Events are sent to, for example, a Sensu Agent CLI argument to specify one or more Handlers `-statsd-event-handler=influxdb,graphite` which sets the value of `Event.Metrics.Handlers=[]` within the transmitted Events.

### Acceptance Criteria

- [ ] `Event.Metrics.Handlers` uses the configured handlers at the agent startup
- [ ] There's no default handlers and handlers can be empty

### Definition of Done

- [ ] Documentation
- [ ] Unit Tests
- [ ] Integration Test
- [ ] E2E Test

nikkiattea:

## What is this change?

Adds the `--statsd-event-handlers` flag to sensu-agent which configures the event handlers for statsd metrics.

Closes #1127.

Yas.

Nah.

Nah.

### DIFF
--- a/agent/cmd/start.go
+++ b/agent/cmd/start.go
@@ -10,12 +10,12 @@ import (
 	"sync"
 	"syscall"
 
-	"github.com/sirupsen/logrus"
 	"github.com/sensu/sensu-go/agent"
 	"github.com/sensu/sensu-go/types/dynamic"
 	"github.com/sensu/sensu-go/util/path"
 	"github.com/sensu/sensu-go/util/url"
 	"github.com/sensu/sensu-go/version"
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
@@ -47,6 +47,7 @@ const (
 	flagRedact                = "redact"
 	flagSocketHost            = "socket-host"
 	flagSocketPort            = "socket-port"
+	flagStatsdEventHandlers   = "statsd-event-handlers"
 	flagStatsdFlushInterval   = "statsd-flush-interval"
 	flagStatsdMetricsHost     = "statsd-metrics-host"
 	flagStatsdMetricsPort     = "statsd-metrics-port"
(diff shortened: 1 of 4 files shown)


agent/cmd/start.go
```diff
@@ -156,6 +157,14 @@ func newStartCommand() *cobra.Command {
 				cfg.Subscriptions = viper.GetStringSlice(flagSubscriptions)
 			}
 
+			// Get a single or a list of statsd event handlers
+			handlers := viper.GetString(flagStatsdEventHandlers)
+			if handlers != "" {
+				cfg.StatsdServer.Handlers = splitAndTrim(handlers)
+			} else {
+				cfg.StatsdServer.Handlers = viper.GetStringSlice(flagStatsdEventHandlers)
```
cyphus: This else clause of the conditional is meant to use the default value for the flag if the user didn't pass in any value (or didn't use the flag at all), correct?

palourde: We could actually replace that whole block (160 to 166) with this single line (165) by defining the flag `flagStatsdEventHandlers` flag as a `StringSlice` instead of a `String` on line  [262](https://github.com/sensu/sensu-go/pull/1288/files#diff-10c3ef0457ad91a0dcfe4a428511a129R262), just like we did for `flagBackendURL`.

nikkiattea: But then a user would have to use the flag multiple times rather than a comma delimited list, correct? For example: `sensu-agent --statsd-event-handler handler1 --statsd-event-handler handler2` vs. `sensu-agent --statsd-event-handlers handler1,handler2`

palourde: @nikkiattea Both works with `StringSlice`. For example, if I print the value of `viper.GetStringSlice(flagStatsdEventHandlers)` and its lenght:

```
$ sensu-agent
value: []
len: 0

$ sensu-agent --statsd-event-handlers foo
value: [foo]
len: 1

$ sensu-agent --statsd-event-handlers foo,bar
value: [foo bar]
len: 2

$ sensu-agent --statsd-event-handlers foo --statsd-event-handlers bar
value: [foo bar]
len: 2
```

Issue #1279, opened 2018-04-10
[Web UI] Switch to React Router v4


jamesdphillips:
Predicated on results of #1278.

PR #1289, opened 2018-04-12
[Web UI] Migrate from Relay to Apollo

### LINKED ISSUE #1278: [Web UI] Explore switch to Apollo from Relay
Explore switch to Apollo from Relay.

- As time has proceeded Apollo has continued to become better supported by the community
- Apollo continues to be simpler, easier to configure and more malleable.
- While not strictly true simplifies; usage maps better to react-router v4's declarative routing approach.
- more...

10xjs:
## What is this change?

Replace [Relay](http://facebook.github.io/relay/) with [Apollo](https://www.apollographql.com/docs/react/api/react-apollo.html).

Apollo serves as a nearly drop-in replacement for Relay that offers us a few advantages.

Apollo is less monolithic and less opinionated. Compared to Relay, more of Apollo's functionality is pluggable and in the open, providing more opportunity to customize things as needed. Its compositional nature allows us to build exactly the system we need.

One major difference worth noting is the lack of coupling between app routes and queries. With Apollo we are free to define root queries at points deep within the component tree, not only at the base router level. Combined with declarative routing from `react-router` v4 (#1279), different portions of the interface will be able to define their own queries. This will allow us to leverage subscription, polling, and refetch on targeted portions of the query graph and will be very useful for real-time lists.

Apollo defaults to injecting a `__typename` property to each level of every query. While this is configurable and can be disabled, doing so causes a ton of problems and prevents Apollo from from correctly understanding queries including fragments. Ideally we would disable this in the interest of minimizing the amount of data transferred on each query, but we probably can't do this while using `apollo-cache-inmemory`. In any case, there is no  evidence so far that including `__typename` adds any significant network/api overhead.

## Further Consideration

Investigate potential advantages of using redux as a cache (see: https://github.com/rportugal/apollo-cache-redux).

Consider using `graphql-tag/loader` to process queries to AST at build time. This could improve the performance of the client, but would necessitate defining all queries and fragments in separate files. I personally think that separating queries into files parallel to each component would be an organizational improvement, but this definitely deserves a separate discussion

Apollo provides the opportunity to manage client-side app state using GraphQL with `apollo-link-state`. This would serve as a replacement for `redux` providing an interface using queries and mutations to access client-side state. (see: https://www.apollographql.com/docs/link/links/state.html)

Investigate using `apollo-link-rest` to interact with REST api endpoints. The `auth` behavior and state could potentially be integrated into the Apollo client.

## Things to Work on Later

- Replace `Loading...` placeholders with nice animated loading indicators.
- Connect error reporting (replace instances of `console.error(error)` and `throw error` with rendered error alerts)
- Replace use of interval polling with proper query subscriptions.
- Create utils to abstract repetition among instances of `<Query>` (error handling and loading state)
- Ensure private data is removed from apollo cache when user logs out.
- Extract only required data from `schema.json` to be included into client build. (see: https://www.apollographql.com/docs/react/advanced/fragments.html#fragment-matcher)

---
Closes #1278


jamesdphillips:
@10xjs when you have a chance can you make sure to create issues for the items listed in "Things to Work on Later"? Thanks! 🙂

PR #1296, opened 2018-04-13
[GraphQL] Expose related entities using simple heuristic

jamesdphillips:
## What is this change?

Expose related entities using simple heuristic

Unblocks #1133 

Mocking each controller is cumbersome. Took a peak at https://github.com/vektra/mockery to possibly go:generate mocks for the controllers.

### DIFF
--- a/backend/apid/graphql/entity.go
+++ b/backend/apid/graphql/entity.go
@@ -1,12 +1,16 @@
 package graphql
 
 import (
+	"context"
+	"sort"
+
 	"github.com/sensu/sensu-go/backend/apid/actions"
 	"github.com/sensu/sensu-go/backend/apid/graphql/globalid"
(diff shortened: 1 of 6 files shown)


backend/apid/graphql/entity.go
```diff
@@ -58,6 +68,38 @@ func (r *entityImpl) Author(p graphql.ResolveParams) (interface{}, error) {
...
+	sort.Slice(entities, func(i, j int) bool {
+		return scores[i] < scores[j]
+	})
+
+	limit := p.Args.Limit
+	if limit > len(entities) {
+		limit = len(entities)
+	} else if limit < 0 {
+		limit = 0
+	}
```
10xjs: Good place to use the `constrainInt` function from #1297.

jamesdphillips: yup!

Issue #1132, opened 2018-03-07
[Web UI] Events Page - Check History Card


jamesdphillips:

jamesdphillips:
<img width="1263" alt="screen shot 2018-04-23 at 11 55 32 am" src="https://user-images.githubusercontent.com/194892/39147018-53f00c0a-46ed-11e8-8891-a3304f9c3b0d.png">

annaplotkin:
@jamesdphillips Can you give me some more info on this when you are back from PTO? There isn't a lot of info in the ticket and I want to make sure I understand what the request is.

jamesdphillips:
@annaplotkin

Some context: Previously in Uchiwa, history was displayed as a comma separated list (eg. `0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 2, 0, 0, 0, 0, 0, 1, 0, 0`), while this was okay, it wasn't exactly friendly for those who were not Sensu experts already. In the interest of reflecting this history in a more meaningful way, @dabria, did some exploration and after some iteration we landed on what you see in the description above.

---

Uchiwa.

<img width="729" alt="screen shot 2018-09-20 at 3 36 06 pm" src="https://user-images.githubusercontent.com/194892/45850452-13c7d580-bceb-11e8-8b6a-ae4c86999a3c.png">

---

The intention was that instead of a comma separated list, the card would visually display the status transitions for the last twenty five executions. For instance in the screenshot below, this would describe a check that was passing and then eventually started failing around ~22:50 and then eventually being resolved ~23:20.

![screen shot 2018-09-20 at 3 22 36 pm](https://user-images.githubusercontent.com/194892/45849967-027dc980-bce9-11e8-9c4d-4b78575c6eda.png)

To see specific details about when the executions occurred we envisioned hovering over the timeline and having a tooltip communicate the specific time and status.

![screen shot 2018-09-20 at 3 24 57 pm](https://user-images.githubusercontent.com/194892/45850035-5688ae00-bce9-11e8-9835-96e92d7a88f2.png)

... and hovering for the next occurrence...

![screen shot 2018-09-20 at 3 31 54 pm](https://user-images.githubusercontent.com/194892/45850270-50470180-bcea-11e8-90fa-f0bb53327173.png)

...and for narrower viewports...

![screen shot 2018-09-20 at 3 35 20 pm](https://user-images.githubusercontent.com/194892/45850381-c9deef80-bcea-11e8-91cc-b32d73cca099.png)

PR #1297, opened 2018-04-13
[GraphQL] Expose an environment's check history

jamesdphillips:
## What is this change?

Expose an environment's check history for use in web UI's history graph. (And more! maybe.)

Unblocks #1132 

Same "complication" as mentioned in #1296.

> Mocking each controller is cumbersome. Took a peak at https://github.com/vektra/mockery to possibly go:generate mocks for the controllers.

### DIFF
--- a/backend/apid/graphql/util.go
+++ b/backend/apid/graphql/util.go
@@ -1,0 +1,11 @@
+package graphql
+
+// constrainInt returns int within range.
+func constrainInt(num, min, max int) int {
+	if num < min {
+		return min
+	} else if num > max {
+		return max
+	}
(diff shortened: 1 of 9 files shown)


backend/apid/graphql/util.go
```diff
@@ -0,0 +1,11 @@
+package graphql
+
+// constrainInt returns int within range.
+func constrainInt(num, min, max int) int {
```
10xjs: Thoughts on calling this `clampInt`?

jamesdphillips: I like.

PR #1305, opened 2018-04-13
Add extension package and extension example.

### LINKED ISSUE #341: Golang framework for Sensu Extensions
Much like the Ruby EM+gRPC Shim, we need something similar that makes building Go extensions simple.

We probably want to go one step further here, though:

We may want an interface that duplicates something like the gRPC service so that we can have in-process enterprise extensions without requiring a gRPC client.

echlebek:
The extension package contains convenience types and functions for
writing third-party Sensu extensions in Go. An example is also provided.

Closes #341 

Yes

### DIFF
--- a/rpc/extension/server.go
+++ b/rpc/extension/server.go
@@ -1,0 +1,93 @@
+package extension
+
+import (
+	"context"
+	"net"
+
+	"github.com/sensu/sensu-go/rpc"
+	"google.golang.org/grpc"
+)
+
+var _ rpc.ExtensionServer = &Server{}
+
+// Server is a Sensu extension.
+//
+// Create new extensions that will run on a local port with NewServer, and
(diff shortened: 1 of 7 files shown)


rpc/extension/server.go
```diff
@@ -0,0 +1,93 @@
...
+		server: grpc.NewServer(),
+		ext:    extension,
+	}
+}
+
+// Start starts the extension service.
+func (s *Server) Start() error {
+	rpc.RegisterExtensionServer(s.server, s)
+	go func() {
+		// TODO: add error handling once gRPC errors are a little less
```
mercul3s: Out of curiosity, what do the gRPC errors look like now?

echlebek: I think it's this: https://github.com/grpc/grpc-go/issues/1737

Not sure, since it's been merged in.

PR #1306, opened 2018-04-14
Optimize check marshaling and add a benchmark for subPump

echlebek:
## What is this change?

```
    This commit significantly reduces the amount of time spent
    marshaling json in agentd. It replaces stdlib encoding/json with
    json-iterator, and it avoids using dynamic.Marshal when there are no
    dynamic attributes to be marshaled.
    
    Profiling performed during load testing shows that json encoding
    and decoding can comprise a substantial portion of time spent in
    the backend. This is most evident when there are many thousands of
    checks being executed per second.
    
    The benchmark results below are for a single iteration of subPump.
    subPump is part of agentd; it receives messages on a channel, and
    forwards them along to the agent. Profiling shows that the majority
    of time in subPump is spent dealing with json rendering. For every
    check that an agent executes, it must process a single subPump
    iteration.
    
    Before:
    26323 ns/op   16770 B/op  149 allocs/op
    
    After:
    5863  ns/op   3822  B/op  40  allocs/op
    
    Note that memory allocations are significantly reduced as well,
    reducing GC pressure by over a factor of 2.
```

At scale, Sensu backends will need to process tens of thousands of requests per second. This commit should significantly reduce request latency, while preserving json compatibility.

Yes

Should we start using protobuf marshaling for backend<->agent communication?

### DIFF
--- a/backend/agentd/benchmark_test.go
+++ b/backend/agentd/benchmark_test.go
@@ -1,0 +1,58 @@
+package agentd
+
+import (
+	"testing"
+
+	"github.com/sensu/sensu-go/backend/messaging"
+	"github.com/sensu/sensu-go/testing/mockring"
+	"github.com/sensu/sensu-go/testing/mockstore"
+	"github.com/sensu/sensu-go/transport"
+	"github.com/sensu/sensu-go/types"
+	"github.com/stretchr/testify/mock"
+)
+
+func BenchmarkSubPump(b *testing.B) {
+	conn := &testTransport{
+		sendCh: make(chan *transport.Message, 10),
+	}
+
+	bus, err := messaging.NewWizardBus(messaging.WizardBusConfig{
+		RingGetter: &mockring.Getter{},
+	})
+	if err != nil {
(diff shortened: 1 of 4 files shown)


backend/agentd/benchmark_test.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+	session, err := NewSession(cfg, conn, bus, st)
+	if err != nil {
+		b.Fatal(err)
+	}
+
+	go func() {
+		for range session.sendq {
+		}
+	}()
```
cyphus: Looping through `session.sendq` is what causes the messages to be processed? Or is this just a no-op?

echlebek: It just drains the channel that subPump fills - otherwise the loop would block indefinitely.

cyphus: Okay, after reading through the subPump implementation this makes more sense. I still have more to learn on Go's concurrency model though.

PR #1318, opened 2018-04-17
Generate interface mocks

### LINKED ISSUE #1315: Generate interface mocks
Writing interface mocks by hand is both time consuming and arguably brittle. As such, ideally we should generate our mocks.

If we were to use a an existing project [vektra/mockery](https://github.com/vektra/mockery) seems like a good candidate as it uses `testify/mock` under the hood.

### Proof of Concept

Example implementation uses `vektra/mockery` and generates mocks for the `Transport` and one of the controllers.

Script: https://github.com/sensu/sensu-go/commit/0a0209a99b8e75c6b38cc073d4b63791fef8029d
Generate Transport: https://github.com/sensu/sensu-go/commit/d9b28e14dd00b531838ed7ce08037bb9ccd4bd08
Generate Controller: https://github.com/sensu/sensu-go/commit/24cb0c1103aa60c7ecf037c8cb77a79713e73458
Full diff: <code>[compare](https://github.com/sensu/sensu-go/compare/prototype/mockery?expand=1)</code>

jamesdphillips:
## What is this change?

Writing interface mocks by hand is both time consuming and arguably brittle. This pull request adds https://github.com/vektra/mockery to effortlessly generate our interface mocks.

Closes #1315 

Unlikely. Nothing customer facing.


jamesdphillips:
Sorry for the size of the pull request, I wanted to update all the mock packages in fell swoop instead of leaving the job partially finished. Each commit should be reviewable on it's own, however, if it's too tricky to review I can split into separate pulls.

scripts/make_mock/main.go
```diff
@@ -0,0 +1,75 @@
...
+	interfaces []string
+	outDir     string
+	outPkg     string
+}
+
+func main() {
+	cfg := parseConfigFromArgs(os.Args)
+
+	if len(cfg.interfaces) < 0 {
+		fmt.Fprintln(os.Stderr, "must specift at least one interface to mock")
```
echlebek: typo

jamesdphillips:
> ...it seems that since controller dependencies are already mockable...

I'm not convinced that this is true. If you imagine writing a test that depends on a controller you now need to understand the implementation of the controller to properly mock calls to the store, bus, etc. This will lead to tests in a consumer's package failing after the implementation of the provider's package changes.

I feel as if the implementation of the producer should remain opaque.

> Note that this testing strategy directly contravenes the code review comments... we should seek to limit it as much as possible.

This is fair. The approach outlined in the review guidelines is also the approach- to date- I've took with mocks in the `graphql` package. However, I've noticed a couple pain points (also what motivated these changes.)

1. It is time-consuming to pump out and maintain redundant interfaces and mocks.
1. Worse, since both the `graphql` and `router` packages are consumers of the package these mocks are (largely) duplicated in two places.

jamesdphillips:
Since the controller mocks clearly need more discussion I'll remove them from this PR and open a separate issue to discuss alternate approaches.

PR #1319, opened 2018-04-17
[Web UI] Add a mechanism to subscribe to auth token changes

10xjs:
## What is this change?

This allows the auth storage utils to behave like a flux store for any components that need the state.

Components that depend on the state of user auth (specifically auth-based routes) need to have a mechanism with which to subscribe to auth state changes

### DIFF
--- a/dashboard/src/utils/authentication/tokens.js
+++ b/dashboard/src/utils/authentication/tokens.js
@@ -1,5 +1,9 @@
 import moment from "moment";
 
+import createDispatcher from "../dispatcher";
+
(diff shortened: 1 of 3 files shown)


dashboard/src/utils/authentication/tokens.js
```diff
@@ -20,8 +24,11 @@ export function get() {
 
 // Swap single authTokens instance for new one
 export function swap(t) {
-  authTokens = t;
+  if (Object.prototype.toString.call(t) !== "[object Object]" || !t) {
+    throw new TypeError("Expected Object");
+  }
+  authTokens = Object.freeze(t);
```
jamesdphillips: 👍

Issue #1175, opened 2018-03-16
Deleting silenced entries with sensuctl no longer works


palourde:
## Expected Behavior
When deleting a specific silenced entry with `sensuctl`, I expect the entry to be removed.

## Current Behavior
The entry is not deleted:

```bash
$ sensuctl silenced list
         ID            Subscription     Check    Expire     ExpireOnResolve   Creator               Reason               Organization   Environment
 ─────────────────── ───────────────── ─────── ─────────── ───────────────── ───────── ──────────────────────────────── ────────────── ─────────────
  entity:i-424242:*   entity:i-424242           41h39m15s   false             admin     Server upgrade                   default        default

$ sensuctl silenced delete 'entity:i-424242:*'
? Are you sure you would like to delete resource 'entity:i-424242:*'? Yes
OK

$ sensuctl silenced list
         ID            Subscription     Check    Expire     ExpireOnResolve   Creator               Reason               Organization   Environment
 ─────────────────── ───────────────── ─────── ─────────── ───────────────── ───────── ──────────────────────────────── ────────────── ─────────────
  entity:i-424242:*   entity:i-424242           41h39m15s   false             admin     Server upgrade                   default        default
```
## Steps to Reproduce (for bugs)

1. Create the following silenced entry:

```
sensuctl silenced create \
--subscription 'entity:i-424242' \
--begin 'Mar 18 2018 1:00AM' \
--expire 3600 \
--reason 'Server upgrade'
```

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): Master (https://github.com/sensu/sensu-go/commit/ee858323a6c85d0d166bf7b9362565fe08855c5b)
* Installation method (packages, binaries, docker etc.): Manually built
* Operating System and version (e.g. Ubuntu 14.04): macOS High Sierra 10.13.3

PR #1331, opened 2018-04-18
Silenced entries with an asterisk in their ID can be deleted

palourde:

## What is this change?

It unescapes the ID parameter when deleting a silenced entry, in case an asterisk (`*`) is used.

I also did a small cleaning.

Closes https://github.com/sensu/sensu-go/issues/1175

Added!

Non!

Non!


backend/apid/actions/silenced.go
```diff
@@ -217,6 +218,11 @@ func (a SilencedController) Destroy(ctx context.Context, params QueryParams) err
 		}
 	}
 
+	// Unescape the ID parameter, in case we have something like an asterisk (*)
+	if unescapedID, err := url.QueryUnescape(id); err == nil {
```
echlebek: There's a bit of a discrepancy in the destroy function here. In most other controller methods we pass in the ID after calling `url.PathUnescape`.

I think that `PathUnescape` is correct here, not `QueryUnescape`, since the ID component is coming from the path. Since we are doing the rest of the unescaping in the router methods, we should do the escaping for destroy there as well.

Issue #1130, opened 2018-03-07
[Web UI] Events Page - Display Check Results


jamesdphillips:
## 🚀 Sketch

![screen shot 2018-04-11 at 12 10 12 pm](https://user-images.githubusercontent.com/194892/38637900-5e65c87a-3d81-11e8-8e47-8841b1ca160b.png)
![screen shot 2018-04-11 at 12 10 02 pm](https://user-images.githubusercontent.com/194892/38637901-5e7dbb06-3d81-11e8-9b56-51ec9e643efe.png)

PR #1334, opened 2018-04-18
[GraphQL] Change return type of issued field

jamesdphillips:
## What is this change?

Change the return type of the check type's `issued` field.

Unblocks #1130

### DIFF
--- a/backend/apid/graphql/check.go
+++ b/backend/apid/graphql/check.go
@@ -84,6 +84,12 @@ func (r *checkImpl) Executed(p graphql.ResolveParams) (time.Time, error) {
 	return time.Unix(c.Executed, 0), nil
 }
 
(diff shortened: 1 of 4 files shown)


10xjs:
I highly recommend serializing timestamps in ISO 8601 format which includes an explicit timezone offset. 

This is the same format retuned by JavaScript `Date.prototype.toJSON`. Using it means using `Date.parse(value)` (which will assume the local system offset if the provided value does not include one) will not eventually cause bugs in the FE when someone forgets the "right" way to parse a UNIX timestamp in js:  https://stackoverflow.com/questions/847185/convert-a-unix-timestamp-to-time-in-javascript.

Example in go: https://play.golang.org/p/8SXhlGl6_fX

PR #1336, opened 2018-04-18
InfluxDB Handler

nikkiattea:

## What is this change?

Adds an influx-db event handler.

Allows users to configure an event handler which will write sensu metrics to an influx-db of their choice!

Yas.

Nah.

Nah.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -112,7 +112,7 @@ build_tools () {
         build_tool $cmd "tools"
     done
 
-    for cmd in slack; do
+    for cmd in slack influx-db; do
         build_tool $cmd "handlers"
     done
 }
@@ -239,7 +239,7 @@ docker_commands () {
         build_tool_binary linux amd64 $cmd "tools"
     done
 
-    for cmd in slack; do
+    for cmd in slack influx-db; do
         echo "Building handlers/$cmd for linux-amd64"
         build_tool_binary linux amd64 $cmd "handlers"
     done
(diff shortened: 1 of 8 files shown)


build.sh
```diff
@@ -239,7 +239,7 @@ docker_commands () {
         build_tool_binary linux amd64 $cmd "tools"
     done
 
-    for cmd in slack; do
+    for cmd in slack influx-db; do
```
cyphus: This is a small thing, but we should probably make handlers a variable and loop through it where needed in build.sh.

handlers/influx-db/main_test.go
```diff
@@ -0,0 +1,68 @@
...
+	event.Metrics = types.FixtureMetrics()
+
+	var apiStub = httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		body, _ := ioutil.ReadAll(r.Body)
+		expectedBody := `answer,foo=bar value=42`
+		assert.Contains(string(body), expectedBody)
+		w.WriteHeader(http.StatusOK)
+		_, err := w.Write([]byte(`{"ok": true}`))
+		require.NoError(t, err)
+	}))
```
cyphus: This httptest stuff is great!

build.sh
nikkiattea: @cyphus I can probably do that in sh, but not quite sure how to do that for ps.

nikkiattea:
Filed https://github.com/sensu/sensu-go/issues/1379 for Travis failure.

PR #1339, opened 2018-04-18
[GraphQL] Fixup bug where env could not found

jamesdphillips:
## What is this change?

Fix bug where environment could not be found.

Allow web UI (or any consumer of GraphQL service) to fetch environments.

🤔 this might be where a smattering of APId integration tests may be helpful useful.


cyphus:
I can see how using a default environment/organization of "default" can lead to argument ordering issues. LGTM!

PR #1354, opened 2018-04-21
Make sensu-{backend,agent} sysvinit compat for centos 6 family

webframp:
Centos 6, perhaps others, using chkconfig to enable services fail if
missing the "chkconfig" line or LSB in the head comment to set runlevels and
priority. The longer LSB style form is also supported by Ubuntu 14.04 init, in theory

Resolves the following error on centos 6:

  STDERR: service sensu-backend does not support chkconfig

ref:
- https://linux.die.net/man/8/chkconfig (heading: Runlevel Files)

Signed-off-by: Sean Escriva <sean.escriva@gmail.com>

## What is this change?

Updates the sysv init scripts 

Testing for chef cookbook development on centos 6 revealed this issue and it seemed possibly better to fix upstream instead of patching the initscripts with the cookbook.

Maybe? I didn't want to presume.

Don't think so.

Nope.

### DIFF
--- a/packaging/services/sensu-agent/sysvinit/etc/init.d/sensu-agent
+++ b/packaging/services/sensu-agent/sysvinit/etc/init.d/sensu-agent
@@ -1,3 +1,14 @@
+#!/bin/sh
+### BEGIN INIT INFO
+# Provides:          sensu-agent
+# Required-Start:    $remote_fs $syslog
+# Required-Stop:     $remote_fs $syslog
+# Default-Start:     2 3 4 5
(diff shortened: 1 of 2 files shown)


webframp:
Sensu is dope so I won't be offended if this is rejected and you want this solved in some other way.

Looks like @amdprophet and @nikkiattea  have history in this area

amdprophet:
Thanks for the contribution! I'm to test these changes on Ubuntu and I'll merge this once I've confirmed everything looks good.

Issue #1148, opened 2018-03-09
Improve pipelined log entries


palourde:
As I'm writing a documentation guide for filters, I'm realizing the log entries produced by `pipelined` could be a bit more precise.

**When an event is being handled, we get the following log entries**

```
[...]

PR #1362, opened 2018-04-23
Improve pipelined log entries

palourde:

## What is this change?

It improves the log entries produced by pipelined so it's easier to understand what's going on.

More precisely:
- Context helpers from the store now have their own util package so they can be reused.
- Created a util package to produce log fields from an event.
- We now have an environment, organization and optionally handler name and/or mutator name fields in our log entries.
- Cleanup of the log messages.

Closes https://github.com/sensu/sensu-go/issues/1148

Added!

Non!

Non!


palourde:
@echlebek Makes sense!

PR #1363, opened 2018-04-23
[Web UI] Avoid relative imports with module-resolver plugin

10xjs:
## What is this change?

Any absolute es6 import statement is resolved relative to the `src` directory. 
Compared to a set of webpack aliases to specific directories within the project, this approach is more flexible, less fragile, and easier to reason.

This avoids the need to use deep relative paths that reference a file's current location making the app far easier to organize. This effectively allows any file in `src` to be referenced as a namespaced module, akin to any of the npm dependencies.

Nah.

Nah.

In order to use `eslint-import-resolver-babel-module`, `babel-preset-react-app` had to be abandoned. The new `.babelrc` file is a simplified copy of that preset preserving compatibility.

### DIFF
--- a/dashboard/.babelrc
+++ b/dashboard/.babelrc
@@ -1,0 +1,70 @@
+{
+  "presets": [
+    ["babel-preset-env", {
+      "targets": {
+        // React parses on ie 9, so we should too.
+        "ie": 9,
+      },
+      // Disable polyfill transforms.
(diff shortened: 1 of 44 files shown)


jamesdphillips:
Build appears to be failing because https://github.com/golang/lint/issues/397.

PR #1364, opened 2018-04-24
Disable gostatsd internal metric collection

### LINKED ISSUE #1259: [Agent statsd] Disable internal metrics
Fork gostatsd and implement https://github.com/atlassian/gostatsd/issues/167 to disable internal metrics. Update the vendor package and invoke the new functionality in statsd_server.go.

nikkiattea:

## What is this change?

Disables gostatsd internal metric collection by bumping the revision of the dependency and using the new NullStatser.

Closes #1259.

Yas.

Nah.

Nah.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -13,6 +13,8 @@ RACE=""
 
 VERSION_CMD="go run ./version/cmd/version/version.go"
 
+HANDLERS=(slack influx-db)
+
 set_race_flag() {
     if [ "$GOARCH" == "amd64" ]; then
         RACE="-race"
@@ -112,7 +114,7 @@ build_tools () {
         build_tool $cmd "tools"
     done
 
-    for cmd in slack; do
+    for cmd in ${HANDLERS[@]}; do
         build_tool $cmd "handlers"
     done
 }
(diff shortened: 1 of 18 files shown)


palourde:
@nikkiattea You might want to merge master into your branch in order to fix Travis builds and you might have to look into AppVeyor failures!

build.sh
```diff
@@ -42,6 +42,7 @@ install_deps () {
     go get github.com/jgautheron/goconst/cmd/goconst
     go get honnef.co/go/tools/cmd/megacheck
     go get github.com/golang/lint/golint
+    install_golang_dep
```
echlebek: Just noticed that `install_golang_dep` does `go get` on dep. But that gives an unspecified version of dep (master).

We should probably think about _not_ doing that, because it makes our builds not reproducible if people are using different versions of dep.

nikkiattea: Is there another way we can install our vendor dependencies in CI at least? Because it requires `go get` on dep if we want to `dep ensure` right?

nikkiattea: I'll wait for @echlebek PR to get merged that will install our dependencies differently for CI and remove `install_golang_dep` from build.sh and build.ps1 so our builds will be reproducible. I'll rebase from master once that's done 😅

PR #1380, opened 2018-04-25
Pin dep version to 4.1.

echlebek:

## What is this change?

It makes sure we are using a specific dep version to build our PRs.

Previously we were using master of dep.

NEIN

NEIN

There probably will be tbh

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -44,12 +44,6 @@ install_deps () {
     go get github.com/jgautheron/goconst/cmd/goconst
     go get honnef.co/go/tools/cmd/megacheck
     go get github.com/golang/lint/golint
-    install_golang_dep
-}
-
-install_golang_dep() {
-    go get github.com/golang/dep/cmd/dep
-    echo "Running dep ensure..."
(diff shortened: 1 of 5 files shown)


build.sh
```diff
@@ -44,12 +44,6 @@ install_deps () {
     go get github.com/jgautheron/goconst/cmd/goconst
     go get honnef.co/go/tools/cmd/megacheck
     go get github.com/golang/lint/golint
-    install_golang_dep
-}
-
-install_golang_dep() {
-    go get github.com/golang/dep/cmd/dep
```
nikkiattea: What if a user (not CI) does not have dep installed? They'll just get an error?

echlebek: That's right. They will have to install dep on their own.

echlebek: Yeah, we should add it to the getting started instructions.

The problem is that we can't install dep for people without making assumptions or adding complexity. Similarly to how we can't install Go for people.

echlebek:
I added some notes about Go and dep to the readme.

PR #1390, opened 2018-04-25
Allow nullable flap thresholds.

### LINKED ISSUE #1235: Low/High flap thresholds should have a default of null
Low & high flap thresholds currently default to 0. We cannot determine if the value is unset or was intentionally set to 0.

## Expected Behavior
Low / high flap thresholds should allow for (and default to) null values.

## Current Behavior
Low / high flap thresholds default to 0.

## Possible Solution
Change to a type that allows for null values (int pointer?).

N/A

## Context
N/A

## Your Environment
N/A

echlebek:
This commit allows users to indicate negative presence of a flap
threshold by supporting 'null' as a valid JSON value. It does this
by defining a proto message called FlapThreshold, which can then
be assigned nil.

This is necessary because in proto3, scalar values cannot be
pointers in generated code.

In GraphQL, a value of -1 will indicate the absence of a flap
threshold value.

Closes #1235 

Yes

Do we need to support flap threshold validation in parity with 1.x?

Yes. Protobuf is fairly inflexible about how things get generated, and I had to implement a wrapper type to achieve the desired behaviour.


echlebek:
Closing this so I can take it in a different direction. See the issue for details.

Issue #1387, opened 2018-04-25
[Metric Extraction] Agent opentsdb_line transformer


nikkiattea:
Child of https://github.com/sensu/sensu-go/issues/745
Depends on #1383 

Create a function `Transform(string) []MetricPoint, error` from the interface `Transformer` for opentsdb_line. The function should accept a string (check output) and parse/transform the metrics to Sensu Metric Format. Return an error if the line is incorrectly formatted. The `opentsdb_line` transformer will implement the [OpenTSDB](http://opentsdb.net/docs/build/html/user_guide/writing.html#data-specification) line protocol, a single line of values, to extract the MetricPoints:

```
sys.cpu.user 1356998400 42.5 host=webserver01 cpu=0
```

Additionally, add agent logic in check_handler.go to implement this transformer if the check’s `metric_format` matches the const `opentsdb_line`.

Issue #1386, opened 2018-04-25
[Metric Extraction] Agent influxdb_line transformer


nikkiattea:
Child of https://github.com/sensu/sensu-go/issues/745
Depends on #1383 

Create a function `Transform(string) []MetricPoint, error` from the interface `Transformer` for influxdb_line. The function should accept a string (check output) and parse/transform the metrics to Sensu Metric Format. Return an error if the line is incorrectly formatted. The `influxdb_line` transformer will implement the [InfluxDB](https://docs.influxdata.com/influxdb/v1.4/write_protocols/line_protocol_tutorial/#measurement) line protocol, a single line of values, to extract the MetricPoints:

```
weather,location=us-midwest temperature=82 1465839830100400200
```

Additionally, add agent logic in check_handler.go to implement this transformer if the check’s `metric_format` matches the const `influxdb_line`.

Issue #1385, opened 2018-04-25
[Metric Extraction] Agent nagios_perfdata transformer


nikkiattea:
Child of https://github.com/sensu/sensu-go/issues/745
Depends on #1383

Create a function `Transform(string) []MetricPoint, error` from the interface `Transformer` for nagios_perfdata. The function should accept a string (check output) and parse/transform the metrics to Sensu Metric Format. Return an error if the line is incorrectly formatted. The `nagios_perfdata` transformer will implement the [Nagios Perf Data](https://assets.nagios.com/downloads/nagioscore/docs/nagioscore/3/en/perfdata.html) line protocol, a single line of human readable text, to extract the MetricPoints:

```
PING ok - Packet loss = 0%, RTA = 0.80 ms
```

or

```
PING ok - Packet loss = 0%, RTA = 0.80 ms | percent_packet_loss=0, rta=0.80
```

Additionally, add agent logic in check_handler.go to implement this transformer if the check’s `metric_format` matches the const `nagios_perfdata`.

Issue #1384, opened 2018-04-25
[Metric Extraction] Agent logic for metric_format, metric_handlers, and graphite_plaintext


nikkictl:
Child of https://github.com/sensu/sensu-go/issues/745
Depends on #1381 and #1383 

On the agent side in check_handler.go, invoke the graphite_plaintext transformer if the check’s `metric_format` matches the const `graphite_plaintext`. Pass the check output into it’s `Transform(string)` function, and store the resulting []MetricPoint in event.Metrics.Points. Inject event.Metrics.Handlers with the `metric_handlers` from the check. Send a failure if the transformer is unable to extract metrics.

PR #1404, opened 2018-04-27
Add Graphite Plain Text transformer.

### LINKED ISSUE #1383: [Metric Extraction] Agent graphite_plaintext transformer
~Depends on #1381~

Create a `Transformer` interface with the function `Transform(string) []MetricPoint, error`. The function should accept a string (check output) and parse/transform the metrics to Sensu Metric Format. Return an error if the line is incorrectly formatted. The `graphite_plaintext` transformer will implement the [Graphite Plain Text](http://graphite.readthedocs.io/en/latest/feeding-carbon.html#the-plaintext-protocol) line protocol, a single line containing 3 values, to extract the MetricPoints:

```
<metric path> <metric value> <metric timestamp>
```

nikkiattea:

## What is this change?

Creates the Transformer interface and the Graphite Plain Text transformer.

Closes #1383.

Yas.

Nah.

Nah.

### DIFF
--- a/agent/transformers/graphite.go
+++ b/agent/transformers/graphite.go
@@ -1,0 +1,53 @@
+package transformers
+
+import (
+	"errors"
+	"strconv"
+	"strings"
+
+	"github.com/sensu/sensu-go/types"
+)
+
+// Graphite contains values of graphite plain text metric format
+type Graphite struct {
+	Path      string
+	Value     float64
+	Timestamp int64
+}
+
+// Transform transforms a metric in graphite plain text format to Sensu Metric
+// Format
+func (g Graphite) Transform() []types.MetricPoint {
+	mp := types.MetricPoint{
+		Name:      g.Path,
+		Value:     g.Value,
+		Timestamp: g.Timestamp,
+		Tags:      []*types.MetricTag{},
+	}
+	return []types.MetricPoint{mp}
+}
+
+// ParseGraphite parses a graphite plain text string into a Graphite struct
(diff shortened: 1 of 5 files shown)


agent/transformers/graphite.go
```diff
@@ -0,0 +1,53 @@
...
+		Name:      g.Path,
+		Value:     g.Value,
+		Timestamp: g.Timestamp,
+		Tags:      []*types.MetricTag{},
+	}
+	return []types.MetricPoint{mp}
+}
+
+// ParseGraphite parses a graphite plain text string into a Graphite struct
+func ParseGraphite(metric string) (Transformer, error) {
```
echlebek: It's better to return a Graphite type here. It will be a Transformer since it implements the Transformer method set.

nikkiattea: You got it!

nikkiattea:
@echlebek I made that small change you suggested! Yes, the transformer interface is not needed by a consumer yet.. but it will be required for #1384, #1385, #1386, and #1387. I was trying to eliminate that dependency so those tickets wouldn't be dependent on each other, rather just this one. I could move the transformer interface to #1384 if you would rather?

echlebek:
@nikkiattea I think the interface itself is well spec'd, but I'm wondering if it would be better for it to exist in the package where it will be used, rather than in the transformers package?

nikkiattea:
@echlebek And avoid `transformers.Transformer`, why yes, yes indeed.

Issue #837, opened 2018-01-08
[Web UI] Events Page: Delete


jamesdphillips:
## 🕸 Related

Sketch: https://github.com/sensu/sensu-go/issues/814

PR #1406, opened 2018-04-27
[GraphQL] Use string to represent ID scalar

jamesdphillips:
## What is this change?

The `ID` scalar type represents a unique identifier, often used to
refetch an object or as key for a cache. The ID type appears in a JSON
response as a String; however, it is not intended to be human-readable.
When expected as an input type, any string (such as ) or integer
(such as `4`) input value will be accepted as an ID.

As such, when resolving a field we can safely assume the value
encountered will always be a string value.

Spec: http://facebook.github.io/graphql/October2016/#sec-ID

Unblocks #837 

Unlikely

nada 

nada

### DIFF
--- a/graphql/generator/object.go
+++ b/graphql/generator/object.go
@@ -773,9 +773,7 @@ func genFieldResolverTypeCoercion(t ast.Type, i info, nullable bool) jen.Code {
 		return jen.Qual(defsPkg, "Int").Op(".").Id("ParseValue").Call(jen.Id("val")).Assert(jen.Int())
 	case graphql.Float.Name():
 		return jen.Qual(defsPkg, "Float").Op(".").Id("ParseValue").Call(jen.Id("val")).Assert(jen.Float64())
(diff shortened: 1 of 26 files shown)


graphql/generator/object.go
```diff
@@ -770,6 +773,8 @@ func genFieldResolverTypeCoercion(t ast.Type, i info, nullable bool) jen.Code {
 		return jen.Qual(defsPkg, "Int").Op(".").Id("ParseValue").Call(jen.Id("val")).Assert(jen.Int())
 	case graphql.Float.Name():
 		return jen.Qual(defsPkg, "Float").Op(".").Id("ParseValue").Call(jen.Id("val")).Assert(jen.Float64())
+	case graphql.ID.Name():
+		fallthrough
```
echlebek: I think this is redundant, since nothing is changed before the fallthrough.

jamesdphillips: 👍 replaced

Issue #1369, opened 2018-04-24
[Web UI] List pagination


10xjs:
Add pagination to table list views (events, entities, checks)

PR #1410, opened 2018-04-30
[GraphQL] Use offset-based pagination for large connections

jamesdphillips:
## What is this change?

Implement offset-based pagination strategy for check, entities and event connections.

The existing cursor-based pagination worked well with infinite scrolling and
when a resource's position in a sorted set was predictable (eg. when sorted
by creation date, elements are only ever shifted a new element is added).
However, when implementing a page based pagination solution on a client,
with multiple sorting options, the product of requested page quickly becomes
unintuitive to an end-user.

This PR switches the checks, entities and events connections to use a simplistic
offset-based pagination solution.

Existing supporting code has been removed but may re-introduced if we
need an event stream or activity feed in the future.

Unblocks #1369 

## Example Usage

screen shot 2018-04-30 at 11 26 17 am

unlikely

nada

nada


grepory:
@jamesdphillips Anything I can do to help move this along?

backend/apid/graphql/schema/pagination.graphql
```diff
@@ -0,0 +1,19 @@
...
+  "When paginating forwards, are there more items?"
+  hasNextPage: Boolean!
+
+  "When paginating backwards, are there more items?"
+  hasPreviousPage: Boolean!
+
+  "Next offset to use when paginating forwards; null if there are no more items."
+  nextOffset: Int
+
+  "Previous offset to use when paginating forwards; null if at terminal."
```
10xjs: "when paginating forwards" typo

PR #1412, opened 2018-04-30
Support metric extraction for graphite_plaintext transformer

### LINKED ISSUE #1384: [Metric Extraction] Agent logic for metric_format, metric_handlers, and graphite_plaintext

nikkictl:
## What is this change?

Adds support for metric extraction from check output for `graphite_plaintext` transformer in the agent's check handler.

Closes #1384 

Yas.

Nah.

Nah.

### DIFF
--- a/agent/check_handler.go
+++ b/agent/check_handler.go
@@ -118,7 +118,10 @@ func (a *Agent) executeCheck(request *types.CheckRequest) {
 	if check.MetricFormat != "" {
 		event.Metrics.Points = extractMetrics(event)
 	}
-	event.Metrics.Handlers = check.MetricHandlers
+
+	if check.MetricHandlers != nil {
+		event.Metrics.Handlers = check.MetricHandlers
+	}
 
 	msg, err := json.Marshal(event)
 	if err != nil {


agent/check_handler.go
```diff
@@ -179,3 +188,22 @@ func (a *Agent) sendFailure(event *types.Event, err error) {
...
 	}
 }
+
+func extractMetrics(event *types.Event) []*types.MetricPoint {
+	var transformer Transformer
+	var err error
+	switch event.Check.MetricFormat {
+	case types.GraphiteMetricFormat:
+		transformer, err = transformers.ParseGraphite(event.Check.Output)
+	}
```
palourde: So in 1.x, the graphite format is the default one in case no metric format is specified: https://github.com/sensu/sensu-enterprise/blob/057f5af0d4960b70241fdf5181ce604d3afb4262/lib/sensu/enterprise/extensions/metric_parsers.rb#L215.

Should we have a `default` statement here too, instead of throwing a `metric format is not supported` error?

nikkiattea: Hm, that's a good point, but I find that strange that we would still attempt to extract a metric (in graphite format) even if the user does not specify a metric format. It seems like we'd be putting in some extra work on the agent for _every_ single check.

I would propose the following logic: if a metric_format is not set, we do not attempt to extract metrics. That's what would make sense to _me_ as a user, but of course I am open to thoughts and abiding by 1.x logic if thats what makes more sense too! Thoughts.. @palourde @portertech?

mercul3s: @nikkiattea's logic makes sense to me as well. As a user, I'd see a default metric extraction that I haven't configured an endpoint for as a silent failure.

palourde: I believe a sane default (e.g. `graphite`) is better than just giving up; convention over configuration. Nothing prevents us to add a log entry to let the user know that no format was specified so the default one was used, and also add an error in the logs if we couldn't extract it. I'm also concerned about this behaviour change for users migrating from 1.x to 2.x.

nikkiattea: Yeah, I can definitely agree with you on the concern for our users migrating from 1.x. I would be curious under which use case is preferable though. If a user is dependent on checks that they are extracting graphite metrics from, but they have not explicitly set their metric_format, that seems rather unsafe to me. Additionally, the perf impact on _all_ other checks that do not have the intent of outputting graphite metrics, I wish we could protect those checks from being subjected to extraction if they truly don't need it. But yes, I'd love to hear their thoughts on overall UX in this case.. wink wink 😉.

portertech: This attribute is currently also acting as the feature flag (not enabled if an output format is not specified), so we cannot have a default format, unless we are adding another check attribute to enable output extraction.

PR #1424, opened 2018-05-01
Fix bug in topic done chan closed on subscribe

### LINKED ISSUE #1407: Agent check execution appears to miss some checks on a tight interval
Recently, testing has revealed that on tight intervals, such as 1 second, agents can "miss" check executions, even when the check executed is very short.

I suspect that there may be some inefficiency or blocking call in the agent code that executes the check, that is triggering the 'check execution still in progress' error to occur.

Verify if the error is really occurring, and if it is, determine why and fix the problem.

mercul3s:
## What is this change?

Fixes an issue in which agents re-connecting may experience missed or skipped check requests.

Closes #1407 

Added under `Fixed`.

Nope!

This was a really gnarly bug that took the debugging efforts of myself, @cyphus, and @echlebek to deduce.

### DIFF
--- a/backend/messaging/wizard_bus.go
+++ b/backend/messaging/wizard_bus.go
@@ -116,11 +116,18 @@ func (b *WizardBus) Subscribe(topic string, consumer string, sub Subscriber) (Su
 	b.topicsMu.Lock()
 	defer b.topicsMu.Unlock()
 
-	if _, ok := b.topics[topic]; !ok {
-		b.topics[topic] = b.createTopic(topic)
+	t, ok := b.topics[topic]
+	if !ok {
+		t = b.createTopic(topic)
+		b.topics[topic] = t
 	}
 
-	subscription, err := b.topics[topic].Subscribe(consumer, sub)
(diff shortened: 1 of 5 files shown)


backend/messaging/wizard_bus.go
```diff
@@ -116,11 +116,18 @@ func (b *WizardBus) Subscribe(topic string, consumer string, sub Subscriber) (Su
 	b.topicsMu.Lock()
 	defer b.topicsMu.Unlock()
 
-	if _, ok := b.topics[topic]; !ok {
-		b.topics[topic] = b.createTopic(topic)
+	t, ok := b.topics[topic]
+	if !ok {
```
echlebek: just noticed we can refactor a bit here:
```
t, ok := b.topics[topic]
if !ok || t.IsClosed() {
    t = b.createTopic(topic)
    b.topics[topic] = t
}
```

mercul3s: Oh, that cleans up nicely 👍

PR #1425, opened 2018-05-01
Support for Nagios performance data metrics

palourde:
## What is this change?

It adds support for Nagios performance data metrics.

Closes https://github.com/sensu/sensu-go/issues/1385

Added!

Non!

Non!


agent/transformers/nagios.go
```diff
@@ -0,0 +1,86 @@
...
+		parts := strings.Split(metric, ";")
+		parts = strings.Split(parts[0], "=")
+		if len(parts) != 2 {
+			return nil, fmt.Errorf("invalid nagios perfdata metric: %s", metric)
+		}
+
+		// Make sure we don't have any whitespace in our label
+		label := strings.Replace(parts[0], " ", "_", -1)
+
+		// Remove all non-numeric characters from the value
```
nikkiattea: 😍 your comments are 💣

agent/transformers/nagios.go
```diff
@@ -0,0 +1,86 @@
...
+	// Fetch the perfdata and remove leading & trailing whitespaces
+	perfdata := strings.Trim(output[1], " ")
+
+	// Split the perfdata into a slice of metrics
+	metrics := strings.Split(perfdata, " ")
+
+	// Create a Nagios metric for each perfdata metrics
+	for _, metric := range metrics {
+		// Clear everything after ';' and split the label and the value
+		parts := strings.Split(metric, ";")
```
nikkiattea: I can't find anything in the nagios documentation about this so sorry if I'm missing something, but what does `;` specify and why are we ignoring what comes after it?

palourde: To be honest, I have absolutely no idea, I just blindly copied this logic from https://github.com/sensu/sensu-enterprise/blob/057f5af0d4960b70241fdf5181ce604d3afb4262/lib/sensu/enterprise/extensions/metric_parsers.rb#L55 😆. Git tells me Sean wrote that, I'll try to poke him today!

nikkiattea: Haha you're fine! I was honestly just curious because I didn't see a test case for where we would have to concatenate the line before `;` so I didn't know what it indicated!

agent/transformers/nagios.go
palourde: :awthanks:

agent/transformers/nagios.go
palourde: @nikkiattea TIL https://nagios-plugins.org/doc/guidelines.html#AEN200

palourde:
The build failed because of https://github.com/sensu/sensu-go/issues/1208#issuecomment-386663948.

Issue #1402, opened 2018-04-27
[Build stability] Appveyor fatal error: out of memory


nikkiattea:
Failure in Appveyor:
```
[00:10:19] Running e2e tests...
[00:12:37] go : runtime: out of memory: cannot allocate 1073741824-byte block (2223890432 in use)
[00:12:37] At C:\gopath\src\github.com\sensu\sensu-go\build.ps1:214 char:5
[00:12:37] +     go test $REPO_PATH/testing/e2e
[00:12:37] +     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[00:12:37]     + CategoryInfo          : NotSpecified: (runtime: out of...3890432 in use):String) [], RemoteException
[00:12:37]     + FullyQualifiedErrorId : NativeCommandError
[00:12:37]  
[00:12:37] fatal error: out of memory
[00:12:37] 
[00:12:37] 
[00:12:37] runtime stack:
[00:12:37] runtime.throw(0x7e9af0, 0xd)
[00:12:37] 	C:/workdir/go/src/runtime/panic.go:619 +0x6b
[00:12:37] runtime.largeAlloc(0x3ffffe00, 0x128d0101, 0x144cb340)
[00:12:37] 	C:/workdir/go/src/runtime/malloc.go:828 +0xdc
[00:12:37] runtime.mallocgc.func1()
[00:12:37] 	C:/workdir/go/src/runtime/malloc.go:721 +0x39
[00:12:37] runtime.systemstack(0x0)
[00:12:37] 	C:/workdir/go/src/runtime/asm_386.s:464 +0x5e
[00:12:37] runtime.mstart()
[00:12:37] 	C:/workdir/go/src/runtime/proc.go:1170
[00:12:37] 
[00:12:37] goroutine 13 [running]:
[00:12:37] runtime.systemstack_switch()
[00:12:37] 	C:/workdir/go/src/runtime/asm_386.s:418 fp=0x14582e50 sp=0x14582e4c pc=0x44c050
[00:12:37] runtime.mallocgc(0x3ffffe00, 0x7855c0, 0x14582e01, 0xa7e)
[00:12:37] 	C:/workdir/go/src/runtime/malloc.go:720 +0x67b fp=0x14582ea4 sp=0x14582e50 pc=0x40ea5b
[...]

PR #1427, opened 2018-05-01
InfluxDB Transformer Implementation

### LINKED ISSUE #1386: [Metric Extraction] Agent influxdb_line transformer

nikkiattea:

## What is this change?

Creates a transformer for InfluxDB Line protocol metrics. Allows the InfluxDB handler to parse a Sensu metric name for a tag field and measurement.

Closes #1386.

Yas.

Nah.

Nah.


agent/transformers/influx_db.go
```diff
@@ -0,0 +1,126 @@
...
+			}
+			fieldList = append(fieldList, field)
+		}
+		i.FieldSet = fieldList
+
+		t, err := strconv.ParseInt(args[2], 10, 64)
+		if err != nil {
+			return InfluxList{}, errors.New("metric timestamp is invalid, third argument must be an int")
+		}
+		i.Timestamp = t
```
palourde: We should probably convert the influx timestamp, which uses nanosecond-precision, to a more standard 10 characters unix timestamp, so handlers know exactly what to expect. It seems we also did this in 1.x extensions: https://github.com/sensu/sensu-enterprise/blob/057f5af0d4960b70241fdf5181ce604d3afb4262/lib/sensu/enterprise/extensions/metric_parsers.rb#L143

agent/transformers/influx_db.go
```diff
@@ -0,0 +1,126 @@
...
+// Influx contains values of influx db line metric format
+type Influx struct {
+	Measurement string
+	TagSet      []Tag
+	FieldSet    []Field
+	Timestamp   int64
+}
+
+// Tag is a key value pair representing an influxdb tag set
+type Tag struct {
```
palourde: We should probably call that `InfluxTag` or move that into a `transformers.go` file so it can shared across the transformers!

agent/transformers/influx_db.go
```diff
@@ -0,0 +1,126 @@
...
+	Key   string
+	Value string
+}
+
+// Field is a key value pair representing an influxdb field set
+type Field struct {
```
palourde: Same as `Tag`

agent/transformers/influx_db.go
palourde: I'm actually wondering if we couldn't use `[]*types.MetricTag{}` in `Influx` and simply avoid this struct? Thoughts?

nikkiattea: Yeah that probably eliminates some complexity if we use `MetricTag` instead!

agent/transformers/influx_db.go
nikkiattea: Yeah, we could probably move this to a `transformers.go` file, I like that idea.

nikkiattea:
Weird travis error for that failure, doesn't look like the cache is related here.
```
(98/98) Failed to write github.com/docker/docker@cbde00b44273a584bed2ee3513db68924b4a6dba
grouped write of manifest, lock and vendor: error while writing out vendor tree: failed to write dep tree: failed to export golang.org/x/time: https://go.googlesource.com/time does not exist in the local cache and fetching failed: unable to get repository: Cloning into '/home/travis/gopath/pkg/dep/sources/https---go.googlesource.com-time'...
POST git-upload-pack (141 bytes)
error: RPC failed; HTTP 502 curl 22 The requested URL returned error: 502 Bad Gateway
fatal: The remote end hung up unexpectedly
: command failed: [git clone --recursive -v --progress https://go.googlesource.com/time /home/travis/gopath/pkg/dep/sources/https---go.googlesource.com-time]: exit status 128
```

palourde:
@nikkiattea I canceled the AppVeyor because it was stuck on https://github.com/sensu/sensu-go/issues/1402. LGTM.

PR #1429, opened 2018-05-02
Support for OpenTSDB metrics

palourde:
## What is this change?

It adds support for OpenTSDB metrics.

Closes https://github.com/sensu/sensu-go/issues/1387

Added!

Non!

Non!


agent/transformers/opentsdb.go
```diff
@@ -0,0 +1,100 @@
...
+	// Split each line of the output into its own metric
+	metrics := strings.Split(output, "\n")
+
+	for _, metric := range metrics {
+		parts := strings.Split(metric, " ")
+
+		// Ensure we have all the required components. A single metric requires a
+		// name, timestamp, value and at least one tag.
+		if len(parts) < 4 {
+			return nil, fmt.Errorf("invalid opentsdb metric: %s", metric)
```
nikkiattea: It might be helpful to have a more descriptive error message here, pertaining to minimum # of fields required of an opentsdb metric.

agent/transformers/opentsdb_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+func TestParseAndTransformOpenTSDB(t *testing.T) {
+	testCases := []struct {
+		name    string
+		output  string
+		want    []*types.MetricPoint
+		wantErr bool
+	}{
+		{
+			name:   "happy path",
```
nikkiattea: I like that we're testing the happy path here! It might be useful to have at least one test case with `wantErr: true`.

agent/transformers/opentsdb.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+
+		name := parts[0]
+
+		// Convert the timestamp to a unix timestamp with second resolution
+		timestamp, err := strconv.ParseInt(parts[1], 10, 64)
+		if err != nil {
+			return nil, fmt.Errorf("invalid opentsdb metric timestamp, must be an integer: %s", parts[1])
+		}
+		if len(parts[1]) == 13 {
```
nikkiattea: We're diluting the timestamp resolution here right? Can you explain the logic a little? We want to convert the timestamp to a 10 digit timestamp correct?

palourde: Exactly; opentsdb specs allow both timestamp with second and millisecond precision, so in order to standardize the format, we convert to second resolution if required. That way, handlers know exactly what to expect. Makes sense?

agent/transformers/opentsdb_test.go
palourde: Done!

agent/transformers/opentsdb.go
palourde: Done!

PR #1431, opened 2018-05-02
Don't add namespace flags for create command.

### LINKED ISSUE #1233: Remove organization and environment flags from sensuctl create.
`sensuctl create` has organization and environment flags by virtue of the CLI framework. However, those flags are ineffectual in the tool. They should be removed from it.

We may need to modify the framework to achieve this.

echlebek:
This commit removes the namespace flags (--organization and
--environment) from the global flags, and instead uses local flags
to implement those features.

This was done so that we could avoid having these flags for the
create command, which does not use them and should not.

Closes #1233 

It was difficult to decide how to do this. I would have preferred to leave the namespace and organization flags as global flags, but there was no way to do that and also drop those flags from the create sub-command. I settled on adding them as local flags to every command with a for loop.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,7 @@ and yarn are now dependencies for building the backend.
 - Improved log entries produced by pipelined.
 - Allow the InfluxDB handler to parse the Sensu metric for an InfluxDB field tag
 and measurement.
+- Removed organization and environment flags from create command.
 
 ### Fixed
 - Terminate processes gracefully in e2e tests, allowing ports to be reused.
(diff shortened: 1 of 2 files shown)


echlebek:
Unfortunately this _completely doesn't work_.

The `SensuCli` type parses its config from flags on the root command only, and we use that config throughout the CLI in many many places.

Since we use this config object instead of the flags directly, it would take a complete rework of the CLI framework to achieve the goal of the PR.

tl;dr: we can't move the organization and environment flags out of the root command without a massive refactor.

I propose that we simply give up on this issue and not do it. It's probably not worth the churn.

mercul3s:
@echlebek well that's a bummer. I agree, it's not worth delving into that work right now, but let's keep the issue open in case we decide to do some cli refactoring later.

echlebek:
@palourde was able to find a workaround for this! Thanks Simon.

palourde:
\o/

PR #1439, opened 2018-05-03
Instantiate empty metrics in the check event

nikkiattea:

## What is this change?

Instantiates an empty Metrics struct inside the check handlers event.

Fixes a nil pointer error.

Nah.

Nah.

No, I think this was the result of a bad merge.

### DIFF
--- a/agent/check_handler.go
+++ b/agent/check_handler.go
@@ -65,7 +65,8 @@ func (a *Agent) executeCheck(request *types.CheckRequest) {
 	check.Executed = time.Now().Unix()
 	check.Issued = request.Issued
 	event := &types.Event{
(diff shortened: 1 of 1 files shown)


agent/check_handler.go
```diff
@@ -65,7 +65,8 @@ func (a *Agent) executeCheck(request *types.CheckRequest) {
 	check.Executed = time.Now().Unix()
 	check.Issued = request.Issued
 	event := &types.Event{
-		Check: check,
+		Check:   check,
+		Metrics: &types.Metrics{},
```
palourde: Let's add a quick test in `TestExecuteCheck`!

Issue #841, opened 2018-01-08
[Web UI] Entity details page: Display configuration


jamesdphillips:
## 🕸 Related

Sketch: https://github.com/sensu/sensu-go/issues/816

PR #1441, opened 2018-05-03
[GraphQL] Fixup issues with resolving entity fields

jamesdphillips:
## What is this change?

- Ensure entity fields are non-nullable
- Ensure `addresses` field is represented as list
- Remove extraneous isTypeOf impl on Entity's descendants
- Ensure `os` field can be resolved
- Ensure `mac` field can be resolved

Unblocks #841 

`null`

`undefined`

`undefined`


jamesdphillips:
> Too bad we can't select extended attributes, I wonder if there's a way we could implement a shim for that.

Totes.

The problem isn't so much implementing a shim on the server, that's easy enough, it is more that any addition/shim/hack we did on the server would be need to be replicated on each every client that wanted to consume our service.

GraphQL is pretty similar to protocol buffers in this regard; there is no built-in map or "any" type. However, like protocolbuffers we could define our own dictionary... of sorts...

```graphql
union All = String | Int | Float | Check | Entity ...

object KVPair {
  key: String!
  val: All!
}

object Dictionary {
  "Retrieve value associated with the given key"
  get(key: String!): All

  "Retrieve all pairs associated with the given keys" 
  pick(keys: [String!]!): [KVPair!]!

  "Retrieve all pairs except those associated with the given keys"
  omit(keys: [String!]!): [KVPair!]!

  "Return all pairs in dictionary"
  pairs: [KVPair!]!
}

object MyWonderfulRecord {
  meta: Dictionary!
}
```

...anyways...

As far as things go at the moment, we simply need to expose the extended attributes associated with an entity, so I'm shying away from going too far down that rabbit hole until we _really_ need to.

Issue #1389, opened 2018-04-25
[Metric Extraction] e2e test for metric extraction


nikkiattea:
Child of https://github.com/sensu/sensu-go/issues/745
Depends on #1384 

Write an e2e test for check output metric extraction. Create a check that simply prints a metric (according to a specific line protocol), and configure the check’s metric_format and metric_handlers. Confirm that the check output populates an event in Sensu Metric Format.

PR #1446, opened 2018-05-04
Add e2e test for metric extraction

palourde:

## What is this change?

It adds an e2e test for metric extraction functionality.

Closes https://github.com/sensu/sensu-go/issues/1389

Added!

Non!

Non!

### DIFF
--- a/testing/e2e/metric_extraction_test.go
+++ b/testing/e2e/metric_extraction_test.go
@@ -1,0 +1,56 @@
+package e2e
+
+import (
+	"encoding/json"
+	"testing"
+	"time"
+
+	"github.com/sensu/sensu-go/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestMetricExtraction(t *testing.T) {
+	t.Parallel()
+
+	// Start the backend
+	backend, cleanup := newBackend(t)
+	defer cleanup()
+
+	// Initializes sensuctl
+	sensuctl, cleanup := newSensuCtl(backend.HTTPURL, "default", "default", "admin", "P@ssw0rd!")
+	defer cleanup()
+
+	// Start the agent
+	agentConfig := agentConfig{
+		ID:          "TestMetricExtraction",
+		BackendURLs: []string{backend.WSURL},
+	}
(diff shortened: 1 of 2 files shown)


nikkiattea:
😢 
```
--- FAIL: TestMetricExtraction (18.33s)
	assertions.go:239: 
                          
	Error Trace:	metric_extraction_test.go:53
		
	Error:      	Should not be zero, but was 0
panic: runtime error: index out of range [recovered]
panic: runtime error: index out of range
```

testing/e2e/metric_extraction_test.go
```diff
@@ -0,0 +1,56 @@
...
+		"--command", "echo 'PING ok - Packet loss = 0% | percent_packet_loss=0'",
+		"--metric-format", "nagios_perfdata",
+	)
+	require.NoError(t, err, string(out))
+
+	// FIXME: Give it few seconds to make sure we're not publishing check requests.
+	time.Sleep(15 * time.Second)
+
+	// There should be a stored event for our metric
+	out, err = sensuctl.run("event", "info", agent.ID, "nagios-metric")
```
nikkiattea: I think your agent id is `TestMetricExtraction` not `nagios-metric`, hence no metric points showing up in the failed test.

testing/e2e/metric_extraction_test.go
```diff
@@ -0,0 +1,56 @@
...
+	agent, cleanup := newAgent(agentConfig, sensuctl, t)
+	defer cleanup()
+
+	// Create a check that publish check requests
+	out, err := sensuctl.run(
+		"check", "create", "nagios-metric",
+		"--publish",
+		"--interval", "1",
+		"--subscriptions", "test",
+		"--command", "echo 'PING ok - Packet loss = 0% | percent_packet_loss=0'",
```
nikkiattea: This is the failure on windows only: `echo`

nikkiattea: We could create a go program in tools https://github.com/sensu/sensu-go/tree/master/tools that will allow us to write a string to stdout. That way, the check will be executed properly on both platforms

palourde: Ah, good catch!

PR #1450, opened 2018-05-04
Attempt to solve the problem with nightly failing.

echlebek:

## What is this change?

Trying to get the nightly working again.


echlebek:
Not needed since #1539 changes how all of this works.

PR #1455, opened 2018-05-07
Improve version handling for nightly and dev builds

### LINKED ISSUE #1320: Add version number to version output
Include the version number in the version output.

## Expected Behavior
Expect that the version number would be included.

## Current Behavior

```
# sensu-agent version
sensu-agent version dev-nightly#d46cb6d, build d46cb6df3d2750a1a7b60fb27293c3bc14c7d6b8, built 2018-04-04T18:18:19+0000

# sensu-backend version
sensu-backend version dev-nightly#d46cb6d, build d46cb6df3d2750a1a7b60fb27293c3bc14c7d6b8, built 2018-04-04T18:17:16+0000
```

Running a nightly build of 2.0.0, though the version output does not include that. Would have to look at the packages themselves, which defeats the purpose of querying the software to output the version.

```
# rpm -qa |grep -i sensu
sensu-agent-2.0.0~nightly+20180404-1.x86_64
sensu-cli-2.0.0~nightly+20180404-1.x86_64
sensu-backend-2.0.0~nightly+20180404-1.x86_64
```

## Possible Solution
Include the number. Do so toward the beginning of the string and ensure it is in the same place, so we know it's always the fifth field or whatever. This makes it easy for other tools to grab that info without ugly regex's.

cyphus:
## What is this change?

Improve version handling for nightly and dev builds.

More specifically, it:
- Adds the current version to the output of `sensu-agent version` and `sensu-backend version` (see #1320 )
- Adds CI detection to properly detect dev builds, along with making dev an official build type.
- Improves version detection by directly querying git for the latest tag belonging to this commit.
- Removes the spurious 'build type not supported for prerelease: "nightly"' errors for local builds

Closes #1320


cyphus:
Let's not worry about this PR until after QA session.

CHANGELOG.md
```diff
@@ -104,6 +104,8 @@ swallowed. The events are sent through the pipeline.
 - Include zero-valued integers in JSON output for all types.
 - Check event entities now have a last_seen timestamp.
 - Improved silenced entry display and UX.
+- Add version output for dev and nightly builds (#1320).
+- Improve git version detection by directly querying for the most recent tag.
```
palourde: Since we just made a release, maybe try to merge master back into your branch, just to make sure these lines don't get appended to an already released version!

cyphus: Agreed, I've rebased off of master and updated the changelog.

cyphus:
Travis failed with one error during `dep ensure`, and one error in the silencing tests (#1208). Both unrelated to these changes.

version/version.go
```diff
@@ -60,53 +65,39 @@ func Semver() string {
...
+// BuildEnv provides methods for determining version info from the current env.
+type BuildEnv interface {
+	IsCI() bool
+	IsNightly() bool
+	GetMostRecentTag() string
+}
+
+// ParseBuildEnv discovers the most recent tag and BuildType using the current
+// build environment
+func ParseBuildEnv(env BuildEnv) (string, BuildType) {
```
echlebek: This is kind of the opposite of parsing, you're taking a data structure and returning string data.

maybe `FindBuildType` or something?

version/cmd/version/version.go
```diff
@@ -56,34 +40,54 @@ func main() {
...
+// buildType getter func that matches the signature of other version funcs
+func buildType(_ string, bt version.BuildType) (string, error) {
+	return string(bt), nil
+}
+
+// implements the version.BuildEnv interface using the real build environment
+type BuildEnv struct{}
+
+// Returns true if we are building from a CI (rather than a local or dev build)
+func (b *BuildEnv) IsCI() bool {
```
echlebek: You don't actually need a pointer here, or even a receiver name, since these methods only exist to satisfy an interface.

You could have
```
func (BuildEnv) IsCI() bool {}
```
for instance.

version/cmd/version/version.go
```diff
@@ -56,34 +40,54 @@ func main() {
...
+// Returns the most recent tag belonging to the current commit
+func (b *BuildEnv) GetMostRecentTag() string {
+	// --abbrev=0 disables tag annotation, returning unmodified most recent tag
+	cmd := exec.Command("git", "describe", "--abbrev=0", "--tags", "HEAD")
 	out, err := cmd.Output()
 	if err != nil {
-		if _, ok := err.(*exec.ExitError); !ok {
-			return nil, err
-		}
+		log.Fatal(err)
```
echlebek: `log.Fatal` is generally best used only in main. The function is a log.Println followed by os.Exit.

That means deferred functions will never be run, and it may be hard to determine the source of the error.

The error should be part of the method and passed through.

version/cmd/version/version.go
```diff
@@ -56,34 +40,54 @@ func main() {
...
+	return os.Getenv("CI") == "true"
+}
+
+// Returns true if this is a nightly release by checking whether the current
+// HEAD is one or more commits ahead of the latest tag.
+func (b *BuildEnv) IsNightly() bool {
 	cmd := exec.Command("git", "describe", "--exact-match", "--tags", "HEAD")
-	out, err := cmd.Output()
-	if err != nil {
-		if _, ok := err.(*exec.ExitError); !ok {
```
echlebek: The reason this distinction exists is to separate non-zero exit status from all other errors that can occur while executing a command. (OOM for instance.)

In these circumstances, the error should be propagated up to the caller.

cyphus: The modified version of this function still distinguishes between exit error and other errors as you describe. It however does not propagate errors to the caller, which will be fixed shortly.

version/cmd/version/version.go
cyphus: These functions define a struct that satisfies the `BuildEnv` interface for use by main, and only by main, which is defined in the same file. Therefore I've considered them to be conceptually part of main (from an abstraction layer perspective) even if technically they are separate.

I was not aware of the implications of using `log.Fatal` exiting without any possibility of cleanup though. I'll update this propagate errors to callers instead.

version/version.go
cyphus: There is some string processing in the function, but overall you're right it's not really a parser. Will update naming to match.

PR #1458, opened 2018-05-07
Trim whitespace from check output in opentsdb transformer

nikkiki:

## What is this change?

Fixed a small bug in the opentsdb transformer so that it trims trailing whitespace characters.

Check output containing one or multiple metrics in opentsdb metric format can also contain trailing whitespace or newline characters. We should not attempt to parse the newline character for a metric, rather we should ignore it.

Yas.

Nah.

Nah.


nikkiki:
@palourde yes I added test cases for all transformers with trailing whitespaces. I believe the nagios transformer is unaffected since we do not attempt to parse multiple lines of it.

PR #1474, opened 2018-05-07
[Web UI] Fixup resolve event

jamesdphillips:
## What is this change?

Fixup resolve event.

Whoops.


10xjs:
🤦‍♂️

Issue #1498, opened 2018-05-10
Statsd is broken on windows


nikkiki:
TCP is not a supported network by `net.ListenPacket` which we attempt to force upon due to https://github.com/sensu/sensu-go/issues/1402: https://github.com/sensu/sensu-go/commit/0895d84bb486b0cc2a6560f2fa6077d236fc070a.

Since we don't attempt to catch the error here, the statsd server silently fails: https://github.com/sensu/sensu-go/blob/0895d84bb486b0cc2a6560f2fa6077d236fc070a/agent/agent.go#L391

We'll need to investigate another alternative to the many issues we've experienced with statsd and windows, and re-enable statsd tests on windows to confirm.

nikkiki:
Filed https://github.com/atlassian/gostatsd/issues/178 to address Windows UDP support in [gostatsd](https://github.com/atlassian/gostatsd).

nikkiki:
https://github.com/atlassian/gostatsd/pull/179

PR #1499, opened 2018-05-10
Log statsd errors

### LINKED ISSUE #1473: Setting --statsd-metrics-host to an invalid address does not cause an error
When an invalid address is set, the agent starts up and does not report any error.

## Expected Behavior
The agent reports an error, or even crashes, when an invalid statsd listen address is specified.

## Current Behavior
The agent starts up and does not report an error.

1. `./sensu-agent start --statsd-metrics-host 192.168.5.5`

nikkiki:
## What is this change?

Sensu-agent logs an error if the statsd listener is unable to start due to an invalid address or is stopped due to any other error.

Closes #1473.

Yas.

Nah.

Found https://github.com/sensu/sensu-go/issues/1498 and working on it next.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -481,46 +481,32 @@ func (a *Agent) addHandler(msgType string, handlerFunc handler.MessageHandlerFun
 // StartStatsd starts up a StatsD listener on the agent, logs an error for any
 // failures.
 func (a *Agent) StartStatsd() {
-	statsdErr := make(chan error, 1)
+	var runStatsd func() error
 	logger.Info("starting statsd server on address: ", a.statsdServer.MetricsAddr)
 
 	// We need to force a TCP connection for Windows. See
 	// https://github.com/sensu/sensu-go/issues/1402
 	if runtime.GOOS == "windows" {
-		// TODO: https://github.com/sensu/sensu-go/issues/1498
-		conn, err := net.ListenPacket("tcp", a.statsdServer.MetricsAddr)
-		if err != nil {
-			logger.WithError(err).Errorf("error with statsd server on address: %s, statsd listener will not run", a.statsdServer.MetricsAddr)
-			return
-		}
-		socketFactory := func() (net.PacketConn, error) {
-			return conn, err
-		}
-
-		go func() {
-			defer close(statsdErr)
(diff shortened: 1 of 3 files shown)


agent/agent.go
```diff
@@ -489,3 +477,50 @@ func (a *Agent) Stop() {
 func (a *Agent) addHandler(msgType string, handlerFunc handler.MessageHandlerFunc) {
 	a.handler.AddHandler(msgType, handlerFunc)
 }
+
+// StartStatsd starts up a StatsD listener on the agent, logs an error for any
+// failures.
+func (a *Agent) StartStatsd() {
```
echlebek: So I think this function is a bit more complex than it needs to be.

Consider:
```
var runStatsd func() error
if runtime.GOOS == "windows" {
    runStatsd = func() error {
        // windows specific stuff
    }
} else {
    runStatsd = func() error {
        // linux specific stuff
    }
}
if err := runStatsd(); err != nil {
    logger.WithError(err).Errorf("error running statsd on %s", addr)
}
```

There's also no need to handle errors with channels here, because the channels are not connecting different goroutines; the chan send and recv are in the same goroutine, so you can just use regular variable assignment and avoid the select.

Finally, keep in mind than an error can occur after a span of time has passed, it might not be from starting the server but another type of error. (Just for the sake of the error message being truthy)

nikkiki: @echlebek Yes, I did try to simplify this, but the caveat here is that `a.statsdServer.Run(a.context)` returns an error and must be run in a goroutine. From the research I did (ex. https://www.atatus.com/blog/goroutines-error-handling/) it seemed as though retrieving an error from a goroutine using channels was the safest way.

> Whether can we directly return values from goroutines similar to a normal function? This is a big no-no in the world of concurrency.

But if this is just referring to assigning variables between _different_ goroutines, then I can definitely go ahead and simplify this, since the two goroutines are _not_ connected in this case!

echlebek: Yeah, you definitely need to start up a new goroutine to handle this stuff, just don't need to use select mechanisms and channels here. :)

nikkiki: @echlebek it is done. 
giphy

PR #1510, opened 2018-05-11
add output to grpc handlers & add additional handler logging

### LINKED ISSUE #1395: Handlers do not return output or status
In Sensu 1.x, handlers returned string output and a status code. Our 2.x handlers only return an error value.

Modify pipelined and the extension RPC service so that handlers produce string values and status codes.

amdprophet:

## What is this change?

Adds an `Output` field to GRPC handlers.

Fixes #1395.

Yes.

I ended up not adding a `Status` field to GRPC handlers as I'm not sure it makes sense. GRPC handlers can provide richer information on failures via the `Error` field.

Ran into issues when running the E2E tests; @echlebek was able to diagnose the issue.


backend/pipelined/handle.go
```diff
@@ -83,7 +84,7 @@ func (p *Pipelined) handleEvent(event *types.Event) error {
 				logger.WithFields(fields).Error(err)
 			}
 		case "grpc":
-			if err := p.grpcHandler(u.Extension, event, eventData); err != nil {
+			if _, err := p.grpcHandler(u.Extension, event, eventData); err != nil {
```
echlebek: Why return the response if we aren't going to use it here?

PR #1530, opened 2018-05-14
Add vendor directory & remove dep ensure from build scripts

amdprophet:

## What is this change?

Re-adds the vendor directory.

It is needed for reproducible builds, it will cut down build times & reduce the failures due to external dep sources failing. It will not fix the failures we see during `go get`.

Probably not.

Nope!

Nope!


nikkiki:
@amdprophet do we also want to remove `Gopkg.toml` and `Gopkg.lock` or are we still maintaining dep?

amdprophet:
@nikkiki we're still using dep. If you need to bump any versions, edit `Gopkg.toml` and then run `dep ensure`. Commit any changes to `Gopkg.toml`, `Gopkg.lock`, and anything within the `vendor` directory.

PR #1532, opened 2018-05-14
Install yarn during travis before_install step

### LINKED ISSUE #1531: CI deploy step resolves to unknown yarn version
## Current Behavior
The `deploy` travis CI build stage does not have a declared dependency on a specific version of yarn, and resolves to whatever may be provided by the travis env.

This is a suspected cause of intermittent yarn related build failures.

## Expected Behavior
The CI build should install and use a defined version of yarn.

## Possible Solution
`run npm install --global yarn@VERSION` before any CI tasks that depend on yarn.

A strict yarn dependency declared under `engines` in `dashboard/package.json` causes the build process to fail if the required version is not present, but doesn't trigger anything to install that yarn version.

10xjs:
## What is this change?
Install a deterministic version of `yarn` during CI.

closes #1531

nope

nope

nope

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,8 @@ before_install:
   - ulimit -s 1082768
   - curl -L -s https://github.com/golang/dep/releases/download/v0.4.1/dep-linux-amd64 -o $GOPATH/bin/dep
   - chmod +x $GOPATH/bin/dep
+  - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.6.0
+  - export PATH="$HOME/.yarn/bin:$PATH"
 install:
 - "git fetch --tags"
 - "./build.sh deps"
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -22,6 +22,8 @@ before_install:
   - ulimit -s 1082768
   - curl -L -s https://github.com/golang/dep/releases/download/v0.4.1/dep-linux-amd64 -o $GOPATH/bin/dep
   - chmod +x $GOPATH/bin/dep
```
jamesdphillips: tagential but... @amdprophet can this be removed now?

amdprophet: I've left it for now and was planning to see if it can do a check to ensure `Gopkg.toml/lock` are in sync with the vendor directory as part of the CI process. You can remove it for now though.

jamesdphillips: Makes sense 👍

PR #1539, opened 2018-05-15
[Web UI] Define client schema extensions

jamesdphillips:
## What is this change?

🏗 Queries referencing client additions to the schema are now properly validated!

**summary**

* Adds `src/apollo/schema/client.graphql` with client specific directions, types, and extensions. 
* Rewrote `update-schema` script to combine client and server schemas. 
* Adds `.graphqlconfig` which points to combined schema.
* Re-enables `FieldsOnCorrectType` and `KnownDirectives` linter rules.
* Unlike introspection results, combined schema is committed to repo, and as such `update-schema` no longer is run in `precompile`, `prestart`, `postinstall` stages. This, to avoid unintended side-effects when simply installing or starting the dev-server. `update-schema` becomes an explicit task.
* By removing the precompile step this likely supersedes #1450. 

**drawbacks**

Updates to server's or client's schemas now require engineer to run `yarn update-schema` if they would like the changes to be reflected in the front-end tooling. As the schema doesn't change overly often I think this is an acceptable trade-off. 

unlikely

nada

1. I could not get webpack to consume the latest version of `graphql-js`; spent far too long fighting with this before simply rolling back to `v0.13.0`. [Nothing of value is lost](https://github.com/graphql/graphql-js/releases). 
2. ~~Tried using the `macroLoader` for pumping a minified copy of the schema. While minified schema is less than 4 kB, the approach definitely isn't quite complete or correct...~~ All good now. Thanks Neal.


dashboard/config/macroLoader.js
```diff
@@ -6,9 +6,18 @@ export default function(content) {
...
+      //   outputPath,
+      // )};
+
+      // ¯\(°_o)/¯
+      // ¯\(°_o)/¯
+      // ¯\(°_o)/¯
+      // ¯\(°_o)/¯
+      // ¯\(°_o)/¯
+      // ¯\(°_o)/¯
+      // ¯\(°_o)/¯
```
oferrigni: One of my life lessons is "Be funny out loud, don't be funny in code". Largely because once its in SCM, its suuper hard to take back.

jamesdphillips: Haha, this is definitely very much WIP and not something I intend to find it's way into master. :)

dashboard/src/apollo/schema/combinedTypes.macro.js
```diff
@@ -0,0 +1,39 @@
...
+  );
+  const result = execute(schema, queryAST);
+  if (result.errors) throw result.errors;
+
+  // https://github.com/apollographql/apollo-client/blob/2701b0acb89711864bde28341cb5cfcf909d2caf/packages/apollo-cache-inmemory/src/fragmentMatcher.ts#L149-L155
+  const filteredTypes = result.data.__schema.types.filter(
+    type => type.kind === "UNION" || type.kind === "INTERFACE",
+  );
+  result.data.__schema.types = filteredTypes;
+  return emitFile("combinedTypes.json", JSON.stringify(result));
```
10xjs: You don't need to call `emitFile` at all here. You can simply return the module contents.

10xjs:
@jamesdphillips I'm encountering an issue running this locally. The syntax here 
https://github.com/sensu/sensu-go/blob/5999049927a8b70c71156eed888552cb60456351/backend/apid/graphql/schema/silenced.graphql#L4 is causing a parse error with `graphql/buildSchema`.

Apparently it needs to be an ampersand instead of a comma (https://stackoverflow.com/questions/49521575/implement-multiple-interfaces-in-graphql/49521662#49521662) 🤔

10xjs:
With running `build` in the `prepare` script, we could also remove https://github.com/sensu/sensu-go/blob/4f0cbfc629caeafd3dcc95b27e4eb619cc346eaf/dashboard/assets_generate.go#L58-L59

jamesdphillips:
Added `yarn build` to `"test"` but as `update-schema` has the side-effect of updating the combined schema I'd rather keep that out of test / prepare and instead be an explicit action.

Issue #1448, opened 2018-05-04
Timezones are not properly handled when creating a silenced entry


palourde:
## Expected Behavior
When specifying a beginning date with a specific timezone, I expect this timezone to be used.

## Current Behavior
The timezone is ignored and we don't throw any error. In this example, I expect the entry to start at `May 04 2018 5:30PM EST` but when displaying it, I have `04 May 18 23:30 EDT` instead!

```
$ date
Fri May  4 17:18:58 EDT 2018
$ sensuctl silenced create \
--subscription 'entity:i-424242' \
--begin 'May 04 2018 5:30PM EST' \
--expire 3600 \
--reason 'Testing'
OK
$ sensuctl silenced list
         ID            Subscription     Check          Begin          Expire   ExpireOnResolve   Creator   Reason    Organization   Environment
 ─────────────────── ───────────────── ─────── ───────────────────── ──────── ───────────────── ───────── ───────── ────────────── ─────────────
  entity:i-424242:*   entity:i-424242           04 May 18 23:30 EDT   59m59s   false             admin     Testing   default        default
```

## Steps to Reproduce (for bugs)

1. Create a silencing entry, with a date in the future:
```
sensuctl silenced create \
--subscription 'entity:i-424242' \
--begin 'May 04 2018 5:30PM EST' \
--expire 3600 \
--reason 'Testing
```
2. List the silenced entries:
```
sensuctl silenced list
```
3.
4.

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): `master`
* Installation method (packages, binaries, docker etc.): manually
* Operating System and version (e.g. Ubuntu 14.04): macOS 10.13.4

jamesdphillips:
What is the timestamp on the entry? In other words, is the issue on create or when we format the date?

palourde:
@jamesdphillips It's stored with the wrong date. For example, using `--begin 'May 04 2018 5:30PM EST'`, it gives the following:

```
$ sensuctl silenced list --format json
[
  {
    "id": "entity:i-424242:*",
    "expire": 24310,
    "creator": "admin",
    "reason": "Testing",
    "subscription": "entity:i-424242",
    "organization": "default",
    "environment": "default",
    "begin": 1525491000
  }
]
```

Where `1525491000` is:
[...]

PR #1545, opened 2018-05-15
Add support for additional time formats and better handle timezones

palourde:
## What is this change?

It adds support for additional time format. So we now support these formats:

- 12-hour kitchen (zone): `3:04PM UTC`
- 24-hour kitchen (zone & offset): `15:04 UTC` or `07:04 -08:00`
- Sensu alpha legacy format (zone): `May 10 2018 3:04PM UTC`
- RFC3339 (offset): `2018-05-10T15:04:00Z`
- RFC3339 with space delimiters (offset): `2018-05-10 07:04:00 -08:00` 

Closes https://github.com/sensu/sensu-go/issues/1448

Added!

Nope!

The `time` pkg is hard.

### DIFF
--- a/subdue.json
+++ b/subdue.json
@@ -1,1 +1,0 @@
-{"days":{"all":[{"begin":"12:00 AM","end":"11:59 PM"},{"begin":"11:00 PM","end":"1:00 AM"}]}}
--- a/subdue.json
+++ b/subdue.json
@@ -1,0 +1,1 @@
+{"days":{"all":[{"begin":"12:00 AM","end":"11:59 PM"},{"begin":"11:00 PM","end":"1:00 AM"}]}}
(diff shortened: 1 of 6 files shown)


echlebek:
Seems that something is going wrong on Windows though. :thinking:

palourde:
@echlebek Yeah, TIL https://github.com/golang/go/blob/dbd66fd3d07433a03da5dc6c6def8c53951ada77/src/time/zoneinfo_windows.go#L20-L21

subdue.json
```diff
@@ -0,0 +1 @@
+{"days":{"all":[{"begin":"12:00 AM","end":"11:59 PM"},{"begin":"11:00 PM","end":"1:00 AM"}]}}
```
jamesdphillips: Should this be here?

PR #1546, opened 2018-05-15
Entity/create cli

### LINKED ISSUE #1461: sensuctl cannot update entity
## Expected Behavior
Able to update entities subscriptions through `sensuctl entity update [entity-id]`

## Current Behavior
Command exits early after prompting for new subscription list:

```
$ bin/sensuctl entity update t-sensu
? Subscriptions: entity:t-sensu
Error: not found
```

## Possible Solution
Looks like we're missing a route for updating entities.

## Context
Discovered while running through release QA checklist.

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): commit e91bfb3
* Installation method (packages, binaries, docker etc.): local build
* Operating System and version (e.g. Ubuntu 14.04): macOS 10.13.4

### LINKED ISSUE #1479: sensuctl entity create
Allow for the creation of custom entities with sensuctl entity create and also sensuctl create -f

mercul3s:
## What is this change?

Adds entity creation and update support to the CLI.

Closes #1479 and closes #1461 

Added.

Our docs state that an entity `class` should be either `proxy` or `agent`, yet it isn't restricted as entity uses the ValidateName method on class. This means that a user can set `class` to be values other than `proxy` or `agent` through the CLI and API. I think this might be ok, but wanted to call it out in case we should restrict it to just those two types.

I updated `opts` to `entityOpts` to be more consistent with the naming in the cli package.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -40,8 +40,6 @@ install:
   - go version
   - go env
   - mkdir %GOPATH%\bin
-  - appveyor DownloadFile https://github.com/golang/dep/releases/download/v0.4.1/dep-windows-amd64.exe
-  - move dep-windows-amd64.exe %GOPATH%\bin\dep.exe
   - ps: .\build.ps1 deps
   - ps: .\build.ps1 build_tools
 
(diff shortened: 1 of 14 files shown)


mercul3s:
I still need to test create -f against an entity json definition, will do this tomorrow.

palourde:
Let's make sure the CI is passing before merging 😨

PR #1549, opened 2018-05-16
Fix the URIPath method of Event.

### LINKED ISSUE #1548: Can't create events with sensuctl create
While refactoring the e2e tests, I also removed the deprecated HTTP client in favour of sensuctl commands.

This exposed an issue with creating events via `sensuctl create`. The URIPath method is wrong, and is missing the `/events` prefix.

echlebek:
This allows events to be created with sensuctl create.

Closes #1548

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ commands
 - Fix a bug where environments could not be created with sensuctl create
 - StatsD listener on Windows is functional
 - Fixed `sensuctl create -f` for `Role`
+- Fixed `sensuctl create -f` for `Event`
 
 ## [2.0.0-beta.1] - 2018-05-07
 ### Added
--- a/types/event.go
+++ b/types/event.go
@@ -240,5 +240,5 @@ func (e *Event) URIPath() string {
 	if !e.HasCheck() {
 		return ""
 	}
-	return fmt.Sprintf("/%s/%s", url.PathEscape(e.Entity.ID), url.PathEscape(e.Check.Name))
+	return fmt.Sprintf("/events/%s/%s", url.PathEscape(e.Entity.ID), url.PathEscape(e.Check.Name))
 }


palourde:
Should we also create a `sensuctl event create` subcommand?

palourde:
@echlebek Maybe I'm missing something but why would we need to allow events creation via `sensuctl create` then? The only thing I could see is the ability to restore from a backup but I'm not even sure this is something relevant nor desired.

echlebek:
@palourde because we want to treat `Event` as a resource, and because it allows us to deprecate the `Client` type in the e2e tests in favour of `sensuctl`.

I think it was either @portertech or @grepory who told me we don't want to support creating events via sensuctl commands. But I think allowing creation via `create -f` does make sense.

palourde:
> it allows us to deprecate the Client type in the e2e tests in favour of sensuctl.

AFAIK, the only reason why we still have `sensuClient` is simply because we couldn't update a check without using the interactive mode, which is no longer the case.

> we don't want to support creating events via sensuctl commands. But I think allowing creation via create -f does make sense.

I don't see why it makes sense honestly. And I'm worried that you could use that to create an event for a missing check & entity, for example:

```
$ cat event.json
{
  "type": "Event",
  "spec": {
    "check": {
      "name": "missing_check",
      "command": "echo 0",
      "interval": 60
    },
    "entity": {
      "id": "missing_entity",
      "class": "foo",
      "subscriptions": ["linux"],
      "environment": "default",
      "organization": "default"
    }
  }
}
$ sensuctl create -f event.json
$ sensuctl event info missing_entity missing_check
=== missing_entity - missing_check
Entity:    missing_entity
Check:     missing_check
Output:
Status:    0
History:
Silenced:  false
Timestamp: 1969-12-31 19:00:00 -0500 EST
$ sensuctl check info missing_check
Error: not found
$ sensuctl entity info missing_entity
Error: {"error":"not found","code":2}
```

echlebek:
The extension tests rely on being able to create an event to function. They use the API to do this.

If we support PUT for Events in the API, we should support creation with `sensuctl create`.

If we should not allow creation of events with `sensuctl create`, then we should also disallow creating them through the API.

PR #1553, opened 2018-05-16
Refactor end-to-end tests.

### LINKED ISSUE #1500: Use a single backend for end-to-end tests
Our current testing strategy suffers due to spinning up multiple backends, and using random ports. This causes a small percentage of our builds to fail during the e2e testing phase.

Modify the test infrastructure to run the e2e tests against a single backend. Use a pattern like the following to do test setup and teardown:
```
func TestMain(m *testing.M) {
    be, cleanup, err := newBackend()
    if err != nil {
        log.Fatal(err)
    }
    status := m.Run()
    cleanup()
    os.Exit(status)
}
```

echlebek:
This commit significantly changes how our end-to-end tests work.
Instead of each test spinning up its own backend, the tests all
share a single backend, with two exceptions.

The reason why we want to constrain the number of backends running
is that allocating ports to these backends is somewhat tricky, and
with a static number of backends, ports can be allocated
statically.

Initially, it was planned that all tests would run on the same
backend, but that didn't pan out for the agent reconnection tests
and the RBAC tests.

This commit also removes the HTTP client in favour of sensuctl.
sensuctl is now used for all e2e tests.

Tests are isolated from one another via RBAC - each test executes
within its own organization and environment.

It remains to be seen if this strategy will lead to a more stable
e2e testing environment. I had a lot of issues in implementing this
work.

Closes #1500 

No

* refactoring was required
* test results were unclear
* test behaviour was inconsistent


testing/e2e/agent_reconnection_test.go
```diff
@@ -7,25 +7,32 @@ import (
...
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
 func TestAgentReconnection(t *testing.T) {
 	t.Parallel()
 
-	// Start the backend
-	backend, cleanup := newBackend(t)
+	backend, cleanup, err := newBackendProcess(40005, 40006, 40007, 40008, 40009)
```
cyphus: Using explicit port numbers for testing will really help when tracing port issues down to the originating test. It does mean that we would have to manually avoid port collisions though. I wonder if there's a better way to coordinate this.

testing/e2e/agent_reconnection_test.go
```diff
@@ -34,32 +41,43 @@ func TestAgentReconnection(t *testing.T) {
...
 	require.NoError(t, json.Unmarshal(output, &event1))
-	assert.NotNil(t, event1)
+	require.NotNil(t, event1)
 
 	// Now terminate the backend
 	require.NoError(t, backend.Terminate())
 
+	if true {
+		return
+	}
```
cyphus: 🤔

testing/e2e/process_test.go
```diff
@@ -176,21 +164,28 @@ type agentProcess struct {
...
 }
 
 // newAgent abstracts the initialization of an agent process and returns a
 // ready-to-use agent or exit with a fatal error if an error occurred while
 // initializing it
 func newAgent(config agentConfig, sensuctl *sensuCtl, t *testing.T) (*agentProcess, func()) {
-	agent := &agentProcess{agentConfig: config}
+	agent := &agentProcess{agentConfig: config, backendURL: sensuctl.wsURL}
+	agent.APIPort = int(atomic.AddInt64(&agentPortCounter, 1))
+	agent.SocketPort = int(atomic.AddInt64(&agentPortCounter, 1))
```
cyphus: Ah, so this is how ports are going to be coordinated! More deterministic this way.

nikkiki:
A new e2e test has been merged 😬 refactoring might be required for that one to run on the same backend, too: https://github.com/sensu/sensu-go/blob/48faf86b2d938522fffc813ef8b3c682ed426f3f/testing/e2e/sensuctl_create_test.go

cyphus:
I had a couple nitpicks, but overall this PR looks good!

testing/e2e/agent_reconnection_test.go
echlebek: I've tried a few approaches in the past. It's pretty tricky to do right, because the only way you can "reserve" a port is by binding to it.

Since the backend takes port numbers, and not listeners, there is no way to just ask the OS for a free port.

This is one of the reasons I was considering docker for the e2e tests.

testing/e2e/agent_reconnection_test.go
echlebek: oops

echlebek: fixed

PR #1557, opened 2018-05-17
[Web UI] Update material-ui dependency to latest RC

jamesdphillips:
## What is this change?

The dependency was previously updated along with our recent build changes but none of the breaking changes were not addressed, this lead to some (relatively minor) issues.

https://github.com/mui-org/material-ui/releases
https://github.com/mui-org/material-ui/compare/v1.0.0-beta.37...v1.0.0-rc.1

Content is borked.

Unlikely

nada

`material-ui` moved to `@material-ui/core`.

mui-org/material-ui#11310
mui-org/material-ui#11330

### DIFF
--- a/dashboard/src/components/NamespaceSelectorMenu.js
+++ b/dashboard/src/components/NamespaceSelectorMenu.js
@@ -2,11 +2,13 @@ import React from "react";
 import PropTypes from "prop-types";
 import gql from "graphql-tag";
 import partition from "lodash/partition";
-import { withStyles } from "material-ui/styles";
+import { withStyles } from "@material-ui/core/styles";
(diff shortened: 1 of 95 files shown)


dashboard/src/components/NamespaceSelectorMenu.js
```diff
@@ -85,9 +88,10 @@ class NamespaceSelectorMenu extends React.Component {
           organization.environments.map(environment => (
             <Link
               to={`/${organization.name}/${environment.name}`}
+              key={`${organization.name}${environment.name}`}
```
10xjs: The key should be separated by a slash here. Otherwise `ab/c` would have the same key as `a/bc`

Issue #1550, opened 2018-05-16
[Build stability] - integration test phantom failure


mercul3s:
Integration tests failed on both Appveyor and Travis. Interestingly, there were no actual test failures, but a lot of keepalived update failure messages. The tests seemed to succeed, but the build script exited with 1, which failed the build. The build succeeded after a retry. The logs were lengthy, so I added them as gists.
[Travis build log](https://gist.github.com/mercul3s/0e632ffd4f0d0d6b7c441248794437b5)
[Appveyor build log](https://gist.github.com/mercul3s/e638a51710a8c9de432689695cb2a0ba)

nikkiki:
Hit this on a recent appveyor build: [log.txt](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/5a00d6eb9a65e90012552ede/125516c66e82c728ace21e0d46e49b9d61cec1b99de6b546a749b2ce6b137f707957a67c2687e27cba112b585a5f4fad120a5f15afeb9539e4e76224d50152a7842f4d7ed43c18a6b9871db907bd218b4f82b5b5825e8d.txt)

nikkiki:
Reopening.. the phantom failure is back https://ci.appveyor.com/project/sensu/sensu-go/build/1.0.3018/job/3rdik0p1cuhpr3ps

annaplotkin:
Closing this one, we will open another issue with more details if we encounter this again.

PR #1561, opened 2018-05-18
Add the wrapped-json format to sensuctl

### LINKED ISSUE #1221: Produce JSON resources compatible with sensuctl create subcommand
Related to https://github.com/sensu/sensu-go/pull/1187

Now that we are adding the ability to create resources with `sensuctl create`, using a new `Wrapper` struct, we should make sure that JSON resources printed by sensuctl are compatible, so users don't have to modify the JSON structure or don't have to start from scratch when creating a new resource. For example:

```bash
$ sensuctl check info check-cpu --format json > check-cpu.json
$ vim check-cpu.json # modify some attributes etc.
$ sensuctl create -f check-cpu.json
```

As per [Eric suggestion](https://github.com/sensu/sensu-go/pull/1187/files#r176783752), a possible solution:

> the API would wrap all of its responses?

nikkiki:

## What is this change?

Adds the format `wrapped-json` to sensuctl `configure`, `list` and `info`commands, which is compatible with `sensuctl create`.

Closes #1221.

Yas.

Nah.

Yes, refactoring was required. And reflection is hard.

### DIFF
--- a/cli/commands/helpers/json.go
+++ b/cli/commands/helpers/json.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"reflect"
 	"strings"
 
 	"github.com/sensu/sensu-go/cli/client/config"
@@ -32,18 +31,12 @@ func PrintJSON(r interface{}, io io.Writer) error {
 	return err
 }
 
-// PrintWrappedJSON takes a record(s) and an io.Writer, converts the record to
+// PrintWrappedJSON takes a record(s) and Resource, converts the record to
 // human-readable JSON (pretty-prints), wraps that JSON using types.Wrapper, and
 // then prints the result to the given writer. Unescapes any &, <, or >
 // characters it finds.
-func PrintWrappedJSON(r types.Resource, io io.Writer) error {
-	v := reflect.ValueOf(r)
-	i := reflect.Indirect(v)
-	t := i.Type()
-	w := types.Wrapper{
-		Type:  t.Name(),
-		Value: r,
-	}
+func PrintWrappedJSON(r types.Resource, wr io.Writer) error {
+	w := types.WrapResource(r)
 
 	buf := new(bytes.Buffer)
 	encoder := json.NewEncoder(buf)
@@ -55,14 +48,14 @@ func PrintWrappedJSON(r types.Resource, io io.Writer) error {
 	}
 
 	s := htmlReplacer.Replace(buf.String())
-	_, err := fmt.Fprint(io, s)
+	_, err := fmt.Fprint(wr, s)
 	return err
 }
 
-// PrintWrappedJSONList takes a record(s) and an io.Writer, converts the record to
-// human-readable JSON (pretty-prints), wraps that JSON using types.Wrapper, and
-// then prints the result to the given writer. Unescapes any &, <, or >
-// characters it finds.
+// PrintWrappedJSONList takes a resource list and an io.Writer, converts the
+// record to human-readable JSON (pretty-prints), wraps that JSON using
+// types.Wrapper, and then prints the result to the given writer. Unescapes
+// any &, <, or > characters it finds.
 func PrintWrappedJSONList(r []types.Resource, io io.Writer) error {
 	for _, res := range r {
 		err := PrintWrappedJSON(res, io)
--- a/cli/commands/helpers/json.go
+++ b/cli/commands/helpers/json.go
@@ -5,7 +5,11 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"reflect"
 	"strings"
+
+	"github.com/sensu/sensu-go/cli/client/config"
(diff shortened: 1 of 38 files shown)


cli/commands/helpers/json.go
```diff
@@ -27,3 +31,64 @@ func PrintJSON(r interface{}, io io.Writer) error {
...
+	if err := encoder.Encode(w); err != nil {
+		return err
+	}
+
+	s := htmlReplacer.Replace(buf.String())
+	_, err := fmt.Fprint(io, s)
+	return err
+}
+
+// PrintWrappedJSONList takes a record(s) and an io.Writer, converts the record to
```
echlebek: Similar comments as above here.

nikkiki:
In addition to CI tests, I've QA tested the `info` and `list` commands on each resource in each format (`json`, `wrapped-json`, and `tabular`).

cli/commands/helpers/json_test.go
```diff
@@ -22,3 +27,111 @@ func TestPrintJSON(t *testing.T) {
...
+
+	w := types.Wrapper{
+		Type:  "CheckConfig",
+		Value: check,
+	}
+	output, err := json.Marshal(w)
+	assert.NoError(err)
+
+	buf := new(bytes.Buffer)
+	writer := io.Writer(buf)
```
echlebek: You can remove this, because buf is already an io.Writer.

cli/commands/helpers/json_test.go
```diff
@@ -22,3 +27,111 @@ func TestPrintJSON(t *testing.T) {
...
+
+	require.NoError(t, PrintWrappedJSONList([]types.Resource{check1, check2}, writer))
+	// trim \n and white space for equal comparison
+	output3 := strings.Replace(buf.String(), "\n", "", -1)
+	output3 = strings.Replace(output3, " ", "", -1)
+	assert.Equal(string(output1)+string(output2), output3)
```
echlebek: This is a bit unsafe because the output might differ. We should use the JSONEq function.

We can do it line by line by using strings.Split.

types/user.go
```diff
@@ -23,6 +24,11 @@ func (u *User) Validate() error {
 	return nil
 }
 
+// URIPath returns the path component of a User URI.
```
echlebek: I'm not so sure we want to allow creating Users via RBAC just yet.

This would mean people would be storing raw passwords in CM, which isn't very secure.

I want to have a discussion about how we can support this, but I think we'll have to change some things about User first.

nikkiki:
Appveyor failure has been documented in #1550

cli/commands/helpers/json_test.go
nikkiki: @echlebek JSONEq does not recognize the output of `PrintWrappedJSONList` as valid json:
```
{check1}
{check2}
```
I think we'll have to fudge the expected value in some other way then. In what case would the output differ with all stripped whitespace? I would _expect_ the contents to be the same.

types/user.go
nikkiki: Okay, I've changed this and addressed the other feedback. I've also made a note that `wrapped-json` is unsupported for `User` because of this!

cli/commands/helpers/json_test.go
echlebek: Right; that's why you would split the lines and compare each one individually. :)

echlebek: The contents can differ; serializing JSON is not necessarily a deterministic process. (Map iteration order can change things a lot.)

nikkiki: How would we split `buf.String()`:
```
{
  "type": "CheckConfig",
  "spec": {
    "command": "command",
    "environment": "default"
  }
}
{
  "type": "CheckConfig",
  "spec": {
    "command": "command",
    "environment": "default"
  }
}
```
So that it is equal to output1 or output2 without trimming whitespace or newline characters:
```
{"type": "CheckConfig","spec": {"command": "command","environment": "default"}}
```

nikkiki: A json decoder perhaps?

nikkiki: This feels weird but works.
```
	output3 := strings.Split(buf.String(), "}\n{")
	assert.JSONEq(string(output1), output3[0]+"}")
	assert.JSONEq(string(output2), "{"+output3[1])
```

echlebek: Oh, good point, I forgot that it tabs the output.

I like what you have above a bit better. Let's go with that for now!

nikkiki: Done!

nikkiki:
@palourde I absolutely agree, an e2e test would be super useful here. Mind if I file a separate issue for it? I was trying to be conscientious of the large diff with the feature implementation and refactor required.

nikkiki:
@palourde YOLO. I pushed an e2e test. Hopefully it won't require much refactoring after #1500!

PR #1563, opened 2018-05-18
[Web UI] Update browser polyfills

### LINKED ISSUE #1357: [Web UI] [Error Handling] Globally polyfill Promise with Bluebird
Epic: #1299

Polyfill `Promise` with `bluebird` in all browsers to provide `uhnandledrejection` events everywhere.

10xjs:
## What is this change?

Upgrade the browser polyfill system to serve cacheable, conditionally-loaded bundle chunks using Modernizr for feature detection.

Replace the conditional `Promise` polyfill with a constant alias to `bluebird` better error handling and consitent general behavior in all browsers.

image
_async polyfill chunks in green_

closes #1357 

Since polyfills are only needed on some clients, it makes sense to serve them conditionally. This keeps the main bundle size to an absolute minimum.

Splitting the bundle into conditional chunks also moves more code into assets that can be cached long term. Polyfill chunks are 100% comprised of "vendor" code, and will be deterministic and cacheable across many builds and updates to the app over a long period of time. This approach improves the app load-time for all browsers.

The decision was made to force a `Promise` polyfill, even in browsers that provide a native `Promise` implementation. This is done for consistency - we are less likely to encounter promise-related bugs, and for debugability - `bluebird` offers the important `onunhandledrejection` that some browsers don't provide (this ensures our global error state is always triggered for unhandled errors) and offers better error logging in general.

nah

nope

I haven't tested the build in IE and other browsers for which we are providing polyfills (although I have tested in Chrome by stubbing the feature detection).


dashboard/src/modernizr.macro.js
```diff
@@ -0,0 +1,31 @@
...
+            .replace(/^!(function\()/m, "module.exports=$1")
+            .replace(/;[a-z]+\.Modernizr=Modernizr/, ";return Modernizr")
+        : output
+            .replace(/^;(\(function\()/m, "module.exports=$1")
+            .replace(/window.Modernizr = Modernizr;$/m, "return Modernizr");
+
+      emitFile("modernizr.js", patched);
+      done(null, patched);
+    },
+  );
```
jamesdphillips: itsagooda

PR #1585, opened 2018-05-23
Make metric event logs more specific with additional fields

nikkiki:

## What is this change?

Provides additional context to metric event logs included # of metrics and first metric.

Metric event logs are vague.

Yas.

Nah.

Nah.

### DIFF
--- a/util/logging/logging.go
+++ b/util/logging/logging.go
@@ -27,10 +27,10 @@ func EventFields(event *types.Event) map[string]interface{} {
 
 	if event.HasMetrics() {
 		count := len(event.Metrics.Points)
-		fields["metric count"] = count
+		fields["metric_count"] = count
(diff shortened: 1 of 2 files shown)


util/logging/logging.go
```diff
@@ -23,16 +23,15 @@ func EventFields(event *types.Event) map[string]interface{} {
...
 	}
 
 	if event.HasMetrics() {
-		fields["metrics"] = true
-	}
-
-	if event.HasMetrics() {
-		fields["metrics"] = true
+		count := len(event.Metrics.Points)
+		fields["metric count"] = count
```
palourde: We should probably avoid whitespaces in the fields key, since it's harder to manipulate programatically. Instead `metric_count`, for example, would be easier!

PR #1606, opened 2018-05-24
[Web UI] Add filtering to the entities page

### LINKED ISSUE #839: [Web] Entities page: filter
## 🕸 Related

Sketch: https://github.com/sensu/sensu-go/issues/815

jamesdphillips:

Closes #839 

`undefined`

`undefined`

* I fought with throttling input for a while but decided it probably was never going to be the most ideal experience. Search box now requires the user to explicitly hit enter to apply the filter.
* Fixed bug where selected order was not being applied to the query.

## Additional Notes

Menu items are collapsed at the `sm` breakpoint so that they are still available.


dashboard/src/components/CollapsingMenu/CollapsingMenu.js
```diff
@@ -0,0 +1,81 @@
...
+
+    const buttons = React.Children.map(children, child =>
+      React.cloneElement(child, { renderAs: "button" }),
+    );
+    const menuId = `collapsed-menu-${this._id}`;
+    const menuItems = React.Children.map(children, child =>
+      React.cloneElement(child, {
+        renderAs: "menu-item",
+        onClick: ev => {
+          if (child.props.onClick) child.props.onClick(ev);
```
10xjs: Nit: Avoid one-liner if statements. They are a little more difficult to set breakpoints on when debugging.

dashboard/src/components/CollapsingMenu/CollapsingMenu.js
```diff
@@ -0,0 +1,81 @@
...
+  };
+
+  render() {
+    const { breakpoint, children } = this.props;
+
+    const buttons = React.Children.map(children, child =>
+      React.cloneElement(child, { renderAs: "button" }),
+    );
+    const menuId = `collapsed-menu-${this._id}`;
+    const menuItems = React.Children.map(children, child =>
```
10xjs: I'm on the fence about this approach. I dislike seeing the coupling to each button child, this might not work if a button is a wrapped component.|

dashboard/src/components/CollapsingMenu/MenuItemBase.js
```diff
@@ -0,0 +1,31 @@
...
+    renderAs: PropTypes.oneOf(["button", "menu-item", null]),
+  };
+
+  static defaultProps = {
+    renderAs: null,
+  };
+
+  render() {
+    const { renderMenuItem, renderButton, renderAs } = this.props;
+    warning(
```
10xjs: I strongly advise against using `warning`. It is only small part of the system facebook uses to "minify" errors in react.

dashboard/src/components/QueryParams/QueryParams.js
```diff
@@ -0,0 +1,68 @@
...
+  static defaultProps = {
+    keys: null,
+  };
+
+  constructor(props) {
+    super(props);
+    this.params = new URLSearchParams(props.location.search);
+  }
+
+  componentWillReceiveProps(nextProps) {
```
10xjs: Nit: Use `getDerivedStateFromProps` for better forward compatibility.

dashboard/src/components/QueryParams/QueryParams.js
```diff
@@ -0,0 +1,68 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { withRouter } from "react-router-dom";
+
+function expandParams(params, keys) {
+  return Array.from(params).reduce((acc, entry) => {
```
10xjs: `Array.from` doesn't work in any version of IE

jamesdphillips: Good call out, IE11 ruins everything.

Added `core-js/es6/array` which might be overkill but should only ever be pulled down for IE users and means we can safely use `isArray`, `slice`, etc. as well.

dashboard/src/components/CollapsingMenu/CollapsingMenu.js
jamesdphillips: Instead of changing this single occurrence.. https://github.com/sensu/sensu-go/pull/1625.

PR #1607, opened 2018-05-24
[GraphQL] Support request batching

jamesdphillips:
## What is this change?

Adds support for batching of GraphQL operations.

Greatly simplifies batch operations by clients.

`undefined`

`null`

`null`


dashboard/src/apollo/client.js
```diff
@@ -38,6 +41,7 @@ const createClient = () => {
       introspectionLink(),
       stateLink({ cache }),
       authLink({ getClient }),
+      dedupLink,
```
10xjs: The docs for `apollo-link-dedup` mention that it is included by default. Has adding it been confirmed to be necessary?

jamesdphillips: Definitely should not required: https://github.com/apollographql/apollo-client/blob/9ac90e0c1729d11a41663c9a7032200369f3305d/packages/apollo-client/src/core/QueryManager.ts#L110

A leftover from something else I was hacking on. Good eye. 😬

PR #1609, opened 2018-05-25
Remove linux/386 & windows/386 CI jobs

amdprophet:

## What is this change?

Removes the linux/386 windows/386 e2e jobs from CI.

We are not planning on supporting Sensu Backend on linux/386 or windows/386. As a result, the e2e tests for those platforms do not provide us any value.

Yes.

Nope!

Nope!


echlebek:
Closes #1610

Issue #1573, opened 2018-05-22
Murder the schedulerd check cache


grepory:
## Expected Behavior

The API and Schedulerd remain in sync such that checks are scheduled are correctly configured.

Checks that are newly created via the API are schedule ~immediately.

## Current Behavior

The schedulerd check cache causes a number of issues, primarily it causes the scheduler and API to be out of sync for a indeterminate amount of time after an API update--wherein the check scheduled doesn't reflect the current configuration for that check.

After creating a check, a check is scheduled some indeterminate amount of time in the future when schedulerd updates its cache.

## Possible Solution

Stop caching checks / managing state in schedulerd. Each check scheduler goroutine should read from etcd when the check's timer fires. This will result in the check being executed as it's currently configured in etcd.

In order to accommodate updates to check config and the creation of new checks, we'll need a goroutine that watches check configs using a [check config watcher](https://github.com/sensu/sensu-go/blob/master/backend/store/etcd/watchers.go#L31). There are three events that we will receive from the watcher.

Create - we need to spin up a new CheckScheduler for the newly created check

Update - Interrupt the check scheduler, causing the check to execute and the timer to be reset.

Delete - Call stop on the scheduler.

(Note: in check_timer.go we aren't adhering to the interface as specified in the [godoc](https://godoc.org/time#Timer.Reset) for Timer.Reset or Timer.Stop).

At Schedulerd startup, build a map of a unique identifier for a check to the check scheduler (schedulerd already has this in the form of the ScheduleManager). Hand the simplified ScheduleManager off to the watcher goroutine. The watcher goroutine then uses the ScheduleManager only to interact with schedulers instead of as a cache.

We can then remove the StateManager entirely.

grepory:
Made good progress today.

Things we realized we didn't think through:

- Instead of caching assets and hooks, we will look them up when we build the check request at scheduling time.
- We need to load all check configs at check scheduler manager startup

PR #1614, opened 2018-05-28
Refactor/check scheduler

echlebek:
## What is this change?

This branch substantially reworks schedulerd, removing the caching layer in favour of direct accesses to etcd. This should resolve bugs related to caching in the scheduler, as well as resolve ambiguity about what the current state of the scheduler is.

The scheduler is difficult to reason about, and the code is poorly factored.

Yes


backend/schedulerd/executor.go
```diff
@@ -233,7 +238,7 @@ func (a *AdhocRequestExecutor) publishProxyCheckRequests(entities []*types.Entit
 }
 
 func (a *AdhocRequestExecutor) execute(check *types.CheckConfig) error {
-	request := a.buildRequest(check)
+	request, _ := a.buildRequest(check)
```
nikkiki: Would it be beneficial to catch and return the error here from `buildRequest`?

util/time/time.go
```diff
@@ -1,105 +0,0 @@
...
-	}
-
-	return current.After(beginTime) && current.Before(endTime), nil
-}
-
-// InWindows determines if the current time falls between the provided time
-// windows. Current should typically be time.Now() but to allow easier tests, it
-// must be provided as a parameter. The function returns a positive value as
-// soon the current time falls within a time window
-func InWindows(current time.Time, timeWindow types.TimeWindowWhen) (bool, error) {
```
nikkiki: Was this a duplicate function? You still invoke `subdue.InWindows` but this PR appears to remove, not replace, the function, as it exists in the `types` package already.

backend/schedulerd/executor.go
echlebek: In this instance, nil is always returned by AdhocRequestExecutor.buildRequest.

util/time/time.go
echlebek: I moved this code over some time ago, but did not quite get around to removing all existing uses of it. Now, it is only called from the types package.

nikkiki: Got it.. thanks for the extra cleanup!

backend/schedulerd/executor.go
nikkiki: Hopefully if that changes in the future, we remember to catch it.. I don't foresee that happening though 😉

echlebek:
Closes #1573

PR #1619, opened 2018-05-29
[Web UI] Create wrapped <Query> component

10xjs:
## What is this change?

This component enhances the original `<Query>` component from `react-apollo` giving us a few extra features. OOB it is compatible with the original component interface.

This new component gives us an `onError` callback for handling errors within the original request continuation. By default `<Query>` will now globally reject errors not handled within `onError`, fixing the issue of `<Query>` silencing errors.

A new `QueryAbortedError` class allows us to "abort" a GraphQL query request without triggering a failure state. The aborted status is returned to the consumer as an `aborted` boolean among the query render callback props. Once a query has been aborted, it can be be manually restarted with the `refetch` render prop.

no

**Why not use `apollo-link-error` to handle errors as part of the request continuation?** Handling errors at the apollo-client level is not a great approach. It does not allow the consumer of `<Query>` to handle errors representing a failure to delegate control to the consumer.

### DIFF
--- a/dashboard/src/components/util/Query.js
+++ b/dashboard/src/components/util/Query.js
@@ -1,0 +1,61 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { ApolloError } from "apollo-client";
+import { Query as BaseQuery } from "react-apollo";
+
+import QueryAbortedError from "/errors/QueryAbortedError";
+
+class Query extends React.PureComponent {
+  static propTypes = {
+    onError: PropTypes.func.isRequired,
+    children: PropTypes.func.isRequired,
+  };
+
+  static defaultProps = {
+    onError(error) {
+      throw error;
+    },
+  };
+
+  componentDidMount() {
+    const { onError } = this.props;
+    this.querySubscription = this.queryRef.current.queryObservable.subscribe({
+      next(result) {
+        if (result.errors && result.errors.length > 0) {
+          throw new ApolloError({ graphQLErrors: result.errors });
+        }
+      },
+      error(error) {
+        if (!(error.networkError instanceof QueryAbortedError)) {
+          onError(error);
+        }
+      },
+    });
+  }
+
+  componentWillUnmount() {
+    if (this.querySubscription) {
+      this.querySubscription.unsubscribe();
+    }
+  }
+
+  queryRef = React.createRef();
+
+  render() {
+    const { onError, children, ...props } = this.props;
+
+    return (
+      <BaseQuery ref={this.queryRef} {...props}>
+        {queryResult => {
+          const { error, ...rest } = queryResult;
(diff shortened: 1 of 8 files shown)


dashboard/src/errors/QueryAbortedError.js
```diff
@@ -0,0 +1,10 @@
+import ExtendableError from "es6-error";
+
+class QueryAbortedError extends ExtendableError {
```
jamesdphillips: Without a concrete use case at the moment it's difficult for me to envision when we might throw this. May be worth documenting either here or in the `Query` component.

dashboard/src/components/views/EnvironmentView/EventDetailsContent.js
```diff
@@ -39,11 +41,18 @@ class EventDetailsContent extends React.PureComponent {
         fetchPolicy="cache-and-network"
         variables={{ ...match.params, ns }}
       >
-        {({ client, data: { event } = {}, loading }) => {
-          if (!loading && (!event || event.deleted)) return <NotFoundView />;
+        {({ client, data: { event } = {}, loading, aborted }) => {
+          if (!loading && !aborted && (!event || event.deleted)) {
```
jamesdphillips: `!loading && !aborted` and `loading || aborted ` seem ubiquitous across every use of the Query container. As a convenience what if we called the render prop with a extra prop. Eg.

```
// Query#render
const pending = loading || aborted;
children({ pending, aborted: false, ...props });

// Ex usage ...
<Query ...>
  ({ data, pending }) => {
    if (!pending && !data) {
      return <NotFoundView />;
    }
    
    return <Container loading={pending} ... />
  }}
</Query>
```

10xjs: We will want to render proper aborted states with a refetch button. This would keep the loading and pending wholly separate making a calculated pending state not worth it.

dashboard/src/components/util/Query.js
```diff
@@ -0,0 +1,61 @@
...
+      throw error;
+    },
+  };
+
+  componentDidMount() {
+    const { onError } = this.props;
+    this.querySubscription = this.queryRef.current.queryObservable.subscribe({
+      next(result) {
+        if (result.errors && result.errors.length > 0) {
+          throw new ApolloError({ graphQLErrors: result.errors });
```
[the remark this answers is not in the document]
10xjs: This would be the case where the GraphQL endpoint returns an `errors` field on the response. This will cause them to be globally thrown (as we discussed last week)

dashboard/src/errors/QueryAbortedError.js
10xjs: Concrete case coming Soon™...

PR #1623, opened 2018-05-30
Fixed a bug where environments could not be deleted.

### LINKED ISSUE #1567: Unable to delete environment
## Expected Behavior

I'd expect to delete an environment that has only been created during tests.

## Current Behavior

```
/usr/bin/sensuctl environment delete test --skip-confirm

{"error":"environment is not empty","code":0}
```

## Possible Solution

It would be handy if the resource keeping an environment from being deleted is printed.  I can imagine this would be very expensive or verbose with lots of resources but in this case there should be nothing, see below.

Unsure how to reproduce as my acceptance testing will create resources then delete them so nothing should be persisting long enough to stay in a test environment.

## Context

I am running acceptance tests for Puppet sensu module and an environment is created, modified then deleted and the deletion is failing.  This only manifests itself when other resources before this environment are created.  What's making this strange is nothing is added or modified while between environment creation and deletion attempt.  

I'm unsure how to debug this issue as seems like this environment should have no resources associated with it and thus should be allowed to be deleted.

CC: @ghoneycutt

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): Nightly as of 05/20/2018
* Installation method (packages, binaries, docker etc.): RPM
* Operating System and version (e.g. Ubuntu 14.04): EL7

echlebek:
This commit fixes a bug where environments could not be deleted if
the organization they reside in has one or more assets.

Assets do not exist within environments, so there is no reason to
consider them when deleting environments.

This commit also stops the environment store from reporting an
error when an environment that doesn't exist is deleted.

Closes #1567


backend/store/etcd/environment_store_test.go
```diff
@@ -71,3 +67,38 @@ func TestEnvStorage(t *testing.T) {
 		assert.Equal(t, 1, len(envs))
 	})
 }
+
+func TestDeleteEnvironment_GH1567(t *testing.T) {
```
nikkiki: Perhaps I'm missing some context here, but what does `GH1567` mean?

backend/store/etcd/environment_store_test.go
```diff
@@ -71,3 +67,38 @@ func TestEnvStorage(t *testing.T) {
...
+		ctx := context.Background()
+		ctx = context.WithValue(ctx, types.OrganizationKey, "default")
+		ctx = context.WithValue(ctx, types.EnvironmentKey, "foo")
+
+		env := types.FixtureEnvironment("foo")
+		require.NoError(t, store.UpdateEnvironment(ctx, env))
+
+		// Second delete should succeed
+		require.NoError(t, store.DeleteEnvironment(ctx, env))
+		require.NoError(t, store.DeleteEnvironment(ctx, env))
```
nikkiki: Could you help me understand why this is the desired action? If an environment has been deleted and does not exist, why wouldn't it return an error when you attempt to delete it again?

backend/store/etcd/environment_store_test.go
echlebek: That's the issue this test was written to reproduce. (GH stands for Github)

backend/store/etcd/environment_store_test.go
echlebek: It's because HTTP DELETE needs to be idempotent. See https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html for more information.

backend/store/etcd/environment_store_test.go
nikkiki: Ah very cool!

echlebek:
@cyphus I think Sean and Greg designed assets to be scoped to Organizations so that Environments could reuse them.

PR #1627, opened 2018-05-30
[Web UI] Display a modal when authentication fails

10xjs:
## What is this change?

In the event that automatic session refresh fails, show the user a message instead of redirecting directly to the sign in page.

image

This provides a nicer experience than forcefully redirecting to the sign-in page and gives us a platform for further enhancements such as a login form modal.

no

nah

naw

### DIFF
--- a/dashboard/src/components/AppRoot.js
+++ b/dashboard/src/components/AppRoot.js
@@ -12,12 +12,14 @@ import ThemeStyles from "/components/ThemeStyles";
 
 import AuthenticatedRoute from "/components/util/AuthenticatedRoute";
 import UnauthenticatedRoute from "/components/util/UnauthenticatedRoute";
-
+import AuthInvalidRoute from "/components/util/AuthInvalidRoute";
 import DefaultRedirect from "/components/util/DefaultRedirect";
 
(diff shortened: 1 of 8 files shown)


dashboard/src/components/AppRoot.js
```diff
@@ -49,6 +51,10 @@ class AppRoot extends React.PureComponent {
               />
               <Route component={NotFoundView} />
             </Switch>
+            <Switch>
+              <UnauthenticatedRoute exact path="/signin" />
+              <AuthInvalidRoute component={AuthInvalidDialog} />
+            </Switch>
```
jamesdphillips: 🤔 can be deferred until later but I'm not exactly in love with this. The intent is not immediately clear to me and will break the next time we add another unauthenticated route to the application.

10xjs: This approach appears to be the only option for creating a route with exceptions. (https://stackoverflow.com/questions/45591165/react-route-component-handle-all-routes-except-for-login)

10xjs: Our other option would be to render `<AuthInvalidRoute component={AuthInvalidDialog} />` somewhere on every page in the main switch.

PR #1634, opened 2018-06-01
Test metric extraction with an integration test

### LINKED ISSUE #1597: Test metric extraction with an integration test
Test that metric extraction is working with an integration test. The test should work similarly to the e2e test, and validate that check output is being transferred to metrics events as appropriate.

nikkiki:

## What is this change?

Removes check output metric extraction e2e test, in favor of more detailed integration coverage.

Closes #1597 

Yas.

Nah.

Yas.


cyphus:
Would it be possible to simplify this further by changing the command from `cat` to `echo`, thereby avoiding creating the temporary file?

I'm always a fan of net code loss 😎

nikkiki:
@cyphus We could, if we wrote a go program `echo` in [sensu-go/tools](https://github.com/sensu/sensu-go/tree/882d062bf337350a5344e94e6e9fd2a4bf313e0b/tools). These "commands" are actually executables that mimic the original shell commands. The program itself would be fairly simple, but arguably a small code gain 😆. It's not a bad idea though, especially for future tests!

nikkiki:
I think we actually could write to stdin to avoid the temp file, without having to write an `echo` command.

nikkiki:
_Remembers why the temp file was easier.._

nikkiki:
giphy-1

PR #1650, opened 2018-06-06
[GraphQL] Gracefully handle failed type assertions

### LINKED ISSUE #1556: Check with an asset on a windows agent breaks events page on dashboard
I am testing using assets with the windows agent.  I set up an asset and set up a check to use that asset.  After the check fires, the events page no longer loads.  

## Expected Behavior
This is intertwined with another issue I'm having with the Windows agent - it doesn't seem to have access to the PATH.  I expect to see an error saying something like `The argument 'check-site.ps1' to the -File parameter does not exist`.

## Current Behavior
What I see currently is a white page.  I see the following error in sensu-backend.log:
```
{"component":"apid.routers","errors":[{"message":"interface conversion: interface {} is nil, not int","locations":[]}],"level":"error","msg":"error(s) occurred while executing GraphQL operation","time":"2018-05-17T12:42:57Z"}
```

1.  Create an asset (threw a little ps1 script up on nofile):
```
Sensuctl asset create check-site.tar -u https://nofile.io/g/Y2YlEd8RBGs96KdSMX83GX8HQ0BbPi2usy8nHtD729Zqf1fCs60v8LFwiim8Az92/checksite.tar/ --sha512 98890ae970a2fcfa95f2023ef65bdf224f2e72c49e7f824e6790c4b9229212f54cf28c3e383ade85327385c1dd73741e3c34efffa49bce7b1b136a01bbd53ea8
```
2.  Create a check that uses that asset:
```
sensuctl check create check-google --command "C:\Windows\System32\WindowsPowerShell\v1.0\Powershell.exe -NonInteractive -NoProfile -ExecutionPolicy Bypass -NoLogo -f check-site.ps1 https://google.ca" --subscriptions "GoogleProxyAgent" --interval 60 --runtime-assets check-site.tar --proxy-entity-id google
```
Note: I do see the agent having downloaded this asset.
3.  Load events page - white screen.

## Context
We'll be using windows agents quite a bit, and assets seem like a great feature, so I'd like to get them working.

## Your Environment
* sensuctl version: sensu-ctl version #f70b9f8, build f70b9f8a14ea926e4ce85467beecfe5cf731513d, built 2018-05-08T03:51:45+0000
* sensu-backend version: sensu-backend version dev-nightly#9bba4cf, build 9bba4cf6af864bbabf0a3120ec2ad6d187552a8b, built 2018-05-06T09:22:49+0000
* sensu-agent version: sensu-agent version #f70b9f8, build f70b9f8a14ea926e4ce85467beecfe5cf731513d, built 2018-05-08T04:01:02+0000
* Installation method (packages, binaries, docker etc.): binaries
* Operating System and version (e.g. Ubuntu 14.04): agent: Windows 10, backend: Amazon Linux

jamesdphillips:
## What is this change?

Avoid panic when selecting an event with an empty check by handling failed type assertions in generated aliases.

Closes #1556.

Ran into what initially seemed like some strange behavior when checking if a value was nil. Probably best most easily described [by this snippet](https://play.golang.org/p/NYdHDn4PLc5).


jamesdphillips:
> I think this makes sense to me. If I'm understanding correctly, it ensures we return an empty object in case of an error to avoid nil cases, right?

@mercul3s Yup. To help narrow things down, the crux of the PR is in this commit: https://github.com/sensu/sensu-go/pull/1650/commits/7a32fdcb089a7c41f77f9c3fde8e88f931ef520b.

Prior to this when a request was being executed we ended up with something like this:

```go
// imagine the service received a request with a fragment like
... on CheckConfig {
  name
  proxyRequest {
    splay
  }
}

// now imagine the check that were referencing does not have a proxyRequest
check := Check{
  Name: "frankie",
  ProxyRequest: nil
}

// when resolving the 'proxyRequests' field previously it would have looked something
// like this.
source := reflect.ValueOf(check)
proxyRequestField := sourceVal.FieldByName("ProxyRequest")
proxyRequest := proxyRequestField.Interface()
return proxyRequest

// next the executor would need to determine if it should resolve the 'splay' field; 
// it does this by simply checking if the value is nil. Eg.
if proxyRequest == nil {
  // we're done here, move on!
}
// continue on...

// before this change, problematically, the result of resolving the proxyRequest
// field is not `nil` but in-fact a reference to `check.ProxyRequest`. Essentially:
check := Check{}
var proxyRequest interface{} = check.ProxyRequest
fmt.Println(check.ProxyRequest == nil) // true
fmt.Println(proxyRequest == nil) // false
fmt.Println(proxyRequest.(*ProxyRequest) == nil) // true

// As a result previously the executor would continue on and try to resolve
// the 'splay' field, leading to a panic when we would try to access fields on `nil`.
```
 
https://play.golang.org/p/NYdHDn4PLc5

PR #1656, opened 2018-06-07
[Web UI] Filter entities by given subscription

### LINKED ISSUE #1582: [Web UI] Entities Page: Filter by subscriptions
As a user I should be able to filter the list of entities by a given subscription.

---

Uchiwa.

screen shot 2018-05-22 at 3 17 37 pm

---

Blocked by #922.

jamesdphillips:

Closes #1582 

`undefined`

`null`

`0`


dashboard/src/components/partials/EntitiesList/EntitiesListHeader.js
```diff
@@ -31,18 +34,36 @@ const styles = theme => ({
...
   static defaultProps = {
     onClickSelect: () => {},
+    onChangeFilter: () => {},
     selectedCount: 0,
+    subscriptions: { entries: [] },
+  };
+
+  static fragments = {
+    subscriptions: gql`
+      fragment EntitiesListHeader_subscriptions on SubscriptionSet {
```
10xjs: Does this run the risk of returning a dangerously large result without a limit?

10xjs: Does this run the risk of returning a dangerously large result without a limit?

dashboard/src/components/partials/EntitiesList/EntitiesListHeader.js
```diff
@@ -31,18 +34,36 @@ const styles = theme => ({
...
+        entries(orderBy: FREQUENCY)
```
10xjs: Does this run the risk of returning a dangerously large result without a limit?
 	 
I think making this it's own query would be a good idea. With query batching there wouldn't be any overhead, and it would set us up for fetching this list in chunks if we need to or refetching it independently of the main page content.

dashboard/src/components/views/EnvironmentView/EntitiesContent.js
```diff
@@ -70,8 +70,9 @@ class EntitiesContent extends React.PureComponent {
                 />
               </Content>
               <EntitiesList
-                loading={loading || aborted}
                 environment={environment}
+                loading={loading || aborted}
+                onQueryChange={this.props.setQueryParams}
```
10xjs: Use `onChangeQuery` for consistency with other event handler props: `on{EventName}{TargetName}`

dashboard/src/components/partials/EntitiesList/EntitiesListHeader.js
jamesdphillips: > Does this run the risk of returning a dangerously large result without a limit?

From looking around the set could probably be in the hundreds, so it is probably worth adding a limit (and offset.) 👍

> I think making this it's own query would be a good idea.

I like your thinking here, with a limit implemented, it would make it easy to implement react virtualized and friends. However, with the design in flux, I think I'll leave it as is for the time being. Lest I build a more involved solution all for naught.

10xjs: cool cool cool

PR #1658, opened 2018-06-07
Return an error from configure on invalid env/org.

### LINKED ISSUE #1515: sensuctl allows users to switch to namespaces that do not exist
I can configure sensuctl to switch to a namespace that does not exist. No error occurs on configure or subsequent commands.

## Expected Behavior
An error occurs when switching to a namespace that does not exist.

## Current Behavior
No error occurs.

## Possible Solution
Make sure the organization and environment exist when configuring sensuctl.

1. `sensuctl configure`, set an environment that does not exist.

echlebek:
The commit modifies the behaviour of sensuctl configure, so that
if a user attempts to configure an organization or environment that
does not exist, an error will be returned.

Previously, sensuctl would allow an organization or environment
to be configured even if they did not exist.

Closes #1515

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,8 @@ when creating resources with `sensuctl create`.
 - Provided additional context to metric event logs.
 - Updated goversion in the appveyor configuration for minor releases.
 - Use a default hostname if one cannot be retrieved.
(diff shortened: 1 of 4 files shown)


echlebek:
Weird, appveyor checks never ran on this branch.

PR #1660, opened 2018-06-08
[Web UI] Create ListController util component

10xjs:
## What is this change?

Pull the common list render and selection logic from the events and entities pages into a shared `ListController` component.

We have the same selection logic requirements on the events, entities, and checks pages.

nope

no

nah


dashboard/src/components/TableList/TableListBody.js
```diff
@@ -17,15 +17,16 @@ const styles = theme => ({
   },
 });
 
-export class TableList extends React.Component {
+export class TableListBody extends React.Component {
```
jamesdphillips: 😬👌

10xjs:
@jamesdphillips This should receive another once-over not that I've resolved a _ton_ of upstream conflicts.

PR #1664, opened 2018-06-11
Add environment variables to checks

### LINKED ISSUE #1551: Add env_vars to check attributes
Like the handler and mutator attributes, I see no good reason not to offer the ability to set environment values for a check, which would be in addition to the overrides done for assets.

## Expected Behavior
When env_vars is set on a check definition, invocations of the check export those variables with the check command.

## Current Behavior
The environment passed to the check command mirrors the environment of the agent, with the asset env overrides only.

## Possible Solution
Extend proto for checks to include env_vars like handlers and mutators.  Update command handler in the agent to look for this configuration and further extend the environment passed to the command accordingly.

not applicable

## Context
 How has this issue affected you? What are you trying to accomplish? 

I'm not 100% sure what I'm trying to accomplish, but I feel like this could be very useful as an alternative to using command flags for very long commands.

Arbitrary example of how this would be done currently
```json
{ "command": "ruby-runner.sh --version 2.5.0 -- my_check.rb --critical 500 --warn 100 --host blah.blah --arbitrary_option 37 --some-other-thing nope" }
```
And how the same could be done with env_vars.  This looks nicer, no?
```json
{ "command": "ruby-runner.sh -- my_check.rb",
  "env_vars": [
    "RUBY_VERSION=2.5.0",
    "CRITICAL=500",
    "WARN=100",
    "CHECK_HOST=my.host.internal",
    "ARBITRARY_OPTION=37",
    "SOME_OTHER_THING=nope" ] }
```

Having looked around in the baseline, this doesn't seem completely outside the realm of possibility, but I am certainly no good at golang (nor do I want to be).  But, if this seems easy enough to implement, I feel it would be worthwhile to do so. 

## Your Environment
N/A.  This came up during a slack conversation regarding current functionality in Sensu 2.0.

Historically, for my use case, the ruby installed with Sensu 1.X is used, but this is no longer possible with 2.0, so I am left considering alternatives, as every system doesn't have the same ruby installed.  As such, I came up with the thought of writing a bash script which ultimately ran ruby checks and took care of the boilerplate task(s) of managing a ruby environment (i.e. setting up the rbenv environ and whatnot) which could result in very long command definitions.  In retrospect, I realize that, in the ruby case, I could also use a .version and Gemfile file included in a check asset to define the expected environ (I wouldn't need the RUBY_VERSION env var above).  This being said, I still contend that an env_vars option to a check definition would be useful, particularly for a command that requires a lot of options to be set.

echlebek:
## What is this change?

Add environment variables to checks.

Closes #1551 

Should we file a followup ticket to add support for env vars to sensuctl? Right now, users will have to use `sensuctl create` to make use of this feature.


echlebek:
I'm happy the windows tests passed, I didn't actually run them locally :D

PR #1668, opened 2018-06-11
Remove --org parameter from environment command.

### LINKED ISSUE #1514: sensuctl does not respect the --org parameter when creating environments
## Expected Behavior
When I use the --org flag when creating an environment, the environment should be created within that org.

## Current Behavior
The environment is created within the default org.

1. `sensuctl create environment bar --org foo`
2. `sensuctl environment list`

## Context
Discovered this bug while refactoring e2e tests to use unique namespaces.

echlebek:
In testing, it was discovered that the --org parameter is actually
not necessary for the environment create command. Users can specify
the --organization parameter instead.

This commit removes the parameter.

Closes #1514


cyphus:
Bug fixed and one less `TODO` LGTM!

PR #1696, opened 2018-06-20
Don't reset check occurrences to 1 on check resolution.

### LINKED ISSUE #1624: Check occurrences property doesn't update
I'm trying to set up a filter to activate a handler to send an alert to OpsGenie.  I want to use the hourly example filter from here: https://docs.sensu.io/sensu-core/2.0/guides/reduce-alert-fatigue/.  If I set this up with my check, it gets handled every time.  I think this is because when I run `sensuctl event list` the occurrences property of the check is always 1.  The code here: https://github.com/sensu/sensu-go/blob/master/backend/eventd/eventd.go#L227 implies to me that it should be the count of checks run that have returned the most recent status.

## Expected Behavior
I expect this handler to be run when the check switches states, or once an hour.

## Current Behavior
The handler is run every time the check fires.

## Possible Solution
I believe the issue is with the occurrences property.

1. `sensuctl filter create hourly --action allow --statements "event.Check.Occurrences == 1 || event.Check.Occurrences % (3600 / event.Check.Interval) == 0"`
2. Any handler using the hourly filter
3. Any check that you can toggle for testing
4. Attach the handler to the check
5. The handler will fire each time

## Context
We want to create alerts in OpsGenie when these checks fail, then clear them when the check goes back to a good status.

## Your Environment
1. sensu-backend version 2.0.0-nightly#82c3128, build 82c31284eb9a378c60869544d95a2cc889a0d686, built 2018-05-23T09:22:02+0000
2. sensu-agent version #f70b9f8, build f70b9f8a14ea926e4ce85467beecfe5cf731513d, built 2018-05-08T04:01:02+0000
3. Windows 10 agent, Amazon Linux backend

echlebek:
Check occurrences are the number of times a check has had the same
status, in the case where checks are failing.

When checks are resolved, the occurrences field will stop
receiving updates. Occurrences only apply to failing status codes.

This is somewhat unintuitive, but it makes 2.x compatible with 1.x
in this respect.

Closes #1624

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,8 @@ lockfiles to not be cleaned up.
 - Fixed a bug where the CLI default for round robin checks was not appearing.
 - Missing custom attributes in govaluate expressions no longer result in
 an error being logged. Instead, a debug message is logged.
+- Fixed a bug in check occurrences where the behaviour was not in line with
+1.x behaviour. Check occurrences now only update on failing check statuses.
 
 ### Removed
 - Removed Linux/386 & Windows/386 e2e jobs on Travis CI & AppVeyor
--- a/backend/eventd/eventd.go
+++ b/backend/eventd/eventd.go
@@ -228,20 +228,27 @@ func updateOccurrences(event *types.Event) {
 	if !event.HasCheck() {
 		return
 	}
-	if len(event.Check.History) > 1 && (event.IsIncident() || isFlapping(event)) {
-		historyLen := len(event.Check.History)
-		if event.Check.History[historyLen-1].Status == event.Check.History[historyLen-2].Status {
-			event.Check.Occurrences++
-		} else {
+
+	defer func() {
+		if event.Check.Occurrences == 0 {
 			event.Check.Occurrences = 1
 		}
+		if event.Check.Occurrences > event.Check.OccurrencesWatermark {
+			event.Check.OccurrencesWatermark = event.Check.Occurrences
+		}
+	}()
+
+	if len(event.Check.History) < 2 || !(event.IsIncident() || isFlapping(event)) {
+		return
+	}
+
+	hist := event.Check.History
+	if hist[len(hist)-1].Status == hist[len(hist)-2].Status {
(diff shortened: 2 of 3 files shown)


palourde:
> This is somewhat unintuitive, but it makes 2.x compatible with 1.x in this respect.

I don't believe we should make 2.x occurrences compatible with 1.x since the concept of _events_ by itself completely changed; in Sensu 1.x, only check results with non zero statuses could create events, whereas Sensu 2.x creates events for every check result received.

Additionally, I don't think it would solve the problem described in https://github.com/sensu/sensu-go/issues/1624; let's say we have the following history: `0, 1, 0, 0, 0`. The `occurrences` field would still be set to `1`, and therefore the handler would still be triggered on `0` statuses.

The problem described in https://github.com/sensu/sensu-go/issues/1624 is actually because of a significant change in Sensu 2.x pipeline; all events (and therefore all check results) are passing through the pipeline, whereas only non-zero results would go through in 1.x. In order to prevent non-failing events from triggering handlers, the `is_incident` built-in filter must also be assigned to relevant handlers, as described in our documentation: https://docs.sensu.io/sensu-core/2.0/guides/reduce-alert-fatigue/#assigning-the-filter-to-a-handler.

echlebek:
ccing @portertech as we discussed 1.x vs 2.x occurrences yesterday.

I think @palourde is right about this. Thoughts?

portertech:
@palourde "whereas only non-zero results would go through in 1.x" is not true, 1.x OK events with check `"type": "metric"` would go through the pipeline (that annoying 1.x built in pipeline logic).

portertech:
Occurrences is typically used in combination with other attributes when filtering, such as status. I think I need to discuss this via voice @echlebek @palourde.

palourde:
@echlebek Do you want to close this PR in favour of https://github.com/sensu/sensu-go/issues/1766?

palourde:
Closing in favour of https://github.com/sensu/sensu-go/issues/1766

echlebek:
Closing in favour of #1766

PR #1699, opened 2018-06-20
Refactor/monitor

### LINKED ISSUE #1674: Replace monitor.go with a leased key and watcher
The current [monitor](https://github.com/sensu/sensu-go/blob/master/backend/monitor/monitor.go) is incompatible with a clustered sensu backend as the monitors are implemented in an in-memory map. We should instead use a key in etcd with a lease for each monitor, and a watcher for that key's expiration/deletion. We'll need to update the monitor implementation (replace the in-memory map) in keepalived and eventd once this work is done.

mercul3s:
## What is this change?

Adds an etcd backed monitor implementation.

Closes #1674 

Maybe not yet? This code is not yet integrated and doesn't change any system  behavior. Happy to add one if necessary, or can update the changelog when integrated.

Nope!

This refactor exists alongside the current implementation, and shares some naming. Due to that, there are a few long names for functions or variables that exist in both files that should be changed when we factor out the in-memory monitor.

Nope!

### DIFF
--- a/backend/monitor/monitors.go
+++ b/backend/monitor/monitors.go
@@ -3,6 +3,7 @@ package monitor
 import (
 	"context"
 	"fmt"
+	"strconv"
 
 	"github.com/coreos/etcd/clientv3"
 	"github.com/coreos/etcd/mvcc/mvccpb"
@@ -27,22 +28,39 @@ type EtcdGetter struct {
 	Client *clientv3.Client
 }
 
-// MonitorUpdateHandler provides an event update handler.
-type MonitorUpdateHandler interface {
-	HandleUpdate(e *types.Event) error
+// Service is the monitors interface.
+type Service interface {
+	// GetMonitor starts a new monitor.
+	GetMonitor(ctx context.Context, name string, entity *types.Entity, event *types.Event, ttl int64) error
 }
 
-// MonitorFailureHandler provides a failure handler.
+// Factory takes an entity and returns a Monitor interface so the
+// monitor can be mocked.
+type Factory func(*clientv3.Client, UpdateHandler, FailureHandler) Service
+
+// MonitorFailureHandler provides a failure handler. TODO: rename this to
+// FailureHandler when we remove the other monitor code.
 type MonitorFailureHandler interface {
 	HandleFailure(entity *types.Entity, event *types.Event) error
 }
 
-// Monitors is a monitor based on a leased key in etcd. Monitor leases can be
-// extended, and a watcher on the key will run a handler when the lease expires
-// and the key is deleted.
-type Monitors struct {
-	FailureHandler MonitorFailureHandler
-	UpdateHandler  MonitorUpdateHandler
+type ErrorHandler interface {
+	HandleError(error)
+}
+
+// ErrorHandler provides a handler for errors from WatchMon.
+type ErrorHandlerFunc func(error)
+
+func (e ErrorHandlerFunc) HandleError(err error) {
+	return e(err)
+}
+
+// EtcdService is an etcd backed monitor service monitor based on a leased key.
+// Monitor leases can be extended, and a watcher on the key will run a handler
+// when the lease expires and the key is deleted.
+type EtcdService struct {
+	failureHandler MonitorFailureHandler
+	errorHandler   ErrorHandler
 	client         *clientv3.Client
 }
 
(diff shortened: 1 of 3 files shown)


backend/monitor/monitors_test.go
```diff
@@ -0,0 +1,185 @@
...
+	testEvent := types.FixtureEvent(testEntity.ID, "testCheck")
+
+	handler := &testMonitorsHandler{}
+	monitorService := NewService(client, handler, handler)
+
+	err = putKeyWithLease(client, monitorPath, 15)
+	require.NoError(t, err)
+	response, err := client.Get(context.Background(), monitorPath)
+	require.NoError(t, err)
+	fmt.Println("put response in test:", response)
```
echlebek: Stray Println

backend/monitor/monitors_test.go
```diff
@@ -0,0 +1,185 @@
...
+	monitorService := NewService(client, handler, handler)
+	_, err = client.Put(context.Background(), testMon.key, fmt.Sprintf("%d", testMon.ttl))
+	require.NoError(t, err)
+
+	mon, err := monitorService.getMonitor(context.Background(), testMon.key)
+	require.NoError(t, err)
+	assert.EqualValues(t, testMon.key, mon.key)
+}
+
+func TestWatchMonDelete(t *testing.T) {
```
echlebek: Given that these tests don't rely on assertions, they probably would benefit from some comments explaining them.

backend/monitor/monitors.go
```diff
@@ -0,0 +1,179 @@
...
+
+// EtcdGetter provides access to the etcd client.
+type EtcdGetter struct {
+	Client *clientv3.Client
+}
+
+// Service is the monitors interface.
+type Service interface {
+	// GetMonitor starts a new monitor.
+	GetMonitor(ctx context.Context, name string, entity *types.Entity, event *types.Event, ttl int64) error
```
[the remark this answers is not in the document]
mercul3s: That sounds reasonable to me.

backend/monitor/monitors_test.go
mercul3s: :woman_facepalming:

backend/monitor/monitors_test.go
mercul3s: I knew I was forgetting something.

backend/monitor/monitors_test.go
nikkiki: it happens 🤷‍♀️

backend/monitor/monitors.go
```diff
@@ -0,0 +1,170 @@
...
+)
+
+var (
+	monitorPathPrefix = "monitors"
+	monitorKeyBuilder = store.NewKeyBuilder(monitorPathPrefix)
+)
+
+// Service is the monitors interface.
+type Service interface {
+	// RefreshMonitor starts a new monitor.
```
echlebek: // RefreshMonitor starts a new monitor or resets an existing monitor

Would that be more accurate?

backend/monitor/monitors.go
```diff
@@ -0,0 +1,170 @@
...
+	RefreshMonitor(ctx context.Context, name string, entity *types.Entity, event *types.Event, ttl int64) error
+}
+
+// Factory takes an entity and returns a Monitor interface so the
```
echlebek: The docs should be a little more truthy here :)

backend/monitor/monitors.go
```diff
@@ -0,0 +1,170 @@
...
+// monitor can be mocked.
+type Factory func(*clientv3.Client, UpdateHandler, FailureHandler) Service
+
+// MonitorFailureHandler provides a failure handler. TODO: rename this to
+// FailureHandler when we remove the other monitor code.
+type MonitorFailureHandler interface {
+	HandleFailure(entity *types.Entity, event *types.Event) error
+}
+
+type ErrorHandler interface {
```
echlebek: Needs docs

backend/monitor/monitors.go
```diff
@@ -0,0 +1,170 @@
...
+	HandleError(error)
+}
+
+// ErrorHandler provides a handler for errors from WatchMon.
```
echlebek: ErrorHandlerFunc implements ErrorHandler

nikkiki:
How did "Don't break the build" actually break the build 😆 
```
go : # github.com/sensu/sensu-go/backend/monitor
166At C:\gopath\src\github.com\sensu\sensu-go\build.ps1:215 char:5
167+     go test -timeout=200s -tags=integration $(go list ./... | Select- ...
168+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
169    + CategoryInfo          : NotSpecified: (# github.com/se...backend/monitor:String) [], RemoteException
170    + FullyQualifiedErrorId : NativeCommandError
171 
172backend\monitor\monitors_test.go:98:9: no new variables on left side of :=
```

Issue #1688, opened 2018-06-15
Unable to create Metrics Event via the Backend API w/o Check


portertech:
The Backend API POST /events endpoint requires Events contain a Check or they are considered invalid. This limitation stops users from sending Events containing an Entity and Metrics into Sensu via the Backend API.

## Expected Behaviour

- The Event must contain an Entity
- The Event must contain a Check or Metrics or both

## Current Behavior

- The Event must contain an Entity
- The Event must contain a Check
- The Event may contain Metrics

## Possible Solution

Update:

https://github.com/sensu/sensu-go/blob/8a73065eee1e385122b67835327e59bfcd578fe7/backend/apid/actions/events.go#L118-L120

and

https://github.com/sensu/sensu-go/blob/8a73065eee1e385122b67835327e59bfcd578fe7/backend/apid/actions/events.go#L160-L162

PR #1714, opened 2018-06-21
Allow creation of metric events via backend API

palourde:
## What is this change?

It allow users to use the backend API to submit events containing metrics, without a check.

Additionally, I removed the `Update` method for the Events controller, which was no longer used.

Closes https://github.com/sensu/sensu-go/issues/1688

Added!

Nope!

Nope!

Nope!


testing/e2e/main_test.go
```diff
@@ -19,7 +19,6 @@ var (
 	agentPortCounter                     int64 = 20000
 	agentPath, backendPath, sensuctlPath string
 	binDir                               = filepath.Join("..", "..", "bin")
-	toolsDir                             = filepath.Join(binDir, "tools")
```
echlebek: Curious why this is in the PR, seems unrelated?

palourde: Oh my bad, related to https://github.com/sensu/sensu-go/pull/1713, forgot to merge master back.

PR #1733, opened 2018-06-22
Return result of error check immediately

mercul3s:
Don't need to check that the error is not nil before returning it.

Nikki helpfully pointed out an error handling simplification, so I fixed it!


nikkiki:
Travis puked but I think rerunning it should be alright..
```
Installing deps...
# cd .; git clone https://go.googlesource.com/lint /home/travis/gopath/src/golang.org/x/lint
Cloning into '/home/travis/gopath/src/golang.org/x/lint'...
fatal: git fetch-pack: expected ACK/NAK, got '?error: Git repository not found'
fatal: The remote end hung up unexpectedly
package golang.org/x/lint: exit status 128
```

mercul3s:
That is not the first time I've hit a failure on travis due to dep. Trying again!

PR #1734, opened 2018-06-25
Fix a bug where checks created with `sensuctl create` would never fail

### LINKED ISSUE #1732: `sensuctl create` with Check type leads to aliasing between CheckConfig ExtendedAttributes and Check attributes
## Expected Behavior

I create a "failing" check and it returns a "failing" status.

## Current Behavior

I create a "failing" check and it returns a "passing" status. (WAT?)

Create a Check using `sensuctl create` with the following spec:

```
{
    "type": "Check",
    "spec": {
        "name": "should-fail",
        "command": "echo lol && exit 2",
        "subscriptions":[
            "workstation"
        ],
        "publish": true,
        "interval": 10,
        "organization": "default",
        "environment": "default"
    }
}
```

Observe related Events reporting with `"status": 0` 😡 

The Agent receives check requests containing extended attributes for the Check type attributes. The Agent then executes the Check and sends an Event with two "status" attributes 😢, one reporting the correct execution status, `2`, and the other reporting `0`.

screen shot 2018-06-22 at 1 42 27 pm

## Possible Solution

Using the type `CheckConfig` fixes the issue.

```
{
    "type": "CheckConfig",
    "spec": {
        "name": "should-fail",
        "command": "echo lol && exit 2",
        "subscriptions":[
            "workstation"
        ],
        "publish": true,
        "interval": 10,
        "organization": "default",
        "environment": "default"
    }
}
```

We need to rethink our data types and extended attributes.

echlebek:
## What is this change?

This commit fixes a bug in the dynamic marshaler that could lead to JSON objects being written out with duplicate object names. Through a series of unfortunate events, this was causing checks created with `sensuctl create` (vs `sensuctl check create`) to never fail.

This is a critical bug and after merging this PR we should do a release.

This took a while to debug. If a user created a check with the `Check` type instead of `CheckConfig`, the `Status` field of a Check would become a custom attribute, as the message would be unmarshaled into a `CheckConfig` on the other side.

This resulted in the extended attributes being blindly copied over to the Check when the agent created a Check from the CheckConfig, and shadowing the Status of the Check. As a result, the Check could never fail.

Closes #1732

### DIFF
--- a/types/check.go
+++ b/types/check.go
@@ -58,7 +58,6 @@ func NewCheck(c *CheckConfig) *Check {
 		Publish:              c.Publish,
 		RuntimeAssets:        c.RuntimeAssets,
 		Subscriptions:        c.Subscriptions,
-		ExtendedAttributes:   c.ExtendedAttributes,
 		ProxyEntityID:        c.ProxyEntityID,
 		CheckHooks:           c.CheckHooks,
 		Stdin:                c.Stdin,
@@ -72,6 +71,9 @@ func NewCheck(c *CheckConfig) *Check {
 		OutputMetricHandlers: c.OutputMetricHandlers,
 		EnvVars:              c.EnvVars,
 	}
+	dummyCheck := &Check{}
+	_ = dynamic.Unmarshal(c.ExtendedAttributes, dummyCheck)
+	check.ExtendedAttributes = dummyCheck.ExtendedAttributes
 	return check
 }
 
--- a/types/check.go
+++ b/types/check.go
@@ -71,9 +71,12 @@ func NewCheck(c *CheckConfig) *Check {
 		OutputMetricHandlers: c.OutputMetricHandlers,
 		EnvVars:              c.EnvVars,
 	}
-	dummyCheck := &Check{}
-	_ = dynamic.Unmarshal(c.ExtendedAttributes, dummyCheck)
-	check.ExtendedAttributes = dummyCheck.ExtendedAttributes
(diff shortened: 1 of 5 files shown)


types/check.go
```diff
@@ -72,6 +71,9 @@ func NewCheck(c *CheckConfig) *Check {
 		OutputMetricHandlers: c.OutputMetricHandlers,
 		EnvVars:              c.EnvVars,
 	}
+	dummyCheck := &Check{}
+	_ = dynamic.Unmarshal(c.ExtendedAttributes, dummyCheck)
+	check.ExtendedAttributes = dummyCheck.ExtendedAttributes
```
grepory: Kill dummyCheck?

types/check_test.go
```diff
@@ -341,3 +341,10 @@ func TestIsSubdued(t *testing.T) {
 		})
 	}
 }
+
```
grepory: 💪

types/check_test.go
```diff
@@ -341,3 +341,10 @@ func TestIsSubdued(t *testing.T) {
...
+func TestAliasedExtendedAttributeBehaviour_GH1732(t *testing.T) {
```
grepory: 💪

types/check.go
echlebek: No, this fixes the bug.

portertech: Can we name this something else?

echlebek: Sure, although I'm having trouble coming up with something better. Any suggestions?

Issue #1008, opened 2018-02-13
Revisit our approach to RBAC in sensuctl


palourde:
Related to https://github.com/sensu/sensu-go/pull/1004.

There are a couple of issues around RBAC in sensuctl, which lead me to believe that we need to clarify our approach.

Right now, we have the `environment` & `organization` global flags defined in `start.go`, which use the `default` value for both flags, without regard to the actual configured values in the profile: https://github.com/sensu/sensu-go/blob/15e0c19e7567824246ac7cf8fdd48670c75dd9c1/cli/cmd/start.go#L51-L52

Also, only some of our interactive prompts ask for the environment & organization. Additionally, we sometime have both `--org` & `--organization` flags defined at the same time (e.g. `sensuctl environment create` subcommand).

I think we first need to decide whether or not we want users, at least for the initial beta, to be able to manipulate resources across organizations (using flags and/or interactive questionnaire).

Considering the fact that the beta release is coming soon, I think we should remove that functionality so users can only interact with resources on their configured env & org (maybe allow something like `--all-environments` & `--all-organizations` when querying) until we improve the UX around manipulating resources across environments and organizations.

Any thoughts @grepory @portertech?

echlebek:
@grepory, @palourde, @portertech since we are about to work on enterprise RBAC plugins, do we need to address this issue?

grepory:
The RBAC drivers for enterprise aren't really related to this, but it would be nice to address this experience sooner rather than later. So--while it's not really in scope, we can still schedule it.

PR #1740, opened 2018-06-26
Use profile env & org values as default when creating resources

palourde:
## What is this change?

- The profile environment & organization values are now used by default when
creating a resource with sensuctl, instead of _default_ (which was the default value configured on the global flags). 
- Added the `sensuctl config view` subcommand.

Closes https://github.com/sensu/sensu-go/issues/1008

Added!

Nope!

Nope!

Nope!


cli/commands/config/view.go
```diff
@@ -0,0 +1,74 @@
...
+	"io"
+
+	"github.com/sensu/sensu-go/cli"
+	"github.com/sensu/sensu-go/cli/commands/helpers"
+	"github.com/sensu/sensu-go/cli/commands/hooks"
+	"github.com/sensu/sensu-go/cli/elements/list"
+	"github.com/spf13/cobra"
+)
+
+// ViewCommand ...
```
nikkiki: 😏...

palourde:
@nikkiki Added a test file!

PR #1749, opened 2018-06-27
Fix a goroutine leak in Ring.Next.

### LINKED ISSUE #1746: Ring leaks goroutines
The ring is leaking goroutines. The key watcher is selecting from the wrong context, and is never being cancelled.

Select from watchCtx instead of ctx to fix the bug.

https://github.com/sensu/sensu-go/blob/master/backend/ring/ring.go#L234

echlebek:
Next creates a goroutine that watches an etcd key for deletes.
After Next returns, the goroutine is supposed to be shut down, but
isn't, because the context it is using to determine cancellation is
the wrong one.

This commit refactors the watcher into a separate function, and
passes the correct context into it.

Closes #1746

### DIFF
--- a/backend/ring/ring.go
+++ b/backend/ring/ring.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"encoding/binary"
 	"errors"
+	"fmt"
 	"path"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


backend/ring/ring.go
```diff
@@ -238,29 +230,44 @@ func (r *Ring) Next(ctx context.Context) (string, error) {
...
+	if wc == nil {
+		return "", ctx.Err()
+	}
+	wait := notifyDeletes(watchCtx, wc)
 	for {
-		select {
-		case <-ctx.Done():
-			return "", ctx.Err()
-		default:
+		if err := ctx.Err(); err != nil {
```
palourde: I'm probably missing something (the _context_ pkg is hard haha) but shouldn't it be `watchCtx.Err()`?

echlebek: It doesn't matter too much in this case. Contexts are hierarchical, and the cancellation we are looking for comes from outside the function. watchCtx will inherit any cancellation from ctx, aside from its own CancelFunc.

However, now that you mention it, it seems a bit inconsistent to create a new context and then not use it here, so I think I will change it to watchCtx.

echlebek: I updated the PR to use a single context.

Issue #1757, opened 2018-06-28
sensuctl extension list does not properly display fields


nikkiki:
`sensuctl extension list` does not display extension name or URL properly in table format:
```
$ sensuctl extension register "extension1" "127.0.0.1:31000"
OK
$ sensuctl extension list
  Name     URL    
 ────── ───────── 
         ///.../  
$ sensuctl extension list --format json
[
  {
    "name": "extension1",
    "url": "127.0.0.1:31000",
    "organization": "default"
  }
]
```

palourde:
Huh https://github.com/sensu/sensu-go/blob/b1f5d9ab29aa8ee4febc1dab0a09422911b6cd3b/cli/commands/extension/list.go#L54

nikkiki:
@palourde excellent find 🤦‍♀️😆

PR #1758, opened 2018-06-28
Fix table formatting for 'sensuctl extension list'

palourde:

## What is this change?

It makes sure the name of extensions are properly displayed when showing all extensions.

Closes https://github.com/sensu/sensu-go/issues/1757

Added!

Nope!

Nope!

Nope!

### DIFF
--- a/cli/commands/extension/list.go
+++ b/cli/commands/extension/list.go
@@ -51,17 +51,17 @@ func printToTable(results interface{}, writer io.Writer) {
 			Title:       "Name",
 			ColumnStyle: table.PrimaryTextStyle,
 			CellTransformer: func(data interface{}) string {
-				asset, _ := data.(types.Asset)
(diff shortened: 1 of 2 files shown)


cli/commands/extension/list.go
```diff
@@ -51,17 +51,17 @@ func printToTable(results interface{}, writer io.Writer) {
...
 		{
 			Title: "URL",
 			CellTransformer: func(data interface{}) string {
-				asset, _ := data.(types.Asset)
-				u, err := url.Parse(asset.URL)
+				extension, _ := data.(types.Extension)
+				u, err := url.Parse(extension.URL)
 				if err != nil {
-					return ""
+					return extension.URL
```
palourde: I thought we should display the raw URL extension instead of nothing 🤷‍♂️

PR #1765, opened 2018-06-29
Address staging env resources to include more use-cases

nikkiki:

## What is this change?

Addresses the following in the staging env resources:
- Configure splay & splay_coverage for “proxy-check” check
- We need a check that uses TTL
- TCP/UDP handlers
- Handler set
- Replace “handle-failing-only” filter with built-in “is_incident” filter
- Two checks are using the “silenced-check” name, so the first definition is overriden.
- Create a “silenced-cat” handler, using the “not_silenced” filter so silenced events are not handled, and modify the “silenced-check” to use that handler
- Remove “cat” handler from “feed-statsd” check. This handler is unnecessary, it is handled by the statsd-metrics-handler
- Extension handlers should be of type “grpc” not “pipe”

To cover more use cases and features in test.

Nah.

Nah.

Nah.

Manual staging instructions have been updated for the tcp handler.

### DIFF
--- a/testing/tcphandler/main.go
+++ b/testing/tcphandler/main.go
@@ -5,11 +5,10 @@ import (
 	"fmt"
 	"log"
 	"net"
-	"os"
 )
 
 var (
-	port = flag.Int("port", 3333, "port to run extension server on")
+	port = flag.Int("port", 3333, "port to run tcp server on")
 	conn = flag.String("type", "tcp", "type of connection to establish (ex. tcp)")
 )
 
@@ -18,7 +17,6 @@ func main() {
 	l, err := net.Listen(*conn, fmt.Sprintf("127.0.0.1:%d", *port))
 	if err != nil {
 		log.Fatal(err)
-		os.Exit(1)
 	}
 	defer l.Close()
 	log.Print(fmt.Sprintf("Listening on %s 127.0.0.1:%d", *conn, *port))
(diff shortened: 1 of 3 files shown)


testing/tcphandler/main.go
```diff
@@ -0,0 +1,43 @@
...
+	if err != nil {
+		log.Fatal(err)
+		os.Exit(1)
+	}
+	defer l.Close()
+	log.Print(fmt.Sprintf("Listening on %s 127.0.0.1:%d", *conn, *port))
+	for {
+		conn, err := l.Accept()
+		if err != nil {
+			log.Fatal(err)
```
echlebek: log.Fatal is a log.Println followed by os.Exit :)

testing/tcphandler/main.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+func handleRequest(conn net.Conn) {
+	buf := make([]byte, 4096)
+	_, err := conn.Read(buf)
+	if err != nil {
+		log.Print("Error reading: ", err.Error())
+	}
+	log.Print(string(buf))
+	conn.Close()
```
echlebek: defer conn.Close() at the top of the function

testing/tcphandler/main.go
```diff
@@ -0,0 +1,43 @@
...
+			os.Exit(1)
+		}
+		go handleRequest(conn)
+	}
+}
+
+func handleRequest(conn net.Conn) {
+	buf := make([]byte, 4096)
+	_, err := conn.Read(buf)
+	if err != nil {
```
echlebek: Return early if err != nil

testing/tcphandler/main.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"flag"
+	"fmt"
+	"log"
+	"net"
+	"os"
+)
+
+var (
+	port = flag.Int("port", 3333, "port to run extension server on")
```
echlebek: Not an extension server

nikkiki: lol

PR #1781, opened 2018-07-05
Docker changes

grepory:
## What is this change?

This changes how we tag nightly images to be "sensu/sensu:nightly" and removes the build type from all docker tags (no more 2.0.0-alpha-alpha.1). It also moves us over the sensu/sensu from sensu/sensu-go.

We'll be flipping to 2.0 being The Sensu eventually, and I'd rather start as early as possible with a small change like the docker image.

There will be a corresponding docs change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,7 @@ or environment does not exist.
 - The profile environment & organization values are used by default when
 creating a resource with sensuctl.
 - Migrated docker image to sensu Docker Hub organization from sensuapp.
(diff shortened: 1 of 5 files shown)


palourde:
Do you want to already prepare a subsequent PR to update Dockerfile, docker-compose etc.?

grepory:
Oh I forgot about the dockerfile stuff and docker-compose. Let me add that.

palourde:
LGTM!

PR #1789, opened 2018-07-07
Use constants instead of hardcoded strings for all rules, all orgs, all envs #1285

bitmonk:
## What is this change?

First pass at implementing the change outlined in #1285 by defining necessary constants and using them throughout the code where seemed appropriate.

Constants are better than literals, there might be some risk of issues around text encoding.

Yes, added.

I am a first-time contributor to Sensu, and super open to any input on how I should engage as a contributor or if I need to sign a contributor agreement!

Nah, it was my first golang work, and I felt pretty comfy, just needed to find the right time and mood to lock it all together, had thought about this for weeks.

Would happily update documentation that folks should use these constants instead of literal splat, and would welcome input and guidance on that!

@mercul3s

### DIFF
--- a/backend/store/etcd/event_store_test.go
+++ b/backend/store/etcd/event_store_test.go
@@ -36,8 +36,8 @@ func TestEventStorage(t *testing.T) {
 		assert.EqualValues(t, event, events[0])
 
 		// Get all events with wildcards
-		ctx = context.WithValue(ctx, types.OrganizationKey, "*")
-		ctx = context.WithValue(ctx, types.EnvironmentKey, "*")
+		ctx = context.WithValue(ctx, types.OrganizationKey, types.OrganizationTypeAll)
+		ctx = context.WithValue(ctx, types.EnvironmentKey, types.EnvironmentTypeAll)
 		events, err = store.GetEvents(ctx)
 		assert.NoError(t, err)
 		assert.Equal(t, 1, len(events))
@@ -50,7 +50,7 @@ func TestEventStorage(t *testing.T) {
 
 		// Get all events from an unexisting org
 		ctx = context.WithValue(ctx, types.OrganizationKey, "acme")
-		ctx = context.WithValue(ctx, types.EnvironmentKey, "*")
+		ctx = context.WithValue(ctx, types.EnvironmentKey, types.EnvironmentKeyAll)
 		events, err = store.GetEvents(ctx)
 		require.NoError(t, err)
 		require.Equal(t, 0, len(events))
--- a/backend/store/etcd/event_store_test.go
+++ b/backend/store/etcd/event_store_test.go
@@ -50,7 +50,7 @@ func TestEventStorage(t *testing.T) {
 
 		// Get all events from an unexisting org
 		ctx = context.WithValue(ctx, types.OrganizationKey, "acme")
-		ctx = context.WithValue(ctx, types.EnvironmentKey, types.EnvironmentKeyAll)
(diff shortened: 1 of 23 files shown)


types/rbac_test.go
```diff
@@ -1,6 +1,7 @@
 package types
 
 import (
+	"sensu-go/types"
```
grepory: Just "types" here.

echlebek: @bitmonk this file is already in the `types` package, so there shouldn't be an import of `types` here, and you don't need to refer to anything with the `types`package qualifier.

Try running `go build` in the package. If the command succeeds, your imports are correct. You can also automatically manage your imports with goimports. (`go get golang.org/x/tools/cmd/goimports`)

types/rbac_test.go
```diff
@@ -1,6 +1,7 @@
 package types
 
 import (
+	"types"
```
palourde: You can remove this line; we are already in the `types` package so no need to import it, otherwise it cause Go to fail!

bitmonk: interesting! addressed, but the suite passed for me. what failure does this cause?

types/rbac_test.go
bitmonk: @echlebek thanks!

types/rbac_test.go
palourde: @bitmonk See https://travis-ci.org/sensu/sensu-go/builds/401827783

types/rbac_test.go
```diff
@@ -47,7 +47,7 @@ func TestRuleValidate(t *testing.T) {
 	assert.NoError(t, r.Validate())
 
 	// Wildcard org
-	r.Organization = "*"
+	r.Organization = types.OrganizationTypeAll
```
palourde: @bitmonk Ha, looks like we also missed that one. The tests are actually complaining about that: https://travis-ci.org/sensu/sensu-go/jobs/401827784#L584

Basically, you should do `r.Organization = OrganizationTypeAll` since it's within the same package, so no need to explicitly refer to it.

bitmonk: gotcha! thanks. :)

looks like i was only running e2e, and i should also run `./build.sh lint` ?

mercul3s: @bitmonk You can do `./build.sh quality` to run the linter and unit tests, and `./build.sh integration` for integration tests. That should run the same suite of tests that Travis is failing on.

backend/store/etcd/event_store_test.go
```diff
@@ -50,7 +50,7 @@ func TestEventStorage(t *testing.T) {
 
 		// Get all events from an unexisting org
 		ctx = context.WithValue(ctx, types.OrganizationKey, "acme")
-		ctx = context.WithValue(ctx, types.EnvironmentKey, "*")
+		ctx = context.WithValue(ctx, types.EnvironmentKey, types.EnvironmentKeyAll)
```
palourde: Looks like `types.EnvironmentKeyAll` does not exist, I assume it's a typo and should be `types.EnvironmentTypeAll`?

types/rbac_test.go
palourde: Otherwise you could just run `./build.sh`, without any argument, if you want to keep it simple; it will run every possible tests and also install all dependencies and tools for the tests.

backend/store/etcd/event_store_test.go
bitmonk: easy enough, and kinda not surprised I missed this in the course of typing EnvironmentTypeAll like 100 times.

PR #1794, opened 2018-07-09
[Web UI] Complete checks list styles

### LINKED ISSUE #1678: [Web UI] Checks Page: Filtering
As a user I should be able to provide a query (in the format of an SQE) to filter the results of the list.

10xjs:
## What is this change?

Finish styles for checks page list.

Standardize and simplify much of the shared style and behavior between the different list views.

image

Closes #1678

no

no

no

no


dashboard/src/components/partials/ListItem/ListItem.js
```diff
@@ -0,0 +1,145 @@
...
+        </TableCell>
+        <TableCell className={classes.cell} padding="none">
+          <div className={classes.flex}>
+            {icon && <div className={classes.icon}>{icon}</div>}
+            <div className={classes.item}>
+              <div className={classes.title}>{title}</div>
+              <div className={classes.details}>{details}</div>
+            </div>
+          </div>
+        </TableCell>
```
jamesdphillips: As per previous IRL discussion: it would be nice if we were to decompose ListItem a little. Ideally the title and details are broken down into their own cell component and a row could be composed of more than one cell.

PR #1818, opened 2018-07-10
Use clustered sensu in docker-compose.yaml

echlebek:
This commit adds two additional backend nodes to the docker compose
as well as two additional agents. Each agent connects to one
backend.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,8 +1,8 @@
 version: '3'
 services:
-  sensu-backend1:
+  backend1:
     image: sensu/sensu:master
-    command: sensu-backend start --listen-client-urls http://0.0.0.0:2379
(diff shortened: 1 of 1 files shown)


docker-compose.yaml
```diff
@@ -1,19 +1,47 @@
 version: '3'
 services:
-  sensu-backend1:
+  backend1:
     image: sensu/sensu:master
```
nikkiki: Out of curiosity, why are we using `sensu/sensu:master` in some backends and agents, and `sensu/sensu-go:master` in others?

echlebek: Oh, that's a mistake, it should be sensu/sensu:master in all cases.

nikkiki: No problem! I think it's just L14 and L24: https://github.com/sensu/sensu-go/blob/7dace965851e6f53f54c2b43dba102f6f649ff04/docker-compose.yaml#L14 and https://github.com/sensu/sensu-go/blob/7dace965851e6f53f54c2b43dba102f6f649ff04/docker-compose.yaml#L24

PR #1826, opened 2018-07-12
Improve TestExecuteCheck

### LINKED ISSUE #1825: [Build stability] Intermittent failure in TestExecuteCheck
```
--- FAIL: TestExecuteCheck (7.85s)
	assertions.go:239: 
                          
	Error Trace:	check_handler_internal_test.go:136
		
	Error:      	Not equal: 
		
	            	expected: 2
		
	            	actual: 0
panic: runtime error: index out of range [recovered]
panic: runtime error: index out of range
```

nikkiki:

## What is this change?

Improves TestExecuteCheck, add debug print statement in the event future failures happen, avoid panic by using `require`.

Closes #1825 

Nah.

Nah.

🤞 Hoping the behavior is the same on windows.

### DIFF
--- a/agent/check_handler_internal_test.go
+++ b/agent/check_handler_internal_test.go
@@ -2,6 +2,7 @@ package agent
 
 import (
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -117,7 +118,7 @@ func TestExecuteCheck(t *testing.T) {
 	metrics := "metric.foo 1 123456789\nmetric.bar 2 987654321"
 	f, err := ioutil.TempFile("", "metric")
 	assert.NoError(err)
-	err = ioutil.WriteFile(f.Name(), []byte(metrics), 0644)
+	_, err = fmt.Fprintln(f, metrics)
 	assert.NoError(err)
 	f.Close()
 	defer os.Remove(f.Name())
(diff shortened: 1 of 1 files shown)


agent/check_handler_internal_test.go
```diff
@@ -117,7 +118,7 @@ func TestExecuteCheck(t *testing.T) {
 	metrics := "metric.foo 1 123456789\nmetric.bar 2 987654321"
 	f, err := ioutil.TempFile("", "metric")
 	assert.NoError(err)
-	err = ioutil.WriteFile(f.Name(), []byte(metrics), 0644)
+	_, err = fmt.Fprintln(f, metrics)
 	assert.NoError(err)
```
echlebek: require.NoError here

agent/check_handler_internal_test.go
```diff
@@ -133,7 +134,10 @@ func TestExecuteCheck(t *testing.T) {
 	assert.NoError(json.Unmarshal(msg.Payload, event))
 	assert.NotZero(event.Timestamp)
 	assert.True(event.HasMetrics())
-	assert.Equal(2, len(event.Metrics.Points))
```
echlebek: require.Equal(2, len(event.Metrics.Points), string(event)) should print it out in case of failure

nikkiki: ahh 😅 that is helpful! `msgAndArgs ...interface{}`

PR #1834, opened 2018-07-16
Add enterprise feature flag to sensuctl

palourde:
## What is this change?

It basically adds a configuration value in the cluster config that indicates the Sensu edition, which can be either `core` or `enterprise`, and that acts as a feature flag. This value is configured whenever the user runs `sensuctl configure` and if it's not configured, we use `core` as the default value. 

More specifically:
- We annotate the enterprise commands to mark them as enterprise
- We also mark these commands as hidden unless we have an enterprise cluster
  - A hidden command can still be used, which is okay I believe because the Core API will not implement enterprise routes
- If we have an enterprise cluster, the help usage will display the specific enterprise commands:

```
$ sensuctl -h
sensuctl controls Sensu instances

Usage:	sensuctl COMMAND

Flags:
[...]

Commands:
[...]

Management Commands:
[...]

Enterprise Commands:
  license      Manage enterprise license
```

Closes https://github.com/sensu/sensu-enterprise-go/issues/42

Added!

Nope!

Nope!

Nope!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Added
 - Added unit test coverage for check routers.
+- Added an enterprise feature flag in sensuctl.
 
 ### Changed
 - The Backend struct has been refactored to allow easier customization for the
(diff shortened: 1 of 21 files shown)


palourde:
Closing this PR, see https://github.com/sensu/sensu-enterprise-go/issues/42.

PR #1850, opened 2018-07-19
Add sensuctl cluster member-add command.

### LINKED ISSUE #1719: Add sensuctl cluster member add command
Add a command to sensuctl that will allow an administrator to add a member to a sensu cluster.

The command should be backed by support in apid. See https://godoc.org/github.com/coreos/etcd/clientv3#Cluster for more information on how to add members to an etcd cluster with etcd clientv3.

echlebek:

## What is this change?

Closes #1719

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 - Added unit test coverage for check routers.
 - Added API support for cluster management.
 - Added sensuctl cluster member-list command.
+- Added sensuctl cluster member-add command.
 
 ### Changed
 - The Backend struct has been refactored to allow easier customization in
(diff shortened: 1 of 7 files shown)


echlebek:
@palourde, cluster operators need a way to modify existing clusters, and that is the intent of the commands we are adding. For instance, you start with a 3 node cluster, and one node becomes unstable, so you remove it from the cluster, and add another.

The difference is the initial cluster state: existing vs new.

See the docs here: https://coreos.com/etcd/docs/latest/etcd-live-cluster-reconfiguration.html

echlebek:
I've amended the commit with a (hopefully) more clear description for the command.

oferrigni:
LGTM

echlebek:
Thanks for the prompt reviews everyone!

PR #1852, opened 2018-07-20
Added support for the standard day period type.

phistuck:
Chrome has a bug - it returns the type as dayperiod instead of dayPeriod and this is probably going to change. Support both of the strings for now.


echlebek:
Hi @phistuck, thanks for the PR.

All developers that contribute to sensu-go need to sign their commits with DCO. You can do this with the git tool.

You can use the following command to amend this commit with sign-off:
```
git commit --amend -s
```

phistuck:
Done.

PR #1857, opened 2018-07-22
Issue 1730: Tools (and related infra) to start posting CircleCI artifacts to S3

preed:
This PR:

* Starts using the `Makefile` to build, instead of direct `build.sh` calls previously used
* Posts the created artifacts to the `sensu-ci-builds` S3 bucket


preed:
Note: this build is red because it was kicked off via SSH... so I had to cancel it when it was done, which is marked as an error.

preed:
There is one part of functionality that I thought was in this PR, but which is not: for CI builds, we want to use S3's lifecycle policies to delete builds older than some time (60 days?); unfortunately, lifecycles work on object _tags_, not object _metadata_, so https://github.com/sensu/sensu-release/blob/master/post-packages-s3.sh#L38 doesn't actually do what we want.

I can work around it using `awscli`'s `s3api` command, but it's kinda messy. It's a shame that `awscli` doesn't have this built in, like `s3cmd` does...

PR #1864, opened 2018-07-23
Export store within the backend

nikkiki:

## What is this change?

Exports store within the backend.

Required by https://github.com/sensu/sensu-enterprise-go/issues/48

Nah.

Nah.

Nah.


oferrigni:
Looks good to me

PR #1867, opened 2018-07-24
[Web UI] Ensure silenced state is reflected where applicable

### LINKED ISSUE #1858: [Web UI] Reflect whether a event / check has been silenced

jamesdphillips:
## What is this change?

Ensure silenced state is reflected where applicable.

screen shot 2018-07-24 at 10 59 04 am

Closes #1858 

`undefined`

`undefined`

`undefined`

`undefined`

### DIFF
--- a/backend/apid/graphql/schema/entity.graphql
+++ b/backend/apid/graphql/schema/entity.graphql
@@ -35,6 +35,12 @@ type Entity implements Node {
   "All events associated with the entity."
   events(orderBy: EventsListOrder = SEVERITY): [Event!]!
 
+  "isSilenced return true if the entity has any silences associated with it."
+  isSilenced: Boolean!
+
+  "All silences associated with the entity."
+  silencedBy: [Silenced!]!
+
   "Extended attributes includes arbitrary user-defined data"
   extendedAttributes: JSON!
 }
--- a/backend/apid/graphql/schema/entity.graphql
+++ b/backend/apid/graphql/schema/entity.graphql
@@ -38,8 +38,11 @@ type Entity implements Node {
   "isSilenced return true if the entity has any silences associated with it."
   isSilenced: Boolean!
 
-  "All silences associated with the entity."
-  silencedBy: [Silenced!]!
+  """
+  All silences matching the entity's subscriptions and where the silence
+  matches all checks.
+  """
+  silences: [Silenced!]!
 
   "Extended attributes includes arbitrary user-defined data"
   extendedAttributes: JSON!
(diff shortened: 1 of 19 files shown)


backend/apid/graphql/schema/entity.graphql
```diff
@@ -35,6 +35,12 @@ type Entity implements Node {
   "All events associated with the entity."
   events(orderBy: EventsListOrder = SEVERITY): [Event!]!
 
+  "isSilenced return true if the entity has any silences associated with it."
+  isSilenced: Boolean!
+
+  "All silences associated with the entity."
+  silencedBy: [Silenced!]!
```
dabria: I'm having a hard time wrapping my head around the naming. I'm not sure if this is something you've come up with or it's established, so forgive any ignorance.

`silencedBy` makes it seem like it's an array of sources that have silenced this entity. Is that the case? Or are we listing the silenced events that belong to this entity?

dashboard/src/components/partials/ChecksList/ChecksListItem.js
```diff
@@ -32,7 +32,6 @@ class CheckListItem extends React.Component {
       fragment ChecksListItem_check on CheckConfig {
         name
         command
-        source
```
dabria: Just confirming this single line removal was intentional?

backend/apid/graphql/schema/entity.graphql
jamesdphillips: > I'm having a hard time wrapping my head around the naming. I'm not sure if this is something you've come up with or it's established, so forgive any ignorance.

This is something I've come up with.

> `silencedBy` makes it seem like it's an array of sources that have silenced this entity. Is that the case? Or are we listing the silenced events that belong to this entity?

The former is the intent, however, I can see your point about it being unclear. Another possibility that the field's name might suggest is that it may be a list of user's that silenced the entity.

I think I may switch it to `silences` to match the other relationships and expand the field's description for clarity.

dashboard/src/components/partials/ChecksList/ChecksListItem.js
jamesdphillips: I was. `source` was deprecated causing this line upset the linter.

backend/apid/graphql/schema/entity.graphql
dabria: If it's the former then I think it's clear enough, unless you really want to switch it. I just don't have the same context around the relationships and my monkey brain was just wanting to be clear.

PR #1868, opened 2018-07-24
[Web UI] Add option to filter out silenced events from list

### LINKED ISSUE #1576: [Web UI] Events Page: Filter by silenced
As a user I should be able to hide events with silenced checks or entities.

--- 

screen shot 2018-05-22 at 2 45 52 pm

jamesdphillips:
## What is this change?

Add option to filter out silenced events from list.

Closes #1576 

`undefined`

`undefined`

Adding another filter to the list header made things untenable when the viewport was narrow. To solve, I re-worked the collapsable menu component to work with nested menus.

`undefined`


dabria:
Does it make sense to have an indicator of some sort to see what you've chosen to hide? Like a check mark or having the text highlighted inside the dropdown (whatever works)? I just imagine a scenario where I forget I've chosen to hide silenced and I scroll awhile to see if any are appearing, eventually coming to the realization. It's not a huge problem, but it was an initial thought I had.

dashboard/src/components/partials/CollapsingMenu/Button.js
```diff
@@ -0,0 +1,59 @@
...
+
+    return (
+      <Item
+        pinned={pinned}
+        renderMenuItem={({ close }) => {
+          const onClick = ev => {
+            if (onClickProp) {
+              onClickProp(ev);
+            }
+            close(ev); // TODO: Don't automatically handle this.
```
dabria: TODO to be handled in another PR?

jamesdphillips:
> Does it make sense to have an indicator of some sort to see what you've chosen to hide? Like a check mark or having the text highlighted inside the dropdown (whatever works)?

@dabria Agreed. A selection state is a glaring omission in the current implementation of the list filters. Unfortunately, getting it "right" is a slightly more effort than I think is worth investing at this very moment. (Or, at least in this set of changes.)

To try to give an example of the effort involved: imagine a user selecting "silenced" from the "hide" menu. To reflect that the "silenced" option was selected we could store the selection in the component's state/context/store what have you. However, if the user then changes the query in the search box above, we then need to either clear that state entirely (inconsistent), or try to determine if the user modified the portion of the query associated with that the selection (involved) and store the new state (complex).

To me, I think the right solution would be to parse the query and have the selection state determined by matching associated tokens. By not having the state into two places it is now less complex and by not having to the clear the selection state when the query changes, we are less likely to arrive at an inconsistent state.

This approach would also give us the ability to "compose" queries instead of simply swapping the filter as we current do.

At the moment though we haven't invested any time into parsing the query expressions on the frontend. As it stands while the lack of a selection state sorta sucks, I don't think is a significant determent to the use of filtering.

jamesdphillips:
> Basically, when arriving to the events page, I only see the event with a warning status. If I filter out the silenced events, now I can see events with status passing.

@palourde hmm! (Assuming that "check-random" is silenced) this is "expected" but I do see your point that it is sorta strange behaviour. The default query when you land on the events page is `IsIncident`, which is to say by default all events with the status of `0` are omitted.

I suppose their may be no harm in including the passing events by default.

PR #1881, opened 2018-07-28
Add CircleCI build status badge.

preed:
Well, that was easy.


preed:
Regarding Slack integrations, I'm not able to do that, because I don't have admin access on Slack; I will ping @grepory or @oferrigni about this on Monday; I won't need admin access, but I think setting that up requires a two-step process (one on the Slack side, and one on the CircleCI side), so we can pair up to get it done.

From my understanding, it, too, should be pretty simple.

grepory:
I setup the slack integration today.

PR #1885, opened 2018-07-30
Sensuctl/health

### LINKED ISSUE #1717: Add sensuctl health command
Add a command to sensuctl called `health`. The command should retrieve information from the health endpoint and display it to the user.

Depending on the information added in #1716, this may optionally include cluster health information.

mercul3s:
## What is this change?

Adds health command to sensuctl, displays cluster health output:
```
$ sensuctl health
         ID            Name     Error   Healthy  
 ────────────────── ────────── ─────── ───────── 
  3437d4749d114afe   backend3   <nil>   true     
  37807cb0bf7500f6   backend1   <nil>   true     
  ad3247ac1d3065f5   backend2   <nil>   true  
```

Closes #1717 

Indeedly doodly.

Nope!

Nope!

I think existing sensuctl documentation _should_ cover this change.


cli/commands/health/LICENSE
```diff
@@ -0,0 +1,20 @@
+Copyright (c) 2017 Sensu Inc.
```
oferrigni: 2018?

palourde:
@mercul3s The new command appears under the `Commands` section, was that intended?

```
Commands:
  completion   Output shell completion code for the specified shell (bash or zsh)
  configure    Initialize sensuctl configuration
  create       create new resources from file or STDIN
  health       get sensu health status <-----------------------------------------
  logout       Logout from sensuctl
  version      Show the sensuctl version information

Management Commands:
  asset        Manage assets
  check        Manage checks
  cluster      manage sensu cluster
  config       Modify sensuctl configuration
  entity       Manage entities
  environment  Manage environments
  event        Manage events
  extension    Manage extension registry
  filter       Manage filters
  handler      Manage handlers
  hook         Manage hooks
  mutator      Manage mutators
  organization Manage organizations
  role         Manage roles
  silenced     Manage silenced subscriptions and checks
  user         Manage users
```

mercul3s:
@palourde yeah, @echlebek had a similar comment, so I'm going to move this under the cluster command.

echlebek:
While you're at it, would you mind correct the capitalization of the `  cluster      manage sensu cluster` command help? My OCD senses are tingling.

cli/commands/cluster/member-list.go
```diff
@@ -33,7 +33,7 @@ func MemberListCommand(cli *cli.SensuCli) *cobra.Command {
 	return cmd
 }
 
-func printToTable(result interface{}, w io.Writer) {
+func printMemberListToTable(result interface{}, w io.Writer) {
 	memberList := result.(*clientv3.MemberListResponse)
```
nikkiki: Should we check for an error on this type assertion? `memberList, ok := result.(*clientv3.MemberListResponse)`

cli/commands/health/health.go
```diff
@@ -0,0 +1,67 @@
...
+				return err
+			}
+			return helpers.Print(cmd, cli.Config.Format(), printHealthToTable, nil, result)
+		},
+	}
+	return cmd
+}
+
+func printHealthToTable(result interface{}, w io.Writer) {
+	table := table.New([]*table.Column{
```
nikkiki: Same thing here, maybe check that `clusterHealth, ok := data.(*types.ClusterHealth)` is ok since we depend on that in each column?

cli/commands/cluster/health_test.go
```diff
@@ -0,0 +1,30 @@
...
+	cli := newCLI()
+	cmd := HealthCommand(cli)
+
+	assert.NotNil(cmd, "cmd should be returned")
+	assert.NotNil(cmd.RunE, "cmd should be able to be executed")
+	assert.Regexp("health", cmd.Use)
+	assert.Regexp("get sensu health status", cmd.Short)
+}
+
+func newCLI() *cli.SensuCli {
```
nikkiki: I think this function is repeated across commands. Perhaps we can move it to the `helpers` package to avoid code duplication?

cli/commands/cluster/health.go
```diff
@@ -0,0 +1,67 @@
...
+				return err
+			}
+			return helpers.Print(cmd, cli.Config.Format(), printToTable, nil, result)
+		},
+	}
+	return cmd
+}
+
+func printToTable(result interface{}, w io.Writer) {
+	table := table.New([]*table.Column{
```
nikkiki: Same thing here, maybe check that clusterHealth, ok := data.(*types.ClusterHealth) is ok first, since we depend on that in each column below?

cli/commands/cluster/health_test.go
mercul3s: That sounds reasonable, but I'd want to do it in a different PR because refactoring that would touch all the cli tests.

PR #1892, opened 2018-08-01
Add a deadline to extension methods.

echlebek:
This commit adds a hardcoded 5 second deadline to extension
methods. It has been observed that extensions can cause a lockup
in the backend, and this commit is designed to mitigate that.

More investigation is needed into why extensions are locking up in
the first place, but this is a starting point that makes the
situation less critical.

## What is this change?

Placeholder PR for testing. Do not merge.

### DIFF
--- a/rpc/extension_executor.go
+++ b/rpc/extension_executor.go
@@ -4,10 +4,13 @@ import (
 	"context"
 	"errors"
 	"sync"
+	"time"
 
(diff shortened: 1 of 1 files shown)


echlebek:
The panic that's posted above is a result of a different bug. I've filed https://github.com/sensu/sensu-go/issues/1893 for tracking it.

PR #1894, opened 2018-08-01
Improve Sensu edition header handling

palourde:
## What is this change?

It inspects the headers returned by the Sensu API on each request to verify the Sensu edition, and updates sensuctl configuration if required.

It ensures a user has access to the enterprise version of sensuctl whenever they upgrade Sensu Backend.

Added

Nope

Nope

Nope

### DIFF
--- a/cli/client/client.go
+++ b/cli/client/client.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/go-resty/resty"
 	"github.com/sensu/sensu-go/cli/client/config"
+	"github.com/sensu/sensu-go/types"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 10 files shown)


cli/client/client.go
```diff
@@ -113,6 +114,23 @@ func New(config config.Config) *RestClient {
 		return nil
 	})
 
+	// Verify the Sensu edition and update the sensuctl configuration if required
+	restyInst.OnAfterResponse(func(c *resty.Client, resp *resty.Response) error {
+		// Retrieve the Sensu edition from the response header
+		headerEdition := resp.Header().Get(types.EditionHeader)
+		if headerEdition == "" {
```
nikkiki: Will there ever be a case that we don't return an edition header?

palourde: I could see the case where someones run a recent sensuctl binary with an older sensu-backend.

palourde:
@nikkiki Indeed, showing the current edition with `sensuctl config view` would be useful!

cli/client/client.go
nikkiki: Ah good call!

palourde:
@nikkiki `config view` subcommand has been updated!

PR #1895, opened 2018-08-01
Gracefully handle type assertions in cli

nikkiki:

## What is this change?

Gracefully handles type assertions in cli list commands. Discovered in https://github.com/sensu/sensu-go/pull/1885.

Avoid panics in type assertions. This shouldn't happen the way we call these print functions, but its probably safer that we check first.

Nah.

Nah.

Nah.

We good.


nikkiki:
@echlebek I agree and I've updated the empty strings with `TypeError` instead.

PR #1897, opened 2018-08-01
RBAC bugfixes

palourde:
## What is this change?

This PR introduces several bug fixes for RBAC, which I discovered with Jason. Each bug fix has its own commit for simplicity.

- I renamed the `sensuctl role list-rules` subcommand to `sensuctl role info` so it's consistent with other commands. I also added an alias to this subcommand for backward compatibility so you can still run the subcommand with its previous name.
- The allowed values for a rule type are now directly displayed in the help usage in `sensuctl role create` & `sensuctl role add-rule`, so users don't have to look in the docs.
- The rules are now implicitly granting read access to their organization and environment, so a user can list the environments & organization it has access to without having to explicitly give these permissions.

It fixes small UX problems we discovered while doing a demo of RBAC.

Added

Nope

Nope

I will make a PR to sensu-docs to document the implicit read access behaviour.

### DIFF
--- a/cli/commands/role/info.go
+++ b/cli/commands/role/info.go
@@ -13,11 +13,12 @@ import (
 	"github.com/spf13/cobra"
 )
 
-// ListRulesCommand defines new command to list rules associated w/ a role
-func ListRulesCommand(cli *cli.SensuCli) *cobra.Command {
+// InfoCommand defines new command to list rules associated w/ a role
+func InfoCommand(cli *cli.SensuCli) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:          "list-rules [ROLE]",
-		Short:        "list rules associated with a role",
+		Use:          "info [ROLE]",
+		Aliases:      []string{"list-rules"}, // backward compatibility
+		Short:        "show detailed role information",
 		SilenceUsage: true,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if len(args) != 1 {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ enterprise edition.
 the enterprise edition.
 - Upgrade dep to v0.5.0
 - Added cluster health information to /health endpoint in sensu-backend.
+- Rename list-rules subcommand to info in sensuctl role commmand with alias
+for backward compatibility.
 
 ### Fixed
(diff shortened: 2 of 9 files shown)


palourde:
@obfuscurity So a user with a single role that explicitly grants read-only access to events in the `default` organization & `default` environment will be able to see them when running `sensuctl environment list` & `sensuctl organization list` but they will not have access to any other resources within this organization/environment.

For example, if I have the following role `read-events`, that grants `read` access to `events`:

```
$ sensuctl role info read-events
   Type     Org.      Env.     Permissions
 ──────── ───────── ───────── ─────────────
  events   default   default   read
```

Any user with this role will be able to see the environment and its organization and there implicitly granted by the role:

```
$ sensuctl environment list
  Organization    Name         Description
 ────────────── ───────── ─────────────────────
  default        default   Default environment
```

But it ends there, the user will not have access to any resources within the environment unless explicitly granted:

```
$ sensuctl check list # check can't be read
  Name   Command   Interval   Cron   Timeout   TTL   Subscriptions   Handlers   Assets   Hooks   Publish?   Stdin?   Metric Format   Metric Handlers
 ────── ───────── ────────── ────── ───────── ───── ─────────────── ────────── ──────── ─────── ────────── ──────── ─────────────── ─────────────────

$ sensuctl-ro event list # event can be read
     Entity          Check       Output   Status   Silenced             Timestamp
 ────────────── ─────────────── ──────── ──────── ────────── ───────────────────────────────
  scotch.local   check-random                  1   true       2018-07-27 10:25:07 -0400 EDT
```

Makes sense?

obfuscurity:
Yes, that makes sense.

cli/commands/role/info.go
```diff
@@ -13,11 +13,12 @@ import (
...
 
-// ListRulesCommand defines new command to list rules associated w/ a role
-func ListRulesCommand(cli *cli.SensuCli) *cobra.Command {
+// InfoCommand defines new command to list rules associated w/ a role
+func InfoCommand(cli *cli.SensuCli) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:          "list-rules [ROLE]",
-		Short:        "list rules associated with a role",
+		Use:          "info [ROLE]",
+		Aliases:      []string{"list-rules"}, // backward compatibility
```
nikkiki: TIL about cobra aliases 👍

PR #1901, opened 2018-08-01
Only allow creating gRPC handlers with /extensions

### LINKED ISSUE #1893: Users should only be able to create gRPC handlers with the /extensions route
Currently, users can accidentally create gRPC handlers incorrectly with `sensuctl create` by specifying `"type": "grpc"` in the request body.

This is not how gRPC handlers are created - they are supposed to be created through the extensions API.

This can result in panics in the backend.

Make sure this state of affairs cannot occur.

echlebek:
This commit restricts users from accidentally creating invalid
gRPC handlers by checking the /handlers API for handlers of type
"grpc".

gRPC handlers should only be created with the /extensions API.

Closes #1893

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ generated correctly.
 future.
 - Fixed a bug where clustered round robin check execution executed checks
 too often.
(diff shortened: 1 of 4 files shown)


cyphus:
LGTM!

PR #1910, opened 2018-08-01
[Web UI] Use cron value in check list

### LINKED ISSUE #1909: [Web UI] Description of check's scheduled using cron field is incorrect
## Expected Behavior

Properly reflect state.

## Current Behavior

Check's scheduled using cron field are said to be "executed every 0 seconds".

## Possible Solution

Use `<CronDescription />` component.
 
1. Create check with `cron` field set.
2. View in check's list.

jamesdphillips:
## What is this change?

Reflect cron value in the check list.

screen shot 2018-08-01 at 4 32 10 pm

Fixes #1909 

Introduced after last release.

`undefined`

`undefined`

`undefined`

### DIFF
--- a/dashboard/src/components/partials/CheckDetailsContainer/CheckDetailsContainer.js
+++ b/dashboard/src/components/partials/CheckDetailsContainer/CheckDetailsContainer.js
@@ -102,7 +102,7 @@ class CheckDetailsContainer extends React.PureComponent {
     if (interval > 0) {
       return `Every ${interval}s`;
     } else if (cron && cron.length > 0) {
-      return <CronDescriptor expression={cron} />;
+      return <CronDescriptor capitalize expression={cron} />;
     }
     return "Never";
   }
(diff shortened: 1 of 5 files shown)


palourde:
@jamesdphillips Somehow I can't reproduce this on my local machine. I'm running Chrome with disabled cache on this branch, and I get the following:

screen shot 2018-08-02 at 10 58 54 am

```
 $ sensuctl check info check-cron
=== check-cron
Name:            check-cron
Interval:        0
Command:         echo 0
Cron:            * * * * *
Timeout:         0
TTL:             0
Subscriptions:   linux
Handlers:
Runtime Assets:
Hooks:
Publish?:        true
Stdin?:          false
Proxy Entity ID:
Organization:    default
Environment:     default
Metric Format:
Metric Handlers:
```

jamesdphillips:
@palourde isn't `every minute` correct given the cron statement? Prior to this change it would have simply stated `every 0s seconds`.

palourde:
Ah I see, thanks

jamesdphillips:
My PR description wasn't very clear, sorry about that.

PR #1915, opened 2018-08-02
[Web UI] Tweaks & minor fixes for issues found in QA session

### LINKED ISSUE #1918: Don't display LastSeen for proxy entities in the API
The LastSeen value for proxy entities is currently wrong; currently, it will display "49 years ago".

We need to tackle https://github.com/sensu/sensu-go/issues/1434 before this will be resolved.

Until it's resolved, let's not display the LastSeen value at all.

jamesdphillips:
## What is this change?

Quick tweaks and fixes for nits encountered during QA and VQA sessions.

- `<RelativeDate />` Always use current time when calculating delta.
- Use `<ButtonIcon />` everywhere to ensure spacing is always consistent.
- Fixup Preferences menu and remove unused theme.
- Enable backdrop filter not just for preferences menu but for all modals.
- Move entity subscriptions out of the details page's caption and into the table to match checks page.
- Only display a check's extended attributes if they are present.
- Enable polling on all pages. With some tweaks to the severity sorting and list items it isn't quite as disruptive as it once was. In time we'll see how user's feel however...
- Collapse list header items on narrow screen sizes.
- *Bug* fix delete confirmation dialog on list pages. 
- *Bug* fix vertical disclosure menu on the silences list.
- Sets default page size to 25; with polling we start to see negative perf with larger page sizes. Friendly default for most users; option to increase page size is intact.
- Ensure that platform and platformFamily pair is properly displayed on the Entity details page. Previously if no platformFamily field was present user might see `linux / `.
- Add bottom margin to all list pages to be consistent.
- Proxy entities will never have the lastSeen field set, as such hide it when the value is `null`. Closes #1918.

Improves overall experience.

`undefined`

`undefined`

`undefined`

`undefined`


dashboard/src/components/RelativeDate/RelativeDate.js
```diff
@@ -67,10 +67,10 @@ class RelativeDate extends React.Component {
 
   render() {
     const { dateTime, capitalize, unit, style, ...props } = this.props;
-    const { now } = this.state;
 
     const formatter = new IntlRelativeFormat("en", { style });
     const dateValue = new Date(dateTime);
+    const now = new Date();
```
10xjs: This makes render impure. The correct way to have the component always render relative to the current date is to store `now` in state and update it on a timer.

It might be useful to create a util helper component to provide the current date to its children:

```jsx
<CurrentDateProvider interval={1000}>
 {(now) => <RelativeDate now={now} dateTime={dateTime}/>}
</CurrentDateProvider>
```

This could solve the time-shift problem for any component that references the current date.

jamesdphillips: good idea 👍

jamesdphillips: updated

PR #1925, opened 2018-08-03
Remove linting from build.sh.

echlebek:
We're going to start linting PRs with golangci, instead of doing
linting in build.sh. This should improve build reliability by
removing the need to `go get` packages.

Also remove the `deps` build command.


grepory:
I'll hazard a guess that this will work based on the following diff:

https://github.com/sensu/sensu-go/pull/1940/files#diff-76eab2536f3d96776474d7553fbdac4c

:)

PR #1939, opened 2018-08-08
Dockerfile: add etcdctl

schu:
The Sensu operator (work in progress) expects `etcdctl` to be available
in the base image. `etcdctl snapshot restore ...` is used from an init
container when a Sensu cluster is restored from a backup.

https://github.com/kinvolk/sensu-operator/blob/6e46ed44cc6a09daad45ad8557012819290b8cc9/pkg/util/k8sutil/k8sutil.go#L127

Also, update the Alpine base image (3.6 -> 3.8)

Signed-off-by: Michael Schubert <michael@kinvolk.io>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,12 @@
-FROM alpine:3.6
+FROM alpine:3.8 as etcdctl-fetch
+
+RUN apk add --no-cache curl
+
+RUN curl -fsSLO https://github.com/coreos/etcd/releases/download/v3.3.9/etcd-v3.3.9-linux-amd64.tar.gz && \
+  tar xvf etcd-v3.3.9-linux-amd64.tar.gz --strip-components 1 etcd-v3.3.9-linux-amd64/etcd etcd-v3.3.9-linux-amd64/etcdctl
+
+
+FROM alpine:3.8
 MAINTAINER Sensu, Inc. Engineering <engineering@sensu.io>
 
 LABEL name="sensu/sensu" \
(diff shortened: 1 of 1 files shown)


schu:
I'm not sure how a restore via API would work: for the API to be available Sensu needs to be running but etcd can be started only after the snapshot was restored (as far as I know): https://coreos.com/etcd/docs/latest/op-guide/recovery.html#restoring-a-cluster

Do you have something in mind how to do that?

Maybe a `sensu-backend --restore <snapshot.db> <other args>...` flag would be an option to tell Sensu to first restore a snapshot and then start etcd.

grepory:
Oh! My bad. I actually didn't even think about restore having to be offline.

This is an interesting one. I actually like the `--restore` option idea, but I'd almost rather we just have a command that didn't initiate startup since this is something that needs to happen outside of Kubernetes and must work well with service management frameworks. E.g. the UX would be something like:

```
sudo -U sensu sensu-backend restore --snapshot /path/to/snapshot
sudo service restart sensu-backend
```

Or something. Regardless, I'll throw this in our product backlog, and we can carry on with etcdctl for now.

As far as online snapshotting goes:

We could have an endpoint for creating a snapshot in the API, but we would have to figure something else out for restoring a snapshot. Sensu could actually do an online snapshot restore by shutting down etcd temporarily, but we would have to have the [relevant bits](https://github.com/coreos/etcd/blob/master/clientv3/snapshot/v3_snapshot.go#L230) in Sensu to do the snapshot restore (which isn't a problem).

I'll say that's out of scope for the time being, and we can run with just etcdctl. Don't worry about filing tickets.

PR #1941, opened 2018-08-08
Fixup redirect loop encountered when backend is missing dashboard

### LINKED ISSUE #1908: [Web UI] Redirect loop encountered when backend is built without dashboard
/cc @amdprophet 

## Expected Behavior

User should be presented with the fallback message describing to the user that the dashboard isn't present in their current build.

## Current Behavior

When trying to navigate to the dashboard the user encounters a redirect loop.

1. Build backend without dashboard.
2. Navigate to `http://backendhost:dashboardport`.

jamesdphillips:
## What is this change?

Eliminates redirect loop encountered when backend is missing dashboard and user attempts to navigate to the root path.

Closes #1908 

Primarily DX related so it's unlikely.

nada

nada

nada

### DIFF
--- a/dashboard/fallback.go
+++ b/dashboard/fallback.go
@@ -13,24 +13,35 @@ type fallbackFS struct {
 }
 
 func (f *fallbackFS) Open(path string) (http.File, error) {
+	if len(path) == 0 || path == "/" {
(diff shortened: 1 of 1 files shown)


jamesdphillips:
Thanks guys, I forgot Simon was out this week.

PR #1958, opened 2018-08-14
[Web UI] Redirect to user to last visited environment

### LINKED ISSUE #1329: [Web UI] Redirect to next available environment if 'Environment not found' occurs after deletion.
Create an `EnvironmentNotFoundContent` component to be used within each `EnvironmentView` content component when the requested environment does not exist  (in place of currently used `NotFoundView`).

It should show an environment switcher UI (like in the `Drawer` component) to give the user a quick way to visit an environment that exists.

jamesdphillips:
## What is this change?

When landing on the app's root path or after authentication redirect user to the last environment they accessed. When not last environment is not set, app falls back to first organization user is authorized to view.

Closes #1329

Primarily improves user experience when the default organization has been deleted (something we've identified as a best practice.)

`undefined`

`null`

`0`

`[]`


dashboard/src/components/util/SigninRedirect.js
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import { Route, Redirect } from "react-router-dom";
+
+const signinPath = "/signin";
+const redirectKey = "redirect-to";
```
10xjs: "redirect-to" is defined in several places. I suggest defining it in a single location.

This would extend nicely to _any_ query params - since it is a single namespace, we can define all param names in a single location to avoid potential conflicts.

dashboard/src/components/util/LastEnvironmentRedirect.js
```diff
@@ -0,0 +1,83 @@
...
+const primaryQuery = gql`
+  query LastEnvironmentQuery {
+    lastEnvironment @client {
+      organization
+      environment
+    }
+  }
+`;
+
+const fallbackQuery = gql`
```
10xjs: Would it make sense to combine these into a single query? I don't suspect there is any real perf advantage of splitting them.

dashboard/src/components/createStyledComponent/createStyledComponent.js
```diff
@@ -0,0 +1,51 @@
...
+  mapProps,
+  defaultProps,
+  componentFromProp,
+} from "recompose";
+
+let id = 0;
+const getNextId = () => {
+  id += 1;
+  return id;
+};
```
10xjs: This little thing should have it's own happy place in a util dir at this point.

jamesdphillips: for sure 🤔

dashboard/src/components/util/LastEnvironmentRedirect.js
jamesdphillips: One hits the network and the other doesn't.

dashboard/src/apollo/client.js
```diff
@@ -44,6 +44,18 @@ const createClient = () => {
     ]),
   });
 
+  localStorageSync(
```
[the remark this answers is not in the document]
jamesdphillips: Good thought. While technically no UI elements listen to updates, it definitely could lead to some confusing behavior if we ever do.

Addressed regardless.

dashboard/src/components/util/SigninRedirect.js
jamesdphillips: Good thinking, addressed.

PR #1962, opened 2018-08-15
Bugfix for sensuctl cluster health error handling

nikkiki:

## What is this change?

Prevents panic and closes the client.

Fixes bug in `sensuctl cluster health` so that the proper error is handled.

Nah.

Nah.

Nah.

We good.


backend/store/etcd/health_store.go
```diff
@@ -30,12 +30,13 @@ func (s *Store) GetClusterHealth(ctx context.Context) []*types.ClusterHealth {
 			Endpoints:   member.ClientURLs,
 			DialTimeout: 5 * time.Second,
 		})
+		defer cli.Close()
```
golangci: should check returned error before deferring cli.Close()

PR #1976, opened 2018-08-16
[Web UI] Refresh tokens when GraphQL request produces 401

### LINKED ISSUE #1975: [Web UI] Exception thrown when access token is invalid
When an access token is has been revoked on the server or if the authorization token pair is no longer recognized this leads to an `UnauthorizedError` being thrown.

We should better handle this case by displaying the invalid authorization dialog.

jamesdphillips:
## What is this change?

Better handles case where server has invalidated the authorization
token pair or the server's secret has changed.

Closes #1975 

`undefined`

`null`

my poor debugging skills..?

`[]`


jamesdphillips:
Maybe related to #2009.

Issue #1997, opened 2018-08-21
Cluster health table format overflows on error field


mercul3s:
I noticed in testing an assert.Contains(<error message>) my assertions were failing on any string longer than one word in a test error message. Error messages appear to overflow this table field:

![display_table_cluster_error](https://user-images.githubusercontent.com/2100610/44372947-5bb0ce00-a49b-11e8-9b7f-1f1da3b099f0.png)

I dug through our table writer and didn't see any obvious clues as to why the row is overflowing. This isn't a major issue but it is annoying.

nikkiki:
I've found that `table.SetAutoWrapText(false)` gives us the desired effect for the cluster health command. This will prevent auto text wrapping for all other commands as well (unless the string contains a newline). I can move forward with this solution if it seems sufficient!

```
$ sensuctl cluster health
         ID           Name                    Error                   Healthy
 ────────────────── ───────── ────────────────────────────────────── ─────────
  8927110dc66458af   default   cluster error runs off screen: <nil>   true
```

Issue #1979, opened 2018-08-16
Implement "dead man's switch" on etcd alarms


mercul3s:
Add a watcher for a PUT on alarms keys (alarms have their own store) and fire a handler on an event. This might be similar to a keepalive handler, but for sensu-backend, and can act as a dead man's switch in case of any etcd alarms (indicating trouble with an etcd cluster).

echlebek:
One thing we could potentially do is fire events, that are not persisted, when etcd alarms occur. This could allow users to handle alarm events like other events.

annaplotkin:
This requires pipelined to function without etcd.

echlebek:
It doesn't seem likely that Sensu will be able to function when etcd is not. Going forward, a "monitor Sensu with Sensu" approach, requiring two etcd installations, seems much more resilient.

> This requires pipelined to function without etcd.

We would have no way of looking up handlers without etcd, and this is core to the design of Sensu, so I'm going to close this issue.

Issue #1724, opened 2018-06-21
Investigate the etcd alarms API


echlebek:
Etcd provides an alarms API. It is basically undocumented. Find out what it does, and figure out if we can use it to provide some information to the sensu backend.

mercul3s:
@cyphus I summarized a couple bullet points from our earlier conversation and the research I did - feel free to add comments if you've got 'em.

cyphus:
I think that the presence of any alarms is valuable info in the sensuctl health response, and that dead man's switch would make a great catch-all for any unexpected error we encounter. @mercul3s if you're happy with the level you've investigated etcd alarms, I'd suggest closing this and making issues for what you want to see implemented.

mercul3s:
@cyphus I've added #1978 and #1979 for these enhancements.

PR #1998, opened 2018-08-21
Etcd alarms

### LINKED ISSUE #1978: Add etcd alarms to health api
Modify the cluster health struct to also return any etcd alarms in an array. Update the cli to also display the alarm data. For further info on etcd's alarms api, check out #1724.

mercul3s:
## What is this change?

Adds etcd alarms to the health api and cluster health command.

Closes #1978 

Added note under changed.

Nope!

I ran into a minor formatting issue with the table print that I was unable to track down. I filed  #1997 to look into this issue.

I don't think documentation is needed for this change.

### DIFF
--- a/cli/commands/cluster/health.go
+++ b/cli/commands/cluster/health.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/coreos/etcd/etcdserver/etcdserverpb"
 	"github.com/sensu/sensu-go/cli"
 	"github.com/sensu/sensu-go/cli/commands/helpers"
 	"github.com/sensu/sensu-go/cli/elements/table"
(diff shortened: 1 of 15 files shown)


echlebek:
Would you mind dropping a note here describing what the etcd alarms API is useful for? Or anything else you learned of interest in implementing this. :)

cli/commands/cluster/health.go
```diff
@@ -78,3 +95,31 @@ func printToTable(result interface{}, w io.Writer) {
...
+		},
+		{
+			Title:       "Alarm Type",
+			ColumnStyle: table.PrimaryTextStyle,
+			CellTransformer: func(data interface{}) string {
+				alarm, ok := data.(*etcdserverpb.AlarmMember)
+				if !ok {
+					return cli.TypeError
+				}
+				return fmt.Sprintf("%s", alarm.Alarm.String())
```
golangci: the argument is already a string, there's no need to use fmt.Sprintf

mercul3s:
@echlebek I wrote up some initial notes in #1724. For the purposes of this PR, we're just displaying any alarms that may be triggered when an etcd cluster becomes unstable. Looking at [etcd's AlarmTypes](https://godoc.org/github.com/coreos/etcd/etcdserver/etcdserverpb#AlarmType), they can have 3 states - `NONE`, `NOSPACE`, or `CORRUPT`. I did create two alarm states in the test, and they display as such:
```
   ID      Name     Error   Healthy  
 ────── ────────── ─────── ───────── 
  3039   backend0   <nil>   true     
  3039   backend1   error   false    
   ID    Alarm Type  
 ────── ──────────── 
  ddd5   CORRUPT     

   ID      Name     Error   Healthy  
 ────── ────────── ─────── ───────── 
  3039   backend1   error   false    
   ID    Alarm Type  
 ────── ──────────── 
  ddd5   NOSPACE     

```
In order to fully test an alarm state I'd need to break a cluster; I'd like to do that, and also explore if we can/should create our own alarms, as part of #1979.

PR #2000, opened 2018-08-22
[Web UI] Create AppLayout component

### LINKED ISSUE #1372: [Web UI] AppLayout component
Create a new `AppLayout` component to takeover the responsibility of `AppFrame` and `AppContent`.

`/src/component/layout/AppLayout` will implement the outermost layout of the page, taking `topBar`, `content`, and `quickNav` as props.

`/src/component/layout/` will serve as a location for other "layout" components to live (`SidebarLayout`, `FullPageLayout` etc.)

10xjs:
## What is this change?

AppLayout defines the outermost structure common to all pages, serving as a foundation for forthcoming notification toasts and banners.

Closes #1372

No

No

No

No


dashboard/src/components/AppBar.js
```diff
@@ -9,34 +9,38 @@ import { withStyles } from "@material-ui/core/styles";
...
 
       ${EnvironmentLabel.fragments.environment}
+      ${Drawer.fragments.environment}
     `,
   };
 
   static styles = theme => ({
-    container: {
-      paddingTop: "env(safe-area-inset-top)",
-      backgroundColor: theme.palette.primary.dark,
```
jamesdphillips: This padding was for devices with a notch, I think looks like it's been lost to the re-organization.

dashboard/src/components/AppLayout.js
```diff
@@ -0,0 +1,172 @@
...
+    flexDirection: "column",
+    alignItems: "stretch",
+
+    marginLeft: "auto",
+    marginRight: "auto",
+
+    maxWidth: 1224,
+
+    paddingLeft: 8,
+    paddingRight: 8,
```
jamesdphillips: Why not use the theme's spacing unit here? It's nice to avoid magic numbers when we can.

jamesdphillips: Between this and removing the padding prop from the `Content` component, we've lost a little in translation. 

Previously we have no padding in the content container on narrower viewports, this allowed lists and other elements to expand to the edge of the screen matching platform conventions.

dashboard/src/components/AppLayout.js
```diff
@@ -0,0 +1,172 @@
...
+
+  handleTopBarResize = rect => {
+    this.setState(state => {
+      if (state.topBarHeight === rect.height) {
+        return null;
+      }
+
+      return { topBarHeight: rect.height };
+    });
+  };
```
jamesdphillips: neat solution 👍

dashboard/src/components/AppLayout.js
```diff
@@ -0,0 +1,172 @@
...
+            <div className={classes.topBarObserver}>
+              <ResizeObserver onResize={this.handleTopBarResize} />
+              {topBar}
+            </div>
+            <div className={classes.quickNavContainer}>
+              <div className={classes.quickNav}>{quickNav}</div>
+            </div>
+            <div className={classes.alertContainer}>
+              <div className={classes.alert}>{alert}</div>
+            </div>
```
jamesdphillips: Where is this being used?

dashboard/src/components/AppBar.js
10xjs: My mistake. I Thought `env(safe-area-inset-top)` was broken functionality in the style build system. I can revert this.

PR #2003, opened 2018-08-23
[WIP] Fix issue with multi-asset checks

amdprophet:

## What is this change?

Fixes a bug where a check with multiple assets would only ever install
the first asset. Also updates the assetmanager store to use the entire
asset SHA as an identifier rather than the first 7 digits.

Checks with multiple assets will only ever install the first asset. Also, using the first 7 digits of a SHA as an identifier may cause collisions.

Updated.

Nope!

(Sometimes) the use of pointers can lead to some not so fun debugging sessions.

N/A


calebhailey:
Thanks for the fix, @amdprophet & @jamesdphillips!

amdprophet:
Closing this for now. I'll reopen if the agent is still affected once assets are rewritten in the agent.

Issue #1829, opened 2018-07-13
[Build stability] TestInfo


nikkiki:
`host.Info()` can return an error in Appveyor due to `context deadline exceeded`. The context is provided by https://github.com/shirou/gopsutil/blob/master/host/host_windows.go#L43-L45
```
--- FAIL: TestInfo (3.00s)
	assertions.go:239: 
                          
	Error Trace:	system_test.go:13
		
	Error:      	Received unexpected error:
		
	            	context deadline exceeded
	assertions.go:239: 
                          
	Error Trace:	system_test.go:14
		
	Error:      	Should NOT be empty, but was 
	assertions.go:239: 
                          
	Error Trace:	system_test.go:15
		
	Error:      	Should NOT be empty, but was 
	assertions.go:239: 
                          
	Error Trace:	system_test.go:16
		
	Error:      	Should NOT be empty, but was 
	assertions.go:239: 
                          
	Error Trace:	system_test.go:17
		
	Error:      	Should NOT be empty, but was 
	assertions.go:239: 
                          
	Error Trace:	system_test.go:21
		
	Error:      	Should NOT be empty, but was 
	assertions.go:239: 
                          
	Error Trace:	system_test.go:22
		
	Error:      	Should NOT be empty, but was []
```

amdprophet:
At a glance it looks like this may be due to gopsutil hanging on `windows/386`. There have been a few updates to the windows code between `2.18.04` and `2.18.06`:
https://github.com/shirou/gopsutil/compare/v2.18.04...v2.18.06. It may be worth trying the latest version to see if it resolves the issue.

amdprophet:
This is happening due to WMI queries taking too long https://github.com/shirou/gopsutil/blob/v2.18.04/internal/common/common_windows.go#L119. The timeout is 3 seconds which feels long already. I'm not sure we'll want to increase the timeout as it will have a direct effect on how long it takes to send a keepalive.

We'll need to figure out how to guard against cases where we hit a timeout for WMI queries. We may want to support keepalive messages that may be missing system information - so long as we make sure not to override any previously stored system information with nothingness.

One other thing we may want to do is collect system information out-of-band of the `getAgentEntity()` function in the agent so that we aren't holding up keepalive messages when trying to fetch system info.

Any thoughts?

grepory:
[...]

PR #2017, opened 2018-08-30
Separate system info collection from getAgentEntity()

### LINKED ISSUE #1947: Separate system info collection from getAgentEntity()
## Expected Behavior
The collection of system info should not affect an agent's ability to send keepalives.

## Current Behavior
It is possible for system info collection to take up to 3 seconds, or outright fail, which has a direct effect on how long a keepalive takes to send or whether it is sent at all.

## Possible Solution
Move system info collection into a new goroutine that starts up with an agent that will cache the latest fetched system info. Update `getAgentEntity()` to retrieve system info from the cache.

The asset manager will also need to be updated to retrieve system info from this cache but should be handled under another issue.

ccressent:
## What is this change?

Instead of `agent.getAgentEntity()` being responsible for gathering system info, the agent itself now spawns a goroutine responsible for periodically updating a local cache of system info.

`agent.getAgentEntity()` now reads that cache whenever it needs to retrieve system information.

Closes #1947.
Might help #1829. 

Yes, added as part of of the change.

- Initial keepalive message might not have system info. That doesn't change how
  it used to work, but is that OK?
- Unsure how/if I should add tests for the change. It might lead to some refactoring?
- Any potential impact I might be missing? Would you have done it differently?
  
- `lsb_release` needed for some distros to pass tests in `system` package (eg:
  Arch). I'll open a separate issue for which the "fix" would probably simply be
  an update to the README.
- unsure about testing (reproducing #1947 easily, need for test in the first
  place, ease of testing without needlessly refactoring)
  
Nothing relevant to change that I could find in the "Sensu Core" doc set for
v2.0.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -63,6 +63,9 @@ const (
 	DefaultStatsdMetricsHost = "127.0.0.1"
 	// DefaultStatsdMetricsPort specifies the default metrics port for statsd server
 	DefaultStatsdMetricsPort = 8125
+	// DefaultSystemInfoRefreshInterval specifies the default refresh interval
+	// (in seconds) for the agent's cached system information.
+	DefaultSystemInfoRefreshInterval = 20
 	// DefaultUser specifies the default user
 	DefaultUser = "agent"
 )
@@ -429,8 +432,7 @@ func (a *Agent) Run() error {
 	}
 
 	go func() {
-		// TODO(cyril): create a separate config knob for system info refresh interval?
-		systemInfoTicker := time.NewTicker(time.Duration(a.config.KeepaliveInterval) * time.Second)
+		systemInfoTicker := time.NewTicker(time.Duration(DefaultSystemInfoRefreshInterval) * time.Second)
 		for {
 			select {
 			case <-systemInfoTicker.C:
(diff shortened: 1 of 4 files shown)


agent/agent.go
```diff
@@ -404,6 +428,21 @@ func (a *Agent) Run() error {
 		logger.WithError(err).Error("error sending keepalive")
 	}
 
+	go func() {
+		// TODO(cyril): create a separate config knob for system info refresh interval?
```
mercul3s: Would this be a user-configurable agent config attribute, or a default based on the keepalive value (as it is now)?

ccressent: I'm assuming there is little value for the user to be able to configure it, except maybe in extreme edge cases such as the user "needing" to change/reconfigure network interfaces more frequently than the system information is refreshed?

I would be happy with a default value of our choosing, say `DefaultSystemInfoRefreshInterval = 60`, that wouldn't be exposed in the user config:
- we can change it at will at compile time
- if it really becomes a valid need for users we can then expose it

ccressent:
@echlebek: I'm also all for testing! It sounds like the we have quite a bit of refactoring and additional testing to do on that `agent` package. I agree, let's make that work part of a future tech debt task then.

@amdprophet: that sounds like another valid approach, but I don't have enough high level understanding of the system yet to make a choice here. I made that choice based on the expected behavior described in #1947 ("The collection of system info should not affect an agent's ability to send keepalives.") and a quick read of #1829. I'm happy to revisit later if the current choice proves to not be the best!

grepory:
Definitely block on sending the first keepalive until we have system info.

PR #2020, opened 2018-08-31
Remove race flag from build.sh integration tests

amdprophet:

## What is this change?

Removes race flag from `build.sh` integration tests.

Just an attempt to fix our CircleCI builds from OOM-ing.

### DIFF
--- a/backend/backend_test.go
+++ b/backend/backend_test.go
@@ -1,4 +1,4 @@
-// +build integration,race
+// +build integration
 
 package backend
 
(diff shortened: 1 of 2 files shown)


echlebek:
In backend/backend_test.go, the tests only run if the race detector is enabled. (Some other tests did this too, but I've been changing that as I've found them as I think that behaviour is wrong.)

Can you remove that build tag as part of this change?

amdprophet:
@echlebek updated!

Issue #2010, opened 2018-08-28
AssetManager lock file isn't appropriate for use in sensu agent.


grepory:
## Expected Behavior
If two competing goroutines attempt to install an asset, only one of them installs the asset.

## Current Behavior
Two competing goroutines will attempt to install assets on top of each other.

If two competing goroutines attempt to lock an asset for installation, the second goroutine will remove the lock file and create a new lock file.

https://github.com/nightlyone/lockfile/blob/master/lockfile.go#L137-L156

## Possible Solution
Replace lockfiles with mutexes.

grepory:
I've started down the path of reimplementing assets entirely rather than patching it with mutexes.

Each sensu component that utilizes assets will store state about those assets in a local instance of boltdb--rather than relying on a database without any sensible locking, such as a filesystem. Goroutines requesting assets from the asset manager will block getting the asset while the asset manager installs the asset. Installations will be serialized.

We'll first open a read-only transaction to boltdb to check for the existence of a local asset, allowing goroutines requesting already-installed assets to continue without blocking on other asset installations. If the asset doesn't exist locally, we then go into the asset fetch-verify-expand process, storing all necessary data in boltdb and finally returning handing the caller everything it needs to know to utilize the asset (just paths, really).

This will be completely decoupled from the agent and should be able to be dropped into place in place of the existing asset manager.

PR #2024, opened 2018-09-04
Rewrite/redesign asset implementation

grepory:
## What is this change?

This implements a PoC for a BoltDB-driven, reusable asset manager.

Assets do not currently work as intended as described in #2010. We can't guarantee that competing installations of assets won't install over one another without external coordination--that only coincidentally is happening. Furthermore, we are opening ourselves up to potential locking issues depending on the filesystem to maintain our locks.

This is not to be merged, but I would like to discuss the approach.

This does not impact documentation.


asset/boltdb_manager.go
```diff
@@ -0,0 +1,145 @@
...
+
+		return bucket.Put(key, assetJSON)
+	}); err != nil {
+		return nil, err
+	}
+
+	return localAsset, nil
+}
+
+func validate(file *os.File, sha512 string) error {
```
golangcibot: `validate` is unused

asset/boltdb_manager.go
```diff
@@ -0,0 +1,145 @@
...
+
+	return nil
+}
+
+func expand(file *os.File) error {
```
golangcibot: `expand` is unused

asset/boltdb_manager_test.go
```diff
@@ -0,0 +1,270 @@
...
+		log.Printf("expected no error, got: %v", err)
+	}
+
+	d := runtimeAsset.LibDir()
+
+	if d == "" {
+		log.Printf("expected lib directory, got: %s", d)
+	}
+
+	runtimeAsset, err = manager.Get(a)
```
golangcibot: ineffectual assignment to `err`

asset/boltdb_manager.go
```diff
@@ -0,0 +1,135 @@
...
+type Getter interface {
+	Get(*types.Asset) (*RuntimeAsset, error)
+}
+
+// BoltDBAssetManager is responsible for the installing and storing the metadata
+// for assets backed by an instance of BoltDB on the local filesystem. BoltDB
+// provides the serialization guarantee that the asset contract specifies.
+// We rely on long-lived BoltDB transactions during Get to provide this
+// mechanism for blocking.
+type BoltDBAssetManager struct {
```
echlebek: I think this type is adding more API surface than is necessary.

BoltDBAssetManager exports all its internals, but that isn't a requirement for any other package right now. I think the LocalStorage and DB fields can be unexported.

Fetcher, Expander and Verifier likely don't need to be anonymously embedded unless there is reason to call these things outside of Get(). I think these things should be unexported also.

Consider unexporting BoltDBAssetManager altogether. If you decide that BoltDBAsset manager should be exported, then it should also have a New function that creates a BoltDBAssetManager with its dependencies as arguments.

asset/boltdb_manager.go
```diff
@@ -0,0 +1,135 @@
...
+			return err
+		}
+
+		localAsset = &RuntimeAsset{
+			Path: assetPath,
+		}
+
+		assetJSON, err := json.Marshal(localAsset)
+		if err != nil {
+			return err
```
echlebek: If marshaling assets results in an error, then this will cause the asset downloader to infinitely loop, downloading assets over and over, erroring on marshal.

Marshaling an asset probably won't result in an error - the RuntimeAsset would need to contain an incompatible type for that to happen.

We should panic if the error is non-nil, and add a test that ensures that RuntimeAssets can be marshaled without error.

asset/boltdb_manager_test.go
```diff
@@ -0,0 +1,274 @@
+package asset_test
```
echlebek: The tests use logging extensively, but should probably use the methods from the testing package.

See https://golang.org/pkg/testing/#T.Fatalf

It is generally a mistake to use log.Fatal in tests - this will cause the test runner to exit without running any other tests.

asset/boltdb_manager.go
grepory: In that case, I will move the asset_test package into the asset package so that I can manipulate the dependencies of BoltDBAssetManager.

If we're not going to implement a different asset manager, then I think `asset.NewManager()` should be sufficient.

If we want to add an integration test later, we can re-add the asset_test package and test against an actual HTTP resource, but I wanted to avoid that because externally available dependencies, even if s3 is "stable enough."

asset/boltdb_manager.go
grepory: Oh yeah, that's a good point. Panicking instead.

asset/boltdb_manager_test.go
grepory: Thanks. I wasn't sure about half of the things I was doing in testing. I got so used to testify, that I was like "what do I do here again? I will wait and have Eric tell me." <3

asset/boltdb_manager.go
grepory: @echlebek -- So I was thinking about this this morning... I see _some_ merit in leaving these things exported in case we want to write a utility for interacting with assets outside of the context of the boltdb manager that's being used in the agent (and eventually backend).

I have thought of a couple of approaches here:

- continue exporting these and allow them to be used piecemeal elsewhere
- unexport them and create a new manager type that doesn't use boltdb so that, for example, the CLI can fetch, verify, and expand an asset

For the latter, however, we may want a command line utility to bypass expanding or we may want to skip verification. There's no mechanism to do that in the interfaces we expose or in the Get interface itself.

We could add an options hash later--or we could do option 1.

Regardless, I am inclined to not take action (potentially requiring unnecessary rework later). If you have a counterargument, I am all ears.

asset/verifier.go
```diff
@@ -0,0 +1,36 @@
...
+
+// Verify that the file matches the desired SHA-512 sum.
+func (v *SHA512Verifier) Verify(rs io.ReadSeeker, desiredSHA string) error {
+	// Generate checksum for downloaded file
+	h := sha512.New()
+	if _, err := io.Copy(h, rs); err != nil {
+		return fmt.Errorf("generating checksum for asset failed: %s", err)
+	}
+
+	if _, err := rs.Seek(io.SeekStart, io.SeekStart); err != nil {
```
golangcibot: the first argument of io.Seeker is the offset, but an io.Seek* constant is being used instead

asset/util.go
```diff
@@ -0,0 +1,12 @@
+package asset
+
+import "os"
+
+// resetFile ensures file contents are synced and rewound
+func resetFile(f *os.File) error {
```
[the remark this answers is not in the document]
grepory: It's used in a couple of different places. You have to read the first 512 bytes to determine the archive type. I moved it there anyway.

There or `util.go`. Either is fine with me.

asset/expander.go
```diff
@@ -0,0 +1,76 @@
...
+	header := make([]byte, headerSize)
+	if _, err := f.Read(header); err != nil {
+		return filetype_types.Type{}, fmt.Errorf("unable to read asset header: %s", err)
+	}
+	ft, err := filetype.Match(header)
+	if err != nil {
+		return ft, err
+	}
+
+	if _, err := f.Seek(io.SeekStart, io.SeekStart); err != nil {
```
golangcibot: the first argument of io.Seeker is the offset, but an io.Seek* constant is being used instead

asset/fetcher.go
```diff
@@ -0,0 +1,85 @@
...
+	if _, err = io.Copy(tmpFile, resp); err != nil {
+		return nil, fmt.Errorf("error downloading asset")
+	}
+
+	if err := tmpFile.Sync(); err != nil {
+		tmpFile.Close()
+		return nil, err
+	}
+
+	if _, err := tmpFile.Seek(io.SeekStart, io.SeekStart); err != nil {
```
golangcibot: the first argument of io.Seeker is the offset, but an io.Seek* constant is being used instead

asset/boltdb_manager.go
grepory: I went and thought some more. You're right, re: API surface area. Let's just not give so many options. I'll do what we originally discussed.

grepory: Unexporting everything makes integration testing this pretty hard without a web server.

asset/boltdb_manager.go
```diff
@@ -0,0 +1,163 @@
...
+	"encoding/json"
+	"os"
+	"path/filepath"
+
+	"github.com/boltdb/bolt"
+	"github.com/sensu/sensu-go/types"
+)
+
+const (
+	assetDBName = "assets.db"
```
golangcibot: `assetDBName` is unused

PR #2030, opened 2018-09-05
Fix a bug in sensuctl check update.

### LINKED ISSUE #2023: sensuctl check update won't let me unset the optional output metric format
when using sensuctl check update
  output_metric_format won't accept ""  as a valid format. 

Effectively this makes it impossible to unset the optional output_metric_format attribute using check update in interactive mode

## Expected Behavior
I should be able to unset the metric format attribute disabling agent's metric processing

## Current Behavior
check update  doesn't ;et you set metric format to ""  

## Possible Solution
allow check update to accept "" or some keyword which means none/nil as one of its valid options in the interactive session.

1. create a check, set the metric output to nagios_perfdata
2. check update   try to unset the metric_output
3. cry

## Context
since metric_format is considered optional, there should be a way to unset it in the check update interactive session

## Your Environment
centOS 7.x running  sensu 2 beta4 packages from beta repository.

echlebek:
## What is this change?

This commit fixes a bug in sensuctl check update where once a
metrics output format had been set, it could no longer be unset.

Closes #2023

Yes

No

It seems likely that a slew of similar bugs exists in the interactive mode of sensuctl. We should discuss how to resolve this on a product-wide basis.

### DIFF
--- a/cli/commands/check/interactive.go
+++ b/cli/commands/check/interactive.go
@@ -236,13 +236,18 @@ func (opts *checkOpts) administerQuestionnaire(editing bool) error {
 		},
 		{
 			Name: "output-metric-format",
-			Prompt: &survey.Input{
+			Prompt: &survey.Select{
 				Message: "Metric Format:",
-				Help:    "Optional output metric format used to parse check output for metric extraction. Valid formats include: nagios_perfdata, graphite_plaintext, opentsdb_line, and influxdb_line",
-				Default: opts.OutputMetricFormat,
+				Options: []string{
+					"nagios_perfdata",
+					"graphite_plaintext",
+					"opentsdb_line",
+					"influxdb_line",
+					"none",
+				},
 			},
 			Validate: func(val interface{}) error {
-				if val.(string) != "" {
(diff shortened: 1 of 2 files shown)


cli/commands/check/interactive.go
```diff
@@ -236,13 +236,18 @@ func (opts *checkOpts) administerQuestionnaire(editing bool) error {
 		},
 		{
 			Name: "output-metric-format",
-			Prompt: &survey.Input{
+			Prompt: &survey.Select{
 				Message: "Metric Format:",
-				Help:    "Optional output metric format used to parse check output for metric extraction. Valid formats include: nagios_perfdata, graphite_plaintext, opentsdb_line, and influxdb_line",
-				Default: opts.OutputMetricFormat,
+				Options: []string{
+					"nagios_perfdata",
```
palourde: We have some constants for these formats, see https://github.com/sensu/sensu-go/blob/6996d8ccb36ba9a9b6750da016ce3b5ea118bcfd/types/check.go#L40

echlebek: Oops, I merged before I noticed that. :man_facepalming:

PR #2034, opened 2018-09-06
Resolve some etcd configuration and operation issues

### LINKED ISSUE #1965: --listen-client-urls does not parse multiple urls
According to etcd, [--listen-client-urls](https://coreos.com/etcd/docs/latest/v2/configuration.html#-listen-client-urls) supports multiple comma separated urls, ex: `"http://localhost:2380,http://localhost:7001"`. However, when we initialize embedded etcd, we don't parse for the comma separated values, and instead [pass a single url string](https://github.com/sensu/sensu-go/blob/d43643add913f31201af6e37c1ec78398cfe2ccc/backend/etcd/etcd.go#L143).

### LINKED ISSUE #2016: Don't bind etcd listeners to random ports on localhost
At first, we were not going to expose an etcd port to any interface other than localhost, but that obviously won't work with clustering.

Since we will need the port to be available on external interfaces in all cases, we may as well not bind to the localhost port.

echlebek:
## What is this change?

* --listen-client-urls can now be a comma-separated list.
* The etcd service no longer forcibly binds to localhost.

There were a couple of etcd configuration issues that I thought were closely related enough to be worthwhile folding together into a single commit.

Closes #1965
Closes #2016 

Yes

No

No

No documentation requires updating, as clustering documentation is still an outstanding effort.

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"runtime/debug"
+	"strings"
 
 	"github.com/coreos/etcd/clientv3"
 	"github.com/sensu/sensu-go/backend/agentd"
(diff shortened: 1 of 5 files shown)


grepory:
Can you make sure that listen client urls and listen peer urls default to *:2379 and :2380 respectively?

backend/backend.go
```diff
@@ -66,7 +67,7 @@ func Initialize(config *Config) (*Backend, error) {
 	// the Wizard bus, which requires etcd to be started.
 	cfg := etcd.NewConfig()
 	cfg.DataDir = config.StateDir
-	cfg.ListenClientURL = config.EtcdListenClientURL
+	cfg.ListenClientURLs = strings.Split(config.EtcdListenClientURL, ",")
```
mercul3s: 👍

PR #2037, opened 2018-09-07
Support using sensu-backend with external etcd instances

### LINKED ISSUE #1963: Sensu TLS configuration requires custom CA
The sensu-backend connects to its own embedded etcd via a localhost TCP connection. [Existing code](https://github.com/sensu/sensu-go/blob/d43643add913f31201af6e37c1ec78398cfe2ccc/backend/etcd/etcd.go#L151-L180) in etcd startup ensures that the embedded etcd always listens on localhost so that this connection always succeeds.

In order for a TLS enabled sensu-backend to work, it must be provided with a TLS certificate that is valid for localhost. No trusted Certificate Authority will let you create a TLS certificate with localhost, which forces the user to create their own CA in order to enable TLS in sensu. This adds configuration complexity, and for some organizations may be impossible due to security policy.

## Possible solution

Make embedded etcd listen over http on localhost, even in TLS enabled configurations.

### LINKED ISSUE #2015: Allow Sensu to function without embedded etcd
Many users have existing etcd installations, and enabling them to use Sensu with these installations would help them a lot.

Add a command line flag to the backend that indicates whether or not to run embedded etcd. The default mode should be to run embedded.

echlebek:
## What is this change?

This commit adds the --no-embed flag to the backend. When this flag
is specified, sensu-backend will dial whatever URLs are specified
in the --listen-client-urls flag, and not start up etcd.

Closes #2015 
Closes #1963

Yes

No

This change required some refactoring of the queue package, which
depended on the cluster member ID for operation. This ID has been
replaced by a client lease ID which is created on backend startup.

As a result, this made queue garbage collection easier to
guarantee, because now all queue keys are created with a lease.
The test that tested queue garbage collection has been removed,
because the logic that removes the keys is no longer part of the
queue package.

Testing was done by starting up an external etcd instance and
pointing the backend at it. No issues were observed, and the etcd
instance was correctly populated on startup.

I don't believe we need to explicitly mention this in the docs for now, given what we have currently. I think we should mention this in the upcoming clustering documentation however.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ keepalive events.
 - The backend no longer requires embedded etcd. External etcd instances can be
 used by providing the --no-embed option. In this case, the client will dial
 the URLs provided by --listen-client-urls.
+- The sensu-agent binary is now located at ./cmd/sensu-agent.
 
 ### Fixed
 - Fixes a bug in `sensuctl cluster health` so the correct error is handled.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@ used by providing the --no-embed option. In this case, the client will dial
 the URLs provided by --listen-client-urls.
 - The sensu-agent binary is now located at ./cmd/sensu-agent.
 - Sensuctl no longer uses auto text wrapping.
+- The backend no longer requires embedded etcd. External etcd instances can be
+used by providing the --no-embed option. In this case, the client will dial
+the URLs provided by --listen-client-urls.
 
 ### Fixed
 - Fixes a bug in `sensuctl cluster health` so the correct error is handled.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ text.
 keepalive events.
 - The sensu-agent binary is now located at ./cmd/sensu-agent.
 - Sensuctl no longer uses auto text wrapping.
+- The backend no longer requires embedded etcd. External etcd instances can be
+used by providing the --no-embed option. In this case, the client will dial
+the URLs provided by --listen-client-urls.
 
 ### Fixed
 - Fixes a bug in `sensuctl cluster health` so the correct error is handled.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ keepalive events.
 used by providing the --no-embed option. In this case, the client will dial
 the URLs provided by --listen-client-urls.
 - The sensu-agent binary is now located at ./cmd/sensu-agent.
+- Sensuctl no longer uses auto text wrapping.
 
 ### Fixed
 - Fixes a bug in `sensuctl cluster health` so the correct error is handled.
--- a/backend/etcd/id.go
+++ b/backend/etcd/id.go
@@ -1,0 +1,45 @@
+package etcd
+
+import (
+	"context"
+	"fmt"
+	"path"
+
+	"github.com/coreos/etcd/clientv3"
+	"github.com/sensu/sensu-go/backend/store"
+)
+
+var (
+	backendIDKeyPrefix         = store.NewKeyBuilder("backends").Build()
+	backendIDLeasePeriod int64 = 60
+)
+
+// BackendID creates and returns a new leased backend ID.
+func BackendID(ctx context.Context, client *clientv3.Client) (int64, error) {
+	// Grant a lease for 60 seconds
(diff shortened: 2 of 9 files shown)


backend/etcd/id.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/sensu/sensu-go/backend/store"
+)
+
+var (
+	backendIDKeyPrefix         = store.NewKeyBuilder("backends").Build()
+	backendIDLeasePeriod int64 = 60
+)
+
+// BackendID creates and returns a new leased backend ID.
+func BackendID(ctx context.Context, client *clientv3.Client) (int64, error) {
```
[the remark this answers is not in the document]
echlebek: Excellent catch. I was discussing this with @ccressent as well.

Your analysis of how this works is correct.

I needed to think about whether we wanted to crash or retry here, and I think we want to retry. I'll add some retry logic.

backend/queue/queue.go
```diff
@@ -195,17 +191,27 @@ func (q *Queue) swapLane(ctx context.Context, currentKey string, currentRevision
...
+		result = append(result, string(kv.Value))
+	}
+	return result, nil
+}
+
 // Enqueue adds a new value to the queue. It returns an error if the context is
 // canceled, the deadline exceeded, or if the client encounters an error.
 func (q *Queue) Enqueue(ctx context.Context, value string) error {
-	resp, err := q.cluster.MemberList(ctx)
+	backendIDs, err := getAllBackendIDs(ctx, q.kv)
```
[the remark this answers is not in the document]
echlebek: The problem with this approach is that a cold start would require iterating these keys and doing puts on them to warm the caches. While I've been thinking about going down that road, I think I would want it to be done as a consistent approach in several places. Caching is easy to do wrong, and it would require a well thought-out design.

backend/queue/queue.go
```diff
@@ -223,43 +229,22 @@ func (q *Queue) Enqueue(ctx context.Context, value string) error {
 	}
 }
 
-// cleanup stale queue entries belonging to backends that no longer exist
-func (q *Queue) cleanup(ctx context.Context, members []*etcdserverpb.Member) error {
```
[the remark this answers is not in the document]
echlebek: I think it was actually more efficient and precise from a GC standpoint, but it's always nice to reduce the amount of logic in our codebase. :grinning:

backend/queue/queue.go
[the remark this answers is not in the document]
cyphus: I agree, let's keep the design simple for now, and look into caching only if it becomes a performance bottleneck (which I doubt will happen)

backend/queue/queue.go
```diff
@@ -46,9 +45,9 @@ func (e EtcdGetter) GetQueue(path ...string) types.Queue {
 
 // Queue is a durable FIFO queue that is backed by etcd.
```
cyphus: We may want to qualify "durable" a bit here.

backend/queue/queue.go
```diff
@@ -215,11 +201,6 @@ func (q *Queue) enqueueOps(backendIDs []string, value string) ([]clientv3.Cmp, [
 // Dequeue gets a value from the queue. It returns an error if the context
 // is cancelled, the deadline exceeded, or if the client encounters an error.
 func (q *Queue) Dequeue(ctx context.Context) (types.QueueItem, error) {
-	err := q.nackExpiredItems(ctx, q.itemTimeout)
```
cyphus: With this commit, items never expire from the queue? If that's the case, I believe ad-hocs checks would eventually be ran even in the event of a backend crashing/restarting and rejoining the queue with a new ID.

echlebek: They would not, because they would rejoin with a different ID, and therefore not receive the message.

backend/queue/queue.go
echlebek: You're right, it is no longer durable.

CHANGELOG.md
```diff
@@ -18,6 +18,14 @@ text.
 keepalive events.
 - The sensu-agent binary is now located at ./cmd/sensu-agent.
 - Sensuctl no longer uses auto text wrapping.
+- The backend no longer requires embedded etcd. External etcd instances can be
+used by providing the --no-embed option. In this case, the client will dial
```
ccressent: Very minor: would `--no-embedded-etcd` be clearer and more "future proof" (as in, no-embed could mean many things ultimately)?

echlebek: I changed it to be --no-embed-etcd

PR #2046, opened 2018-09-10
Moved the sensu-agent binary to ./cmd/sensu-agent.

### LINKED ISSUE #2044: Move sensu agent entry point
The sensu agent main is in ./agent/cmd, but it should be in ./cmd/sensu-agent. Then users will be able to type `go get github.com/sensu/sensu-go/cmd/sensu-agent` and the sensu-agent binary will be placed in `$GOPATH/bin`.

echlebek:

## What is this change?

I'm doing some work on the sensu agent, so I thought I'd take care of some tech debt at the same time. We were supposed to move the sensu agent to the cmd package a while ago, but it seems that nobody ever filed an issue for that at the time.

Closes #2044 

Yes

No

No


echlebek:
@amdprophet they're already in ./cmd/sensu-backend and ./cmd/sensuctl :)

PR #2047, opened 2018-09-10
Disable auto text wrapping in sensuctl

### LINKED ISSUE #1997: Cluster health table format overflows on error field

nikkiki:

## What is this change?

Disables auto text wrapping in sensuctl.

Closes #1997 

Yas.

Nah.

Nah.

We good.


ccressent:
What's up with all the `boltdb` stuff?

nikkiki:
@ccressent looks like running `dep ensure` purged some vendored `boltdb` code. We're not explicitly using the code (its not managed in `Gopkg.toml`), so I believe this was just some unintended cleanup.

nikkiki:
@ccressent Alright, I've added `boltdb` into `Gopkg.toml`, but the diff is still going to be large (removing vendored `boltdb` test files). We [prune tests and unused packages](https://github.com/sensu/sensu-go/blob/67cb5caae2047530296a81597aedbbafbd53f22d/Gopkg.toml#L11-L13), so I think dep is doing what its supposed to be doing here.

PR #2056, opened 2018-09-13
[Web UI] Rework toolbar menus

### LINKED ISSUE #1938: [Web UI] List improvements

- [x] Menu items in our list item's vertical disclosures ("kebab" menus) are missing icons.
- [x] Pages are missing margin at the bottom.
- [ ] Checkboxes should not be present on narrower viewports.
  - Currently there is very little room for a little item's content especially on smaller devices,
  - Especially for those on mobile devices who additionally are left handed it is too easy to unintentionally select items.
  - Down the road we maybe could reimplement in a way that follows conventions of the platform, eg. long touch or gestures.
- [x] Font is hardcoded and not consistent with theme(s)?

jamesdphillips:
## What is this change?

Reworks menus for consistency, to include icons and improvements to their visibility.

The new pattern closely mirrors the Android toolbar menu implementation,
where buttons are automatically collapsed based on the available space.
Behaviour of individual items is configurable, so that an implementor can
declare if / when / how an item is collapsed or how to make room.

screen shot 2018-09-12 at 11 33 50 amscreen shot 2018-09-12 at 11 34 18 am

Closes #1938


dashboard/src/components/partials/ToolbarMenu/Menu.js
```diff
@@ -1,180 +1,187 @@
...
 
-      const marginLeft = parseFloat(styles.marginLeft, 10);
-      const marginRight = parseFloat(styles.marginRight, 10);
+      return { buttonsWidth: rect.width };
+    });
+  };
 
-      widths[i] = rect.width + marginLeft + marginRight;
-    }
+  handleWindowResize = debounce(ev => {
```
10xjs: Would using RAF instead of a `debounce` dep work here?

Issue #1935, opened 2018-08-07
[Web UI] Checks page improvements


jamesdphillips:
Targeted page improvements for issues discovered during QA and VQA process.

- [ ] Filter checks by subscription. (Would mirror entities page, as a user I would be somewhat conditioned to expect this.)
- [ ] If a check has been unpublished this state should be reflected in the list.
  - in particular "executed every X" should be omitted from copy to reduce any confusion.
- [ ] Add option to publish / unpublish checks.
  - Additionally it would be nice to collapse the delete action by default. We should be guiding users toward soft deletion when possible.

dabria:
A seperate PR for adding a fragment to the CheckListHeader should be created

PR #2075, opened 2018-09-19
Allow updateCheck mutation to update individual fields

jamesdphillips:
## What is this change?

Allow updateCheck mutation to update individual fields of given check.

---

Unblocks #1935.

unlikely

n/a

Without any mocks / stubs for the actions package we continue to accrue redundant code when writing field resolver tests.

n/a

### DIFF
--- a/backend/apid/graphql/mutations.go
+++ b/backend/apid/graphql/mutations.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"time"
 
+	"github.com/mitchellh/mapstructure"
 	"github.com/sensu/sensu-go/backend/apid/actions"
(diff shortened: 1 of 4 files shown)


backend/apid/graphql/mutations.go
```diff
@@ -136,15 +148,17 @@ func (r *mutationsImpl) ExecuteCheck(p schema.MutationExecuteCheckFieldResolverP
...
+	ins, ok := in.(map[string]interface{})
+	if !ok {
+		return
+	}
+	props, ok := ins["props"].(map[string]interface{})
+	if !ok {
+		return
+	}
+
+	mapstructure.Decode(props, r)
```
golangcibot: Error return value of `mapstructure.Decode` is not checked

Issue #1999, opened 2018-08-22
Keepalive check / handler ignores filter


brieken:
I'm running the Beta 4 release. What I did is the following

```
systemctl stop sensu-backend
rm -r /var/lib/sensu/etcd
```
then `sensuctl configure` en created users and changed passwords. And then created two handlers

```
sensuctl handler create slack --command 'dummyslacktest' --type pipe --timeout 30 --filters is_incident,not_silenced
sensuctl handler create keepalive --handlers slack --type set
```
and created some checks

```
sensuctl check create DiskCheckWindows --command 'c:\\programdata\\sensu\\checks\\check_disk.exe -u GB -w 10% -c 5%' --handlers slack --interval 1800 --subscriptions Windows --ttl 1860 --timeout 60
sensuctl check create ServerHardwareCheckWindows --command 'C:\\Windows\\system32\\WindowsPowerShell\\v1.0\\powershell.exe -file C:\\programdata\\sensu\\checks\\check_server_hardware.ps1 ;exit $LastExitCode' --handlers slack --interval 3600 --subscriptions Server,Windows --ttl 3660 --timeout 60
```

Everything is working, but i keep receiving resolved slack messages on the keepalive check. Some log lines:

```
Aug 22 12:45:51 monitor.*****.net sensu-backend[28739]: {"check":"keepalive","component":"pipelined","entity":"**************************","environment":"default","handler":"slack","level":"info","msg":"sending event to handler","organization":"default","time":"2018-08-22T12:45:51+02:00"}
Aug 22 12:45:51 monitor.*****.net sensu-backend[28739]: {"component":"pipelined","environment":"default","handler":"slack","level":"info","msg":"event pipe handler executed","organization":"default","output":"dummyslacktest","status":0,"time":"2018-08-22T12:45:51+02:00"}
```

the other checks are filtered like expected:

```
Aug 22 12:07:09 monitor.*****.net sensu-backend[28739]: {"check":"ServerHardwareCheckWindows","component":"pipelined","entity":"**************************","environment":"default","handler":"slack","level":"info","msg":"event filtered","organization":"default","time":"2018-08-22T12:07:09+02:00"}
```

brieken:
Check:     keepalive
Output:
Status:    0
History:   0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,1
Silenced:  false
Timestamp: 2018-08-22 20:37:57 +0200 CEST

[...]

Issue #1924, opened 2018-08-03
Rework keepalives to support history


amdprophet:
Keepalive events do not currently support `CheckHistory` but probably should.

## Expected Behavior
Keepalive events should show their history of received / not received keepalive messages.

## Current Behavior
As keepalive events are not stored, there is no way of fetching the event's history. We do store `KeepaliveFailures` but the type does not support history.

nikkiki:
This issue requires the implementation of a keepalive store. Sending this back until we can spec out some subtasks.

annaplotkin:
Discussion set for 9/4.

PR #2076, opened 2018-09-19
Fix keepalive events filtering

palourde:
## What is this change?

It ensures keepalive events are properly handled or filtered.

More specifically, we first made sure that keepalive events have a proper check history, which wasn't the case because of the bogus event created by `keepalived`.

Then, we fixed the `IsResolution` method which wrongly reported these keepalive events (and other events as a matter of fact) as resolution events. For reference, `eventd` adds the current status at the end of the check history so by the time  `IsResolution` was called, it always used the current status (`len(e.Check.History)-1]`) and not previous one (`len(e.Check.History)-2]`).

Fixes https://github.com/sensu/sensu-go/issues/1999
Fixes https://github.com/sensu/sensu-go/issues/1924

Added!

Nope!

Yes, a keepalive should probably be handled as an event and not as a special snowflake.

Not required.


backend/keepalived/keepalived_test.go
```diff
@@ -246,8 +246,6 @@ func TestCreateKeepaliveEvent(t *testing.T) {
 	assert.Equal(t, "keepalive", keepaliveEvent.Check.Name)
 	assert.Equal(t, uint32(60), keepaliveEvent.Check.Interval)
 	assert.Equal(t, []string{"keepalive"}, keepaliveEvent.Check.Handlers)
-	assert.Equal(t, uint32(1), keepaliveEvent.Check.Status)
+	assert.Equal(t, uint32(0), keepaliveEvent.Check.Status)
```
ccressent: It _appears_ to be fine here, but I'm a bit uncomfortable with changing the semantics of `createKeepaliveEvent()` from "the created event's check status is 1 (failing)" to "the created event's check status is 0 (passing)".

types/event.go
```diff
@@ -128,12 +128,12 @@ func (e *Event) IsResolution() bool {
...
 
 	// Try to retrieve the previous status in the check history and verify if it
-	// was a non-zero status, therefore indicating a resolution
-	isResolution := (len(e.Check.History) > 0 &&
-		e.Check.History[len(e.Check.History)-1].Status != 0 &&
+	// was a non-zero status, therefore indicating a resolution. The current event
+	// has already been added to the check history by eventd so we must retrieve
+	// the second to the last
+	return (len(e.Check.History) > 1 &&
+		e.Check.History[len(e.Check.History)-2].Status != 0 &&
```
ccressent: This is quite confusing. We should either:
- have the history include the current check and not have `Check.Status`
- have the history not include the current check and have `Check.Status`
But not both at the same time, as it duplicates information and can lead to inconsistencies. `TestEventIsResolution()` shows that.

types/event_test.go
```diff
@@ -131,13 +131,13 @@ func TestEventIsResolution(t *testing.T) {
...
 			status:   0,
-			expected: true,
+			expected: false,
 		},
 		{
 			name: "check has transitioned but still an incident",
 			history: []CheckHistory{
-				CheckHistory{Status: 0},
 				CheckHistory{Status: 2},
+				CheckHistory{Status: 0},
```
ccressent: Based on the comment above for `IsResolution()`, I don't see why, if the history includes the current check, `Check.Status` would be different than `Check.History[len(Check.History) - 1].Status`

CHANGELOG.md
```diff
@@ -46,6 +46,9 @@ sending of keepalive messages.
 - Fixed a bug in `sensuctl cluster health` that resulted in an unmarshal
 error in an unhealthy cluster.
 - Fixed a bug in the web ui, removed references to keepaliveTimeout.
+- Keepalive checks now have an history.
```
nikkiki: :fr: 🇨🇦 "an" -> "a" 😉

backend/eventd/silenced.go
```diff
@@ -118,7 +118,10 @@ func silencedBy(event *types.Event, silencedEntries []*types.Silenced) []string
 }
 
 func handleExpireOnResolveEntries(ctx context.Context, event *types.Event, store store.Store) error {
-	if !event.HasCheck() || !event.IsResolution() {
+	// Make sure we have a check that contains at least one silenced ID. Then,
+	// verify if the event is a resolution, which is a more expensive operation to
+	// perform
+	if !event.HasCheck() || len(event.Check.Silenced) == 0 || !event.IsResolution() {
```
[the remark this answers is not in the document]
palourde: Ha, yes. I added this condition to prevent some tracing statements from appearing multiple times but completely forgot to remove that since it's definitely not an optimization.

backend/eventd/silenced_test.go
```diff
@@ -187,6 +187,7 @@ func TestHandleExpireOnResolveEntries(t *testing.T) {
 	resolution := func(e *types.Event) *types.Event {
 		e.Check.History = []types.CheckHistory{
 			types.CheckHistory{Status: 1},
+			types.CheckHistory{Status: 0},
```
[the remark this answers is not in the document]
palourde: No, because this test assumed the last element in the check history was the _previous_ check result and not the current one, which isn't the case. And it was using the check status as the current one.

backend/keepalived/keepalived.go
```diff
@@ -311,7 +305,7 @@ func (k *Keepalived) handleUpdate(e *types.Event) error {
 		return err
 	}
 	event := createKeepaliveEvent(e)
-	event.Check.Status = 0
```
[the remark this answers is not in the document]
palourde: Makes sense!

backend/keepalived/keepalived.go
```diff
@@ -258,14 +258,8 @@ func createKeepaliveEvent(rawEvent *types.Event) *types.Event {
 		Handlers:     []string{KeepaliveHandlerName},
 		Environment:  rawEvent.Entity.Environment,
 		Organization: rawEvent.Entity.Organization,
-		Status:       1,
+		Executed:     time.Now().Unix(),
```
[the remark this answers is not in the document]
palourde: Not really; in the original design, keepalives weren't supposed to be stored but turns out we do, so in order to provide a valid history, we fixed that. That being said, I will definitely add a test condition for that in our QA checklist!

types/event.go
palourde: I think the first option should be avoided since it would be a major breaking change.

The second one is probably the right one but it would definitely be a breaking change, both in our code and for our users since a lot of logic (especially the flapping one) is expecting the current status in the history. Because of that, I would suggest we create a new issue so we can ship this bugfix ASAP and then we can take the time to redesign the check history!

CHANGELOG.md
palourde: 🥖

backend/keepalived/keepalived_test.go
ccressent: Yeah, that's why I said it appears to be fine. Generally speaking though, those kind of changes always trigger paranoid mode in my head :)

Agreed with the naming convention! No need for a change now though: it's clear we're on the same page and that function is not exported or used in a lot of places anyway.

types/event.go
palourde: Just created https://github.com/sensu/sensu-go/issues/2093 so we can keep track of that!

types/event_test.go
palourde: Ha my bad!

PR #2080, opened 2018-09-20
[Web UI] Only display list item toolbar on hover

### LINKED ISSUE #1938: [Web UI] List improvements

jamesdphillips:
## What is this change?

Improves lists by...

a. only displaying an item's toolbar when it is hovered,
b. disabling individual toolbars when one or more items are selected,
c. and disables editing entirely when screen is very narrow.

---

Closes #1938


dabria:
> only displaying an item's toolbar when it is hovered

I like this, but I wonder if it'd be good to have this based on a media query. I feel like a large enough screen would welcome having them displayed normally. 

Do you know how this hover behaviour might affect someone on a tablet?

jamesdphillips:
> Do you know how this hover behaviour might affect someone on a tablet?

Good call out. I've changed it to use the `any-hover` [media query](
https://www.w3.org/TR/mediaqueries-4/#mf-interaction); this way on devices like the iPad all the actions are still visible.

PR #2088, opened 2018-09-20
Set gRPC to use Logrus logger

### LINKED ISSUE #2054: non json WARNING messages being issued on stdout
my sensu-backend is throwing non JSON output WARNING messages to stdout.
Could impact log parsing, if there is a strong expectation that backend will be produce json to stdout.

## Expected Behavior
json output to stdout

## Current Behavior
Getting this message:
WARNING: 2018/09/09 13:25:08 grpc: addrConn.resetTransport failed to create client transport: connection>

The warning (i think) has to do we me trying to start a local sensu-backend service while I had a vagrant managed virtual machine running with some of its ports partially mapped to localhost via NAT.

The warning itself isn't such a big deal, but the format is a break from the expected json output to stdout.

I believe Eric made mention that it has something to do with a library logging and by passing the expected logging facility/formatter.
The log message seems to be from:
google.golang.org/grpc/clientconn.go

1.
2.
3.
4.

## Context
ran across this Warning while trying to help sort out an unrelated log parsing issue associated with journalctl.
ideally it would be nice to be able to take all the stdout message and pass them into jq for log parsing.
this Warning message prevents that.

## Your Environment

sensu-backend-2.0.0~nightly+20180911-1.x86_64 rpms running a a fedora and centos 7 systems.

cyphus:
## What is this change?

Set gRPC to use Logrus logger

Closes #2054 

gRPC-based log messages were being output directly to stderr.

Added

I'm good.

Complications explained in https://github.com/sensu/sensu-go/issues/2054#issuecomment-423290450

I upgraded etcd from 3.3.2 to 3.3.9. I've carefully reviewed the etcd changelog before doing this. While upgrading etcd didn't fix the issue like I intended (the logger removal is part of 3.4). I decided to keep the patch upgrade.

I also found some lint issues unrelated to gRPC logging. The fixes are included in this PR.

No new docs required :)


palourde:
@cyphus Travis is sad :cry:

cyphus:
Ugh, what a mess. The gRPC docs say not to set the logger outside of `init()` functions, but that's exactly what etcd does. If we overwrite the gRPC logger after etcd start, we create a race condition for the gRPC logger variable.

The only real way out of this is to remove the logger setting code from the etcd server init function, which means forking etcd.

I'll update the PR to depend on forked etcd.

cyphus:
https://github.com/sensu/sensu-go/issues/2054#issuecomment-423290450

Issue #2077, opened 2018-09-19
Failing keepalives events are not always emitted


palourde:
## Expected Behavior
When an agent goes away, I expect failing events for this entity/keepalive combinaison to be created once the timeout is reached.

## Current Behavior
Possibly half the time, no failing events are created.

@ccressent and I both noticed this behaviour.

## Possible Solution
I don't have a lot of idea so far. I didn't have time yet to investigate but maybe it has something to do with the number of keepalive messages sent by the agent (just like if the backend didn't had time to create a monitor for that entity).

## Steps to Reproduce (for bugs)

1. Start the backend
2. Start the agent
3. Wait for the backend to receive at least one keepalive message and stop the agent
4. Look at the backend logs/sensuctl to look for a failing keepalive event.

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): Running on master.
* Installation method (packages, binaries, docker etc.): Manually compiled.
* Operating System and version (e.g. Ubuntu 14.04): macOS 10.13.6

palourde:
I wasn't able to identify an exact pattern but what I've found so far to reproduce this issue:

* When the backend is started, the agent's entity must not have a failing keepalive event.
* Once the agent is started, the first time it shutdowns, the backend won't create any failing keepalive event.
* The second time the agent shutdowns (while the backend still runs), then and only then a failing keepalive event will be produced.

palourde:
After digging into the code and a discussion with @mercul3s, I'm pretty sure we identified the problem:

1. The first time ever our agent connects to the backend, the _monitor_ grants a lease on the key `/sensu.io/monitors/{entity_id}`:
https://github.com/sensu/sensu-go/blob/1af5b06711056ab5c780fcd67e6a9c9667e924b3/backend/monitor/supervisor.go#L80-L91
2. If the backend stops/restarts while the agent is still connected (so the agent's entity never creates a keepalive failure), the lease is not deleted but the goroutine watching the lease is stopped:
https://github.com/sensu/sensu-go/blob/1af5b06711056ab5c780fcd67e6a9c9667e924b3/backend/monitor/supervisor.go#L140-L156
3. Once the backend starts again, etcd will set back the lease on our key to its original value (e.g. `60`).
4. If an agent connects to the backend before the lease expires, the Monitor method will only extend the lease without starting a new goroutine (which is responsible for creating an event for the failing keepalive), assuming an another cluster is running this goroutine:
[...]

PR #2091, opened 2018-09-20
Revoke etcd leases on shutdown in monitor

palourde:
## What is this change?

It makes sure a monitor goroutine revoke its lease during a graceful shutdown, so a new goroutine can be created at the next startup.

Also:
* etcd is now the last component to shutdown during a graceful shutdown
* Added some debug logging statements in monitor to help troubleshooting

Closes https://github.com/sensu/sensu-go/issues/2077

Added

There might be a better to do this!

Nope

Nope


echlebek:
This commit has caused a goroutine leak in the monitor (which I am investigating independently of this)

https://github.com/sensu/sensu-go/blob/master/backend/monitor/supervisor.go#L169

This will cause a goroutine to be run for each call to watchMon. The goroutine will never shut down until a signal is issued. So the goroutines grow unbounded.

I'm going to replace this strategy with a single goroutine that will Revoke a slice of leases when the signal is caught.

Issue #2057, opened 2018-09-13
[Windows] Possible issue with Sensu not passing PATH to 'command' runs


absolutejam:
I'm still trying to get to the bottom of this one as there's a few different working parts, but I *think* the 
Windows sensu agent may not be passing the correct PATH to `command` runs.

I have `sensu.exe` running via. NSSM. First I thought it was that, but after creating a test service that simply dumps the `PATH` env var via. Powershell to a log file, I see the correct `PATH` being printed.

This came about when I was trying to run the check `check-ping.rb` (sensu-plugins-network-checks) and it kept failing. Further digging looked like it was failing to find the host's `ping` command (Which the `net-ping` module uses under the covers) and ultimately after more testing, it looks like the `PATH` variable is being ignored. For example, `ping` and `ruby` do not fire unless the full path is used, but are in the `PATH` (`'ping' is not recognized as an internal or external command,`) but `cmd` does work? I also tried firing a simply ruby script that puts the `ENV['PATH']` but this returns an empty string.

If I use a `command` of `cmd /C echo %PATH%` I simply get `%PATH%` returned, but if I do `cmd /C echo %PATHEXT%`, I get the expanded environment variable. Could this be a culprit?

## Expected Behavior
`PATH` should be passed to `command` runs

## Current Behavior
`PATH` doesn't seem to be passed to `command` runs

## Steps to Reproduce (for bugs)
1. Fire a check that uses the `PATH` in some way, eg. `ping 127.0.0.1 -n 1` and check the event output.

## Your Environment
```
sensu-agent version 2.0.0-beta.4-4#a05ad41, build a05ad411199077286f49c06629eed4823331176c, built 2018-08-16T21:12:56+0000
```

absolutejam:
Alright, I don't want to get too far off into the weeds with Go because I know I'll fall flat on my face, buuuuut I was trying to at least lighten the load with some ideas. Please note, this post is an amalgamation of Google code snippets & 'Go for dummies 101' level knowledge.

So, https://github.com/sensu/sensu-go/blob/249a69d1f3b19de832d8753974489ef0bbd3a87d/agent/hook.go#L55-L62 shows the command object being initiated. 

And in https://github.com/sensu/sensu-go/blob/249a69d1f3b19de832d8753974489ef0bbd3a87d/command/command.go#L39-L44 there's the option of `Env` slice of strings. Does this need to be populated?

I read the docs on `os/exec` and it does state:
```
[...]

PR #2098, opened 2018-09-21
Be case-insensitive when dealing with env. vars

palourde:

## What is this change?

Windows uses `Path` instead of `PATH` so be case-insensitive when looking for that env. variable.

Closes https://github.com/sensu/sensu-go/issues/2057

Added!

Nope

Nope

Nope


oferrigni:
Tests?

agent/assetmanager/manager.go
```diff
@@ -77,15 +77,16 @@ func getSystemEnviron() []string {
 
 	for _, e := range env {
 		pair := strings.Split(e, "=")
-		key, _ := pair[0], pair[1]
+		// Transform the key to uppercase because Windows uses "Path" and not "PATH"
+		key := strings.ToUpper(pair[0])
 
-		if presentVars[key] != true {
+		if !presentVars[key] {
 			presentVars[key] = true
```
cyphus: Looks like we can unconditionally set presentVars[key] to true here.

palourde:
@oferrigni We are calling `os.Environ()` within this function, which makes it very hard to test. I'll look into refactoring it so it's easier to test!

agent/assetmanager/manager.go
palourde: Ha, good catch!

PR #2115, opened 2018-09-24
Mock out command execution

### LINKED ISSUE #1827: [Build stability] TestExecuteCheck
```
--- FAIL: TestExecuteCheck (6.73s)
	assertions.go:239: 
                          
	Error Trace:	check_handler_internal_test.go:137
		
	Error:      	Not equal: 
		
	            	expected: 2
		
	            	actual: 0
		
	Messages:   	{"timestamp":1531440760,"entity":{"class":"agent","deregister":false,"deregistration":{},"environment":"default","id":"APPVYR-WIN","keepalive_timeout":120,"last_seen":1531440753,"organization":"default","subscriptions":null,"system":{"hostname":"APPVYR-WIN","os":"windows","platform":"Microsoft Windows Server 2012 R2 Datacenter","platform_family":"Server","platform_version":"6.3.9600 Build 9600","network":{"interfaces":[{"name":"Ethernet 8","mac":"00:15:5d:07:ef:df","addresses":["fe80::6dd2:1287:a208:840e/64","192.168.2.182/21"]},{"name":"Loopback Pseudo-Interface 1","addresses":["::1/128","127.0.0.1/8"]},{"name":"isatap.{0080E500-4205-442C-9729-FDB7CEB0507A}","mac":"00:00:00:00:00:00:00:e0","addresses":["fe80::5efe:c0a8:2b6/128"]}]},"arch":"amd64"},"user":"agent"},"check":{"command":"..\\bin\\tools\\cat.exe C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\metric677696619","environment":"default","handlers":[],"high_flap_threshold":0,"interval":60,"low_flap_threshold":0,"name":"check","organization":"default","publish":true,"runtime_assets":["ruby-2-4-2"],"subscriptions":["linux"],"proxy_entity_id":"","check_hooks":[{"non-zero":["hook1"]}],"stdin":true,"subdue":null,"ttl":0,"timeout":1,"round_robin":false,"duration":1.5272485,"executed":1531440758,"history":null,"issued":1531440753,"output":"Execution timed out\n","status":2,"total_state-change":0,"last_ok":0,"occurrences":0,"occurrences_watermark":0,"output_metric_format":"graphite_plaintext","output_metric_handlers":null,"env_vars":null},"metrics":{"handlers":null,"points":null}}
```

nikkiki:

## What is this change?

This change mocks command execution involving the least amount of changes to `agent.go`. By making the function used to execute a command an argument of the agent, we're able to inject dummy functions and mock execution results for testing.

Closes #1827. Also, quality of life. By mocking out command execution, we don't rely on build tools or actual execution during testing.

yas

nah

nah

we good


grepory:
This is heading in the right direction, but I'd like to figure out a way to make it a bit more forward thinking.

For example, we reuse executions across the agent and pipelined. Is there a way to make the interface more portable?

agent/agent.go
```diff
@@ -129,6 +130,8 @@ type SocketConfig struct {
 	Port int
 }
 
+type executionFunction func(ctx context.Context, execution *command.Execution) (*command.Execution, error)
```
golangcibot: `executionFunction` is unused

grepory:
There is a lot going on in this package that I did not understand at first, but I'm starting to understand now.

There are a few things here that concern me.

The use of an Execution to act on another Execution and returning the Execution that was passed in to ExecuteCommand(). That's pretty confusing, and unnecessary. If I understand this, you're reusing an Execution in order to create a mock of an Execution. That's not quite what I had in mind, but I see where you were going with this.

Let's think through this for a moment.

We have multiple use cases for executing an external command and getting its output. They all take the same configuration data that we see in the Execution struct:

```golang
type Execution struct {
	// Command is the command to be executed.
	Command string

	// Env ...
	Env []string

	// Input to provide the command via STDIN.
	Input string

	// Execution timeout in seconds, will be set to a default if
	// not specified.
	Timeout int
}
```

We are currently reusing this object to record the response of the external process and some metadata as well. We're also overloading this object with a mutex that's used in the agent to avoid parallel check execution. That's some pollution that we should get rid of, but for now, let's carry it forward.

If the goal is to say "whenever we're testing, we want to not actually execute commands, but still pretend to act on a configuration for a thing to execute," then we need a way to take in an Execution (or some object that is a request to execute a command) and then return the response. The actor here, the Executor, is the thing we want to fake.

So there are a couple of interfaces here we could use, but I think the latter requires the fewest changes:

```golang
type Executor interface {
  func Execute(req *ExecutionRequest) *ExecutionResponse
}

// or

type Executor interface {
  func Execute(ex *Execution)
}
```

Everywhere we want to actually shell out and execute a thing, we can use the real Executor from the `command` package (maybe `ExecExecutor` since we use the exec package to do the dirty work), but in tests, we can create our own executor that serves our purpose for testing.

Think about that a bit, and then let's try again. :)

nikkiki:
Plan is to ship as is, and revisit the ExecutorFunc type. @palourde could I get 👍 or 👎 on the review?

PR #2126, opened 2018-09-26
[Web UI] Checks Page Publishing/Unpublishing

### LINKED ISSUE #1935: [Web UI] Checks page improvements

dabria:
## What is this change?
Closes #1935 
Adds the ability to publish and unpublish checks, either individually through the overflow icon on the right side, or in bulk with the checklist. A confirmation toast will appear to let the user know if the action succeeded.

screen shot 2018-09-26 at 3 33 51 pm
screen shot 2018-09-26 at 3 34 03 pm
screen shot 2018-09-26 at 3 38 48 pm

Users should have the ability to [un]publish in the UI

Yes, I added one.

- We should create a new ticket for adding a graphql fragment to the CheckListHeader component.
- We should also probably allow this behaviour on the check details page
- I have not collapsed the delete button from the bulk actions header due to #2056 
- Toasts occur on success for this and execution, do we want silencing to also have a success toast?

I had to remove defaults that were set on check creation (interval and publish). Interval was causing a bug when trying to unpublish/publish checks with cron specified.

There was discussion about adding the ability to "undo" on the notification toast. It seems rather simple to save a list of the checks interacted with, then invert them if the user clicks "undo". However, some discussion was had and it seems like the act of undoing as a whole might want to be implemented in a smarter way. Perhaps deleting should also allow for undoing. There's a conversation to be had there, and for now I chose to not implement the behaviour.


dashboard/src/components/partials/ChecksList/ChecksListItem.js
```diff
@@ -87,25 +90,12 @@ class CheckListItem extends React.Component {
...
-                  )}
-                </strong>{" "}
-                by{" "}
-                <strong>
-                  {check.subscriptions.length}{" "}
-                  {check.subscriptions.length === 1
-                    ? "subscription"
-                    : "subscriptions"}
-                </strong>.
+                {check.publish && <CheckPublishInfo check={check} />}
```
jamesdphillips: 🤔 how would you feel about naming this component CheckSummary or CheckDescriptor and including the unpublished case as well?

dashboard/src/components/partials/ChecksList/CheckPublishInfo.js
```diff
@@ -0,0 +1,38 @@
...
+  };
+
+  render() {
+    const { check } = this.props;
+
+    return (
+      <React.Fragment>
+        Executed{" "}
+        <strong>
+          {check.interval ? (
```
jamesdphillips: Ideally this component would include a query fragment describing the fields it requires. In this case it's referencing the `interval`, `cron`, and `subscription` fields.

dashboard/src/components/partials/ChecksList/ChecksListHeader.js
```diff
@@ -9,24 +9,30 @@ import ListSortMenu from "/components/partials/ListSortMenu";
...
     onClickDelete: PropTypes.func.isRequired,
     onClickExecute: PropTypes.func.isRequired,
+    onClickSetPublish: PropTypes.func.isRequired,
     onClickSilence: PropTypes.func.isRequired,
     rowCount: PropTypes.number.isRequired,
     selectedItems: PropTypes.array.isRequired,
     toggleSelectedItems: PropTypes.func.isRequired,
   };
 
+  // TODO add a fragment
```
jamesdphillips: 😬 my bad

dashboard/src/components/relocation/PublishCheckStatusToast.js
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  render() {
+    const { mutation, onClose, checkName, entityName } = this.props;
+    const { loading } = this.state;
+
+    const subject = (
+      <React.Fragment>
+        <strong>{checkName}</strong>
+        {entityName && (
```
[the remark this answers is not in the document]
dabria: Good catch. Remnants of copying the Execute toast

dashboard/src/components/partials/ChecksList/ChecksListItem.js
dabria: Seems like a good idea, I'll add that.

backend/apid/graphql/schema/mutations.graphql
```diff
@@ -87,7 +87,7 @@ input CheckConfigInputs {
   command: String
 
   "interval is the time interval, in seconds, in which the check should be run. Defaults to 60."
-  interval: Int = 60
+  interval: Int
```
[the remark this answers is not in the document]
dabria: Good catch!

Issue #2072, opened 2018-09-17
Decouple asset manager from agent


cyphus:
Decouple sensu-agent from the asset manager by having [the agent struct](https://github.com/sensu/sensu-go/blob/bb39ab1dccfa8e1e953b0d3c029b3324fce3aaad/agent/agent.go#L188) depend on the [boltdb Getter interface](https://github.com/sensu/sensu-go/blob/cce7a1c02aa93d8d634f6853d714ed8609b9ad80/asset/boltdb_manager.go#L20-L22) instead of an `assetmanager.Manager` directly.

grepory:
duplicate of #2038

PR #2132, opened 2018-09-27
Refactor agent to use boltdb-based asset manager

### LINKED ISSUE #2038: Re-work sensu-agent to use new asset package
Requires #2052 
Requires #2072

### LINKED ISSUE #2128: Default agent cache directory doesn't exist on macOS
## Expected Behavior
Sensu agent uses a reasonable default cache directory for each platform

## Current Behavior
Sensu agent uses /var/cache/sensu, but fails to create the directory on macOS because /var/cache doesn't exist.

## Possible Solution
Create a macOS specific default path in [path.go](https://github.com/sensu/sensu-go/blob/f5b8db4d5d75c9ed9960ff6d5ed5bd905a3f9812/util/path/path.go#L37-L45) e.g. `/Library/Caches/sensu`.

1. Start sensu-agent on macOS with the default cache directory (omit `--cache-dir` from CLI and `cache-dir` from agent config.
1. Attempt to run a check on the agent with an asset.

## Context
Brought up by a user in #sensu2 community slack.

## Your Environment
- macOS 10.13.6
- Latest sensu nightly as of September 18th.

cyphus:
## What is this change?

A new BoltDB-based asset manager package was added in #2024, but the package wasn't being used by any other part of sensu-go. This PR deletes the (now legacy) agent/assetmanager package and replaces it with the asset package.

I also fixed some unrelated lint errors, and made the agent tests use a temporary directory for asset cache (and clean up said temporary directory after testing is finished).

Closes #2038 
Closes #2128

The old asset package was using mutexes and lockfiles to handle concurrent asset installs. BoltDB is a much better solution to this issue.

Added!

I'm good!

I wish I didn't have to include unrelated linter fixes but that's how it goes.

Assets have not changed at all in terms of how they are used. I added tests to ensure assets have the same behavior as before.


cyphus:
I'm going to update this PR per discussion in #2128

asset/filtered_manager.go
```diff
@@ -0,0 +1,65 @@
...
+	if filtered {
+		return nil, nil
+	} else {
+		return f.getter.Get(asset)
+	}
+}
+
+// isFiltered evaluates the given asset's filters and returns true if any of
+// them match the current entity.
+func (f *filteredManager) isFiltered(asset *types.Asset) (bool, error) {
```
grepory: If multiple filters exist, they are to be AND'd, but this ORs them.

asset/filtered_manager.go
```diff
@@ -0,0 +1,65 @@
...
+	return &filteredManager{
+		getter: getter,
+		entity: entity,
+	}
+}
+
+// filteredManager evaluates asset filter(s) and calls the underlying
+// asset Getter based on filter result.
+type filteredManager struct {
+	entity *types.Entity
```
grepory: Rather than coupling Getter to entity, can we abstract this a bit and have a PredicatedGetter where the predicate is supplied and evaluated indirectly?

grepory: You could then test your predicate separately from your predicated getter and your getter.

palourde:
AppVeyor is sad 😢

asset/filtered_manager.go
cyphus: I was so focused on `filteredManager` having a pluggable asset manager backend that I didn't even notice I could make the filter (predicate) pluggable as well. Good catch!

echlebek:
I'll take over this PR.

echlebek:
I rebased the commits onto the latest master and fixed a DCO problem. I'm not sure if there is any more work outstanding here.

grepory:
closes #2072

asset/filtered_manager.go
nikkiki:
```
// isFiltered... returns true if any of them match the current entity
```
description and functionality should actually reflect
```
// isFiltered... returns true if all of them match the current entity
```
Let me update the function and the tests here.

nikkiki:
Okay, I've addressed @grepory PR feedback, but it doesn't look like the assets are actually getting installed. Troubleshooting that now!

```
$ sensuctl event list
        Entity              Check                        Output                  Status   Silenced             Timestamp            
 ──────────────────── ────────────────── ────────────────────────────────────── ──────── ────────── ─────────────────────────────── 
  Nikkis-MacBook-Pro   check-with-asset   sh: echo-asset.sh: command not found      127   false      2018-10-16 09:04:54 -0700 PDT  
```

nikkiki:
I totally lied, assets _are_ being installed properly! User error: I wasn't labelling the cache directory with a absolute path!

```
$ sensuctl event list
        Entity              Check                       Output                  Status   Silenced             Timestamp            
 ──────────────────── ────────────────── ───────────────────────────────────── ──────── ────────── ─────────────────────────────── 
  Nikkis-MacBook-Pro   check-with-asset   you have installed an asset, kudos!        0   false      2018-10-16 09:38:44 -0700 PDT  
```

This PR should be good to go for review!

PR #2162, opened 2018-10-10
[Web UI] Check details page improvements

### LINKED ISSUE #1934: [Web UI] Check details page improvements

- [ ] The command value does not wrap, meaning large values can render past the edge of the row and underneath content to it's right.
- [ ] Environment variables also do no wrap and are clipped by the container, meaning the entire value cannot viably be viewed.
- [ ] Actions are missing for silencing or un-silencing a check. Especially important if we decide to remove checkboxes on narrower viewports.

dabria:
## What is this change?
Closes #1934 

screen shot 2018-10-10 at 3 15 39 pm

- [x] Command value wraps (respects spaces at the end of lines)
- [x] Environment variables wrap and are not clipped by the container.
- [x] Environment variables block respects the parent containers padding
- [x] [Un]Publishing

Improves readability and allows for [un]publishing action

Added one

### DIFF
--- a/dashboard/src/components/Code.js
+++ b/dashboard/src/components/Code.js
@@ -2,11 +2,10 @@ import * as React from "react";
 import PropTypes from "prop-types";
 import { withStyles } from "@material-ui/core/styles";
 import { fade } from "@material-ui/core/styles/colorManipulator";
-import classNames from "classnames";
 
 const styles = theme => ({
   root: {
-    whiteSpace: "nowrap",
+    whiteSpace: "pre-wrap",
     margin: "0 1px",
     padding: `${2 / 16}em ${4 / 16}em`,
     borderRadius: `${5 / 16}em`,
@@ -20,33 +19,16 @@ const styles = theme => ({
         ? theme.palette.secondary.light
         : theme.palette.secondary.dark,
   },
-  preWrap: { whiteSpace: "pre-wrap" },
-  block: {
-    // because this seems to be layered with Monospaced
-    // the background colours overlap otherwise
-    backgroundColor: "unset",
-  },
 });
 
 class Code extends React.PureComponent {
   static propTypes = {
     classes: PropTypes.object.isRequired,
-    block: PropTypes.bool,
-    preWrap: PropTypes.bool,
-  };
-
-  static defaultProps = {
-    block: false,
-    preWrap: false,
   };
 
   render() {
-    const { classes, block, preWrap, ...props } = this.props;
-    const className = classNames(classes.root, {
-      [classes.block]: block,
-      [classes.preWrap]: preWrap,
-    });
-    return <code {...props} className={className} />;
+    const { classes, ...props } = this.props;
+    return <code {...props} className={classes.root} />;
   }
 }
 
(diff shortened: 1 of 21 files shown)


dashboard/src/components/Dictionary/DictionaryValue.js
```diff
@@ -6,9 +6,13 @@ import Typography from "@material-ui/core/Typography";
 
 const styles = theme => ({
   root: {
-    display: "table-cell",
+    display: "grid",
     paddingLeft: theme.spacing.unit,
     userSelect: "text",
+    "& > code": {
```
jamesdphillips: I think a more ideal option would be to either make this an option of the codeblock component or add a wrapper component that provides this functionality. In its current state it may complicate making future changes. For instance if we wanted to have an inline code code block (eg. "The command run was <code>exit 0</code>.") we wouldn't want the implicit margin.

Being a bit more explicit would give us a bit more power tailor the output appropriately.

dashboard/src/components/Dictionary/DictionaryValue.js
```diff
@@ -6,9 +6,13 @@ import Typography from "@material-ui/core/Typography";
 
 const styles = theme => ({
   root: {
-    display: "table-cell",
+    display: "grid",
```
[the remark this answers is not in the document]
dabria: As far as I can tell, the apparent behaviour remains the same. I tested `grid` vs `table-cell` on codepen, and it seems the main difference is that `grid` takes 100% of available space, where `table-cell` stretches to the largest item. It doesn't change anything in our use from what I can see. However, I'll change this to be `inline-grid` instead. This works almost identically to `table-cell`, sizing to the largest element, while still allowing us to keep the horizontal scrollbar.

dashboard/src/components/Dictionary/DictionaryValue.js
dabria: It appears using our custom `Code` component instead of `code` solves this issue.

dashboard/src/components/Dictionary/DictionaryValue.js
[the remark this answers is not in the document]
10xjs: I think we are safe for browser support https://caniuse.com/#feat=css-grid

dashboard/src/components/Code.js
```diff
@@ -19,16 +20,28 @@ const styles = theme => ({
...
+  rightMargin: {
+    whiteSpace: "pre-wrap",
+    marginRight: "24px",
+  },
 });
 
 class Code extends React.PureComponent {
   static propTypes = {
     classes: PropTypes.object.isRequired,
+    dictionaryMargin: PropTypes.bool,
```
jamesdphillips: I probably confused you with my comment before but I think this may be a bit too specific.

dabria:
I removed `in-line grid` from the root of DictionaryValue, returning it to it's original state as `table-cell`. I noticed it rendered Code differently than it would normally look by removing some linespacing on a line break. Instead, since we only need that setting for scrollable CodeBlocks, I've created an additional prop to pass that in instead.

dabria:
I've implemented code highlighting as well. However, more work will be needed for proper theme colour swapping. It doesn't look great in dark mode at the moment. In addition colour-blind themes would be a great bonus.
screen shot 2018-10-24 at 3 51 10 pm

PR #2170, opened 2018-10-11
RequestInfo middleware

ccressent:
This new middleware gives us the ability to update the request context with a lot of details subsequent middlewares, such as authorization and authentication, will use.
Those details are:
- API group
- API version
- Namespace
- Resource type
- Resource name
- Verb

I'm looking for feedback specifically on:
- The name "RequestInfo"
- the format of supported URL paths
- the supported verbs; how do we support "list" and "watch"?
- should `types.RequestInfo` be in protobuf format, like a lot of the other types in the `types' package? I don't think so since it's not stored?
- do I need to regenerate some form of type map to properly add `types.RequestInfo` to package `types`?


backend/apid/middlewares/request_info_test.go
```diff
@@ -0,0 +1,226 @@
...
+		{
+			description: "path valid up to specific resource type",
+			method:      "GET",
+			path:        "/apis/rbac/v1alpha1/namespaces/my-namespace/check",
+			expected: types.RequestInfo{
+				APIGroup:   "rbac",
+				APIVersion: "v1alpha1",
+				Namespace:  "my-namespace",
+				Resource:   "check",
+				Verb:       "get",
```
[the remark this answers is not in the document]
grepory: Yes please. If there's no name, then change the verb to list.

ccressent:
Indeed, I glanced at the gorilla API and you're totally right.
Since we're already using gorilla for the API server, I might as well use it here too!

backend/apid/middlewares/request_info.go
```diff
@@ -0,0 +1,82 @@
...
+				if info.Namespace == "" {
+					info.Verb = "list"
+				}
+
+				// Likewise, if there is no resource name specified, the request
+				// targets /namespaces/{namespace}/{type} and a GET on this endpoint
+				// corresponds to a list verb.
+				if info.Namespace != "" && info.Resource != "" && info.ResourceName == "" {
+					info.Verb = "list"
+				}
```
palourde: I believe this bit could be simplified by first replacing line 38 with the following line, since the resourceName here should be the namespace name:

```
router.Path("/{name}").Methods("GET", "HEAD", "PUT", "DELETE")
```

And then we could replace this whole block with:

```
if info.Verb == "get" && info.ResourceName == "" {
  info.Verb = "list"
}
```

WDYT?

backend/apid/middlewares/request_info.go
```diff
@@ -0,0 +1,82 @@
...
+func (i RequestInfo) Then(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+		info := types.RequestInfo{}
+
+		router := mux.NewRouter().PathPrefix("/apis/{group}/{version}/namespaces/").Subrouter()
+		router.Path("").Methods("GET", "HEAD", "POST")
+		router.Path("/{namespace}").Methods("GET", "HEAD", "PUT", "DELETE")
+		router.Path("/{namespace}/{type}").Methods("GET", "HEAD", "POST")
+		router.Path("/{namespace}/{type}/{name}").Methods("GET", "HEAD", "PUT", "DELETE")
```
palourde: We also need to support resources that are within namespaces (e.g. users).

grepory: That.

Also, I now have a question about these router paths.

I don't think you should be able to GET/HEAD/POST on `/apis/group/version/namespaces`. Was the intent to allow people to easily list everything in all namespaces? I could maybe agree with that, but let's not allow POST here. Let's require API clients to fully qualify the path of the thing they want to create.

As far as GET for this route goes, I'd almost rather API clients had to enumerate the namespaces they had access to and then query all objects of a given type across those namespaces. It'd be a heavier query, but I'd rather put that burden on API clients and not the apiserver.

The same goes for `/apis/group/version/namespaces/namespace`. GET would list all objects for that group in the given namespace at the specified version. Again, I'd rather people had to query specific types instead of us having to iterate through all known types for that version and that group. I don't see this being a compelling use case to support this route.

grepory: Also, you certainly shouldn't be able to DELETE on `/apis/group/version/namespaces/namespace`. I'm not even sure what that means.

grepory: I think what we actually want to support is:

`GET, HEAD, POST /apis/group/version/type` :: operate on a set of a versioned non-namespaced type.

`GET, HEAD, POST /apis/group/version/namespaces/namespace/type` :: operate on a set of a versioned namespaced type.

`GET, HEAD, PUT, DELETE /apis/group/version/type/name` :: operate on a versioned non-namespaced object of a specified type

`GET, HEAD, PUT, DELETE /apis/group/version/namespaces/namespace/type/name` :: operate on a versioned namespaced object of a specified type.

Do we actually need any other routes than those?

backend/apid/middlewares/request_info.go
```diff
@@ -0,0 +1,82 @@
...
+			info.Namespace = match.Vars["namespace"]
+			info.Resource = match.Vars["type"]
+			info.ResourceName = match.Vars["name"]
+
+			if info.Verb == "get" {
+				// If there is no namespace specified, the request targets
+				// /namespaces and a GET on this endpoint corresponds to a list verb.
+				if info.Namespace == "" {
+					info.Verb = "list"
+				}
```
grepory: As I mentioned above, I don't think we should build a route that lists across all namespaces.

backend/apid/middlewares/request_info.go
echlebek: > Also, you certainly shouldn't be able to DELETE on `/apis/group/version/namespaces/namespace`. I'm not even sure what that means.

How will people delete namespaces?

grepory: We have to have a namespace API much like we have an API for organizations and environments now.

grepory: Which reminds me, we need to validate namespaces when they're provided by the user in input.

ccressent: > I don't think you should be able to GET/HEAD/POST on
> `/apis/group/version/namespaces`. Was the intent to allow people to easily
> list everything in all namespaces?

The intent is to be able to list the available namespaces (GET) and create a new
namespace (POST).

`Namespace`, [defined here](https://github.com/sensu/sensu-go/blob/devel/api/internal/apis/core/namespace.go),
appears to be just another object. So a GET on `/namespaces` is a list verb
internally and would return a list of `Namespace` objects, but not all the
objects that are under all the namespaces. A POST on `/namespaces` expects a
`Namespace` object in the request body, which would hold the name of the new
namespace that needs to be created.

> I could maybe agree with that, but let's not allow POST here. Let's require
> API clients to fully qualify the path of the thing they want to create.

If that's also what we do with all the other object types, then yes. If not, I
would disagree and err on the side of consistency for all object types.

My understanding was that, for example, to create a `Check` object, one would
POST the object, with its name in its metadata, to `{...}/checks`. That would
result in an object being available at `{...}/checks/{name}`. I thought all
objects would work the same.

> The same goes for `/apis/group/version/namespaces/namespace`. GET would list
> all objects for that group in the given namespace at the specified version.
> Again, I'd rather people had to query specific types instead of us having to
> iterate through all known types for that version and that group. I don't see
> this being a compelling use case to support this route.

Based on my answer above, GET on
`/apis/{group}/{version}/namespaces/{namespace}`
would only return a single `Namespace` object, not everything that's in the
namespace. It doesn't sound very useful, and it's mostly for consistency with
all the other object types for which a GET on a specific object name gives you
back that object.

> Also, you certainly shouldn't be able to DELETE on /apis/group/version/namespaces/namespace. I'm not even sure what that means.

An HTTP DELETE on `/apis/{group}/{version}/namespaces/{namespace}` deletes the
`Namespace` object with name `{namespace}`.

ccressent:
Let me take a step back and try to clarify a few things:

- the purpose of this middleware is purely to inspect the path of requests
  coming in and populate a `RequestInfo` object in the request's context. The
  goal being that any subsequent middleware can then use that information
  without having to do manual path parsing. Any allowing/disallowing of methods
  on endpoints has no place in this middleware, conceptually.

- the HTTP methods listed here do not allow or disallow anything. They are there
  because they are needed by the gorilla router to produce a path match. View
  those as a "hack" to make the router based approach to path parsing work,
  essentially! For example, a PATCH request would still go through, but the
  request would have an empty `RequestInfo` object in its context because such a
  request would not match anything in the router as currently defined.

- I want to reiterate that my initial intent was to only try to extract as much
  info here as possible, with no regard for endpoint, allowed methods or
  anything of the sort. Unfortunately, although it makes the logic simpler and
  cleaner, using gorilla leaks a bit of that concern into the middleware!
  Again, the router here isn't used for anything but free path parsing.

ccressent:
Regarding resources outside of namespaces, I agree, and I will add support for that!
That is, `RequestInfo` populated properly with `Resource` and `ResourceName` even if no namespace is in the path.

backend/apid/middlewares/request_info.go
grepory: You don't manage Namespaces within, for example, the RBAC API. Namespaces are global objects that are defined via a namespace API (assuming it is in core) `/apis/core/{version}/namespace`. Consequently, "list"-ing namespaces is done via the namespace API, not via `/apis/{group}/{version}/namespaces`.

My comment about POST above is that if you want to create an object of a specific type, you would `POST /apis/{group}/{version}/{type}`. 

These behaviors should be consistent across all object types.

Again, to be clear, interacting with `Namespace` objects only happens within the core API where the `Namespace` object is defined.

As a general rule of thumb: you can only modify objects that belong to an API group _within_ that API group.

grepory: To be very clear, the only ways to interact with `Namespace` objects are:

`GET /apis/core/v1/namespace` - list namespaces
`GET /apis/core/v1/namespace/{name}` - get a specific namespace
`HEAD /apis/core/v1/namespace/{name}` - validate the existence of a namespace
`POST /apis/core/v1/namespace` - create a new namespace
`DELETE /apis/core/v1/namespace` - delete a namespace (except you cannot delete the default namespace

I don't even think we should allow PUT to namespaces.

grepory:
Hmm. That is very confusing, but maybe I just don't understand how Gorilla routers work. 

The comments in request_info.go on lines 20-22 are what I'm taking issue with. I take those to mean that they are the behavior encoded in your router. If that is not the case, then please update the comment.

If that is the case, then this behavior is incorrect.

PR #2172, opened 2018-10-12
[Web UI] Clear filter button

### LINKED ISSUE #2094: [web-ui] Feature request: clear filters button
## Expected Behavior
If filters can be created via the mouse by clicking menu items, they should be able to be cleared via the mouse as well.

This request is for a small `X` near the filter text at the top of the page. Clicking it should clear all filters and update the page accordingly.

## Current Behavior
Right now, if I'm creating filters using the mouse, in order to clear it I have to:
- select all text in the filters text box
- hit `backspace` to clear the text box, then `return` to update the page

In doing so I have to move from mouse to keyboard and back to mouse to continue interacting with the UI. Adding a dedicated button to clear all filters would be a nice shortcut for this use case.

## Context
I noticed this while going through the QA checklist for the frontend and had to repeatedly set and clear filters.

dabria:
## What is this change?
Closes #2094 

Allows the user to clear their filter easier

I don't think so

I did notice that when applying filters (like on the Events List) the `AppLayout-content` div occasionally moves approx. 5px to the right. It looks like it's related to the automatic margins. Setting it to a pixel value would stop this, but be frustrating due to having to add more breakpoints to make it scale nicely. A percent value would help with scaling, but still moves slightly when the data set is empty.

### DIFF
--- a/dashboard/src/components/SearchBox/SearchBox.js
+++ b/dashboard/src/components/SearchBox/SearchBox.js
@@ -7,7 +7,8 @@ import Paper from "@material-ui/core/Paper";
 import Typography from "@material-ui/core/Typography";
 import Icon from "@material-ui/icons/FilterList";
 import IconButton from "@material-ui/core/IconButton";
-import ClearIcon from "@material-ui/icons/Clear";
+import ClearIcon from "@material-ui/icons/Cancel";
+import Slide from "@material-ui/core/Slide";
 
 const Keycap = withStyles(theme => ({
   root: {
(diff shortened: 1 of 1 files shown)


dashboard/src/components/SearchBox/SearchBox.js
```diff
@@ -45,6 +48,7 @@ const SearchInput = withStyles(theme => ({
 
 const styles = theme => ({
   root: {
+    overflow: "hidden",
     minWidth: 300,
     display: "flex",
     border: "1px solid",
```
jamesdphillips: The clear button feels a bit uneven when compared to the filter icon on the left hand side of the container. Might be nice if the button was flush against the edge of the container.

...not a big deal though, not something we necessarily have to deal with right now.

PR #2180, opened 2018-10-16
Add gen-register, a tool for generating type registries

echlebek:
## What is this change?

This PR adds the gen-register tool, and an initial generated registry for the new sensu-go types.

A type registry is necessary for features like `sensuctl create`.


echlebek:
Oops, wrong branch.

PR #2181, opened 2018-10-16
Add gen-register, a tool for generating type registries.

echlebek:
## What is this change?

This PR adds the gen-register tool, and an initial generated registry for the new sensu-go types.

A type registry is necessary for features like `sensuctl create`.

### DIFF
--- a/internal/cmd/gen-register/main.go
+++ b/internal/cmd/gen-register/main.go
@@ -24,8 +24,17 @@ var (
 	outPath     = flag.String("o", "", "Output path")
 )
 
+const TypeMetaName = "TypeMeta"
+
 func main() {
 	log.SetFlags(log.Lshortfile)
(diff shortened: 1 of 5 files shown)


internal/cmd/gen-register/main.go
```diff
@@ -0,0 +1,169 @@
...
+	return false
+}
+
+// walker walks a filesystem recursively, looking for Go packages to parse.
+type walker struct {
+	packages map[string]*ast.Package
+	fset     *token.FileSet
+}
+
+func (w *walker) walk(path string, fi os.FileInfo, err error) error {
```
golangcibot: argument err is overwritten before first use

echlebek:
@palourde it's going to be part of the freeze-api tool, so I'll add some docs for that tool instead if that's OK. :)

PR #2188, opened 2018-10-19
Add Groups for User

ccressent:
## What is this change?

This change set adds support for groups to users: `User` objects now have an extra `Groups` field holding the list of groups the given user is a member of.

Several commands are introduced to manipulate a user's groups, all part of the `sensuctl user` sub-command:
- `add-group`, to add a group to a user's list of groups
- `remove-group`, to remove a group from a user's list of groups
- `set-groups`, to set a user's list of groups to the provided list
- `remove-groups`, to clear a user's list of groups

User groups are part of the new RBAC model.

No; looking to merge into `devel/api`.

No.

No.

No.


grepory:
Can you provide some sample CLI output and examples?

backend/apid/actions/users.go
```diff
@@ -224,6 +224,48 @@ func (a UserController) Enable(ctx context.Context, name string) error {
 	return err
 }
 
+// AddGroup adds a given group to a user
```
echlebek: I hate to be pedantic, but aren't we adding the user to a group, conceptually speaking?

ccressent: I understand what you mean, and I have to say I kept feeling a bit weird about that concept while working on this. Internally, groups are not objects and do not exist per se. Therefore a user cannot be added to a group, technically. What we do instead is "label" a user with the list of groups he/she should be considered a member of and use that information to derive access rights in the authorization middleware using group bindings, which bind a set of `Rule` to a label we call the group name.

And with that in mind, here comes the conundrum: should I say "add a group to a user", which is technically what happens and corresponds clearly with the code below, or "add a user to a group", which is what happens conceptually but seem to be the opposite of what happens in the code below?

Maybe we could use the former phrasing for anything internal, to be truer to what the code does, and the later phrasing for anything user facing, such as the CLI, to be closer to the expected concepts of users and groups?

I'll merge that do `devel/api` for now for the sake of time; good question though!

Issue #2145, opened 2018-10-09
Spike on teasing apart authorization and apid. 


annaplotkin:
Tease apart authorization and apiD to use the new model and auth workflow.

Requires #2107

palourde:
## Todo

- [ ] Create a WithRequestInfo (name TBD) middleware (see https://github.com/kubernetes/apiserver/blob/e9e4beec4b157a40f6c742782d74faf45c2b280e/pkg/endpoints/request/requestinfo.go#L116)
	- [ ] We should store the something like this in the request context: https://github.com/grepory/storage/blob/3886dbfcfedb33caf8af447c78fec71c586a35d1/authorization/interface.go#L13 (minus the User, which should be done by the authentication middleware)
- [ ] Add a group field for users
	- [ ] Add support in sensuctl
- [ ] Store the user's group into the JWT. See https://github.com/sensu/sensu-go/blob/f5b8db4d5d75c9ed9960ff6d5ed5bd905a3f9812/types/token.go#L7 https://github.com/sensu/sensu-go/blob/f5b8db4d5d75c9ed9960ff6d5ed5bd905a3f9812/backend/apid/middlewares/authentication.go#L35
- [ ] Create RBAC types (see https://github.com/grepory/storage/tree/master/apis/rbac)
- [ ] Create new RBAC pkg (see https://github.com/grepory/storage/tree/master/authorization/rbac)
- [ ] Create new authorization middleware (https://github.com/kubernetes/apiserver/blob/e9e4beec4b157a40f6c742782d74faf45c2b280e/pkg/endpoints/filters/authorization.go)
- [ ] Replace environments & organizations by namespaces
	- [ ] update all existing structs
    - [ ] update store (???)
	- [ ] update sensuctl
	- [ ] all the rest
- [ ] Create/rename store methods to GetX or ListX. (**WARNING** This is only required if the new structs/storage is not implemented soon)

PR #2209, opened 2018-10-23
Setup new RBAC model in authorization middleware

palourde:
## What is this change?

It adds the new RBAC model. More specifically...

* Add a new `rbac` package which contains all RBAC logic
* Add the user groups within the JWT claims
* Update the seeding to create all new roles/bindings
* Add a mockstore for the new store (v2).

⚠️ Few notes:
* We add a temporary route for `clusterroles` as a PoC, but it will be replaced soon.
* This PR breaks the API, until the new routes are put in place
* The RBAC & seeding code needs to be updated once the key building stuff is added

Required by https://github.com/sensu/sensu-go/issues/2145

Not yet

Nope

Nope

Nope


backend/apid/actions/generic.go
```diff
@@ -0,0 +1,23 @@
+package actions
+
+import (
+	storev2 "github.com/sensu/sensu-go/storage"
+	"golang.org/x/net/context"
```
echlebek: This should probably be stdlib context

backend/authorization/interface.go
```diff
@@ -0,0 +1,10 @@
+package authorization
+
+import "github.com/sensu/sensu-go/types"
+
+// Authorizer determines whether a request is authorized using the RequestInfo
+// passed. It returns true if the request should be authorized, along with any
+// error encountered
+type Authorizer interface {
+	Authorize(reqInfo *types.RequestInfo) (bool, error)
```
echlebek: :+1:

backend/apid/middlewares/authorization_test.go
```diff
@@ -2,92 +2,100 @@ package middlewares
...
+	cases := []struct {
+		name            string
+		claims          *types.Claims
+		reqInfo         *types.RequestInfo
+		authorizedError error
+		isAuthorized    bool
+		expected        int
+	}{
+		{
+			name:     "missing JWT",
```
ccressent: Maybe rename this to "missing claims and request info" (or something in that vein). When I first read "missing JWT" I assumed it was a test where the request itself is missing an Authorization header and it didn't make sense to me that it would return an internal error.

backend/authorization/interface.go
ccressent: I feel like `Authorize ()` being `RequestInfo -> Bool` is a bit strange. It sounds like "authorizing request information". Maybe it should be `Authorize(req http.Request) (bool, error)` or we might want to find a better name for `RequestInfo`...

backend/apid/actions/generic.go
palourde: Huh, it seems like we are using `golang.org/x/net/context` in other files within the same package so the editor just picked that. I'll remove all mentions of that pkg in order to switch to the built-in one!

backend/authorization/interface.go
palourde: I want to avoid passing `req http.Request` to Authorizers because they are not exclusive to the REST API; we will probably need to implement similar on the agentd side to ensure agents have the right to create events on certain namespaces for example.

I'd like to move the `types.RequestInfo` struct to maybe the `authorization` package and name it something like `Attributes`.

WDYT?

ccressent: `RequestInfo` might be too general for now anyway, as only the Authorization middleware uses it really. And it runs the risk of becoming `BigFatBagOfStuff` as maybe other middlewares piggyback on it in order to not have to define a new type.

I'm happy for you to move it in the `authorization` package. As for its new name... your guess is as good as mine!

`Operation` comes to mind as I feel it's natural to "authorize an operation". But at the same time, `Operation` is also rather general.

PR #2212, opened 2018-10-23
Add RESTKey and StorageKey types

### LINKED ISSUE #2165: Next-gen storage key building
We need to be able to build keys/paths to objects (both for the REST API and for Etcd). 

Create a uniform interface for determining the path to an object or type based on supplied data.

Given an API request of the following form

`/apis/{group}/{version}/namespaces/{namespace}/{type}/{name}`

When we want to interact with an object or type of object in Sensu, we have to be told:
- API Group
- Version
- Namespace (optional)
- Type
- Name

With that we can generate a REST path or an Etcd path by providing the following additional information:

- Whether or not the type is namespaced
- The prefix
  - /apis for REST
  - /sensu.io for Etcd
- Is the object versioned?
  - Yes for REST
  - No for Etcd

echlebek:
## What is this change?

This PR adds the RESTKey and StorageKey types, suitable for discovering the correct path of a given datatype.

This change is necessary for correctly addressing storage.

Closes #2165

Note: I had this PR up before, sans tests. I've now rebased it from devel/new-api, and have hence made a new PR for it. I have also added some unit tests, and fixed some bugs.


storage/key_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+type noNamespaceKeyableFixture struct {
+	keyableFixture
+}
+
+func (noNamespaceKeyableFixture) NoNamespace() {}
+
+func TestRESTKey(t *testing.T) {
+	fixture := keyableFixture{
+		kind:      "frogs",
```
palourde: While working on the generic API, I went down the rabbit hole and realized we have a problem here. Our resources kind are singular, not plural. It means that the REST/store keys in this current form are `[...]/check/foo` instead of `[...]/checks/foo`.

I tried to figure out how K8s does it and I'm still confused.

They have a [`Add` method](https://github.com/kubernetes/kubernetes/blob/7f23a743e8c23ac6489340bbb34fa6f1d392db9d/staging/src/k8s.io/apimachinery/pkg/api/meta/restmapper.go#L96) to register types, which uses a function said to be [broken and unsafe](https://github.com/kubernetes/kubernetes/blob/7f23a743e8c23ac6489340bbb34fa6f1d392db9d/staging/src/k8s.io/apimachinery/pkg/api/meta/restmapper.go#L123) to determine the plural of a kind. There's also a [ResourceSingularizer](https://github.com/kubernetes/kubernetes/blob/7f23a743e8c23ac6489340bbb34fa6f1d392db9d/staging/src/k8s.io/apimachinery/pkg/api/meta/restmapper.go#L149:29) function that doesn't seem to be used anywhere execepted in a test file...

Halp

echlebek: Woof! I'll resolve this in a follow-up PR.

PR #2214, opened 2018-10-24
Add support for +resource-name.

echlebek:
All package-level types that are frozen will now be given a
ResourceName method. By default, the method will return an empty
string.

By annotating a type with the +resource-name option, users can set
what the output of the ResourceName method will be.

This commit adds such annotations to the rbac package. This was
used for testing the functionality, as to give the rbac package
types usable resource names.


echlebek:
@grepory I'm not opposed to the idea, but can you explain why that would be desirable?

@ccressent no, we haven't generated the versioned API yet, and internal types won't have ResourceName().

grepory:
@echlebek so that it’s clear why the annotation is there. So that when something breaks or we need to make a change we know what generator to go digging around in.

internal/cmd/freeze-api/convert.go
```diff
@@ -74,40 +72,23 @@ var convert_{{ $toPackage }}_{{ $t.TypeName}}_To_{{ $t.TypeName}} = func(from *{
 {{ end }}
 `
 
-var converterTmpl = template.Must(template.New("converter").Parse(converterTmplStr))
+const converterTestTmplStr = `package {{ .FromPackage }}
```
golangcibot: `converterTestTmplStr` is unused

internal/cmd/freeze-api/resource.go
```diff
@@ -0,0 +1,103 @@
...
+		if t.Name == kindName {
+			docType = t
+		}
+	}
+	if docType == nil {
+		return ""
+	}
+	scanner := bufio.NewScanner(strings.NewReader(docType.Doc))
+	scanner.Split(bufio.ScanWords)
+	var found bool
```
echlebek: The found variable is unnecessary. The desired control flow can be accomplished with another call to scanner.Scan() within the loop body.

apis/meta/v1alpha1/converters.go
```diff
@@ -0,0 +1,51 @@
...
+func (r *ObjectMeta) ConvertTo(to interface{}) {
+	ptr := to.(*meta.ObjectMeta)
+	convert_ObjectMeta_To_meta_ObjectMeta(r, ptr)
+}
+
+var convert_ObjectMeta_To_meta_ObjectMeta = func(from *ObjectMeta, to *meta.ObjectMeta) {
+	*to = *(*meta.ObjectMeta)(unsafe.Pointer(from))
+}
+
+// ConvertFrom converts the receiver to a *meta.ObjectMeta.
```
ccressent: Should this read "converts the receiver to a *ObjectMeta"?

apis/meta/v1alpha1/converters.go
```diff
@@ -0,0 +1,51 @@
...
+func (r *TypeMeta) ConvertTo(to interface{}) {
+	ptr := to.(*meta.TypeMeta)
+	convert_TypeMeta_To_meta_TypeMeta(r, ptr)
+}
+
+var convert_TypeMeta_To_meta_TypeMeta = func(from *TypeMeta, to *meta.TypeMeta) {
+	*to = *(*meta.TypeMeta)(unsafe.Pointer(from))
+}
+
+// ConvertFrom converts the receiver to a *meta.TypeMeta.
```
ccressent: Should this read "converts the receiver to a *TypeMeta"?

apis/meta/v1alpha1/converters_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"reflect"
+	"testing"
+
+	fuzz "github.com/google/gofuzz"
+	"github.com/sensu/sensu-go/internal/apis/meta"
+)
+
+func Test_convert_ObjectMeta_To_meta_ObjectMeta(t *testing.T) {
```
ccressent: This actually also tests `convert_meta_ObjectMeta_to_ObjectMeta()`, doesn't it?
I would suggest a name that reflects that fact, because I had to read the whole test to figure that out  and not add a comment about missing tests for `meta.ObjectMeta` -> `ObjectMeta` :(

apis/meta/v1alpha1/converters_test.go
```diff
@@ -0,0 +1,32 @@
...
+	var v3 meta.ObjectMeta
+	fuzzer := fuzz.New().NilChance(0)
+	fuzzer.Fuzz(&v1)
+	v1.ConvertTo(&v3)
+	v2.ConvertFrom(&v3)
+	if !reflect.DeepEqual(v1, v2) {
+		t.Fatal("values not equal")
+	}
+}
+func Test_convert_TypeMeta_To_meta_TypeMeta(t *testing.T) {
```
ccressent: ditto, but for `TypeMeta`

internal/cmd/freeze-api/convert.go
```diff
@@ -58,7 +56,7 @@ func (r *{{ $t.TypeName }}) ConvertTo(to interface{}) {
 }
 
 var convert_{{ $t.TypeName }}_To_{{ $toPackage }}_{{ $t.TypeName}} = func(from *{{ $t.TypeName }}, to *{{ $toPackage}}.{{ $t.TypeName }}) {
-	{{ if $t.Simple }} *to = *(*{{ $toPackage }}.{{ $t.TypeName }})(unsafe.Pointer(from)){{ else }}panic("complex conversion not supported yet"){{ end }}
+	{{ if $t.Simple }}*to = *(*{{ $toPackage }}.{{ $t.TypeName }})(unsafe.Pointer(from)){{ else }}panic("complex conversion not supported yet"){{ end }}
 }
 
 // ConvertFrom converts the receiver to a *{{ $toPackage }}.{{ $t.TypeName }}.
```
ccressent: If my other, related comments make sense, then this should be "converts the receiver to a *{{ $t.TypeName }}."

internal/cmd/freeze-api/convert.go
```diff
@@ -69,45 +67,45 @@ func (r *{{ $t.TypeName}}) ConvertFrom(from interface{}) {
...
-	for _, v := range packages {
-		return v, nil
-	}
-	return nil, errors.New("no packages found")
-}
+	fuzz "github.com/google/gofuzz"
+	"{{ .ImportPackage }}"
+)
+{{ $toPackage := .ToPackage }}{{ range $i, $t := .Types }}
+func Test_convert_{{ $t.TypeName }}_To_{{ $toPackage }}_{{ $t.TypeName}}(t *testing.T) {
```
ccressent: I suggest `Test_convert[...]_and_back()`.

internal/cmd/freeze-api/convert.go
```diff
@@ -149,3 +160,7 @@ func typesEquivalent(a, b *ast.TypeSpec) bool {
 
 	return t1 == t2
 }
+
+func packagePath(path string) string {
+	return filepath.Join(build.Default.GOPATH, "src", path)
+}
```
ccressent: This is already defined in the `astutil` package.

hack/update-protobuf.sh
```diff
@@ -6,6 +6,7 @@ PACKAGES=(
     github.com/sensu/sensu-go/internal/apis/meta
     github.com/sensu/sensu-go/internal/apis/rbac
     github.com/sensu/sensu-go/internal/apis/core
+    github.com/sensu/sensu-go/apis/meta/v1alpha1
```
ccressent: It might be a stupid question, but should this touch frozen APIs packages?
I would expect _any_ change to those frozen packages to _only_ come from modifications of the internal API + run of the `freeze-api` tool.

apis/meta/v1alpha1/converters.go
echlebek: No, because the receiver is already a *ObjectMeta. I can try to rearrange the wording if it's confusing!

internal/cmd/freeze-api/convert.go
echlebek: True. I can export it there and remove this definition.

hack/update-protobuf.sh
echlebek: Hmm. That is probably how it _should_ work. But we have some technical debt to clear out of the way first. (i.e., the hack script)

apis/meta/v1alpha1/converters.go
ccressent: What's confusing me is that it says that it converts the receiver, a `*ObjectMeta`, to a `*meta.ObjectMeta`. That's confusing because:
- it sounds like it's what `ConvertTo()` is doing
- it literally calls a function called `convert_meta_ObjectMeta_To_ObjectMeta`

Am I missing something here?

internal/registry/registry.go
```diff
@@ -97,6 +110,15 @@ func scanPackages(packages map[string]*ast.Package) templateData {
 	return td
 }
 
+func lowerKind(kind string) (result string) {
```
palourde: Not sure if we need the first case to be lower (e.g. `roleBinding `) but we'll definitely need it all lowercase for the resource names, and probably the registry too, because the URI will look like `/apis/core/v1alpha1/rolebindings`

internal/cmd/freeze-api/convert.go
```diff
@@ -125,6 +121,7 @@ func createConverters(from, to string) error {
 		fromType := fromTypes[typeName]
 		toType, ok := toTypes[typeName]
 		if !ok {
+			fmt.Println(typeName)
```
palourde: Can it be removed? Otherwise we should probably add more information about this error.

internal/registry/registry.go
grepory: They should be all lowercase, imo.

grepory: When @echlebek asked about snake case vs. camel case, I did not realize that we were talking about URLs. Sorry about the confusion.

internal/cmd/freeze-api/convert.go
```diff
@@ -42,105 +41,173 @@ type conversionType struct {
...
+		return fmt.Sprintf("%s.%s", pkg, typename)
+	},
 }
 
-func createConverters(from, to string) error {
-	fromPackage, err := getPackage(from)
+var (
+	converterTmpl           = template.Must(template.New("converter").Funcs(templateFuncs).Parse(converterTmplStr))
+	converterTestsTmpl      = template.Must(template.New("converter_test").Funcs(templateFuncs).Parse(converterTestTmplStr))
+	conversionInitFuncsTmpl = template.Must(template.New("pkginits").Funcs(templateFuncs).Parse(conversionInitFuncs))
```
golangcibot: `conversionInitFuncsTmpl` is unused

internal/cmd/freeze-api/convert.go
```diff
@@ -42,105 +41,173 @@ type conversionType struct {
 	ComplexFields map[string]reflect.Kind
 }
 
-const converterTmplStr = `package {{ .FromPackage }}
+const conversionInitFuncs = `package conversion
```
golangcibot: const `conversionInitFuncs` is unused

internal/cmd/freeze-api/convert.go
echlebek: Yes, it should be removed.

internal/registry/registry.go
echlebek: I'll change it to be lower-cased.

PR #2217, opened 2018-10-25
Testing/tcp handler

nikkiki:
## What is this change?

Allows the tcp handler to write a single json event to an output file.

Required by https://github.com/sensu/sensu-staging/issues/40 to verify the tcp handler works as expected.

Nah, it's just there for testing purposes.

Nah.

AWS was kind of being a hater. I had to use `aws s3` command line to upload the new binary.

Yas.

### DIFF
--- a/testing/tcphandler/main.go
+++ b/testing/tcphandler/main.go
@@ -1,15 +1,21 @@
 package main
 
 import (
+	"bytes"
+	"encoding/json"
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"log"
 	"net"
+
+	"github.com/sensu/sensu-go/types"
 )
 
 var (
 	port = flag.Int("port", 3333, "port to run tcp server on")
 	conn = flag.String("type", "tcp", "type of connection to establish (ex. tcp)")
+	file = flag.String("file", "/tmp/tcp-handler.json", "optionally write json to file (ex. /tmp/tcp-handler.json")
 )
 
 func main() {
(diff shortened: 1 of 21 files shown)


nikkiki:
Not merging until after the release!

testing/tcphandler/main.go
```diff
@@ -37,5 +43,23 @@ func handleRequest(conn net.Conn) {
...
+		if err != nil {
+			log.Print("Recieved non-event: ", err.Error())
+			return
+		}
+		eventJSON, err := json.Marshal(event)
+		if err != nil {
+			log.Print("Error marshalling event: ", err.Error())
+			return
+		}
+		err = ioutil.WriteFile(*file, eventJSON, 0644)
```
ccressent: The problem with `ioutil.WriteFile()` is that it truncates the given file before writing. Would appending to the file be better?

My concern is that, if you receive 2 or more events in a short time period only the last one will be in the file, potentially creating confusion where one might think the "wrong" event was received while it was in fact "swallowed".

testing/tcphandler/main.go
```diff
@@ -1,15 +1,21 @@
...
 	"log"
 	"net"
+
+	"github.com/sensu/sensu-go/types"
 )
 
 var (
 	port = flag.Int("port", 3333, "port to run tcp server on")
 	conn = flag.String("type", "tcp", "type of connection to establish (ex. tcp)")
+	file = flag.String("file", "/tmp/tcp-handler.json", "optionally write json to file (ex. /tmp/tcp-handler.json")
```
ccressent: I'd call that flag `-out` or `-o` to clearly indicate that the given argument is purely to output things into (and be in line with most other go tools I think?)

nikkiki:
Actually.. I'm going to remove the tcp handler from sensu-go and into sensu-staging.

testing/tcphandler/main.go
ccressent: I see, but doesn't this introduce a form of race condition?
Here is what I imagine can happen:
- the qa test makes it so that the tcp handler receives an expected event, say `event1`
- while the qa test goes to check what was written to the file, another event, say `event2` arrives and overwrite the content of the output file
- the qa test is considered a failure because `event1 != event2`

Or can this situation never occur because of some other mechanism I'm not aware of? Sorry, I don't have much context on where/how this is going to be used!

nikkiki: That's a valid point. For the test I'm concerned with, and the only test invoking this specific handler, the event should be almost identical. The test is only checking for certain key values such as event.check.status and event.check.output.

Let's discuss it when I push the tcp handler to sensu-staging.. it will have much more context regarding the tests.

PR #2218, opened 2018-10-25
Remove tcp handler from sensu-go

nikkiki:

Removes the tcp handler from sensu-go. It will be migrated to sensu-staging in https://github.com/sensu/sensu-staging/issues/40.


portertech:
Remove TCP handler testing bits (the event receiver).

PR #2224, opened 2018-10-26
Use versioned metadata for rbac.

echlebek:

## What is this change?

Use versioned TypeMeta and ObjectMeta in the rbac API.


echlebek:
So I did some crazy hacking here.

* Meta is no longer internal at all.
* There is a special v1 meta package, in the published apis package, that is not generated.
* Fixed the issue where type names were not registered like rbac/v1, but instead would be v1.
* Added a versioned rbac package.

PR #2237, opened 2018-10-29
Add resource name kinds to the registry.

echlebek:

### DIFF
--- a/internal/registry/registry.go
+++ b/internal/registry/registry.go
@@ -22,29 +22,42 @@ const templateText = `package registry
 // automatically generated file, do not edit!
 
 import (
-  "fmt"
-  "reflect"
-
-  metav1 "github.com/sensu/sensu-go/apis/meta/v1"
-  {{ range $i, $import := imports . }}{{ $import }}
-  {{ end }}
+	"fmt"
+	"reflect"
+	
+	metav1 "github.com/sensu/sensu-go/apis/meta/v1"
+	{{ range $i, $import := imports . }}{{ $import }}
+	{{ end }}
 )
 
 type registry map[metav1.TypeMeta]interface{}
 
 var typeRegistry = registry{ {{ range $index, $t := . }}
-  metav1.TypeMeta{APIVersion: "{{ $t.APIVersion }}", Kind: "{{ $t.Kind }}"}: {{ replace $t.APIVersion "/" "" 1 }}.{{ $t.Kind }}{},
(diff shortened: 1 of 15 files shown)


internal/registry/registry.go
```diff
@@ -22,29 +22,42 @@ const templateText = `package registry
...
+	metav1.TypeMeta{APIVersion: "{{ $t.APIVersion }}", Kind: "{{ lower $t.Kind }}"}: {{ replace $t.APIVersion "/" "" 1 }}.{{ $t.Kind }}{}, {{ end }}
+}
+
+func init() {
+	for k, v := range typeRegistry {
+		r, ok := v.(interface{ResourceName() string})
+		if ok {
+			newKey := metav1.TypeMeta{
+				APIVersion: k.APIVersion,
+				Kind: r.ResourceName(),
```
ccressent: Should we also change the resource name tags as currently defined?

Right now for, say, `ObjectMeta`, the resource name is `objectMetas`. 

Don't we want to end up with `objectmetas` and `ObjectMetas` in the registry instead (or at least the all lowercase version)?

echlebek: No, because:

```
apiGroup: core/v1
kind: check
```

is desirable in yaml configuration.

ccressent: I'm sorry, I don't understand.

Based on the content of [resource.go for RBAC](https://github.com/sensu/sensu-go/blob/devel/new-api/apis/rbac/v1alpha1/resource.go), what I'm saying is that this loop effectively adds entries of this form, eg for `ClusterRole`:
```
metav1.TypeMeta{APIVersion: "rbac", Kind: "clusterRoles"}: rbac.ClusterRole{},
```
while we want an entry of this form:
```
metav1.TypeMeta{APIVersion: "rbac", Kind: "clusterroles"}: rbac.ClusterRole{},
```
That is, all lowercase and pluralized. It's fine for `Role` and `Subject`, but not the other types.

echlebek: I see. I've amended the PR to change the resource name declaration in the RBAC internal API.

palourde: While we are here, would it be hard to have `clusterroles` return `[]rbac.ClusterRole{}`?

echlebek: Interesting. I'll look into it. :)

palourde:
Let's merge that for now :)

PR #2271, opened 2018-10-31
Sensuctl/yaml support

### LINKED ISSUE #2251: Add YAML serialization/deserialization to sensuctl
We wish to be able to send YAML documents to the `sensuctl create` command. However, our API speaks only JSON.

The kubernetes project has solved this problem by deserializing YAML documents into a generic map, serializing the map to JSON, and then sending the resulting JSON to the API. Amazingly, this seems to actually work, with some minor caveats.

Adopt a similar approach for sensuctl.

echlebek:
## What is this change?

```
Author: Eric Chlebek <eric@sensu.io>
Date:   Wed Oct 31 14:51:51 2018 -0700

    Add yaml support to sensuctl.
    
    Users can now specify one or more yaml documents when creating
    resources with sensuctl create.
    
    Users can view the output of sensuctl commands in yaml format.
    
    The output yaml format is compatible with sensuctl create.
    
```

Closes #2251 

Yes

### DIFF
--- a/cli/commands/create/create.go
+++ b/cli/commands/create/create.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"regexp"
 
 	"github.com/ghodss/yaml"
 	"github.com/sensu/sensu-go/cli"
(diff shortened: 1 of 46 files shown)


cli/commands/create/create_test.go
```diff
@@ -76,3 +111,35 @@ func TestCreateCommand(t *testing.T) {
...
+	require.NoError(t, err)
+	defer os.RemoveAll(td)
+
+	fp := filepath.Join(td, "input")
+
+	f, err := os.Create(fp)
+	require.NoError(t, err)
+
+	buf := new(bytes.Buffer)
+	yamlSpecTmpl.Execute(buf, yamlResources)
```
golangcibot: Error return value of `yamlSpecTmpl.Execute` is not checked

cli/commands/create/create.go
```diff
@@ -67,26 +71,57 @@ func execute(cli *cli.SensuCli) func(*cobra.Command, []string) error {
 	}
 }
 
+var jsonRe = regexp.MustCompile(`^(\s)*[\{\[]`)
+
+// parseResources is a rather heroic function that will parse any number of valid
+// JSON or YAML resources. Since it attempts to be intelligent, it likely
+// contains bugs.
```
ccressent: That's the kind of comment I like to read before reviewing a function! :)

cli/commands/create/create.go
```diff
@@ -67,26 +71,57 @@ func execute(cli *cli.SensuCli) func(*cobra.Command, []string) error {
...
+	}
+	// Support concatenated yaml documents separated by '---'
+	array := bytes.Split(b, []byte("\n---\n"))
+	for _, b := range array {
+		var jsonBytes []byte
+		if jsonRe.Match(b) {
+			// We are dealing with JSON data
+			jsonBytes = b
+		} else {
+			// We are dealing with YAML data
```
ccressent: You split on `---` and _then_ you try to detect if each document is JSON or YAML.
This is not what the comment above `parseResources()` is describing.

I have no preference for either approach (split then detect JSON, or detect JSON then split), except for the fact that the current implementation makes it possible to have a mixed JSON/YAML input, which sounds... demonic.

cli/commands/helpers/yaml.go
```diff
@@ -0,0 +1,56 @@
...
+	_ = json.Unmarshal(b, &result)
+	return result
+}
+
+// PrintYAML serializes the value v to yaml and writes the result to w.
+func PrintYAML(v interface{}, w io.Writer) (err error) {
+	enc := yaml.NewEncoder(w)
+	defer func() {
+		if err == nil {
+			err = enc.Close()
```
ccressent: If `err != nil` you don't call `enc.Close()`.
Is that OK?

cli/commands/create/create.go
echlebek: Haha. Well, the presence of `\n---\n` in a document indicates that it is not valid JSON. I've only done it that way for programmer convenience.

YAML is a superset of JSON -- so documents that contain a bunch of JSON blobs separated by --- are actually yaml. :)

It seems to me that if the split results in more than a single element, we should assume that the parser is dealing with yaml. I can add some logic around that.

cli/commands/helpers/yaml.go
echlebek: Sure, I don't care what happens to the output stream if serialization has failed.

cli/commands/create/create.go
ccressent: Thanks for the clarification!

echlebek:
Looks like numeric values are getting rendered as floating point numbers. No good. I need to figure out how to fix that.

echlebek:
I've solved the number formatting issue by adding support for json.Number to go-yaml. I've sent them a PR, hopefully they accept it.

For now I've vendored a fork of go-yaml that we can use. We should remove the fork if and when the patch gets merged into go-yaml.

echlebek:
@nikkiki looks like empty slices need some special case handling.

PR #2273, opened 2018-10-31
Remove unused asset metadata field

### LINKED ISSUE #2270: Remove asset metadata field.
The asset's metadata field conflicts with the object metadata field name.

grepory:
## What is this change?

This removes the asset Metadata field--a map[string]string that is currently unused by the agent and backend.

It's a non-functional field and the field name conflicts with the new object metadata field name.

https://github.com/sensu/sensu-docs/pull/863

closes #2270

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Breaking Changes
 - Environments and organizations have been replaced with namespaces.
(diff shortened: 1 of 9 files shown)


grepory:
closes #2270

grepory:
@echlebek Do you think the changelog field should be amended to explain that behavior?

echlebek:
I think it's fine as-is. Users don't need to take any course of action to keep everything working, so there is no breaking change.

Issue #2249, opened 2018-10-30
Update API to use new route layouts anchored at /apis/core/v2


grepory:
We're going to be transitioning Sensu to a versioned API, or, more precisely, a group of versioned APIs.

Initially, there will only be a single API group: `core`. This API will function in the same way that the legacy sensu-go API functions.

The API group will be rooted at `/apis/{apiName}/{apiVersion}/`.

Because the product-family wide version of Sensu is 5, the first `apiVersion` will begin at 5.

Do not remove the existing API routes until all consumers have been updated to use the new routes.

grepory:
Closing #2252

PR #2275, opened 2018-11-01
Add authorization attributes middleware

### LINKED ISSUE #2256: Add authorization attributes middleware
We need to implement a new middleware that prepares the _authorization attributes_ later used by the RBAC authorizer.

- [ ] Implement the Attributes struct (https://github.com/sensu/sensu-go/blob/devel/new-api/backend/authorization/interface.go)
- [ ] Implement the middleware (https://github.com/sensu/sensu-go/blob/devel/new-api/backend/apid/middlewares/authorization_attributes.go)

Child of https://github.com/sensu/sensu-go/issues/2243

ccressent:
Closes #2256.


backend/apid/middlewares/authorization_attributes.go
```diff
@@ -0,0 +1,59 @@
...
+type AuthorizationAttributes struct{}
+
+// Then is the AuthorizationAttrs middleware's main logic, heavily inspired by
+// Kubernetes' WithRequestInfo.
+//
+// The general, expected format of a path one of the following: -
+// /apis/{group}/{version}/namespaces, to create or list namespaces -
+// /apis/{group}/{version}/namespaces/{name} -
+// /apis/{group}/{version}/namespaces/{namespace}/{kind} -
+// /apis/{group}/{version}/namespaces/{namespace}/{kind}/{name}
```
palourde: The problem here is that we need to support the _old_ routes for now, until https://github.com/sensu/sensu-go/issues/2249 is implemented (probably next sprint).

Could you try to see if we could do that without too much effort?

ccressent: Sure, I thought actually plugging the routes and middleware in would be part of another PR. What are the "old" routes supposed to look like? I'm assuming something like:
- `/rbac/rules`
- `/rbac/roles`
- ...

You've added a router for `roles` already; I can do the same with the other resources (`rules`, `clusterroles`). Am I missing anything? I'll also register those new routers in `registerRestrictedResources()` and plug in the `AuthorizationAttributes` middleware.

palourde: Just realized the challenge here is that we will have RBAC routes use the new pattern but the rest of the objects will continue to use the older patterns. Maybe we could split that logic into 2 different middlewares and once we are fully migrated, we just remove the _old_ one?

Let's sync about the routers!

backend/apid/middlewares/authorization_attributes_test.go
```diff
@@ -0,0 +1,544 @@
...
+		{
+			description: "GET /silenced/checks",
+			method:      "GET",
+			path:        "/silenced/checks",
+			urlVars:     map[string]string{},
+			expected: authorization.Attributes{
+				APIGroup:     "core",
+				APIVersion:   "v2",
+				Namespace:    "default",
+				Resource:     "silenced/checks",
```
palourde: We should have `silenced` as the resource here

backend/apid/middlewares/authorization_attributes_test.go
```diff
@@ -0,0 +1,544 @@
...
+			method:      "GET",
+			path:        "/silenced/checks/foo",
+			urlVars: map[string]string{
+				"check": "foo",
+			},
+			expected: authorization.Attributes{
+				APIGroup:     "core",
+				APIVersion:   "v2",
+				Namespace:    "default",
+				Resource:     "silenced/checks",
```
palourde: Idem

backend/apid/middlewares/authorization_attributes_test.go
```diff
@@ -0,0 +1,544 @@
...
+		{
+			description: "GET /silenced/subscriptions",
+			method:      "GET",
+			path:        "/silenced/subscriptions",
+			urlVars:     map[string]string{},
+			expected: authorization.Attributes{
+				APIGroup:     "core",
+				APIVersion:   "v2",
+				Namespace:    "default",
+				Resource:     "silenced/subscriptions",
```
palourde: Idem

backend/apid/middlewares/authorization_attributes_test.go
```diff
@@ -0,0 +1,544 @@
...
+			method:      "GET",
+			path:        "/silenced/subscriptions/foo",
+			urlVars: map[string]string{
+				"subscription": "foo",
+			},
+			expected: authorization.Attributes{
+				APIGroup:     "core",
+				APIVersion:   "v2",
+				Namespace:    "default",
+				Resource:     "silenced/subscriptions",
```
palourde: Idem

PR #2278, opened 2018-11-01
API authentication credential test endpoint

### LINKED ISSUE #1437: sensuctl does not provide a way to get the value of a password for a user
Configuration management tools need both get and set methods to ensure idempotency. We do a get to see if the system is in the desired state and if not, then do a set. If there is no get, then the set always runs, which is not idempotent.

`sensuctl user list` does not have a way to get the password, so there is no idempotent way to set the password.

## Expected Behavior
`sensuctl user list` has a way of showing the actual password or a hashed password. This way it can be matched against desired state and the configuration management tool can decide if the password needs to be updated.

## Current Behavior
`sensuctl user list` does not allow you to see the current state with regards to a user's password.

## Possible Solution
Output a sha256 hash of the password.
Only allow the `admin` role to do so.

## Context
This is needed for configuration management tools to act in an idempotent way.

## Your Environment

`sensu-ctl version dev-nightly#cc2ded6, build cc2ded623f0646610c0d3135430544dc2993169b, built 2018-05-03T09:26:06+0000`

* Operating System and version (e.g. Ubuntu 14.04): EL7

portertech:
Provide a Backend API endpoint for testing User credentials (username & password). This pull-request allows automation tooling to check if user credentials it has on record are still valid and manage the user accordingly (e.g. reset or rotate the user password).

A possible solution for https://github.com/sensu/sensu-go/issues/1437

Signed-off-by: Sean Porter <portertech@gmail.com>

closes #1437

### DIFF
--- a/cli/commands/user/test-creds_test.go
+++ b/cli/commands/user/test-creds_test.go
@@ -1,0 +1,23 @@
+package user
+
+import (
+	"testing"
+
+	//	clientmock "github.com/sensu/sensu-go/cli/client/testing"
+	test "github.com/sensu/sensu-go/cli/commands/testing"
+	"github.com/stretchr/testify/assert"
+	//	"github.com/stretchr/testify/mock"
+	//	"github.com/stretchr/testify/require"
+)
+
+func TestTestCredsCommand(t *testing.T) {
+	assert := assert.New(t)
+
+	cli := test.NewMockCLI()
+	cmd := TestCredsCommand(cli)
+
+	assert.NotNil(cmd, "cmd should be returned")
+	assert.NotNil(cmd.RunE, "cmd should be able to be executed")
+	assert.Regexp("test-creds", cmd.Use)
+	assert.Regexp("test user credentials", cmd.Short)
+}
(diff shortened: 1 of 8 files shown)


portertech:
screen shot 2018-11-01 at 9 38 41 pm

portertech:
Now to get a custom build to @ghoneycutt and @treydock

cli/commands/user/test-creds_test.go
```diff
@@ -0,0 +1,23 @@
+package user
+
+import (
+	"testing"
+
+	//	clientmock "github.com/sensu/sensu-go/cli/client/testing"
```
portertech: Need to add a few more tests.

echlebek:
I'm curious why we need this route. Wouldn't the login method suffice?

treydock:
If this PR uses the `/auth` API endpoint then one thing I discovered is that I'm unable to authenticate disabled users.  I don't know if this PR's subcommand will fail if the tested user is disabled.  If failure to authenticate disabled users is not expected behavior using `/auth` then I can open an issue.

portertech:
@treydock I believe it should fail. I will ask the team what they think.

PR #2283, opened 2018-11-02
Replace extended attributes with labels for Checks and Entities

### LINKED ISSUE #2244: Add labels/annotations to Checks, Handlers, Mutators, and Extensions
Labels and annotations are map[string]string that can be used to add metadata to objects, allowing programmatic interaction with Labels (e.g. labels can be accessed inside of filters) within Sensu, and interactions with annotations through the plugin/extension system.

echlebek:
## What is this change?

This PR removes extended attributes, and replaces them with labels. Labels are key/value string pairs that can be queried with govaluate, and set on check and entities.

Labels are part of the new ObjectMeta datatype, which will eventually be part of every Sensu datatype.

Extended attributes are difficult to use and reason about. They are probably the worst-understood feature of Sensu, both for users and developers. Labels are more restrictive, but much easier to use and reason about.

Whereas extended attributes were serialized JSON data that would be dynamically unmarshaled and queried on demand, labels are plain old data and can be easily included in a JSON or YAML Sensu Resource Definition.

closes #2244

Yes

What needs to be done before we release this change? What documentation will need updating?

Yes. This change was the result of an inordinate amount of discussion and argumentation. I hope that this will be the end of discussion around this feature, and that the feature will serve us well enough to replace the more complex extended attributes.

Documentation needs updating but I have not reviewed it at this time.


types/dynamic/encoding.go
```diff
@@ -28,105 +27,7 @@ func init() {
...
-	for _, any := range sortAnys(anys) {
-		s.WriteMore()
-		s.WriteObjectField(any.Name)
-		any.WriteTo(s)
-	}
-	return nil
-}
-
-func encodeStructFields(v AttrGetter, s *jsoniter.Stream) ([]string, error) {
+func encodeStructFields(v interface{}, s *jsoniter.Stream) ([]string, error) {
```
golangcibot: `encodeStructFields` is unused

PR #2293, opened 2018-11-05
Improve debug logging for event filtering in pipelind

### LINKED ISSUE #1745: Add logging to pipelined filters
Add logging to `pipelined/filters.go` so that we know when filters, silencing, etc. are accepting / rejecting an event.

## Expected Behavior
Log messages for filtering.

## Current Behavior
No log messages.

## Possible Solution
N/A

N/A

## Context
N/A

## Your Environment
N/A

grepory:
## What is this change?

Adds additional debug logging in pipelined.

Aids in debugging filters.

closes #1745 

Yes, added.

Please verify that the log messages match behavior. It can be really hard to reason about these things. I updated some of the comments to make things more clear (at least to me). If you can also make sure that these comments make sense to you and match behavior.

### DIFF
--- a/backend/pipelined/filter.go
+++ b/backend/pipelined/filter.go
@@ -34,8 +34,6 @@ func evaluateEventFilterStatement(event *types.Event, statement string) bool {
 func evaluateEventFilter(event *types.Event, filter *types.EventFilter) bool {
 	fields := utillogging.EventFields(event, false)
 	fields["filter"] = filter.Name
-	fields["check"] = event.Check.Name
-	fields["entity_name"] = event.Entity.ID
 
 	if filter.When != nil {
 		inWindows, err := filter.When.InWindows(time.Now().UTC())
@@ -97,8 +95,6 @@ func (p *Pipelined) filterEvent(handler *types.Handler, event *types.Event) bool
 	// Prepare the logging
 	fields := utillogging.EventFields(event, false)
 	fields["handler"] = handler.Name
-	fields["check"] = event.Check.Name
-	fields["entity_name"] = event.Entity.ID
 
 	// Iterate through all event filters, the event is filtered if
 	// a filter returns true.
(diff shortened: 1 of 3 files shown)


backend/pipelined/filter.go
```diff
@@ -30,22 +30,28 @@ func evaluateEventFilterStatement(event *types.Event, statement string) bool {
 	return result
 }
 
-// Returns true if the event should be filtered.
+// Returns true if the event should be filtered/denied.
 func evaluateEventFilter(event *types.Event, filter *types.EventFilter) bool {
+	fields := utillogging.EventFields(event, false)
+	fields["filter"] = filter.Name
+	fields["check"] = event.Check.Name
+	fields["entity_name"] = event.Entity.ID
```
echlebek: The entity ID and check are already provided by utillogging.EventFields. We may wish to amend the EventFields function to use the check name instead of the entire check.

backend/pipelined/filter.go
```diff
@@ -55,121 +61,123 @@ func evaluateEventFilter(event *types.Event, filter *types.EventFilter) bool {
...
 
 // filterEvent filters a Sensu event, determining if it will continue
-// through the Sensu pipeline.
+// through the Sensu pipeline. Returns true if the event should be filtered/denied.
 func (p *Pipelined) filterEvent(handler *types.Handler, event *types.Event) bool {
 	// Prepare the logging
 	fields := utillogging.EventFields(event, false)
 	fields["handler"] = handler.Name
+	fields["check"] = event.Check.Name
+	fields["entity_name"] = event.Entity.ID
```
echlebek: See above.

grepory:
@echlebek Addressed feedback. Let me know if that's reasonable.

util/logging/logging.go
```diff
@@ -14,15 +14,14 @@ func EventFields(event *types.Event, debug bool) map[string]interface{} {
 	}
 
 	fields := logrus.Fields{
-		"entity":    event.Entity.ID,
-		"namespace": event.Entity.Namespace,
+		"entity_id":        event.Entity.ID,
+		"entity_namespace": event.Entity.Namespace,
+		"check_name":       event.Check.Name,
```
grepory: whoops. this has to be gated by Hascheck. My bad.

PR #2295, opened 2018-11-06
[Web UI] Add clamping mechanism to RelativeDate component 

### LINKED ISSUE #2192: [Web UI] Event prompts - last executed value can be in the future
`Last executed in a few seconds` isn't an accurate representation of when a check was last executed. This value should not be in the future (because English), and it should actually represent the last executed time of the check.

## Expected Behavior
`Last executed just now`, `Last executed 2 minutes ago`, etc...

## Current Behavior
screen shot 2018-10-19 at 10 06 50 am

## Possible Solution
Pull the proper value for the check and calculate the time elapsed since now.

1. Create some checks running at an interval.
2. Open the dashboard - events page

## Context
Issue found in manual qa - staging environment.

## Your Environment
* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): https://github.com/sensu/sensu-go/commit/05af2ca99b4db81a1c20a9cb05a74c7f258db042
* Installation method (packages, binaries, docker etc.): packages deployed from terraform in aws
* Operating System and version (e.g. Ubuntu 14.04): backend: Ubuntu 16.04, dashboard: OSX 10.13.6 (Google Chrome)

jamesdphillips:
## What is this change?

Add option to clamp delta in RelativeDate component to mitigate issues where the dates appear in the future/past when the a user's system clock is (slightly) off.

Closes #2192 

yep

I've made the time window small enough that I don't feel _too_ gross about fudging the numbers; in the long run it would be nice to explore ways to ensure our dates are relatively consistent system wide.


jamesdphillips:
We decided to close this ticket. I do not feel comfortable fudging the numbers with regards to something as important as time. We're far too likely to wallpaper over subtle issues that may hint at bigger problems.

I will open a separate ticket for future investigation.

PR #2305, opened 2018-11-07
[Web UI] MenuToolbar collapsing bug

### LINKED ISSUE #2198: [Web UI] Check details silencing button bug
You can silence an event, however the button moves into the kebab and you can no longer interact with it. The includes unsilencing it as well. Refreshing the page moves the button back into view and it becomes clickable.

## Expected Behavior
Since my viewport has not changed, I would expect my buttons to not hide away after they were selected. In addition, even if they were added to the list, they should still be able to be interacted with. Perfect behaviour would be the button would not hide away, and it would keep the ability to be interacted with

## Current Behavior
After click button hides in kebab and no longer works.

## Possible Solution
This likely has to do with the implementation, we have options for these items to display them, "if room", "always", etc. I think the problem is since the list has been rendered already, we can't re-add them because it thinks the available room has been taken up. We might need to rerender the whole list? Just my initial quick thought.

1. Go to check details page
2. Try to silence
3. Notice button hides in kebab
4. Try to click unsilence

dabria:
## What is this change?
Closes #2198 
Updates the ToolbarMenu to recalculate the button width on visibility change (in addition to id change).

Also fixed some stuff with Toasts which was causing console errors.

The buttons were collapsing into the list when they shouldn't have been. 

Added one

@jamesdphillips expressed interest in rewriting some of this component. At first glance it seems like there's a fair bit going on over multiple components. It could probably be improved and simplified a bit.


dashboard/src/components/relocation/Toast.js
```diff
@@ -208,12 +208,13 @@ class Toast extends React.PureComponent {
...
-                    {closeButton}
-                  </CircularProgress>
-                ))}
+              {showAgeIndicator
+                ? progress => (
+                    <CircularProgress width={4} value={progress} opacity={0.5}>
+                      {closeButton}
+                    </CircularProgress>
+                  )
+                : null}
```
dabria: Not sure if this is the best way to handle it, but Timer wants a child to be a function, previously I was getting console errors about it being a boolean. This was just a really quick change.

jamesdphillips: 👍

PR #2306, opened 2018-11-07
Add support for namespaces into path variable

palourde:
Inject the namespace found in path variable into the context too


palourde:
@ccressent I'll merge this PR so I can get started in the integration but please review it and let me know if you find anything that needs more love!

PR #2313, opened 2018-11-09
Grepory/agent

grepory:
## What is this change?

This reworks the reconnection / connection management path of the agent. Agents now reconnect if their initial connection attempt fails, and the transport is simplified so that it doesn't include any connection reuse logic. This also collapses the number of goroutines required for connection handling in the agent.

The agent's connection handling logic was pretty difficult to follow and spread across a number of goroutines unnecessarily. This simplifies that workflow into two goroutines.

Added.

I don't really need clarification, but I think if this approach to connection handling (sans reconnect logic) is satisfactory with people, then we should give the backend's agent session the same treatment. We could also then dramatically simplify the transport library.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"strings"
 	"sync"
+	"sync/atomic"
 
 	time "github.com/echlebek/timeproxy"
 
@@ -45,7 +46,7 @@ type Agent struct {
 	backendSelector BackendSelector
 	cancel          context.CancelFunc
 	config          *Config
-	conn            transport.Transport
+	connected       int32
 	context         context.Context
 	entity          *types.Entity
 	executor        command.Executor
@@ -70,6 +71,7 @@ func NewAgent(config *Config) *Agent {
 		backendSelector: &RandomBackendSelector{Backends: config.BackendURLs},
 		cancel:          cancel,
 		context:         ctx,
+		connected:       0,
 		config:          config,
 		executor:        command.NewExecutor(),
 		handler:         handler.NewMessageHandler(),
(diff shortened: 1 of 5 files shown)


agent/agent.go
```diff
@@ -319,27 +225,75 @@ func (a *Agent) Run() error {
...
+			}
+		}(conn, done)
+
+	SEND:
+		for {
+			select {
+			case <-done:
+				break SEND
+			case <-a.stopping:
+				break
```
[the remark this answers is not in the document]
grepory: call `conn.Close()`

agent/agent.go
```diff
@@ -319,27 +225,75 @@ func (a *Agent) Run() error {
...
+					break
+				}
+			}
+		}
+	}
+}
+
+// Connected returns true if the agent is connected to a backend.
+func (a *Agent) Connected() bool {
+	return a.connected == 1
```
grepory: `atomic.LoadUint32(&a.connected) == 1`

grepory:
I got a sync review with @echlebek. I'm going to address his feedback, and then I'll push another commit shortly.

grepory:
To QA:

Start an agent (with no backend; debug logging is useful), observe that the agent goes into a reconnect loop.
curl http://localhost:3031/healthz, observe that it says backend unavailable.
Start backend, observe agent connects to backend.
curl http://localhost:3031/healthz, observe that it says ok.
Stop backend, observe agent disconnects and begins to reconnect.
curl http://localhost:3031/healthz, observe that it says backend unavailable.
Start backend, observe agent reconnects.
curl http://localhost:3031/healthz, observe that it says ok.



You should be able to see relevant log messages indicating errors, etc.

agent/agent.go
```diff
@@ -405,3 +364,32 @@ func (a *Agent) StartStatsd() {
...
+
+func connectWithBackoff(url string, tlsOpts *types.TLSOptions, header http.Header) transport.Transport {
+	var conn transport.Transport
+
+	backoff := retry.ExponentialBackoff{
+		InitialDelayInterval: 500 * time.Millisecond,
+		MaxDelayInterval:     10 * time.Second,
+		MaxRetryAttempts:     0, // Unlimited attempts
+		Multiplier:           1.5,
+	}
```
ccressent: Should these be exposed as part of the agent configuration?

grepory: I'd rather not let the scope of this PR creep any more. Can we file an issue instead of pulling this in?

PR #2332, opened 2018-11-13
Replace "kind" with "type" in RBAC

### LINKED ISSUE #2327: Replace "kind" with "type" in RBAC
In order to be consistent with the Wrapper, let's use the attribute `type` instead of `kind`, as discussed in today's sprint planning.

Child of #2243

ccressent:
### Why is this change necessary?

This is for the sake of consistency everywhere, and especially for files
fed to `sensuctl create -f`.

Closes #2327.

### Have you reviewed and updated the documentation for this change? Is new documentation required?

Upcoming RBAC documentation may be impacted @apaskulin.


backend/apid/middlewares/authorization_attributes.go
```diff
@@ -24,8 +24,8 @@ type AuthorizationAttributes struct{}
 // The general, expected format of a path is one of the following:
 // /apis/{group}/{version}/namespaces, to create or list namespaces
 // /apis/{group}/{version}/namespaces/{name}
-// /apis/{group}/{version}/namespaces/{namespace}/{kind}
-// /apis/{group}/{version}/namespaces/{namespace}/{kind}/{name}
+// /apis/{group}/{version}/namespaces/{namespace}/{type}
```
palourde: I think we should actually use `resource` here and make a distinction between a type and the resource; For example, the kind is `entity` but the resource here in the REST request should be `entities`.

ccressent: Good point, and to add to it: the related field in the `AuthorizationAttributes` is called `Resource`, so let's unify all of this!

I'll change `type` to `resource` in the routes.

Issue #2321, opened 2018-11-13
Do not fetch the default namespace upon login


palourde:
When authenticating via sensuctl, we fetch the default namespace. Not all users necessarily have access to it, which provokes a (false) error.

We ~do~ could add the `system:users` group upon login so the user can view itself, but we need to make sure it "survives" modification etc.

Child of #2243

PR #2333, opened 2018-11-13
Use the get user endpoint to validate credentials

palourde:
Closes https://github.com/sensu/sensu-go/issues/2321

Add the `system:users` group upon login and token renewal so all users can _view_ themselves, which is used to validate the users credentials.

### DIFF
--- a/backend/apid/routers/authentication.go
+++ b/backend/apid/routers/authentication.go
@@ -47,6 +47,9 @@ func (a *AuthenticationRouter) login(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Add the 'system:users' group to this user
+	user.Groups = append(user.Groups, "system:users")
+
 	// Create the token and a signed version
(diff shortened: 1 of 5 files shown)


backend/apid/routers/authentication.go
```diff
@@ -194,6 +197,9 @@ func (a *AuthenticationRouter) token(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Make sure to the 'system:users' group is present
+	user.Groups = append(user.Groups, "system:users")
+
```
ccressent: Wouldn't this ultimately create an ever growing number of `system:users` entries in the user's group list?
Same thing in the previous hunk in `login()`.

palourde: No because we never update the actual user in the store with this group; it's only injected in the user claims; it will never show up in the `sensuctl` commands and if you update a user group, it will be re-injected here once we renew the token.

Issue #2323, opened 2018-11-13
Improve the error message when creating an RBAC resource in a missing namespace


palourde:
If you create an RBAC resource (e.g. Role) on a namespace that doesn't exist, it returns an error telling that the resource already exist.

Child of #2243

PR #2335, opened 2018-11-13
Identify the cause of a transaction failure in generic store methods

palourde:
Closes https://github.com/sensu/sensu-go/issues/2323

Upon failure of an etcd transaction, try to fetch the namespace and the key itself to determine why it failed and return the appropriate response. Example:

```
$ sensuctl role-binding create testing --role testing --user foo --group foo
Error: resource already exists
$ sensuctl role-binding create testing --role testing --user foo --group foo --namespace acme
Error: the namespace acme does not exist
$ sensuctl role-binding create testing2 --role testing --user foo --group foo
Created
```

### DIFF
--- a/backend/store/store.go
+++ b/backend/store/store.go
@@ -49,6 +49,16 @@ func (e *ErrEncode) Error() string {
 	return fmt.Sprintf("could not encode the key %s: %s", e.Key, e.Err.Error())
 }
 
+// ErrNamespaceMissing is returned when the user tries to manipulate a resource
+// within a namespaces that does not exist
+type ErrNamespaceMissing struct {
+	Namespace string
+}
+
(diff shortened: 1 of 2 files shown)


backend/store/store.go
```diff
@@ -49,6 +49,16 @@ func (e *ErrEncode) Error() string {
 	return fmt.Sprintf("could not encode the key %s: %s", e.Key, e.Err.Error())
 }
 
+// ErrNamespaceMissing is returned when the user tries to manipulate a resource
+// within a namespaces that does not exist
```
ccressent: s/namespaces/namespace

palourde: Haha good catch!

PR #2336, opened 2018-11-13
Replace govaluate filtering with javascript filtering.

### LINKED ISSUE #2328: Replace govaluate filtering with javascript filtering
The govaluate library which we use to provide support for filter expressions is a bit limited. It seems to be unmaintained, and we have been carrying a fork of it for some time.

Another issue is that govaluate's syntax is rather unique - people using sensu filters will have to learn the quirks of this syntax.

Using a more widely known language for filtering should make using sensu filters easier for newcomers.

We would like to introduce Javascript, powered by the Otto javascript VM, as the language powering sensu filters. The Otto VM provides a much richer language that targets a standard (ECMA5). (Note - there are some Go-specific exceptions regarding regular expressions.)

Introducing Javascript also allows us to add another powerful concept - assets for filters. Users can add javascript libraries to filters (for instance, underscore.js) and take advantage of a richer set of programming abilities.

This will break some, but not all, existing filters. Users should take care that their filters are still working as intended after they upgrade.

echlebek:
## What is this change?

```
    This commit removes govaluate filtering and replaces it with
    javascript. Anywhere govaluate was used before, javascript is used
    now.
    
    This commit also introduces runtime assets for filters. Users can
    add javascript libraries to filter execution context by including
    them in asset tarballs.
    
    The interpreted asset scripts will be cached after the first use.
    
    This commit changes the semantics of dynamic.Synthesize. The
    Synthesize function is now fully-recursive, and will traverse
    slice and map datatypes. Values returned by Synthesize no longer
    contain any pointers to input data.
    
    Synthesize has been altered to return maps keyed by the json field
    name of the struct fields, instead of literal struct field names.
    This is done to maintain a consistent API between our HTTP server,
    sensuctl create, filters, and token substitution. Now, users will
    always refer to names by what is specified in our API, instead of
    what is specified by our struct field naming.
```

Closes #2328

Big time.

This was a large and complex change to implement. To get everything to play well together, I had to make some opinionated technical choices. These choices have trade-offs, but I believe they are the correct ones to make right now.

The documentation has not yet been updated. It will need a thorough review before this change is released.


cli/commands/edit/edit.go
```diff
@@ -0,0 +1,97 @@
...
+				resourceType,
+				"info",
+			}
+			ctlArgs = append(ctlArgs, resourceKeyParams...)
+			ctlArgs = append(ctlArgs, "--format")
+			ctlArgs = append(ctlArgs, format)
+			fmt.Println(ctlArgs)
+			b, err := exec.Command("sensuctl", ctlArgs...).CombinedOutput()
+			if err != nil {
+				cmd.OutOrStdout().Write(b)
```
golangcibot: Error return value of `(io.Writer).Write` is not checked

js/js.go
```diff
@@ -0,0 +1,107 @@
...
+	if err != nil {
+		return false, err
+	}
+	defer releaseOttoVM(jsvm, assets)
+	if params, ok := parameters.(map[string]interface{}); ok {
+		for name, value := range params {
+			if err := jsvm.Set(name, value); err != nil {
+				return false, err
+			}
+			defer jsvm.Set(name, otto.UndefinedValue())
```
golangcibot: Error return value of `jsvm.Set` is not checked

js/vm_cache.go
```diff
@@ -0,0 +1,119 @@
...
+	for k, v := range c.vms {
+		valueTime := time.Unix(v.lastRead, 0)
+		if valueTime.Before(time.Now().Add(-cacheMaxAge)) {
+			delete(c.vms, k)
+		}
+	}
+	defer c.Unlock()
+}
+
+func makeCacheKey(keys []string) string {
```
golangcibot: `makeCacheKey` is unused

cli/commands/edit/edit.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/sensu/sensu-go/cli"
+	"github.com/sensu/sensu-go/cli/commands/create"
+	"github.com/sensu/sensu-go/cli/commands/helpers"
+	"github.com/spf13/cobra"
+)
+
+// vi is the default editor!
+const defaultEditor = "vi"
+
+type unmarshalFunc func([]byte, interface{}) error
```
golangcibot: `unmarshalFunc` is unused

backend/apid/graphql/namespace.go
```diff
@@ -86,23 +87,21 @@ func (r *namespaceImpl) Checks(p schema.NamespaceChecksFieldResolverParams) (int
...
+		filteredChecks := make([]*types.CheckConfig, 0, len(records))
+		for _, record := range records {
+			r := dynamic.Synthesize(record)
+			matched, err := js.Evaluate(filter, r, nil)
+			if err != nil {
+				logger.WithError(err).Debug("unable to filter record")
+				continue
+			}
+			if matched {
+				filteredChecks = append(filteredChecks, record)
```
golangcibot: this result of append is never used, except maybe in other appends

cli/commands/edit/edit.go
```diff
@@ -0,0 +1,95 @@
...
+			}
+			ctlArgs := []string{
+				resourceType,
+				"info",
+			}
+			ctlArgs = append(ctlArgs, resourceKeyParams...)
+			ctlArgs = append(ctlArgs, "--format")
+			ctlArgs = append(ctlArgs, format)
+			fmt.Println(ctlArgs)
+			b, err := exec.Command("sensuctl", ctlArgs...).CombinedOutput()
```
palourde: I'd replace `"sensuctl"` with `os.Args[0]` so we don't assume the name of the executable nor its location (might not necessarily be in $PATH)

cli/commands/edit/edit.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/sensu/sensu-go/cli/commands/helpers"
+	"github.com/spf13/cobra"
+)
+
+// vi is the default editor!
+const defaultEditor = "vi"
+
+func Command(cli *cli.SensuCli) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "edit [RESOURCE NAME] [KEY1] ... [KEYN]",
```
palourde: `[RESOURCE NAME] [KEY1]...` is a bit confusing to me. I'd rather do `[RESOURCE] [NAME1] ... [NAMEN]` or `[RESOURCE TYPE] [NAME1] ... [NAMEN]`. WDYT?

cli/commands/edit/edit.go
```diff
@@ -0,0 +1,95 @@
...
+			if err != nil {
+				return err
+			}
+			if _, err := tf.Write(b); err != nil {
+				return err
+			}
+			if err := tf.Close(); err != nil {
+				return err
+			}
+			editorEnv := os.Getenv("EDITOR")
```
nikkiki: Maybe something for the future, but it would be nice to have a sensuctl command that would set the preferred editor.

cli/commands/edit/edit.go
```diff
@@ -0,0 +1,95 @@
...
+			if len(resources) == 0 {
+				return errors.New("no resources were parsed")
+			}
+			if err := create.ValidateResources(resources); err != nil {
+				return err
+			}
+			if err := create.PutResources(cli.Client, resources); err != nil {
+				return err
+			}
+			fmt.Fprintf(cmd.OutOrStdout(), "Updated %s\n", resources[0].URIPath())
```
nikkiki: Is there a way to detect if the temp file has not been changed? For example, a user attempts to edit, cancels instead, but still receives the confirmation message `Updated`.

cli/commands/edit/edit.go
nikkiki: I agree, I was a little confused at first too. I like `[TYPE] [NAME1] ... [NAMEN]` (and add `RESOURCE` if you find that helpful). It's more consistent with our internal naming conventions.

cli/commands/edit/edit.go
ccressent: The Unix way is to use `$VISUAL`, then `$EDITOR`, then a standard editor (usually vi or emacs) :nerd_face:

cli/commands/edit/edit.go
echlebek: Yeah, I think that's probably reasonable to detect.

cli/commands/edit/edit.go
echlebek: NAMEN :+1:

echlebek: Haha just kidding. I hate NAMEN. :laughing: 

The thing is that it's not a series of names - it's a series of key components. Not sure how best to communicate that.

ccressent: If it helps you make a decision, the typical way in Linux man pages to indicate multiple optional parameters of the same type would be `[KEY]...`. *BSD man pages would typically use `[key ...]`

Gopkg.toml
```diff
@@ -198,3 +193,7 @@ required = [
 [[constraint]]
   name = "github.com/sensu/yaml"
   revision = "704b36ea12f40d32ffe6bd109c2ddc894af28dde"
+
+[[constraint]]
+  branch = "master"
+  name = "github.com/robertkrimen/otto"
```
ccressent: No specific version or revision constraint for this new dependency?
I'm not a fan of "YOLO pulling master" :)

backend/pipelined/filter.go
```diff
@@ -54,8 +46,10 @@ func evaluateEventFilter(event *types.Event, filter *types.EventFilter) bool {
 		}
 	}
 
+	synth := dynamic.Synthesize(event)
+
 	for _, statement := range filter.Statements {
-		match := evaluateEventFilterStatement(event, statement)
+		match := evaluateJSFilter(synth, statement, assets)
```
ccressent: I thought you said you were replacing the "statement" terminology with the "expression" terminology?

backend/pipelined/filter.go
```diff
@@ -5,35 +5,27 @@ import (
...
+	"github.com/sensu/sensu-go/asset"
+	"github.com/sensu/sensu-go/js"
 	"github.com/sensu/sensu-go/types"
-	"github.com/sensu/sensu-go/util/eval"
+	"github.com/sensu/sensu-go/types/dynamic"
 	utillogging "github.com/sensu/sensu-go/util/logging"
 )
 
-func evaluateEventFilterStatement(event *types.Event, statement string) bool {
+func evaluateJSFilter(event interface{}, expr string, assets asset.RuntimeAssetSet) bool {
```
ccressent: Is it strictly necessary to change `event` from being a `*types.Event` to being a `interface{}`?

cli/commands/edit/edit.go
```diff
@@ -0,0 +1,109 @@
...
+			execCmd.Stdout = os.Stdout
+			execCmd.Stderr = os.Stderr
+			if err := execCmd.Run(); err != nil {
+				return err
+			}
+			changedBytes, err := ioutil.ReadFile(tf.Name())
+			if err != nil {
+				return err
+			}
+			if bytes.Equal(originalBytes, changedBytes) {
```
ccressent: I suppose that works and I doubt people will have huge objects, say with sizes on the order of dozens or hundreds of MB, that will make byte by byte comparison a problem.

Generally speaking though, I think your best bet is to use a hashing function over the content of the files (it doesn't have to be cryptographically strong here) and compare the resulting hashes.

js/js_bench_test.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/sensu/sensu-go/types/dynamic"
+)
+
+func BenchmarkCheckEval(b *testing.B) {
+	check := types.FixtureCheck("foo")
+	for i := 0; i < b.N; i++ {
+		synth := dynamic.Synthesize(check)
+		params := map[string]interface{}{
+			"check": synth,
+		}
```
ccressent: Shouldn't `synth` and `params` be defined outside the loop to not skew the benchmark (since they don't depend on `i` and are not part of the function you want to measure, here `Evaluate()`)?

js/js_test.go
```diff
@@ -0,0 +1,12 @@
+package js
+
+import "testing"
+
+func BenchmarkNewOttoVM(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		_, _ = newOttoVM(nil)
+		for k := range ottoCache.vms {
+			delete(ottoCache.vms, k)
+		}
```
ccressent: Should clearing that `ottoCache.vms` be part of what is measured?
You might want to start/stop the clock around `newOttoVM()` only.

types/filter.go
```diff
@@ -42,7 +42,7 @@ func (f *EventFilter) Validate() error {
 		return errors.New("filter must have one or more statements")
 	}
 
-	if err := eval.ValidateStatements(f.Statements, false); err != nil {
+	if err := js.ParseExpressions(f.Statements); err != nil {
```
ccressent: "Statements" here should probably be "Expressions" (just a quick note in case you miss that one)

cli/commands/edit/edit.go
```diff
@@ -0,0 +1,109 @@
...
+				return err
+			}
+			if err := tf.Close(); err != nil {
+				return err
+			}
+			editorEnv := os.Getenv("EDITOR")
+			if editorEnv == "" {
+				editorEnv = defaultEditor
+			}
+			execCmd := exec.Command(editorEnv, tf.Name())
```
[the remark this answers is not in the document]
echlebek: Oh ffs haha

cli/commands/edit/edit.go
echlebek: Individual resources will never even be a MB, we are good!

js/js_bench_test.go
echlebek: No, because a typical filter invocation will require a synthesis, and I wanted to measure worst case performance.

Gopkg.toml
echlebek: The project hasn't made any releases. I think in this case it's better to use master and commit it to vendor instead of pinning a specific SHA. If we notice some bugs we can try to upgrade or fork.

Issue #2262, opened 2018-10-31
Update sensu-docs with the new RBAC


palourde:
We need to update the RBAC reference document & the `How to create a Read Only User with RBAC` guide to reflect the new changes. Also go through the rest of the docs (getting started etc.) to make sure it's all still valid.

Child of https://github.com/sensu/sensu-go/issues/2243

grepory:
closed https://github.com/sensu/sensu-docs/pull/890

PR #2338, opened 2018-11-14
Complete redesign of RBAC

ccressent:
## What is this change?

This is a complete redesign of the Role Based Access Control (RBAC) feature with the aim to make it more flexible and powerful. It is inspired by the Kubernetes RBAC (see [this overview](https://kubernetes.io/docs/reference/access-authn-authz/rbac/)).

In no particular order:

- add new types to model RBAC resources: `Role`, `ClusterRole`, `RoleBinding` and `ClusterRoleBinding`
- completely remove the old concept of "role"
- users can now belong to groups, which are not actual internal objects but rather a way to "reference a set of users".
- the authorization is now all done within the `Authorization` middleware, which relies on the [`LegacyAuthorizationAttributes`](https://github.com/sensu/sensu-go/pull/2338/files#diff-20e3b45c8891151dda820f70cbebaa9aR115) (for "older" routes) or [`AuthorizationAttributes`](https://github.com/sensu/sensu-go/pull/2338/files#diff-20e3b45c8891151dda820f70cbebaa9aR19) middlewares to determine the nature of the request.
- the actual authorization logic now lives in `backend/authorization/rbac`
- we added Store/Router/Controller/CLI command for the new RBAC resources 
- the `sensu` default user no longer exists
- the default roles and bindings are documented in the seeds.go file or [here](https://github.com/sensu/sensu-docs/pull/890#discussion_r233059047)
- we now make a distinction in RBAC between listing all resources of a certain type and retrieving a specific one. Getting a bunch of resources (eg: `GET /checks`) corresponds to an RBAC `list` verb. Getting a particular resource (eg: `GET /checks/check-cpu`) corresponds to an RBAC `get` verb.
- the new RBAC resources use the new routes format (e.g. `/apis/rbac/v5/...`)
- we introduced "generic" store methods (see backend/store/etcd/store.go), which greatly simplifies the new RBAC stores (which only take care of basically validating the input)
- we also introduced new error types for the store (see backend/store/store.go), so the `apid/action`s controllers only has to deal with returning the right HTTP error, when required.
- update `logrus` to support the `trace` level
- update `gorilla/mux` to add the `SetURLVars` function in tests

Definitely.
A single line entry has already been added, but reviewers are very welcome to suggest additional entries for the changelog if they see fit!

New documentation (and updates to what we currently have) is definitely needed. This is being worked on by @apaskulin and @palourde through #2262 and [sensu-docs#890](https://github.com/sensu/sensu-docs/pull/890).


backend/apid/actions/clusterrolebindings_test.go
```diff
@@ -0,0 +1,406 @@
...
+		{
+			name:            "Invalid input",
+			ctx:             context.Background(),
+			argument:        types.FixtureClusterRoleBinding("read-write"),
+			storeErr:        &store.ErrNotValid{},
+			expectedErr:     true,
+			expectedErrCode: InvalidArgument,
+		},
+		{
+			name:            "Already existing",
```
nikkiki:
```suggestion
			name:            "Already exists",
```

backend/apid/actions/clusterroles_test.go
```diff
@@ -0,0 +1,406 @@
...
+		{
+			name:            "Invalid input",
+			ctx:             context.Background(),
+			argument:        types.FixtureClusterRole("read-write"),
+			storeErr:        &store.ErrNotValid{},
+			expectedErr:     true,
+			expectedErrCode: InvalidArgument,
+		},
+		{
+			name:            "Already existing",
```
nikkiki:
```suggestion
			name:            "Already exists",
```

backend/apid/actions/rolebindings_test.go
```diff
@@ -0,0 +1,406 @@
...
+		{
+			name:            "Invalid input",
+			ctx:             context.Background(),
+			argument:        types.FixtureRoleBinding("read-write", "default"),
+			storeErr:        &store.ErrNotValid{},
+			expectedErr:     true,
+			expectedErrCode: InvalidArgument,
+		},
+		{
+			name:            "Already existing",
```
nikkiki:
```suggestion
			name:            "Already exists",
```

cli/commands/clusterrole/help.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"github.com/sensu/sensu-go/cli"
+	"github.com/spf13/cobra"
+)
+
+// HelpCommand defines new parent
+func HelpCommand(cli *cli.SensuCli) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "cluster-role",
+		Short: "Manage ClusterRoles",
```
nikkiki:
```suggestion
		Short: "Manage cluster roles",
```

cli/commands/role/help.go
```diff
@@ -9,17 +9,15 @@ import (
 func HelpCommand(cli *cli.SensuCli) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "role",
-		Short: "Manage roles",
+		Short: "Manage Roles",
```
nikkiki:
```suggestion
		Short: "Manage roles",
```

nikkiki:
Is there any type of validation for role verbs? It looks like I can pass any string as a verb. Without looking at the tests and seeing some examples, I'm not sure what strings are expected.

treydock:
I was looking at this pull request on how it affects Puppet and unsure if this is expected behavior:

```
[root@sensu-backend ~]# sensuctl role list
Error: not found
[root@sensu-backend ~]# sensuctl version
sensuctl version 2.0.0-dev#13d0336, build 13d033614a862eb515c1d7f08bf35ca9bced174c, built 2018-11-22T20:14:05+0000
```

This is fresh etcd data set, I stopped sensu-backend , clearned etcd files, restarted sensu-backend.  Most other resources would return empty array with `list` when nothing to list.

ccressent:
Hi @treydock,

The fact that there are no roles is expected (only some cluster roles are seeded), but I agree with you, the UX is not great: `role list` should behave consistently with the rest of the `* list` commands!

treydock:
Let me know if warrants new issue but seems `--all-namespaces` flag is missing from `sensuctl role list`:

```
[root@sensu-backend ~]# sensuctl role list --all-namespaces
Error: unknown flag: --all-namespaces
```

ccressent:
I've noticed that too when having a quick look at the CLI code to check out the other issue you mentioned. The `role-binding` sub-command is also affected.

I'll take care of addressing this before merging. Thanks @treydock!

nikkiki:
This got lost in the shuffle but I'm still a little curious: https://github.com/sensu/sensu-go/pull/2338#issuecomment-438852393

ccressent:
@nikkiki: sorry I never addressed your comment!

Those verbs are created by the `AuthorizationAttributes` middleware, based on the HTTP verb used and the endpoint being hit. The expected verbs are the RBAC `get,list,create,update,delete`. 

You're right we don't explicitly validate the input, but the Authorization library should tell you your action is unauthorized if the verb is not one of these. Additionally, a user cannot set the RBAC verb manually.

ccressent:
Good point raised by @nikkiki: not validating the verbs in user created `Role` and `ClusterRole` means we potentially have a bad UX: while the backend will just never authorize any action with a verb it doesn't recognize, the CLI will show and send bogus verbs to the backend.

Will address in a separate ticket.

treydock:
I am testing this further and having issues creating RoleBinding using `sensuctl create -f`:

```
[root@sensu-backend ~]# cat role_binding.json
{
  "type": "RoleBinding",
  "spec": {
    "name": "test",
    "roleRef": {
      "type": "Role",
      "name": "test"
    },
    "namespace": "default",
    "subjects": [
      {
        "kind": "User",
        "name": "admin"
      }
    ]
  }
}
[root@sensu-backend ~]# sensuctl create -f role_binding.json
json: unknown field "kind"
Error: some resources couldn't be parsed
```

Maybe this is bug in documentation?  The docs say to use `kind` but rbac.proto expects `type` for a subject.

treydock:
Changing `kind` to `type` fixes my issue.  Should I go by what's in the code or by the docs?  If the code is correct then I'll open issue for docs.

echlebek:
@treydock please open a docs issue. Thanks!

ccressent:
@treydock: this was changed in 501e88336dece9edb86e4fc17bba79de69497759 in order to only use `type` everywhere instead of a mix of `type` and `kind`.

treydock:
@ccressent @echlebek Thanks for clarification, issue opened with docs repo: https://github.com/sensu/sensu-docs/issues/922

Issue #2309, opened 2018-11-08
[Web UI] Address breaking changes to field names


nikkiki:
/cc @jamesdphillips 

See the breaking name changes listed in https://github.com/sensu/sensu-go/pull/2308. This is to address the name changes in graphql and dashboard files.

Requires https://github.com/sensu/sensu-go/issues/2304
Requires https://github.com/sensu/sensu-go/issues/2049

PR #2339, opened 2018-11-15
[Web UI] Update filtering to use JS expressions

jamesdphillips:
## What is this change?

Update filtering to use ES5 expressions instead of govalutate.

Closes https://github.com/sensu/sensu-go/issues/2309
> Move fast and break things

Regression was introduced prior to the last beta release.

### DIFF
--- a/dashboard/src/components/partials/EventsList/EventsListHeader.js
+++ b/dashboard/src/components/partials/EventsList/EventsListHeader.js
@@ -2,18 +2,26 @@ import React from "react";
 import PropTypes from "prop-types";
 import gql from "graphql-tag";
 
-import StatusMenu from "/components/partials/StatusMenu";
-import ListHeader from "/components/partials/ListHeader";
-
 import ConfirmDelete from "/components/partials/ConfirmDelete";
 import DeleteMenuItem from "/components/partials/ToolbarMenuItems/Delete";
 import ExecuteMenuItem from "/components/partials/ToolbarMenuItems/QueueExecution";
+import ListHeader from "/components/partials/ListHeader";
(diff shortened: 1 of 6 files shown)


dashboard/src/components/partials/EventsList/EventsListHeader.js
```diff
@@ -62,36 +70,35 @@ class EventsListHeader extends React.Component {
...
+    this.props.onChangeQuery({ filter: `check.name === "${newValue}"` });
   };
 
   requeryHide = newValue => {
     if (newValue === "passing") {
-      this.props.onChangeQuery({ filter: `Check.Status != 0` });
+      this.props.onChangeQuery({ filter: `check.status !== 0` });
     } else if (newValue === "silenced") {
-      this.props.onChangeQuery({ filter: `!IsSilenced` });
+      this.props.onChangeQuery({ filter: `check.silenced.length > 0` });
```
echlebek: The IsSilenced property has been preserved but is now named `is_silenced`

dashboard/src/components/views/EnvironmentView/EventsContent.js
```diff
@@ -11,7 +11,7 @@ import Query from "/components/util/Query";
 import { withQueryParams } from "/components/QueryParams";
 
 // If none given default expression is used.
-const defaultExpression = "HasCheck";
```
echlebek: You can use the `has_check` property here.

jamesdphillips: Ah, I see now @ `SynthesizeExtras` 👍

PR #2360, opened 2018-11-16
Fix resource leaks in the monitor and ring.

### LINKED ISSUE #2314: Verified leak in Backend
# Problem description

About a month of having sensu-go beta deployed, I've discovered and verified with @grepory and @echlebek on Slack that there are leaking gRPC related goroutines. The environment has about 80 agents connected to a 3-node cluster (2cores, 8GB RAM each). Every few hours however, the kernel kills the process because of the node's memory exhaustion as seen in this graph

Last week of sensu01's available memory

# Debugging resources

- latest beta release from deb repo `sensu-backend version 2.0.0-beta.7-7#0ad392d, build 0ad392d15948e7cc38fdb93a9ada56aa05f42970, built 2018-10-26T17:07:52+0000`
- [Heap profile (bytes) SVG](https://sensucommunity.slack.com/files/U88DUR2HX/FE1KMKS14/heap.svg)
- [Heap profile (objects) SVG](https://sensucommunity.slack.com/files/U88DUR2HX/FDZFLA88G/heap-obj.svg)
- [Trace file](https://www.dropbox.com/s/1i1oio03ccopox8/trace?dl=1)

echlebek:
```
    Fix resource leaks in the monitor and ring.
    
    This commit fixes a goroutine leak in the supervisor, where each
    monitor would create a goroutine that would never terminate until
    the end of the program. It also adds some code that guards against
    watchers that are still waiting on events after the TTL has been
    reached.
    
    This commit also fixes a goroutine leak in the ring, where Next
    calls would build up over time. Now, Next calls will last no
    longer than the check interval duration they are associated with.
```

Closes #2314

### DIFF
--- a/backend/messaging/wizard_bus_test.go
+++ b/backend/messaging/wizard_bus_test.go
@@ -1,6 +1,7 @@
 package messaging
 
 import (
+	"context"
 	"fmt"
 	"math/rand"
 	"testing"
@@ -142,7 +143,7 @@ func TestPublishDirect(t *testing.T) {
 		_, err := bus.Subscribe("topic", id, ch)
 		require.NoError(t, err)
 	}
-	require.NoError(t, bus.PublishDirect("topic", "hello, world"))
+	require.NoError(t, bus.PublishDirect(context.TODO(), "topic", "hello, world"))
 	select {
 	case msg := <-subscribers["a"].Channel:
 		assert.Equal(t, msg, "hello, world")
--- a/backend/monitor/supervisor.go
+++ b/backend/monitor/supervisor.go
@@ -2,11 +2,14 @@ package monitor
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"os/signal"
 	"strconv"
+	"sync"
 	"syscall"
+	"time"
 
 	"github.com/coreos/etcd/clientv3"
 	"github.com/coreos/etcd/mvcc/mvccpb"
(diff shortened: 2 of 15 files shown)


echlebek:
Oops. Do not review yet. It is broken and I didn't notice because I was testing manually.

echlebek:
This is ready for review now :)

backend/monitor/supervisor.go
```diff
@@ -146,38 +178,37 @@ func (m *EtcdSupervisor) getMonitor(ctx context.Context, key string) (*monitor,
...
-		)
-
-		if _, err := cli.Lease.Revoke(ctx, mon.leaseID); err != nil {
-			logger.WithError(err).Warningf("could not revoke the lease %v for the key %s",
-				mon.leaseID, mon.key,
-			)
-		}
-	}()
+	leasesMu.Lock()
+	defer leasesMu.Unlock()
```
ccressent: My 2 cents on style related to critical sections: I prefer to explicitly call `Unlock()` as soon as I'm done with a resource, rather than say the equivalent of "from this point on, hold the lock until this function returns".

You don't want someone else potentially unknowingly adding code in a critical region!

backend/messaging/wizard_bus_test.go
```diff
@@ -142,7 +143,7 @@ func TestPublishDirect(t *testing.T) {
 		_, err := bus.Subscribe("topic", id, ch)
 		require.NoError(t, err)
 	}
-	require.NoError(t, bus.PublishDirect("topic", "hello, world"))
+	require.NoError(t, bus.PublishDirect(context.TODO(), "topic", "hello, world"))
```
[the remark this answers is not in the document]
echlebek: I used TODO because I didn't put much thought into whether or not the context here should be cancelable.

backend/monitor/supervisor.go
echlebek: In this case I agree with you. However in other cases, it is a very good practice to defer the Unlock: if a panic occurs in the CS, then the mutex will still be unlocked. This can make a big difference in code that calls recover().

util/strings/strings.go
```diff
@@ -12,15 +16,58 @@ func InArray(item string, array []string) bool {
...
+func alphaNumeric(s string) bool {
+	for _, r := range s {
+		if !(isDigit(r) || isAlpha(r)) {
+			return false
+		}
+	}
+	return true
+}
+
+func isNormalized(array []string) bool {
```
golangcibot: `isNormalized` is unused

echlebek:
I realized there is still a resource leak present, although it seems to be a slower one. I'm hunting it down now. In doing so, I noticed that one function is accounting for 87% of all allocations in the backend. So I optimized it to try to get a clearer picture of where we are leaking memory.

echlebek:
Heap size has been stable on my test cluster for 6 hours now. This is a snapshot of memory stats for a backend that's servicing 25 agents. The heap has not increased its high water mark of 532 MB. I think this task is complete.

```
# runtime.MemStats
# Alloc = 70241536
# TotalAlloc = 82947490440
# Sys = 568527096
# Lookups = 0
# Mallocs = 483211135
# Frees = 483013305
# HeapAlloc = 70241536
# HeapSys = 532316160
# HeapIdle = 443457536
# HeapInuse = 88858624
# HeapReleased = 402309120
# HeapObjects = 197830
# Stack = 4554752 / 4554752
# MSpan = 1313584 / 5406720
# MCache = 27648 / 32768
# BuckHashSys = 2348350
# GCSys = 20238336
# OtherSys = 3630010
```

PR #2364, opened 2018-11-19
Don't re-use otto VMs

### LINKED ISSUE #2354: Redact objects passed into otto js vms
When we pass events into the otto VM, we should redact values. People shouldn't filter things based on AWS secret keys.

They may want to filter based on the _presence_ of keys, so we should simply redact like we do for logging and not remove the keys altogether.

### LINKED ISSUE #2363: Don't re-use otto VMs
It's too easy for users to cache values in javascript VMs, which could easily lead to memory leaks and undefined behaviour. Return a copy of the cached VMs instead of allowing them to be re-used.

echlebek:
This commit causes the javascript interpreter to always use a new
VM for every filter evaluation. This is much slower, but greatly
increases the safety of the feature, since users cannot cache state
between filter executions.

Before:
```
BenchmarkNewOttoVM-16   2572889 ns/op	 1514648 B/op	   13593 allocs/op
BenchmarkCheckEval-16     70336 ns/op	   18120 B/op	     405 allocs/op
BenchmarkMetricsEval-16   18662 ns/op	    4344 B/op	     120 allocs/op
```

After:
```
BenchmarkNewOttoVM-16   545758 ns/op	  321478 B/op	    3040 allocs/op
BenchmarkCheckEval-16   369857 ns/op	  189656 B/op	    1911 allocs/op
BenchmarkMetricsEval-16 315916 ns/op	  175872 B/op	    1626 allocs/op
```

The change dramatically increases CPU costs and GC pressure. It's
not known at this time if this can be improved. Profiles of the
Otto Copy() function haven't yielded any low hanging fruit for
optimization.

Entity redaction has been re-worked, and entities are now redacted before being filtered.

Closes #2363 
Closes #2354


types/entity.go
```diff
@@ -19,8 +19,15 @@ const (
 
 	// EntityBackendClass is the name of the class given to backend entities.
 	EntityBackendClass = "backend"
+
+	// Redacted is filled in for fields that contain sensitive information
+	Redacted = "REDACTED"
 )
 
+// DefaultRedactFields contains the default fields to redact
+var DefaultRedactFields = []string{"password", "passwd", "pass", "api_key",
```
ccressent: They all seem sensible, default field names to redact to me, except maybe for `pass`: it sounds more like a pass/fail type of field to me, holding a boolean.

js/vm_cache.go
```diff
@@ -87,26 +77,17 @@ func (c *vmCache) Acquire(key string) *otto.Otto {
...
 	}
-	return <-val.vms
+	if val.vm == nil {
+		return nil
+	}
+	return val.vm.Copy()
 }
 
 // Init initializes the value in the cache, creating cacheMaxConcurrency copies
 // of it.
```
ccressent: Outdated comment.

types/entity.go
echlebek: These are already established in the product and documented. I don't want to change them in this PR.

echlebek:
Oops. I forgot to run the integration tests.

echlebek:
@ccressent to address your question:

The VM that the user's code will run on is a copy of a VM that has had assets and utility functions preloaded. It is discarded after execution.

PR #2366, opened 2018-11-20
Merge the assets and {check,handler,mutator} environments

### LINKED ISSUE #2317: Systemd vs sensu-agent start paths are different
In testing out the [monitoring external resources guide](https://docs.sensu.io/sensu-go/5.0/guides/monitor-external-resources/), it seems that there is a difference between the $PATH used by `systemd` and `sensu-agent`:

**systemd**

systemctl-agent-path

**sensu-agent**

sensu-start-path

If one has to use `sensu-agent start`, or chooses to use this over an operating system service manager (or wouldn't have access to one), then this particular check would not work due to the differences in the respective paths. 

Are these expected to be different? Is this simply a case of using the sensu-agent workflow when we should be using an OS service manager instead?

ccressent:
```
The values of the PATH, CPATH and LD_LIBRARY_PATH variables are merged,
with a preference given to the values held in the {check,handler,mutator}'s 
environment. That is, paths coming from a {check,handler,mutator}'s
environment take precedence, but do not overwrite paths coming from the 
assets' environment.

Any other variable is merged, with conflicts resolved in favor of the
{check,handler,mutator} environment.

Signed-off-by: Cyril Cressent <cyril.cressent@sensu.io>
```

Closes #2317.

Yes, added.

No.
Changes may be required.


agent/check_handler.go
```diff
@@ -223,3 +224,44 @@ func extractMetrics(event *types.Event) []*types.MetricPoint {
 
 	return transformer.Transform()
 }
+
+// mergeEnvironments merges env2 into env1, overwriting any existing variable
+// in env1, except for the "special" variables PATH, CPATH and LD_LIBRARY_PATH.
+// These variables are merged by prepending the value from env2 to the value in
+// env1, effectively giving priority to the value from env2.
+func mergeEnvironments(env1, env2 []string) []string {
+	toMap := func(s []string) map[string]string {
```
echlebek: Defining this in the scope of mergeEnvironments is a bit odd. It captures the env1 and env2 parameters, which are not used in the function. Why not move it out to the package scope?

agent/check_handler.go
```diff
@@ -223,3 +224,44 @@ func extractMetrics(event *types.Event) []*types.MetricPoint {
...
+		m := map[string]string{}
+
+		for _, v := range s {
+			split := strings.SplitN(v, "=", 2)
```
echlebek: SplitN with an `n` parameter of 2 does not make sense here, as we expect to find only a single `=`. Also, we should probably do a sanity check here that we actually get a slice with len(s) > 1, as not to inadvertently panic on invalid environment variables.

agent/check_handler.go
```diff
@@ -223,3 +224,44 @@ func extractMetrics(event *types.Event) []*types.MetricPoint {
...
+			m[split[0]] = split[1]
+		}
+
+		return m
+	}
+
+	fromMap := func(m map[string]string) []string {
```
echlebek: Similar to the above closure, this function doesn't really use anything it closes over, so why not make it a stand-alone?

agent/check_handler.go
```diff
@@ -223,3 +224,44 @@ func extractMetrics(event *types.Event) []*types.MetricPoint {
...
+	for k, v := range e2 {
+		switch k {
+		case "PATH", "CPATH", "LD_LIBRARY_PATH":
+			e1[k] = strings.Join([]string{v, e1[k]}, ":")
+		default:
+			e1[k] = v
+		}
+	}
+
+	return fromMap(e1)
```
echlebek: It would be nice if the output of mergeEnvironments produced a predictably ordered slice. Maybe sort the strings?

agent/check_handler_internal_test.go
```diff
@@ -308,3 +308,74 @@ func TestExtractMetrics(t *testing.T) {
 		})
 	}
 }
+
+func TestMergeEnvironments(t *testing.T) {
```
echlebek: Tests look good, but can you add a test case for an invalid environment variable, so that we can get a well defined result for that case?

agent/check_handler.go
ccressent: What was odd to me was to represent the environment with `[]string` :)

Here was my reasoning for not defining `toMap()` and `fromMap()` in the package scope: they are fairly generic, but they have to be restricted to `[]string -> map[string]string` and `map[string]string -> []string` respectively in Go. So they become one off helpers as far as I'm concerned, because I don't want to pollute the package scope with either a function name that imply genericity or a function name that looks like `fromStringSliceToMapStringString()`.

You can probably guess I wasn't in a very Go mood when I wrote that code! I'm happy to go with whatever is "idiomatic Go" though.

agent/check_handler.go
ccressent: I disagree, we do not expect to find a single `=`. An environment variable's value is an arbitrary string. Here is a counter example:

```
MY_SUPER_IMPORTANT_VAR="launch_missile=false && eat_tacos=true"
```

All I think we can expect is that the left side of the _first_ `=` is a variable name (the key in the map) and its left side is the associated value.

I agree with checking the size of `split` though!

echlebek: But shouldn't it at least be SplitN with a parameter of `1` then?

agent/check_handler.go
echlebek:
```
¯\_(ツ)_/¯
```

grepory:
If you want a "rule" for what I would consider "idiomatic" in Go:

If you are not capturing state from the outer scope inside of a closure, do not user a closure.

PR #2374, opened 2018-11-21
Update GraphQL Service for redesigned RBAC

### LINKED ISSUE #2261: Support the new RBAC in the dashboard
Since abilities & policies are no longer a thing, we need to ensure the dashboard supports the new RBAC model.

TODO: define the tasks

Child of https://github.com/sensu/sensu-go/issues/2243
Requires https://github.com/sensu/sensu-go/issues/2260

jamesdphillips:
## What is this change?

graphql-service-diagram

This set of changes updates the GraphQL service to communicate with the REST API instead of directly with the controllers. This has two distinct benefits:

1. Authorization does not have to replicated in multiple places. This way we can continue using the authorization middleware introduced in #2338 without any changes. 
2. The GraphQL service can leverage existing client stubs making it easier to test field resolvers.

This of course comes with the obvious downside that queries are moderately less efficient. However, in tests, I haven't found this to be significantly slower in any way.

**Note:** The follow changes were put together pretty quickly, primarily with the goal of getting the dashboard back up and running after the RBAC redesign. There is tons of room for improvement, such as, generating a lot of the redundant code, adopting [dataloader](https://github.com/graph-gophers/dataloader) to reduce trips to the network and plenty of room for optimization. 

Closes #2261

Arguably, the RBAC redesign entry already encompasses these changes.

In an attempt to reduce the time required to ship this change, I decided to leverage the existing REST client library used by sensuctl. While this made writing tests a great deal easier, the client is a bit of a mess. (As the primary contributor to the package, I can safely say this is my fault. Sorry™)

It would be nice if we can adopt openapi-generator or swagger-gen in the future.


backend/apid/graphql/restclient/client.go
```diff
@@ -0,0 +1,42 @@
...
+		accessToken = token
+	}
+
+	tokens := types.Tokens{
+		Access:    accessToken,
+		ExpiresAt: 10000000000000, // TODO
+	}
+
+	config := inmemory.New(c.url)
+	config.SaveTokens(&tokens)
```
golangcibot: Error return value of `config.SaveTokens` is not checked

backend/apid/graphql/restclient/client.go
```diff
@@ -0,0 +1,42 @@
...
+	config.SaveNamespace(types.ContextNamespace(ctx))
```
golangcibot: Error return value of `config.SaveNamespace` is not checked

backend/apid/graphql/entity.go
```diff
@@ -62,13 +45,15 @@ func (r *entityImpl) LastSeen(p graphql.ResolveParams) (*time.Time, error) {
...
 
 	// fetch
-	ctx := types.SetContextFromResource(p.Context, entity)
-	evs, err := r.eventQuerier.Query(ctx, entity.Name, "")
+	evs, err := fetchEvents(client, src.Namespace, func(obj *types.Event) bool {
+		return obj.Entity.Name == src.Name
+	})
 	if err != nil {
-		return 0, err
+		return evs, err
```
ccressent: Not sure why the return changes here. If `err != nil`, one is not supposed to expect anything out of the other returned parameters anyway. I would say it should return `nil, err`, but I have to admit I don't want to try to fix what's not broken...

backend/apid/graphql/entity.go
```diff
@@ -80,32 +65,14 @@ func (r *entityImpl) Events(p schema.EntityEventsFieldResolverParams) (interface
...
-		// - Since we are dealing with a slice of pointers we explicilty set the
-		//   reference to the last element to nil to ensure it can be GC'd.
-		//
-		// https://github.com/golang/go/wiki/SliceTricks#delete-without-preserving-order
-		//
-		entities[i] = entities[len(entities)-1]
-		entities[len(entities)-1] = nil
-		entities = entities[:len(entities)-1]
-		break
+		return entities, nil
```
ccressent: Why `return entities, nil` if `err != nil`?

backend/apid/graphql/error.go
```diff
@@ -17,7 +18,7 @@ type stdErr struct {
 func newStdErr(input string, err error) stdErr {
 	out := stdErr{code: schema.ErrCodes.ERR_INTERNAL, input: input}
 	switch terr := err.(type) {
-	case actions.Error:
+	case client.APIError: // TODO APIError
```
ccressent: Not TODO after all with the changes to `mapServiceErrCode()` below? Or is it referring to something else?

backend/apid/routers/graphql_test.go
```diff
@@ -7,21 +7,10 @@ import (
...
-	store := &mockstore.MockStore{}
-	queue := &mockqueue.MockQueue{}
-	bus := &mockbus.MockBus{}
-
-	getter := &mockqueue.Getter{}
-	getter.On("GetQueue", mock.Anything).Return(queue)
-
-	router := NewGraphQLRouter(store, bus, getter)
+	router := NewGraphQLRouter("test")
 	return router
```
ccressent: You could just `return NewGraphQLRouter("test")` if you wanted.

Issue #2368, opened 2018-11-20
etcd SSL Peer Handshake fails because missing IP SANs


palourde:
A user reported the following error when trying to configure a TLS cluster:

```
x509: cannot validate certificate for x.x.x.x because it doesn't contain any IP SANs
```

Related issues:
- https://github.com/etcd-io/etcd/issues/3496

I can't reproduce this issue on my local machine (macOS) but @cwjohnston was able using sensu-staging.

palourde:
Finally identified the problem... Somehow, etcd needs to connect to the **first** URL passed in `etcd-listen-client-urls`. If the certificates were issued using only DNSNames in the SANs, it will never work.

Therefore, I was able to fix this issue by doing the following:

1. Add the `etcd-advertise-client-urls` attribute to Sensu, so the proper "private and/or public" URL is passed to other cluster members.
2. Specify the loopback IP address **first** in the `etcd-listen-client-urls` attribute, then the private and or public URL address so etcd uses the first address to connect. E.g. `"http://127.0.0.1:2379,https://10.0.1.10:2379"`

I just need to prepare a PR with what I have!

Issue #1964, opened 2018-08-15
Add initial-advertise-client-urls configuration option to sensu-backend


cyphus:
Etcd has a configuration option, `initial-advertise-client-urls`, that specifies one or more client urls that get advertised to other etcd members. The current implementation advertises `listen-client-urls` to other members, but this is incorrect, as the listening address may be different than the address used by other members to connect.

## Context
This issue was discovered while testing the health endpoint in a clustered environment. Checking etcd health requires sending client requests to all other etcd members, which is done by connecting to the other members' advertised client urls. In the case of sensu-staging, the advertised client url is the listening address, which is a non-routable `0.0.0.0`.

annaplotkin:
Citi came across an issue with clustering w/ ssl and after further investigation it was deemed that this was the culprit. Nominating this to be resolved in Beta 9.

PR #2375, opened 2018-11-21
Improve SSL/TLS clusters support 

palourde:
## What is this change?

This PR fixes a couple of bugs around SSL/TLS clusters. In summary:

- It adds the `etcd-advertise-client-urls` config attribute to sensu-backend
	- This is useful if you are using TLS certificates with only DNSNames in the SANs. Sensu will now use the advertised URLs in our etcd clients, instead of the listen URLS, so the certificates won't throw a validation errors (e.g. the certificate was issued for foo.bar.io, not 127.0.0.1)
- It now passes the Cluster client and etcd Client TLSConfig to the ClusterHealth store (`backend/store/etcd/health_store.go`).
	- The `s.client.MemberList` method won't use the advertised URLs and doesn't have the Client TLSConfig, so certificate validation errors will be thrown otherwise
	- The *clientv3.Client used to query each member now uses the provided etcd Client TLS Config in case HTTPS encryption or authentication is enabled
- I removed the default etcd values configured in `backend/etcd/etcd.go` (lines 33 to 44); the `backend/cmd/start.go` file already sets default values so they were never used, excepted for our tests, so I configured the proper default values for tests in `backend/store/etcd/testutil/store.go`.
- All 4 etcd URLs attributes (`etcd-advertise-client-urls`, `etcd-listen-client-urls`, `etcd-listen-peer-urls` & `etcd-initial-advertise-peer-urls`) should accept a list of URLs, so I modified viper to use `GetStringSlice` & pflag to use `StringSlice` for these attributes, therefore we deal with `[]string` values all the way.
	- I also modified the logic around the validation of these URLs in order to use etcd own validation logic instead of doing it inconsistently in our codebase (see `backend/etcd/etcd.go`, lines 108 to 129)

Closes https://github.com/sensu/sensu-go/issues/1964
Closes https://github.com/sensu/sensu-go/issues/2368

Added!

Nope!

Nope!

I'll need to update the How to Run a Sensu Cluster guide in our docs


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Added
+- Add the `etcd-advertise-client-urls` config attribute to sensu-backend
```
nikkiki: My ocd is suggesting you add a newline here.

palourde: Fixed via https://github.com/sensu/sensu-go/pull/2375/commits/e51fbaebda392863ddbf56543e9d16659c134e28

backend/cmd/start.go
```diff
@@ -74,10 +75,14 @@ const (
...
 	// only)
 	defaultEtcdName = "default"
-	// DefaultEtcdPeerURL is the default URL to listen for Etcd peers (single-node
+	// defaultEtcdPeerURL is the default URL to listen for Etcd peers (single-node
 	// cluster only)
 	defaultEtcdPeerURL = "http://127.0.0.1:2380"
 
+	// defaultEtcdAdvertiseClientURL is the default list of this member's client
+	// URLs to advertise to the rest of the cluster
+	defaultEtcdAdvertiseClientURL = "http://localhost:2379"
```
nikkiki: This is `localhost` intentionally since this field can accept hostnames?

backend/store/etcd/health_store.go
```diff
@@ -10,14 +11,15 @@ import (
 )
 
 // GetClusterHealth retrieves the cluster health
-func (s *Store) GetClusterHealth(ctx context.Context) *types.HealthResponse {
+func (s *Store) GetClusterHealth(ctx context.Context, cluster clientv3.Cluster, etcdClientTLSConfig *tls.Config) *types.HealthResponse {
```
palourde: FYI this method could definitely moved from the store; it doesn't use the store's client anymore. Not sure if it's worth doing now.

backend/cmd/start.go
palourde: Yep, and it's etcd default value too (https://coreos.com/etcd/docs/latest/op-guide/configuration.html)

amdprophet: This makes sense to get certs working.

We have been trying to avoid the use of localhost anywhere because it can resolve to the IPv6 loopback address on some platforms. When localhost is used by a listener, this can mean that nothing is bound to IPv4. When localhost is used by a client, the endpoint it's connecting to may not be bound to IPv6.

We'll need to make sure to document troubleshooting steps for cases where this could happen.

amdprophet: I think so long as etcd is listening on `[::]` by default (which should bind to both IPv4 and IPv6 loopbacks) this shouldn't actually be an issue.

palourde: For reference, this is etcd's default too (https://coreos.com/etcd/docs/latest/op-guide/configuration.html) but yeah, we definitely need to add a mention about that in our docs

palourde:
Test with the following scenarios:

- [ ] Standalone backend without encryption/authentication
- [ ] Standalone backend with Client HTTPS authentication
- [ ] 3 nodes cluster with Client & Peer HTTPS authentication

PR #2377, opened 2018-11-21
Disable check subdue functionality in sensuctl

### LINKED ISSUE #2376: Disable check subdue until after #1955
Check subdue is broken, and we don't have time to fix it before GA. The proposed solution is a breaking change, which we would not want to make soon after the GA release.

So, we are not going to be shipping check subdue with GA. The feature will be disabled in sensuctl until we can provide a 100% correct implementation.

echlebek:

## What is this change?

Disables check subdue set/remove in sensuctl.

Closes #2376 

Yes

No

No

## Doc changes?

Yes, we'll need to make documentation changes.


echlebek:
@nikkiki ah you're right, I completely forgot about the `create` method.

I'm kind of split on the remove command... anyone else have an opinion?

I'll add a shim to make it impossible to create a subdue via create as well.

ccressent:
I see @nikkiki's point... though the CHANGELOG says "remove and re-create your check" so...

echlebek:
I think we can do without the remove-subdue command.

Users can do this:
```
sensuctl check info mycheck --format yaml | sensuctl create
```

to remove any lingering subdues.

echlebek:
I've added a shim to null-out any subdues specified in `sensuctl create`.

cli/commands/create/create.go
```diff
@@ -123,9 +123,26 @@ func ParseResources(in io.Reader) ([]types.Resource, error) {
...
 	}
+
+	// TODO(echlebek): remove this
+	filterCheckSubdue(resources)
+
 	return resources, err
 }
 
+// filterCheckSubdue nils out any check subdue fields that are supplied.
+// TODO(echlebek): TODO(echlebek): TODO(echlebek): this is temporary; remove it after fixing check subdue.
```
ccressent: Wow, I hope you don't forget to remove it.

echlebek: o_O

echlebek: Amended the stutter

PR #2378, opened 2018-11-21
Fix object listing in namespaces

### LINKED ISSUE #2371: Resources from similarly named namespaces are included in fetch results
## Expected Behavior

Fetch results from one namespace should not include results from another. 

## Current Behavior

* Given two namespaces `sensu` and `sensu-devel`.
* The query `GET /checks?namespace=sensu` returns checks not just from the `sensu` namespace but also `sensu-devel`.
* Caveat: Querying checks from the `sensu-devel` namespace does not include checks from the `sensu` namespace.

## Possible Solution

This is likely a prefix issue. We may likely simply need a trailing slash in the query to etcd.

See above.

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): sensu-backend
* Installation method (packages, binaries, docker etc.): docker
* Operating System and version (e.g. Ubuntu 14.04): macOS

ccressent:
## What is this change?
Resources from namespaces that share a common prefix, eg: "sensu" and
"sensu-devel", could be listed together.

This was due to a bug with key construction and key prefix usage in the
etcd client.

Closes #2371.

Yes, added.

No.

### DIFF
--- a/backend/store/key_builder.go
+++ b/backend/store/key_builder.go
@@ -81,7 +81,6 @@ func (b KeyBuilder) BuildPrefix(keys ...string) string {
 		if key == "" {
 			continue
 		}
-		//
 		out = out + keySeparator + key
 	}
 
--- a/backend/store/key_builder.go
+++ b/backend/store/key_builder.go
@@ -49,7 +49,20 @@ func (b KeyBuilder) Build(keys ...string) string {
 		},
 		keys...,
 	)
-	return path.Join(items...)
+
+	key := path.Join(items...)
+
+	// In order to not inadvertently build a key that could list across
+	// namespaces, we need to make sure that we terminate the key with the key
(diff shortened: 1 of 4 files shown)


backend/store/key_builder.go
```diff
@@ -68,6 +81,7 @@ func (b KeyBuilder) BuildPrefix(keys ...string) string {
 		if key == "" {
 			continue
 		}
+		//
```
nikkiki:
```suggestion
```

ccressent: Whoops, left one of my "markers" for things to check out. Good catch!

ccressent:
@nikkiki 
I think this covers it:
https://github.com/sensu/sensu-go/pull/2378/files#diff-7f3840639c3b6eb5237f3c0012451a08R29

PR #2381, opened 2018-11-21
[Web UI] Silencing language & confirmation

### LINKED ISSUE #2195: [Web UI] Deleting/clearing silences should be easier to understand
When doing QA testing, I found the language used for clearing silences to be a bit scary. While we use language like "clear silence" on the check details page and different list pages, when you're on the silences page the language says "delete". The hover text for the bulk actions button also says "Delete this resource". When you click delete through this button, or through the kebab menu the language states "Are you sure you want to permanently delete this resource?". To me, all of this sounds like I'm going to be deleting checks themselves. I think we should change the wording to state our intentions clearer.

## Expected Behavior
Clearly understand I'm only clearing a silence (or deleting a silenced entry, whatever nomenclature) and NOT deleting the resource itself.

## Current Behavior
Described in the summary.

## Possible Solution
Discuss the proper language to use in this scenario and change the confirmation modal, button text, and hover text as needed.

## Context
I'm fairly new to the product on this team and even in a staging environment I was scared to accidentally delete something. I imagine our end users could feel the same.

dabria:
## What is this change?
Closes #2195 
- Changed some language around unsilencing and deleting to be more clear and consistent
- Added confirmation on the Silences List for unsilencing

screen shot 2018-11-21 at 12 49 30 pm

Consistent language is good. I also figured that one might appreciate a confirmation on clearing a silence, in case a lot of people start getting alerted for something because of it. I would like to do something similar for deleting in bulk as well.

I added one in case

So, I realized that the ChecksList page implements this behaviour better than I currently have in SilencesList. I'd like to modify what I have in order to mimic that implementation. It has the dialog in the parent, and passes functions down to open the dialog. Right now I have the dialogs in the children, which is some duplicate code.

There's an issue currently when you clear 100% of the silences - either through a single item in the kebab or through bulk actions. 
```
warning.js:33 Warning: Can't call setState (or forceUpdate) on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
```
The implementation on ChecksList seems to avoid this issue. I know I could likely change my code somewhat easily to fix this with the assistance of someone, but considering I'd like to modify my approach entirely, I don't think it's worth the time given I'm in a bit of a time crunch to get this in. I don't think this is a big issue, and can be merged in otherwise - however I'll leave the final judgement of that up to you.


dashboard/src/components/controller/ListController.js
```diff
@@ -122,6 +127,7 @@ class ListController extends React.PureComponent {
       children: items.length
         ? items.map((item, i) => {
             const key = keys[i];
+            // This works, however you can't unselect an item in the list if there's confirmed items passed in
```
10xjs: Comment is no longer relevant

dashboard/src/components/partials/ClearSilencedEntriesDialog/Dialog.js
```diff
@@ -138,18 +138,17 @@ class ClearSilencedEntriesDialog extends React.PureComponent {
...
             <Loader loading={submitting} passthrough>
               <DialogTitle>Clear Silencing Entries</DialogTitle>
-              <DialogContent style={{ paddingBottom: 8 }}>
-                <DialogContentText>
-                  Select all entries you would like to clear.
-                </DialogContentText>
-              </DialogContent>
+              <Typography style={{ paddingLeft: "24px" }}>
+                Select all entries you would like to clear.
+              </Typography>
```
10xjs: A margin below this text would improve the visual look of the modal

dashboard/src/components/controller/ListController.js
dabria: Whoops, thanks!

dashboard/src/components/partials/ClearSilencedEntriesDialog/Dialog.js
dabria: screen shot 2018-11-21 at 1 00 31 pm
Added a bit of a margin

dashboard/src/components/partials/ConfirmDelete/Dialog.js
```diff
@@ -45,7 +45,7 @@ class ConfirmDeleteDialog extends React.Component {
         <DialogTitle id={titleId}>Confirm</DialogTitle>
         <DialogContent>
           <Typography>
-            Are you sure you would like to permanantly delete{" "}
+            Are you sure you would like to permenantly delete{" "}
```
jamesdphillips: 🤦‍♂️

dabria: I spelled it wrong there, whoops.

dabria: But if you're facepalming there was a different but similar line you could do that on ;)

PR #2391, opened 2018-11-23
Remove a flaky test

echlebek:


echlebek:
Part of the test is testing something we can't rely on, and the other part is represented elsewhere in the suite.

PR #2392, opened 2018-11-23
GraphQL service unauthorized err

jamesdphillips:
## What is this change?

Updates the GraphQL service to drop `PermissionDenied` errors when resolving fields that list resources. Selecting a field shouldn't result in an error unless something unrecoverable occurred.

In the future, we will need a method of retrieving permissions so that the UI can communicate lack of permission properly.

**NOTE** this should not be conflated with displaying resources despite lack of permission; we simply return an empty list in lieu of the error.
 
Without this change the exception modal is displayed in the web application when a user navigates to a page with resources they do not have access to.

unlikely.

### DIFF
--- a/backend/apid/apid.go
+++ b/backend/apid/apid.go
@@ -194,7 +194,10 @@ func registerGraphQLService(router *mux.Router, store store.Store, url string, t
 			router.NewRoute(),
 			middlewares.SimpleLogger{},
 			middlewares.LimitRequest{},
-			// Allow requests without an access token to continue
+			// TODO: Currently the web app relies on receiving a 401 to determine if
+			//       a user is not authenticated. However, in the future we should
+			//       allow requests without an access token to continue so that the
+			//       schema can be queried without authorization.
 			middlewares.Authentication{IgnoreUnauthorized: false},
 			middlewares.AllowList{Store: store, IgnoreMissingClaims: true},
 			middlewares.Edition{Name: version.Edition},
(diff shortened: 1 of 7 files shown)


backend/apid/apid.go
```diff
@@ -194,8 +194,11 @@ func registerGraphQLService(router *mux.Router, store store.Store, url string, t
 			router.NewRoute(),
 			middlewares.SimpleLogger{},
 			middlewares.LimitRequest{},
-			// Allow requests without an access token to continue
-			middlewares.Authentication{IgnoreUnauthorized: true},
+			// TODO: Currently the web app relies on receiving a 401 to determine if
+			//       a user is not authenticated. However, in the future we should
+			//       allow requests without an access token to continue so that the
+			//       schema can be queried without authorization.
```
ccressent: You mean queried without authentication?

jamesdphillips:
> Is there an easy way to find a middle ground?

Unfortunately there is no "easy" middle group at this point in time.

As discussed during the meeting we'll need to update the web app to consume the user's RBAC rules and proactively disable portions of the UI / communicate insufficient permissions better. At this point I don't think we can safely ship something like that today.

It's certainly not ideal that those with very limited sets of permissions will see empty states but at least they can _use_ the application and we have a simple path forward to enhancing the experience.

backend/apid/apid.go
jamesdphillips: Let me update that comment.

Issue #2398, opened 2018-11-23
Re-root API at /api/core/v2


echlebek:
Re-root the REST API at /api/core/v2.

* Change routes in APId
* Update routes in sensuctl client

Refs #2252

PR #2405, opened 2018-11-23
Update CLI client for new API routes

### LINKED ISSUE #2401: Update CLI client for new API routes
- [x] asset
- [x] check
- [x] clustermember
- [x] clusterrole
- [x] clusterrolebinding
- [x] event
- [x] filter
- [x] handler
- [x] hook
- [x] mutator
- [x] namespace
- [x] role
- [x] rolebinding
- [x] silenced
- [x] user

Refs #2252

jamesdphillips:
## What is this change?

Update CLI client for update API paths.

Requires #2398
Closes #2401 

Unlikely.

### DIFF
--- a/cli/client/client.go
+++ b/cli/client/client.go
@@ -4,7 +4,6 @@ import (
 	"crypto/tls"
 	"errors"
 	"fmt"
-	"net/http"
 	"time"
 
 	"github.com/go-resty/resty"
--- a/cli/client/extension.go
+++ b/cli/client/extension.go
@@ -3,7 +3,6 @@ package client
 import (
 	"encoding/json"
 	"fmt"
-	"net/url"
 
 	"github.com/sensu/sensu-go/types"
 )
@@ -33,11 +32,11 @@ func (client *RestClient) DeregisterExtension(name, namespace string) error {
 	path := extPath(namespace, name)
 	res, err := client.R().Delete(path)
 	if err != nil {
-		return fmt.Errorf("DELETE %q: %s", extensionPath, err)
+		return fmt.Errorf("DELETE %q: %s", path, err)
 	}
 
 	if res.StatusCode() >= 400 {
-		return fmt.Errorf("DELETE %q: %s", extensionPath, res.String())
+		return fmt.Errorf("DELETE %q: %s", path, res.String())
 	}
 
 	return nil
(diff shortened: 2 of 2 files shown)


cli/client/path.go
```diff
@@ -0,0 +1,42 @@
...
+func createBasePath(group, version string, resType ...string) func(...string) string {
+	fn := createNSBasePath(group, version, resType...)
+
+	return func(components ...string) string {
+		// first arg "" represents that no namespace has been specified
+		return fn("", components...)
+	}
+}
+
+func createNSBasePath(group, version string, resType ...string) func(string, ...string) string {
```
ccressent: It might be useful to describe how the returned function works, especially regarding how its first argument works: `""` being a special value that bypass namespaces, ...

cli/client/path_test.go
```diff
@@ -0,0 +1,42 @@
...
+			ins:       []string{"enterprise", "v7", "distributed-ledgers"},
+			paths:     []string{"😂"},
+			namespace: "sensu-devel",
+			out:       "/api/enterprise/v7/namespaces/sensu-devel/distributed-ledgers/%F0%9F%98%82",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.out, func(t *testing.T) {
+			fn := createNSBasePath(tc.ins[0], tc.ins[1], tc.ins[2])
```
ccressent: You can use `createNSBasePath(tc.ins...)` here so that you can add more "general" test cases with "multi-words" resource types, such as `ins: []string{"core", "v2", "cluster", "member"}`

cli/client/check.go
```diff
@@ -148,7 +139,7 @@ func (client *RestClient) AddCheckHook(check *types.CheckConfig, checkHook *type
 
 // RemoveCheckHook removes an association between an existing hook and an existing check
 func (client *RestClient) RemoveCheckHook(check *types.CheckConfig, checkHookType string, hookName string) error {
-	path := checksPath(check.Name, "hooks", checkHookType, "hook", hookName)
+	path := checksPath(check.Namespace, check.Name, "hooks", checkHookType, "hook", hookName)
 	res, err := client.R().SetQueryParam("namespace", check.Namespace).Delete(path)
```
ccressent: You need to remove the query parameter here.

cli/client/cluster.go
```diff
@@ -9,12 +9,13 @@ import (
 	"github.com/coreos/etcd/clientv3"
 )
 
-const clusterMembersBasePath = "/cluster/members"
+var clusterMembersPath = createBasePath(coreAPIGroup, coreAPIVersion, "cluster/members")
```
ccressent: You could also say `createBasePath(coreAPIGroup, coreAPIVersion, "cluster", "members")`, which I think is more in line with the intent of `createBasePath()`?

cli/client/user.go
```diff
@@ -2,15 +2,16 @@ package client
 
 import (
 	"encoding/json"
-	"net/url"
 
 	"github.com/sensu/sensu-go/types"
 )
 
+var usersPath = createBasePath(coreAPIGroup, coreAPIVersion, "rbac", "users")
```
ccressent: I think this should be `createBasePath(coreAPIGroup, coreAPIVersion, "users")`, ie: drop the `rbac`.

PR #2408, opened 2018-11-25
Move all Resource types to core/v2 package

### LINKED ISSUE #2397: Move types package to api/core/v2
Move the types from `types` to `api/core/v2`. Maintain Go API compatibility by adding type aliases for the existing types. This will keep external consumers from breaking. (I know of at least 1)

Refs #2252

echlebek:
## What is this change?

This commit moves all of Sensu's Resource types to the core/v2
package. This is part of a larger effort to version Sensu's API.

Type aliases have been added in the types package, to avoid
breaking consumers of the types package.

A side effect of this change is that mock.AnythingOfType no longer
works, as it relies on string matching. To work around this
problem, calls to mock.AnythingOfType have been replaced wherever
the call used a type from the types package.

The types.Wrapper machinery has been updated to rely on the v2
package. In the future, the types.Wrapper machinery should be able
to dynamically determine which package to look for types in.

Closes #2397

No. No users or dependants are affected by this change.


backend/daemon/logger.go
```diff
@@ -0,0 +1,8 @@
+package daemon
+
+import "github.com/sirupsen/logrus"
+
+var logger = logrus.WithFields(logrus.Fields{
```
golangcibot: `logger` is unused

echlebek: This file was mistakenly included in the patch.

api/core/v2/check_test.go
```diff
@@ -257,67 +256,3 @@ func TestSortCheckConfigsByName(t *testing.T) {
 		})
 	}
 }
-
-func TestIsSubdued(t *testing.T) {
```
nikkiki: 💀

api/core/v2/handler.go
```diff
@@ -12,7 +12,7 @@ const (
 	HandlerPipeType = "pipe"
 
 	// HandlerSetType represents handlers that groups event handlers, making it
-	// easy to manage groups of actions that should be executed for certain types
+	// easy to manage groups of actions that should be executed for certain v2
```
nikkiki:
```suggestion
	// easy to manage groups of actions that should be executed for certain v2 types
```

api/core/v2/hook.go
```diff
@@ -73,7 +73,7 @@ func (h *HookList) Validate() error {
 
 	if !(CheckHookRegex.MatchString(h.Type) || isSeverity(h.Type)) {
 		return errors.New(
-			"valid check hook types are \"0\"-\"255\", \"ok\", \"warning\", \"critical\", \"unknown\", and \"non-zero\"",
+			"valid check hook v2 are \"0\"-\"255\", \"ok\", \"warning\", \"critical\", \"unknown\", and \"non-zero\"",
```
nikkiki:
```suggestion
			"valid check hook types are \"0\"-\"255\", \"ok\", \"warning\", \"critical\", \"unknown\", and \"non-zero\"",
```

api/core/v2/resource.go
```diff
@@ -0,0 +1,10 @@
+package v2
+
+// Resource represents a Sensu resource.
+type Resource interface {
```
nikkiki: 💯

api/core/v2/typemap.go
```diff
@@ -7,7 +7,7 @@ import (
 	"reflect"
 )
 
-// typeMap is used to dynamically look up data types from strings.
+// typeMap is used to dynamically look up data v2 from strings.
```
nikkiki:
```suggestion
// typeMap is used to dynamically look up data types from strings.
```

api/core/v2/typemap.tmpl
```diff
@@ -7,7 +7,7 @@ import (
   "reflect"
 )
 
-// typeMap is used to dynamically look up data types from strings.
+// typeMap is used to dynamically look up data v2 from strings.
```
nikkiki:
```suggestion
// typeMap is used to dynamically look up data types from strings.
```

PR #2417, opened 2018-11-27
Make API more uniform

### LINKED ISSUE #2398: Re-root API at /api/core/v2

### LINKED ISSUE #2399: Change RBAC path prefix to /api/core/v2
Move RBAC "api group" into `/api/core/v2` prefix.
ex: `/apis/rbac/v2/` would become `/api/core/v2/rbac`
- backend/apid/middlewares
- backend/apid/routers
- cli/client
- types (or api/core/v2 from #2397)

https://github.com/sensu/sensu-go/blob/cc846c027461b1eeb8dcad5454132175fb60b561/backend/apid/middlewares/authorization_attributes.go#L24-L28

https://github.com/sensu/sensu-go/blob/cc846c027461b1eeb8dcad5454132175fb60b561/backend/apid/routers/roles.go#L29

### LINKED ISSUE #2402: Make namespace part of the URL path
This task includes:

- Getting rid of the `namespace` query parameter
- Introducing the right `PathPrefix` for the route

Basically, the routes for all the resources are to be modeled after the routes for the RBAC resources.

Refs #2252

ccressent:
## What is this change?

```
All resources that can be namespaced are now accessed by providing the
namespace as part of the URL path instead of query parameters.

Additionally the whole API has been re-rooted at /api/core/v2, except
for the /auth, /health, and /graphql endpoints.

A typical path now looks like this:
/api/core/v2/namespaces/default/checks/my-check

The LegacyAuthorizationAttributes middleware has been removed.
It was meant to deal with old, unversioned routes. Now that routes have
a more uniform format, we have greatly reduced the special cases in the
routes and we can merge LegacyAuthorizationAttributes and
AuthorizationAttributes.

All the relevant tests have been updated to reflect those changes.
```

Closes #2398 
Closes #2399
Closes #2402

I don't think so: this changes routes for an API that hasn't even been published.

As far as I can tell, the `namespaces` resource seem to need authorization attributes with a namespace set to `default` to work. I'm not sure this is intended.

While getting rid of the `LegacyAuthorizationAttributes` middleware, I had to merge what was remaining of its logic into the `AuthorizationAttributes` middleware. All tests pass, but I encourage reviewers to pay particular attention to this change.

A lot of tests broke; they were mostly fixed by updating the routes they hit.

No documentation change or update is needed: the API is not public yet.


backend/apid/middlewares/authorization_attributes.go
```diff
@@ -56,129 +55,32 @@ func (a AuthorizationAttributes) Then(next http.Handler) http.Handler {
...
+		attrs.ResourceName = vars["id"]
 
-		// Add the user to the attributes
-		if err := getUser(ctx, attrs); err != nil {
-			http.Error(w, err.Error(), http.StatusUnauthorized)
-			return
+		// Namespaces actually belong to the "default" namespace?
+		// Not forcing the namespace to "default" for namespaces leads to
+		// authorization failures.
+		// TODO: investigate
```
echlebek: Oh dear. I had thought that namespaces would belong to themselves. But when I reflect on what that means for the API, I'm not sure it makes any sense.

backend/apid/middlewares/authorization_attributes.go
```diff
@@ -187,6 +89,12 @@ func (a LegacyAuthorizationAttributes) Then(next http.Handler) http.Handler {
 			attrs.Verb = "list"
 		}
 
+		// Add the user to the attributes
+		if err := getUser(ctx, attrs); err != nil {
+			http.Error(w, err.Error(), http.StatusUnauthorized)
```
echlebek: If getUser is hitting the DB then the error should be inspected a bit further before being displayed. If there are I/O errors the server should probably return 503.

PR #2420, opened 2018-11-27
Add metadata to Wrapper

### LINKED ISSUE #2400: Add metadata to types.Wrapper
Add metadata to the Wrapper. Add or update JSON unmarshaling logic in order to copy metadata from the wrapper to the wrapped datatype.

* When metadata is specified in the wrapper, use that metadata in the resulting object.
* When metadata is not specified in the wrapper, but is specified in the object, use the metadata in the object.
* When metadata is specified in both places, merge the metadata, according to the following rules:
  - Prefer attributes from the wrapper over attributes in the object in the case of collisions.
  - Merge labels together. For colliding keys, prefer attributes in the wrapper's labels.
  - Merge annotations in the same way as labels.
* When displaying wrapped resources with sensuctl, display the metadata in the wrapper, not in the object.
* Update sensu docs with the metadata moved into the wrapper scope.
* Add support for multiple API groups. If the API group is missing in the metadata, assume core/v2. EDIT: implement partial support only.

echlebek:
Add metadata to Wrapper
    
    This commit adds TypeMeta and ObjectMeta to the Wrapper datatype.
    It does not remove ObjectMeta from Sensu resources; rather, it
    copies ObjectMeta into Wrappers when they are displayed, and hides
    ObjectMeta in the resource.
    
    When users specify ObjectMeta in Wrappers, the ObjectMeta will be
    stored in the wrapped resource.
    
    This commit also changes most of the resource protos by using the
    gogoproto "face" extension. This was done so that GetObjectMeta
    methods would be generated for resources embedding ObjectMeta.
    
    The wrapper can now resolve types in multiple API groups, but
    right now, only "core/v2" is supported. As a legacy fallback
    mechanism, an empty API version means "core/v2" as well.
    
    Some files have been updated to use the new v2 package naming
    instead of the alias in the types package.
    
    The Resource interface in the types package has been removed in
    favour of an alias to v2.Resource.
    
Closes #2400


api/core/v2/event.proto
```diff
@@ -27,6 +30,9 @@ message Event {
   // Check describes the result of a check; if event is associated to check execution.
   Check check = 3 [(gogoproto.nullable) = true];
 
-  // Metrics ...
+  // Metrics are zero or more Sensu metrics
   Metrics metrics = 4 [(gogoproto.nullable) = true];
+
+  // Metadata contains name, namespace, labels and annotations
+  ObjectMeta metadata = 5 [(gogoproto.embed) = true, (gogoproto.jsontag) = "metadata", (gogoproto.nullable) = false];
```
nikkiki: Metadata is likely never empty, but I noticed the other resources you refactored have `omitempty`. Is this something we want for events as well?

api/core/v2/rbac.proto
```diff
@@ -26,26 +27,29 @@ message Rule {
 
   // ResourceNames is an optional list of resource names that the rule applies
   // to.
-  repeated string resource_names = 3 [(gogoproto.jsontag) = "resourceNames"];
+  repeated string resource_names = 3 [(gogoproto.jsontag) = "resource_names"];
```
nikkiki: 😄

backend/agentd/entity_test.go
```diff
@@ -41,6 +42,9 @@ func TestGetProxyEntity(t *testing.T) {
 		{
 			name: "The event has a proxy entity with a corresponding entity",
 			event: &types.Event{
+				ObjectMeta: v2.ObjectMeta{
```
nikkiki: I suppose this does the same thing as `ObjectMeta: v2.NewObjectMeta("", "default")`, but I like the consistency that this refactor introduces.

nikkiki: And the same with all the other declarations in entity_test.go and authorization_test.go

api/core/v2/event.proto
echlebek: It turns out that `omitempty` has no effect for structs.

backend/agentd/entity_test.go
echlebek: Done!

api/core/v2/rbac.proto
```diff
@@ -68,28 +72,31 @@ message Subject {
...
 // a set of users
 message RoleBinding {
+  option (gogoproto.face) = true;
+  option (gogoproto.goproto_getters) = false;
+
   // Subjects holds references to the objects the Role applies to
   repeated Subject subjects = 1 [(gogoproto.jsontag) = "subjects", (gogoproto.nullable) = false];
 
   // RoleRef references a Role in the current namespace
   RoleRef role_ref = 2 [(gogoproto.jsontag) = "roleRef", (gogoproto.nullable) = false];
```
nikkiki:
```suggestion
  RoleRef role_ref = 2 [(gogoproto.jsontag) = "role_ref", (gogoproto.nullable) = false];
```

nikkiki: Not totally related.. but you fixed `resource_names` in this pr and this could save us some problems down the road.

api/core/v2/rbac.proto
```diff
@@ -68,28 +72,31 @@ message Subject {
...
 // to a user or a set of users
 message ClusterRoleBinding {
+  option (gogoproto.face) = true;
+  option (gogoproto.goproto_getters) = false;
+
   // Subjects holds references to the objects the ClusterRole applies to
   repeated Subject subjects = 1 [(gogoproto.jsontag) = "subjects", (gogoproto.nullable) = false];
 
   // RoleRef references a ClusterRole in the current namespace
   RoleRef role_ref = 2 [(gogoproto.jsontag) = "roleRef", (gogoproto.nullable) = false];
```
nikkiki:
```suggestion
  RoleRef role_ref = 2 [(gogoproto.jsontag) = "role_ref", (gogoproto.nullable) = false];
```

PR #2428, opened 2018-11-28
Add list namespaces rule to system:user cluster role

### LINKED ISSUE #2418: Add the ability to list all namespaces to the system:user cluster role
As per our discussion, in order to prevent weird behaviours in the dashboard, let's add a rule to the `system:user` cluster role in order to allow `list` on `namespaces`.

ccressent:
## What is this change?

This allows any user to implicitly list the namespaces and prevent
issues in the current implementation of the dashboard.

Closes #2418 

No.

No.

No.

No.

### DIFF
--- a/backend/seeds/seeds.go
+++ b/backend/seeds/seeds.go
@@ -115,7 +115,7 @@ func setupClusterRoleBindings(store store.Store) error {
 	}
 
 	// The system:user ClusterRoleBinding grants permission found in the
-	// cluster-admin ClusterRole to any user belonging to the cluster-admins group
+	// system:user ClusterRole to any user belonging to the system:users group
 	systemUser := &types.ClusterRoleBinding{
 		Name: "system:user",
 		RoleRef: types.RoleRef{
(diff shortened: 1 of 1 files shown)


backend/seeds/seeds.go
```diff
@@ -244,6 +244,10 @@ func setupClusterRoles(store store.Store) error {
 				Verbs:     []string{"get", "update"},
 				Resources: []string{types.LocalSelfUserResource},
 			},
+			types.Rule{
+				Verbs:     []string{"list"},
```
jamesdphillips: Should this apply to GET as well?

ccressent: Good question; I think `list` is enough to solve our problem. `get` would let you retrieve specific namespaces objects, which are [pretty uninteresting](https://github.com/sensu/sensu-go/blob/master/api/core/v2/namespace.proto#L15-L19).

grepory: I mentioned this in Slack, but I think we should add get as well.

PR #2429, opened 2018-11-28
Grepory/monitor panic#2406

### LINKED ISSUE #2406: After a monitor times out, if you restart the backend, it will panic
## Current Behavior

When the timer in the monitor expires, it sends an event and the monitor returns. This is the cause of #2153 -- we actually want to reset the timer and send another event when it times out again (based on the entity's KeepaliveTimeout field).

```
{"component":"monitor","level":"debug","msg":"watchMon: timer expired for monitor /sensu.io/monitors/mother.local","time":"2018-11-23T16:43:46-08:00"}
{"component":"monitor","level":"info","msg":"monitor timed out, for /sensu.io/monitors/mother.local, handling failure","time":"2018-11-23T16:43:46-08:00"}
{"component":"keepalived","entity":"mother.local","level":"info","msg":"keepalive timed out, creating keepalive event for entity","time":"2018-11-23T16:43:46-08:00"}
{"check_name":"keepalive","check_namespace":"default","component":"pipelined","entity_name":"mother.local","entity_namespace":"default","level":"debug","msg":"received event","silenced":[],"time":"2018-11-23T16:43:46-08:00","timestamp":1543020226}
{"check_name":"keepalive","check_namespace":"default","component":"pipelined","entity_name":"mother.local","entity_namespace":"default","level":"info","msg":"no handlers available","time":"2018-11-23T16:43:46-08:00"}
```

At this point the watchMon goroutine has returned. If you want long enough, the lease will expire and the monitor key will no longer exist.

If the agent never reconnects, and you stop the backend, it will end up in a panic and refuse to start up:

```
{"component":"backend","error":"error starting keepalived: monitor ttl cannot be 0","level":"fatal","msg":"error executing sensu-backend","time":"2018-11-23T16:50:52-08:00"}
```

Why does this happen, you ask? 

When starting up, keepalived will fetch the failing keepalives

```
/sensu.io/keepalives/default/default/mother.local
{"metadata":{"name":"mother.local","namespace":"default","labels":null,"annotations":null},"time":1543020346}
```

For each failing keepalive, it attempts to create a new monitor. In order to calculate the TTL for the monitor, we look to see how much time is left before the keepalive should expire again.

https://github.com/sensu/sensu-go/blob/948dd2e6dfe3f21209236942dbd534a2238c328a/backend/keepalived/keepalived.go#L158-L166

The monitor spins up a watchMon goroutine, with a TTL of 0, causing watchMon to panic:

https://github.com/sensu/sensu-go/blob/948dd2e6dfe3f21209236942dbd534a2238c328a/backend/monitor/supervisor.go#L180-L183

## Possible Solution

I think this may require some discussion.

Start a backend.
Start an agent.
Kill the agent.
Wait for the keepalive to expire.
Stop the backend.
Start the backend.

## Your Environment

Commit 948dd2e6dfe3f21209236942dbd534a2238c328a

grepory:
## What is this change?

This allows monitor TTLs to be set to 0, which will cause a monitor to immediately fail and send a failing event on startup instead of panicking.

closes #2406 

Added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ uses the same facility as check subdue for handling time windows.
 period of disconnection from the backend.
 - Fixed a bug where logging events without checks would cause a nil panic.
 - Removed the ability to rerun keepalives on the events list page
+- A panic in keepalive/check ttl monitors causing a panic.
 
 ## [2.0.0-beta.8-1] - 2018-11-15
 
--- a/backend/monitor/supervisor.go
+++ b/backend/monitor/supervisor.go
@@ -2,7 +2,6 @@ package monitor
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
 	"os/signal"
@@ -94,9 +93,6 @@ func NewEtcdSupervisor(client *clientv3.Client, h Handler) *EtcdSupervisor {
 // extended. If the monitor's ttl has changed, a new lease is created and the
 // key is updated with that new lease.
 func (m *EtcdSupervisor) Monitor(ctx context.Context, name string, event *types.Event, ttl int64) error {
-	if ttl == 0 {
-		return errors.New("monitor ttl cannot be 0")
-	}
 	key := monitorKeyBuilder.Build(name)
 	// try to get the monitor from the store
 	mon, err := m.getMonitor(ctx, key)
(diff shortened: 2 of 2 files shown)


echlebek:
Good to go after a rebase.

PR #2436, opened 2018-11-28
Ensure password is not overwritten when users are updated

jamesdphillips:
## What is this change?

Ensure existing password digest is not overwritten when a user is updated.

Prior to this change, updating a user's groups would cause their password to be overwritten.

Yes!

Nah!

🤔 

Unlikely.


palourde:
@jamesdphillips Looks like there's some tests failing after this change!

jamesdphillips:
@palourde 🤦‍♂️ I always seem to forget to run the integration tests. I've fixed them now and it should be ready to review.

Issue #2422, opened 2018-11-27
[Web UI] Set up network status flags in apollo cache 


10xjs:
```
Query {
  LocalNetwork {
    offline: Boolean
    retry: Boolean
  }
}

Mutation {
  setLocalNetworkOffline(offline: Boolean): Boolean
  retryLocalNetwork: Boolean
}
```

Issue #2103, opened 2018-09-23
[Web UI] Offline network error state


annaplotkin:
As a user of the Sensu web UI, I would like a clear indicator of when the app cannot connect to the API backend. 

Acceptance criteria:
1. The web UI should detect the inability to connect to the API backend (offline state) by intercepting network errors.
2. Downstream error handling should continue as normal (detecting offline state should silence encountered network errors)
3. If offline state is detected, a below app bar banner should be displayed including information to help solve connection issues.
  a. Should have a message describing that connection has been lost and the app is retrying 'An error has been detected. Retrying to connect.'
  b. Represents a persistent global state. Only one instance is visible for any number of detected failures. Dismissible by the user.
  c. Does not contain a manual "retry" action. This notification is agnostic to any particular network request details.

annaplotkin:
Sent an invite to Sketch file with proposed new design. 

@10xjs @jamesdphillips to review on Monday.

annaplotkin:
https://sketch.cloud/s/Znr4G/qzqJ1d

annaplotkin:
Remaining task is: 

1. Instrument API calls to the /auth endpoint to detect offline state.

PR #2437, opened 2018-11-28
[Web UI] Network error improvements - offline and retry fields

### LINKED ISSUE #2423: [Web UI] Instrument http-link to toggle offline state in apollo cache

dabria:
This is >probably< to be merged in after GA

## What is this change?
Closes #2423 and #2422
Adds LocalNetwork to the Query component. LocalNetwork contains fields for offline and retry. These fields do nothing at this point in time and are an initial step towards #2103

This is to help track the state of the network better and add appropriate errors for the user.

I don't think so, when this functionality is completed in it's entirety then it would make more sense to add an entry. As of now these fields do nothing.

### DIFF
--- a/dashboard/src/apollo/client.js
+++ b/dashboard/src/apollo/client.js
@@ -72,6 +72,18 @@ const createClient = () => {
     `,
   );
 
+  localStorageSync(
+    client,
+    gql`
+      query SynclocalNetworkQuery {
+        localNetwork @client {
+          offline
+          retry
+        }
+      }
+    `,
+  );
+
   return client;
 };
 
--- a/dashboard/src/apollo/client.js
+++ b/dashboard/src/apollo/client.js
@@ -40,7 +40,7 @@ const createClient = () => {
       introspectionLink(),
       stateLink({ cache }),
       authLink({ getClient }),
-      httpLink(),
+      httpLink({ getClient }),
     ]),
   });
 
@@ -75,9 +75,20 @@ const createClient = () => {
   localStorageSync(
     client,
     gql`
-      query SynclocalNetworkQuery {
+      query SyncLocalNetworkQuery {
         localNetwork @client {
           offline
+          retry
+        }
+      }
+    `,
+  );
+
+  localStorageSync(
+    client,
+    gql`
+      query SyncRetryQuery {
+        localNetwork @client {
           retry
         }
       }
--- a/dashboard/src/apollo/client.js
+++ b/dashboard/src/apollo/client.js
@@ -72,29 +72,6 @@ const createClient = () => {
     `,
   );
 
-  localStorageSync(
-    client,
-    gql`
-      query SyncLocalNetworkQuery {
-        localNetwork @client {
-          offline
-          retry
-        }
-      }
-    `,
-  );
-
-  localStorageSync(
-    client,
-    gql`
-      query SyncRetryQuery {
-        localNetwork @client {
-          retry
-        }
-      }
-    `,
-  );
-
   return client;
 };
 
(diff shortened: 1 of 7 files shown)


dashboard/src/apollo/client.js
```diff
@@ -72,6 +72,29 @@ const createClient = () => {
...
+    client,
+    gql`
+      query SyncLocalNetworkQuery {
+        localNetwork @client {
+          offline
+          retry
+        }
+      }
+    `,
+  );
```
jamesdphillips: Do these need to be persisted? Is this to synchronize the state across tabs?

dashboard/src/apollo/httpLink.js
```diff
@@ -1,10 +1,34 @@
...
+const httpLink = ({ getClient }) =>
   new HttpLink({
     uri: "/graphql",
-    fetch: doFetch,
+    fetch: (url, init) =>
+      doFetch(url, init).then(
+        response => {
+          getClient().mutate({
+            mutation,
+            variables: { offline: true },
```
jamesdphillips: Unless I'm mistaking the intent, do we not want to set the state to false if the query was successful?

dashboard/src/apollo/schema/client.graphql
```diff
@@ -29,6 +29,17 @@ extend type Query {
...
+  localNetwork: LocalNetwork
+}
+
+type LocalNetwork {
+  "The current online state of the client."
+  offline: Boolean!
+
+  "If the client is offline and trying to retry the connection."
+  retry: Boolean!
 }
```
jamesdphillips: 👍

dashboard/src/apollo/httpLink.js
```diff
@@ -1,10 +1,34 @@
...
+          });
+          return response;
+        },
+        error => {
+          getClient().mutate({
+            mutation,
+            variables: { offline: false },
```
10xjs: James is right, these values are swapped.

dashboard/src/apollo/httpLink.js
```diff
@@ -1,10 +1,34 @@
...
+      doFetch(url, init).then(
+        response => {
+          getClient().mutate({
+            mutation,
+            variables: { offline: true },
+          });
+          return response;
+        },
+        error => {
+          getClient().mutate({
```
10xjs: There is a check here that we are missing.

We need to test that the error is an instance of `FailedError` in order to isolate connection errors from other http errors.

(See https://github.com/sensu/sensu-go/blob/master/dashboard/src/utils/fetch.js for the heuristic)

dashboard/src/apollo/client.js
10xjs: I think this is unnecessary

dashboard/src/apollo/httpLink.js
dabria: Whoops, yeah that should be false

dashboard/src/apollo/client.js
dabria: Was following an example previously, I've removed them

dashboard/src/apollo/httpLink.js
dabria: Updated!

dashboard/src/apollo/resolvers/localNetwork.js
```diff
@@ -0,0 +1,43 @@
...
+  defaults: {
+    localNetwork: {
+      __typename: "LocalNetwork",
+      offline: false,
+      retry: false,
+    },
+  },
+  resolvers: {
+    Mutation: {
+      retryLocalNetwork: () => (_: mixed, args: mixed, { cache }: Context) => {
```
10xjs: There's an extra function wrapper here `() =>`

dabria: 🦅 👀

Issue #2431, opened 2018-11-28
Backend can lockup when agents crash during registration


nikkiki:
It appears that a simple round robin check request does not get published if more than one entity is subscribed to the check. This occurs in a single and clustered environment.

## Expected Behavior
When a check's round robin field is set to `true`, I would expect the check to execute in a round robin fashion. For all entities subscribed to that check, only one entity will execute it each time it is published.

## Current Behavior
Schedulerd complains that there is an error publishing the round robin check request:
```
Nov 28 03:46:28 ip-10-0-1-171 sensu-backend[3404]: {"check":"round-robin","component":"schedulerd","error":"context deadline exceeded","level":"error","msg":"error publishing round robin check request","time":"2018-11-28T03:46:28Z"}
```
https://github.com/sensu/sensu-go/blob/3860ab4f37d6dc7b16a8650b040637871b10fa41/backend/schedulerd/round_robin_scheduler.go#L79

## Possible Solution
Investigate the `context deadline exceeded` error thrown by `PublishDirect`.
https://github.com/sensu/sensu-go/blob/3860ab4f37d6dc7b16a8650b040637871b10fa41/backend/schedulerd/round_robin_scheduler.go#L60-L61

## Steps to Reproduce (for bugs)
1. create the following check:
```
spec:
  check_hooks: []
  command: echo round-robin
  cron: ""
  handlers: []
  high_flap_threshold: 0
  interval: 10
  low_flap_threshold: 0
  metadata:
    annotations: null
    labels: null
    name: round-robin
    namespace: default
  output_metric_format: ""
  output_metric_handlers: []
  proxy_entity_name:
  publish: true
  round_robin: true
  runtime_assets: []
  stdin: false
  subdue: null
  subscriptions:
  - round-robin
  timeout: 0
  ttl: 0
type: CheckConfig
```
2. start 2 agents subscribed to `round-robin`
3. observe that no events are created for the check `round-robin`
4. observe the errors logged on the backend(s)

## Context
Round robin checks should work with more than one entity subscribed. That is the defining use case for a round robin check.

## Your Environment
* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): https://github.com/sensu/sensu-go/commit/9f7978204c67a1186847ba9ef80492cebe952048
* Installation method (packages, binaries, docker etc.): packages deployed from terraform in aws
* Operating System and version (e.g. Ubuntu 14.04): backend: Ubuntu 16.04

[...]

PR #2445, opened 2018-11-29
Temporarily disable round-robin scheduling

### LINKED ISSUE #2444: Temporarily disable round robin scheduling
We've found an issue with round robin scheduling, where agents that are involved in the round robin can cause a distributed lock to deadlock.

This can only be resolved by the user by manually deleting the key, so we are going to disable the feature until it can be fixed.

grepory:
## What is this change?

Per https://github.com/sensu/sensu-go/issues/2444, we are temporarily
disabling round-robin scheduling. Issues with round-robin
scheduling, https://github.com/sensu/sensu-go/issues/2431 that made
us uncomfortable shipping this enabled as it can lead to a potential
deadlock in the scheduler.

closes #2444

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ uses the same facility as check subdue for handling time windows.
   values coming from the mutator's environment.
 - Metadata from wrappers and resources is now merged, with a preference given to
 the values coming from the wrapper. Labels and annotations are deep-merged.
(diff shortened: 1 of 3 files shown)


grepory:
@nikkiki We log errors in scheduling in two places.  Once in the executor and another inside the scheduler. We would be doing this anyway if there was an error scheduling the check.

PR #2455, opened 2018-11-29
[Web UI] Network error banner

dabria:
## What is this change?
Adds a banner which can update LocalNetwork retry to `true`. 

Will give a user the ability to attempt a reconnect if they lose connection

No

### DIFF
--- a/dashboard/src/components/relocation/RetryConnectionBanner.js
+++ b/dashboard/src/components/relocation/RetryConnectionBanner.js
@@ -1,25 +1,35 @@
 import React from "react";
 import PropTypes from "prop-types";
-import { withApollo } from "react-apollo";
+import gql from "graphql-tag";
+import { withApollo, graphql } from "react-apollo";
 import Button from "@material-ui/core/Button";
 import Banner from "./Banner";
 import retryLocalNetwork from "../../mutations/retryLocalNetwork";
 
 class RetryConnectionBanner extends React.PureComponent {
-  static propTypes = { client: PropTypes.object.isRequired };
+  static propTypes = {
+    client: PropTypes.object.isRequired,
+    data: PropTypes.object.isRequired,
+  };
 
   retryConnection = () => {
-    console.log("this is running");
     retryLocalNetwork(this.props.client);
   };
 
   render() {
+    const { data } = this.props;
     return (
       <Banner
-        message="You've lost network connection."
+        // TODO: make message better by linking to a troubleshooting guide
+        // or use NetworkInformation API: https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation
+        message="You've lost network connection. Check your network connection or the server."
         variant="warning"
         actions={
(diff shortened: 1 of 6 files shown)


dashboard/src/components/relocation/RetryConnectionBanner.js
```diff
@@ -0,0 +1,31 @@
...
+import { withApollo } from "react-apollo";
+import Button from "@material-ui/core/Button";
+import Banner from "./Banner";
+import retryLocalNetwork from "../../mutations/retryLocalNetwork";
+
+class RetryConnectionBanner extends React.PureComponent {
+  static propTypes = { client: PropTypes.object.isRequired };
+
+  retryConnection = () => {
+    console.log("this is running");
```
10xjs: Remove console log

dashboard/src/components/relocation/RetryConnectionBanner.js
```diff
@@ -0,0 +1,31 @@
...
+    retryLocalNetwork(this.props.client);
+  };
+
+  render() {
+    return (
+      <Banner
+        message="You've lost network connection."
+        variant="warning"
+        actions={
+          <Button color="inherit" onClick={() => this.retryConnection()}>
```
10xjs: Set this button to disabled when `retry` is true to provide subtle feedback that clicking did something

dashboard/src/components/relocation/RetryConnectionBanner.js
```diff
@@ -0,0 +1,31 @@
...
+
+  retryConnection = () => {
+    console.log("this is running");
+    retryLocalNetwork(this.props.client);
+  };
+
+  render() {
+    return (
+      <Banner
+        message="You've lost network connection."
```
10xjs: We could potentially provide a better explanation here. Maybe a link to a help document if one exists (would probably have to create that).

There's even an opportunity to use the `NetworkInformation` API (https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation) which @jamesdphillips was interested in.

dashboard/src/components/relocation/RetryConnectionBanner.js
```diff
@@ -1,25 +1,35 @@
...
+        // TODO: make message better by linking to a troubleshooting guide
+        // or use NetworkInformation API: https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation
+        message="You've lost network connection. Check your network connection or the server."
         variant="warning"
         actions={
-          <Button color="inherit" onClick={() => this.retryConnection()}>
+          <Button
+            color="inherit"
+            onClick={() => this.retryConnection()}
+            disabled={data.localNetwork.offline}
```
10xjs: `data.localNetwork.retry`

dabria: Oh frick haha. Thanks

PR #2456, opened 2018-11-29
[Web UI] Set up GlobalAlert component

10xjs:
## What is this change?

This component provides a mechanism for declaring alert banners with life-cycles coupled to the current app state.

Supports the upcoming offline alert banner.

No

No

No

No

### DIFF
--- a/dashboard/src/components/util/GlobalAlert.js
+++ b/dashboard/src/components/util/GlobalAlert.js
@@ -11,14 +11,6 @@ class GlobalAlert extends React.PureComponent {
     data: PropTypes.object.isRequired,
   };
 
-  static defaultProps = {
-    data: {
-      localNetwork: {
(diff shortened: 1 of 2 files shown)


dashboard/src/components/util/GlobalAlert.js
```diff
@@ -0,0 +1,44 @@
...
+import PropTypes from "prop-types";
+import { graphql } from "react-apollo";
+import gql from "graphql-tag";
+
+import BannerSink from "/components/relocation/BannerSink";
+import Banner from "/components/relocation/Banner";
+
+class GlobalAlert extends React.PureComponent {
+  static propTypes = {
+    data: PropTypes.object.isRequired,
```
dabria: We have a default declaration below, should this be optional?

PR #2463, opened 2018-11-30
API URL

### LINKED ISSUE #2461: Dashboard is broken in the presence of TLS
When the dashboard is configured to use TLS, hostname verification fails, rendering it unusable.

From @palourde:

https://github.com/sensu/sensu-go/blob/f8b13b06b8783821918878b5b91293515e091d71/backend/apid/apid.go#L96. We pass “url”, which is defined at line https://github.com/sensu/sensu-go/blob/f8b13b06b8783821918878b5b91293515e091d71/backend/apid/apid.go#L88. It uses the “listen” host IP

Suggested solution:
We probably just need to make a distinction between the address APId listens on and the advertised one, just like etcd does

palourde:

## What is this change?

It adds a new configuration attribute in the backend for specifying the API URL the dashboard will use.

Closes https://github.com/sensu/sensu-go/issues/2461

Added

Nope

Naming things

Already done

Closes #2461


backend/dashboardd/dashboardd.go
```diff
@@ -221,8 +220,6 @@ func newBackendProxy(host string, port int, TLS *types.TLSOptions) (*httputil.Re
 
 	// Configure TLS
 	if TLS != nil {
-		target.Scheme = "https"
-
```
jamesdphillips: I wonder if we can remove `cfg.InsecureSkipVerify = true` from L227 now? I think this is an artifact from when the dashboard statically defined `localhost` as the APIUrl.

https://github.com/sensu/sensu-go/blob/add72d2f5c001c67ff94ce440b3d642a479795dc/backend/dashboardd/dashboardd.go#L227

palourde: We are now using it EVERYWHERE instead 🤣

jamesdphillips: not sure if I want to laugh or cry

backend/apid/apid.go
```diff
@@ -78,27 +70,23 @@ func New(c Config, opts ...Option) (*APId, error) {
 	var tlsConfig *tls.Config
 	var err error
 	if c.TLS != nil {
+		// TODO(palourde): We should avoid using the loopback interface
+		c.TLS.InsecureSkipVerify = true
```
jamesdphillips: Should we make a copy for specifically for the GraphQL service? Otherwise we might be allowing it on the `apid`.

echlebek: InsecureSkipVerify is only for clients.

palourde: Good point, I applied `InsecureSkipVerify` to `tlsConfig` instead; I believe it only applies to client but I prefer to avoid any potential problem.

nikkiki:
giphy

CHANGELOG.md
```diff
@@ -7,12 +7,13 @@ Versioning](http://semver.org/spec/v2.0.0.html).
...
 ## Unreleased
 
-### [5.0.0] - 2018-10-30
+### [5.0.0] - 2018-11-30
 
 ### Added
 - Add the `etcd-advertise-client-urls` config attribute to sensu-backend
 - Support for multiple API versions added to sensuctl create
 - Support for metadata added to wrapped resources (yaml, wrapped-json)
+- Added the backend configurationa attributes `api-listen-address` & `api-url`.
```
nikkiki:
```suggestion
- Added the backend configuration attributes `api-listen-address` & `api-url`.
```

palourde:
This is me trying to fix multiple things at once

CHANGELOG.md
```diff
@@ -34,6 +35,8 @@ uses the same facility as check subdue for handling time windows.
 - Metadata from wrappers and resources is now merged, with a preference given to
 the values coming from the wrapper. Labels and annotations are deep-merged.
 - Round-robin scheduling has been temporarily disabled.
+- The dashboard now uses the `api-url` configuration attribute to connect to the
```
nikkiki: Maybe add that `api-host` and `api-port` have been deprecated?

CHANGELOG.md
```diff
@@ -53,6 +56,10 @@ period of disconnection from the backend.
 - Monitors are now properly namespaced in etcd.
 - Updating a users groups will no longer corrupt their password
 
+### Breaking Changes
+- The backend configuration attributes `api-host` & `api-port` have been
```
nikkiki: 💯

CHANGELOG.md
echlebek: Should this say "no longer uses"?

PR #2267, opened 2018-10-31
[Web UI] Adopt namespaces

jamesdphillips:
## What is this change?

Update web UI to use new simpler namespaces.

## Why is this change necessary?

Closes #2176.

## Does your change need a Changelog entry?

Might be covered under the existing changelog entry.

PR #2482, opened 2018-12-04
[Web UI] Sort order updates

### LINKED ISSUE #2481: [Web UI] Silences page ascending/descending order
Silences page lists silenced entries in descending (backwards alphabetical) order.

## Expected Behavior
I would expect the silences to appear in ascending (alphabetical) order.

## Current Behavior
Silences appear in descending order (backwards alphabetical) order.
screen shot 2018-12-03 at 3 39 52 pm

## Possible Solution
This was believed to be fixed. I'm not sure how to go about investigating this again.

1. Open silences page
2. Make sure resources are populated with a few silenced entries

## Context
The order which we display resources should be consistent across the dashboard.

## Your Environment
* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): 419c199
* Installation method (packages, binaries, docker etc.): packages via terraform on was
* Operating System and version (e.g. Ubuntu 14.04): Ubuntu 16

jamesdphillips:
## What is this change?

Alleviates two issues with sorting of resources in the web application.

1. Due to a bad merge conflict, the fix introduced in #2235 was overwritten. This branch restores functionality so that silences are now sorted in the appropriate ascending and descending order requested.
1. Improves sorting of events when an event's last ok timestamp are the same. Both the entity and check name are now used to arbitrate.
  Without this fix the events list can shuffle a great deal, making it difficult to select some list items.

Reintroduces #2235 
Closes #2481

Yup!

Regression was likely introduced in #2267


api/core/v2/event.go
```diff
@@ -150,18 +150,41 @@ func EventsByLastOk(es []*Event) sort.Interface {
...
+
+	if ai == bi {
+		return 0
+	} else if ai < bi {
+		return 1
+	}
+	return -1
+}
+
+func cmpByCheckName(a, b *Event) int {
```
golangcibot: `cmpByCheckName` is unused (from `deadcode`)

PR #2502, opened 2018-12-07
Fix a bug where keepalive events only fire once

### LINKED ISSUE #2135: Keepalive events stop processing after the first occurrence
I've been experimenting with alerting on keepalive events the past few days and have managed to get them to alert via a handler. However, I've not managed to figure out how to get them to alert more than once or get past warning state. 

@palourde suggested I raise this bug after discussing it with him on slack.

## Expected Behavior
Keepalive events should continue to run and be handled after the agent has stopped running.

## Current Behavior
I see `event filtered` every 20 seconds whilst the agent is running. When I stop the agent I see `sending event to handler` after 120 seconds, then there are no more events processed until I start the agent back up.

1. set up keepalive handler
```
[root@phbtest ~]# sensuctl handler info keepalive
{
  "type": "Handler",
  "spec": {
    "name": "keepalive",
    "type": "set",
    "timeout": 60,
    "handlers": [
      "cst",
      "cstv2"
    ],
    "filters": null,
    "env_vars": null,
    "environment": "default",
    "organization": "default"
  }
}
[root@phbtest ~]# sensuctl handler info cstv2
{
  "type": "Handler",
  "spec": {
    "name": "cstv2",
    "type": "pipe",
    "command": "/usr/bin/handler-cst-sensu2",
    "timeout": 0,
    "handlers": null,
    "filters": [
      "is_incident"
    ],
    "env_vars": null,
    "environment": "default",
    "organization": "default"
  }
}
[root@phbtest ~]# cat /usr/bin/handler-cst-sensu2
#!/usr/bin/env python3
import datetime
import sys

if __name__ == "__main__":
    event = sys.stdin.read()
    with open('/tmp/sensu-v2handler-dump', 'w') as f:
        f.write("{}: {}".format(str(datetime.datetime.now()), event))
```
2. turn off the agent
3. watch the logs, after the initial occurrence of the failing state the keepalive does not handle again until the agent is restarted. 

## Context
Keepalives do not behave in the same way as they did under Sensu 1.x.

## Your Environment
```
[root@phbtest ~]# rpm -qa sensu-*
sensu-cli-2.0.0~nightly+20180926-1.x86_64
sensu-backend-2.0.0~nightly+20180924-1.x86_64
sensu-agent-2.0.0~nightly+20180926-1.x86_64
```
Scientific Linux 7.5

echlebek:
## What is this change?

This commit fixes a bug where keepalives and check TTL failure
events only fire once. The fix makes failing events occur in
a loop until the next successful event is observed.

This strategy only works for the lifetime of the backend that is
handling the keepalive. If the backend is taken offline, failure
events will not continue to be emitted. A more general solution is
needed, but this is better than what we have now.

Closes #2135 

Yes

A new issue should be filed after this PR is merged, as this PR does not solve the problem between backend restarts.


backend/monitor/supervisor.go
```diff
@@ -174,27 +189,28 @@ func (m *EtcdSupervisor) getMonitor(ctx context.Context, key string) (*monitor,
...
 		defer cancel()
+		ctx, cancelHandler := context.WithCancel(context.Background())
 		for {
 			select {
 			case wresp := <-responseChan:
 				for _, ev := range wresp.Events {
 					if ev.Type == mvccpb.DELETE {
-						failureHandler()
-						return
+						go failureHandler(ctx, mon.ttl)
```
ccressent: What does `mon.ttl` represent?
For keepalives, is that the keepalive interval?
For other objects, like checks, is that their timeout?

echlebek: `mon.ttl` is the time to live for both checks and keepalives.

echlebek:
@nikkiki when the PUT event is observed (line 211), the context that's governing the failure loop is cancelled (line 213). The failure loop will terminate because the select case on line 144 will fire. Then, the next event will be a passing check or keepalive event.

backend/monitor/supervisor.go
ccressent: I was confused so I've re-read the docs for what timeout and ttl mean for Sensu. Now I think it makes sense to me. Thanks!

PR #2504, opened 2018-12-07
Do not build or test the backend for windows

echlebek:
The backend is unsupported on Windows. Also disables 386 tests for
Windows.


brieken:
Does this mean there will also be no 386 windows agent?

Issue #2514, opened 2018-12-10
Windows agent provides incorrect event status when running a PowerShell command


jamesdphillips:
From #2494
Also reported in #2513

> Checks based on PowerShell fail in a manner that provides an error code that the Sensu front end does not like. The back end continues to work without issue, albeit the check status is incorrect.

While the front-end issue was solved in #2503; the agent issue persists and should be resolved.

## Expected Behavior

PowerShell based Sensu checks run without issue.

## Current Behavior

Sensu Agent does not appear to capture the exit codes properly when running PowerShell based Sensu checks.

## Possible Solution

Requires further investigation.

## Steps to Reproduce (for bugs)

1. Create a check with the following command: `powershell -command "exit 0"`.
2. Wait for check to run.
3. Observe the result of the check. `sensuctl event info ...`

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): Windows Agent + Powershell
* Installation method (packages, binaries, docker etc.): n/a
* Operating System and version (e.g. Ubuntu 14.04): Windows

PR #2516, opened 2018-12-10
Change the entity's status field to use uint type

jamesdphillips:
## What is this change?

Updates the GraphQL service's entity status field to use the uint type.

* Windows exit codes are represented as unsigned 32bit integers.
* Discovered while attempting to replicate #2514 

Likely covered by the last entry.

### DIFF
--- a/backend/apid/graphql/entity.go
+++ b/backend/apid/graphql/entity.go
@@ -94,7 +94,7 @@ func (r *entityImpl) Related(p schema.EntityRelatedFieldResolverParams) (interfa
 }
 
 // Status implements response to request for 'status' field.
-func (r *entityImpl) Status(p graphql.ResolveParams) (int, error) {
+func (r *entityImpl) Status(p graphql.ResolveParams) (interface{}, error) {
 	src := p.Source.(*types.Entity)
 	client := r.factory.NewWithContext(p.Context)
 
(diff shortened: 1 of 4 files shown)


backend/apid/graphql/entity.go
```diff
@@ -94,7 +94,7 @@ func (r *entityImpl) Related(p schema.EntityRelatedFieldResolverParams) (interfa
 }
 
 // Status implements response to request for 'status' field.
-func (r *entityImpl) Status(p graphql.ResolveParams) (int, error) {
+func (r *entityImpl) Status(p graphql.ResolveParams) (interface{}, error) {
```
ccressent: Why not one of the `uint` types instead of `interface{}`?

jamesdphillips: While I agree this would be much better, this method is implementing the [`EntityStateFieldResolver` interface](https://github.com/sensu/sensu-go/pull/2516/files#diff-15797ce5fbb9917807da2d97768c8266R80) found in the `graphql/schema` package. A current limitation of the code generator is that it does not have any way of resolving custom defined types and as such falls back to an `interface{}` when generating it's field resolver interfaces. In the future it would be nice to add support for mapping GraphQL types to Go types / primitives, but I think it is out of scope for this change.

ccressent: Thanks for the clarification regarding the code generator!

PR #2519, opened 2018-12-11
Fix issues around command execution in the agent

### LINKED ISSUE #2514: Windows agent provides incorrect event status when running a PowerShell command

jamesdphillips:
## What is this change?

Fixes a couple issues around command execution in the client.

* All environment variables from the parent process are now included when executing checks.
* Environment variables are now joined with the proper platform specific separator. 
* Timeout timer is now stopped when the command immediately bombs out.
* Command arguments are no longer escaped on Windows. See [blog post](http://www.josephspurrier.com/prevent-escaping-exec-command-arguments-in-go/) & [commit](https://github.com/golang/go/commit/f18a4e9609aac3aa83d40920c12b9b45f9376aea).

Closes #2514

Yes.

I am not particular familiar with Windows.


util/environment/environment.go
```diff
@@ -1,33 +1,39 @@
...
 
 import (
+	"os"
 	"sort"
 	"strings"
 )
 
-// MergeEnvironments merges env2 into env1, overwriting any existing variable
-// in env1, except for the "special" variables PATH, CPATH and LD_LIBRARY_PATH.
+var pathSeparator = string(os.PathListSeparator)
```
ccressent: I would call this variable `pathListSeparator`. The first thing that came to my mind when reading `pathSeparator` was `/` vs `\` as opposed to the intended `:` vs `;`.

util/environment/environment.go
```diff
@@ -1,33 +1,39 @@
...
 // The "special" variables PATH, CPATH and LD_LIBRARY_PATH are merged by
 // prepending the value from env2 to the value in env1, effectively giving
 // priority to the value from env2.
 //
 // The expected format for an environment variable definition is VAR=VALUE. Any
 // malformed environment variable definition will be discarded by the merge.
-func MergeEnvironments(env1, env2 []string) []string {
-	e1 := toMap(env1)
-	e2 := toMap(env2)
+func MergeEnvironments(ea []string, es ...[]string) []string {
```
ccressent: We need to rephrase the comment above `MergeEnvironments()` to explain how it merges the arguments "right to left" in succession, instead of just `env2` into `env1`.

util/environment/environment_test.go
```diff
@@ -1,3 +1,5 @@
+// +build !windows
```
ccressent: Tests can probably be made multi-platform by using `os.PathListSeparator`, just like you did in `environment.go`.

agent/check_handler.go
```diff
@@ -89,10 +90,13 @@ func (a *Agent) executeCheck(request *v2.CheckRequest) {
 		return
 	}
 
+	// Prepare environment variables
+	env := environment.MergeEnvironments(syscall.Environ(), assets.Env(), check.EnvVars)
```
ccressent: I would use `os.Environ()` rather than `syscall.Environ()`, even if right now `os.Environ()` simply calls `syscall.Environ()`.

The `syscall` package is meant to be a low level interface; using the `os` package is preferred in most cases.

command/command.go
```diff
@@ -140,12 +140,13 @@ func (e *ExecutionRequest) Execute(ctx context.Context, execution ExecutionReque
...
 			}
 		})
+		defer timer.Stop()
 	}
 
 	if err := cmd.Start(); err != nil {
 		// Something unexpected happended when attepting to
 		// fork/exec, return immediately.
-		return nil, err
+		return resp, err
```
ccressent: Usually, when a function returns `val, err`, if `err != nil` one is not expected to use `val` at all. Why return `resp, err` instead of `nil, err`?

amdprophet:
We're updating the calls to `syscall.Environ()` with `os.Environ()` as the Golang documentation for `syscall` states:

```
The primary use of syscall is inside other packages that provide a more portable interface to the system, such as "os", "time" and "net". Use those packages rather than this one if you can.
```

util/environment/environment.go
```diff
@@ -55,8 +55,10 @@ func toMap(s []string) map[string]string {
 				m[split[0]] = ""
 			}
 		case 2:
+			// On Windows environment variables can be mixed case, eg. Path or PATH.
+			key := strings.ToUpper(split[0])
```
ccressent: That's breaking POSIX!

command/command.go
jamesdphillips: Hm, I feel ya. Unfortunately the agent currently expects, regardless of err, that the response is always present. Meaning we'll encounter a panic if the process fails to start.

I'll revert this change for now and we can handle it another time.

ccressent: I see. The real fix then, in my opinion, is to change the agent to properly handle the fact that when there is an error, it can't also always expect a meaningful response.

util/environment/environment.go
jamesdphillips: Yeah, I stumbled on the documentation. 😞

util/environment/environment_test.go
```diff
@@ -3,11 +3,20 @@
...
 import (
+	"fmt"
+	"os"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 )
 
+// Makes platform complaint list of values
```
ccressent: s/complaint/compliant :)

jamesdphillips: haha, I have a complaint or two about the Windows platform.

jamesdphillips:
@ccressent if you don't mind give this one more once over.

util/environment/environment.go
```diff
@@ -56,7 +56,7 @@ func toMap(s []string) map[string]string {
 			}
 		case 2:
 			// See _windows.go
-			key := strings.ToUpper(split[0])
+			key := coerceKey(split[0])
```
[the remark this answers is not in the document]
ccressent: `runtime.GOOS` is defined as `const`, so not mutable.
I don't think it changes much in term of testing, but I could be wrong. I'll go with whatever you think is best!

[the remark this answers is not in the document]
jamesdphillips: I'm going to stick with it, at this point it is the path of least resistance and I don't feel strongly enough one way or the other.

jamesdphillips:
Thanks for sticking with me @ccressent. 🎉

ccressent:
Thanks for _your_ work!

PR #2546, opened 2018-12-17
Update backend state-dir to /var/lib/sensu/sensu-backend

amdprophet:

## What is this change?

Update backend state-dir to `/var/lib/sensu/sensu-backend`.

We've made this change to the default in packaging but never made a change to the default in sensu backend.

Nope!

Nope!

### DIFF
--- a/backend/etcd/etcd.go
+++ b/backend/etcd/etcd.go
@@ -26,7 +26,7 @@ import (
 
 const (
 	// StateDir is the base path for Sensu's local storage.
-	StateDir = "/var/lib/sensu"
+	StateDir = "/var/lib/sensu/sensu-backend"
 	// ClusterStateNew specifies this is a new etcd cluster
 	ClusterStateNew = "new"
 	// EtcdStartupTimeout is the amount of time we give the embedded Etcd Server


backend/etcd/etcd.go
```diff
@@ -26,7 +26,7 @@ import (
 
 const (
 	// StateDir is the base path for Sensu's local storage.
-	StateDir = "/var/lib/sensu"
+	StateDir = "/var/lib/sensu/sensu-backend"
```
jamesdphillips: I believe we'll need to update `backend/cmd/start.go`

amdprophet:
@annaplotkin this change has the possibility to affect installations not using packaging, configuration management, or a config file with `state-dir` specified (`state-dir` is specified by default in our example configs).

We may want some messaging that we've updated the backend's default state directory to reflect what the default is when using a configuration file (`/var/lib/sensu` is now `/var/lib/sensu/sensu-backend`).

backend/etcd/etcd.go
amdprophet: There is little chance that this will affect anyone and will not conclude in any data loss.

The only scenario in which this will affect anyone is if a user is not using a config file (or has removed the `state-dir` line from the backend config) and has manually created `/var/lib/sensu` themselves and applied the appropriate ownership/permissions changes.

Our current behaviour is causing issues for users as packaging works as intended, but the backend binary does not.

amdprophet: The worst case scenario is that a user will need to either:
1. Specify the state dir when starting the backend with the `-d` flag.
2. Add `state-dir` to their existing backend.yml.
3. Move the contents of `/var/lib/sensu` into `/var/lib/sensu/sensu-backend`.

jamesdphillips: A user adding a configuration file doesn't necessarily mean they also configured the state dir, does it?

palourde: I doubt this PR was ever tested because this line doesn't actually have any effect and should be removed. 

It is only referred in the [`NewConfig` function](https://github.com/sensu/sensu-go/blob/321ff41fe8170b37ae4673cce1df2cf29733f702/backend/etcd/etcd.go#L63), but it gets [immediately replaced](https://github.com/sensu/sensu-go/blob/e089a61c0c710dac81f81a9fbcee5beadf4ead5d/backend/backend.go#L68) with whatever config was provided in the configuration. The [default value](https://github.com/sensu/sensu-go/blob/321ff41fe8170b37ae4673cce1df2cf29733f702/backend/cmd/start.go#L268) is indeed `/var/lib/sensu`, which is provided by the [path util](https://github.com/sensu/sensu-go/blob/321ff41fe8170b37ae4673cce1df2cf29733f702/util/path/path.go#L53).

portertech:
1. Docker deployments use the binary default and run as root
2. Package deployments are working due to example configuration (not ideal)

We need to "break" one of these for consistency.

backend/etcd/etcd.go
amdprophet: Good catch!

portertech:
b3b7e022-390e-11e4-8e2e-be8e525f837f

echlebek:
nervous

Issue #2201, opened 2018-10-19
Adhoc checks with assets cannot find assets


nikkiki:
Check with asset executes with no problem:
```
ubuntu@ip-10-0-1-114:~$ sensuctl event info staging-agent-0 check-with-asset
=== staging-agent-0 - check-with-asset
Entity:      staging-agent-0
Check:       check-with-asset
Output:      you have installed an asset, kudos!
Status:      0
History:     0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0
Silenced:    true
Silenced By: *:check-with-asset
Timestamp:   2018-10-19 19:46:02 +0000 UTC
```
Execute adhoc check request on check with asset:
```
ubuntu@ip-10-0-1-114:~$ sensuctl check execute check-with-asset
Issued
```
Check the event data for the adhoc check, script from asset is not found:
```
ubuntu@ip-10-0-1-114:~$ sensuctl event info staging-agent-0 check-with-asset
=== staging-agent-0 - check-with-asset
Entity:      staging-agent-0
Check:       check-with-asset
Output:      sh: 1: echo-asset.sh: not found
Status:      127
History:     0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,127
Silenced:    true
Silenced By: *:check-with-asset
Timestamp:   2018-10-19 19:46:20 +0000 UTC
```
Check resolves itself on next execution:
```
ubuntu@ip-10-0-1-114:~$ sensuctl event info staging-agent-0 check-with-asset
=== staging-agent-0 - check-with-asset
Entity:      staging-agent-0
Check:       check-with-asset
Output:      you have installed an asset, kudos!
Status:      0
History:     0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,127,0,0
Silenced:    true
Silenced By: *:check-with-asset
Timestamp:   2018-10-19 19:46:32 +0000 UTC
```

PR #2570, opened 2018-12-20
Update build request function for adhoc check requests

nikkictl:

## What is this change?

Updates the build request function for adhoc check requests. This fixes a bug where adhoc check's would not retrieve asset dependencies.

Closes https://github.com/sensu/sensu-go/issues/2201

Yas.

Nah.

Nah.

We good.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 ## Unreleased
 
 ### Fixed
+- Fixed a bug where `sensuctl edit` was not removing the temp file it created.
 - Fixed a bug where adhoc checks were not retrieving asset dependencies.
 
 ## [5.1.0] - 2018-12-18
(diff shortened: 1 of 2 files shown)


backend/schedulerd/executor.go
```diff
@@ -301,3 +279,46 @@ func processCheck(ctx context.Context, executor Executor, check *types.CheckConf
...
+	request := &types.CheckRequest{}
+	request.Config = check
+
+	ctx := types.SetContextFromResource(context.Background(), check)
+
+	// Guard against iterating over assets if there are no assets associated with
+	// the check in the first place.
+	if len(check.RuntimeAssets) != 0 {
+		// Explode assets; get assets & filter out those that are irrelevant
+		assets, err := executor.getStore().GetAssets(ctx)
```
echlebek: The only thing executor is used for here is getStore(). It would be better to pass the store into the function. Then, there would be no need to include getStore() in the methods of executor.

PR #2579, opened 2018-12-30
Substitutes inconsistent messages in sensuctl to an OK message (#2552)

### LINKED ISSUE #2552: Inconsistent confirmation messages in sensuctl
Sometimes when a resource is created, sensuctl prints `OK` as a confirmation message, other times, it prints `Created`. This is a minor, stylistic inconsistency that I noticed when writing automated tests. However, its probably best to print the same confirmation message regardless of the resource.

ex.
```
$ sensuctl asset create test-asset --namespace default --url https://s3-us-west-2.amazonaws.com/sensu-omnibus-cache/assets/slack.tar --sha512 839af73bf03cc94aba11d80e149fd46e265a3a5fcb13e6b11f348adc1a0199e0bd791a5dc84371ee12bd06d1c97e11d075945bcff5569ed6f706af60cbeefd69
$ OK
```
```
$ sensuctl role create test-role --verb=create --resource=check
$ Created
```

The same is true for the delete confirmation.

ex.
```
$ sensuctl check delete test-check --skip-confirm
$ OK
```
```
$ sensuctl cluster-role delete test-cluster-role --skip-confirm
$ Deleted
```

arnaldomf:
Signed-off-by: Arnaldo Mendonca <arnaldomf@user.noreply.github.com>

## What is this change?

This change fixes the messages for sensuctl command, returning an OK message to all succeeded commands.

Following the open issue #2552 stated by @nikkiki "its probably best to print the same confirmation message regardless of the resource."

Closes #2552 

Yes

I was in doubt about the cluster command, if the message with the member's ID returned by the update/delete was really necessary as it is already a cli's argument, so I change the message to just OK in those two cases.

Some test were missing, so I added some simple tests just to force an OK message from the commands changed.

No


cli/commands/cluster/member-remove.go
```diff
@@ -30,7 +30,7 @@ func MemberRemoveCommand(cli *cli.SensuCli) *cobra.Command {
 			if _, err := cli.Client.MemberRemove(id); err != nil {
 				return fmt.Errorf("error removing cluster member: %s", err)
 			}
-			fmt.Fprintf(cmd.OutOrStdout(), "Removed member %x from cluster\n", id)
+			fmt.Fprintln(cmd.OutOrStdout(), "OK")
```
nikkiki: I know you mentioned you were unsure about the cluster commands. I think it's safe to leave those as is, for now.
```suggestion
 			fmt.Fprintf(cmd.OutOrStdout(), "Removed member %x from cluster\n", id)
```

cli/commands/cluster/member-update.go
```diff
@@ -33,7 +33,7 @@ func MemberUpdateCommand(cli *cli.SensuCli) *cobra.Command {
 				return fmt.Errorf("error updating cluster member: %s", err)
 			}
 
-			fmt.Fprintf(cmd.OutOrStdout(), "Updated member with ID %x in cluster\n", id)
+			fmt.Fprintf(cmd.OutOrStdout(), "OK")
```
nikkiki: I know you mentioned you were unsure about the cluster commands. I think it's safe to leave those as is, for now.
```suggestion
 			fmt.Fprintf(cmd.OutOrStdout(), "Updated member with ID %x in cluster\n", id)
```

cli/commands/clusterrole/create_test.go
```diff
@@ -30,3 +34,18 @@ func TestCreateCommandRunEClosureMissingArgs(t *testing.T) {
 	assert.NotEmpty(out)
 	assert.Error(err)
 }
+
+func TestCreateCommandRunEClosureWithDeps(t *testing.T) {
```
nikkiki: Love that you wrote tests for missing coverage 💚

arnaldomf:
Hi @nikkiki , thanks for the feedback! The changes wont be that hard, I'll try again :)

Just to be sure, the messages should be "Created", "Deleted", "Updated", and so on?

nikkiki:
@arnaldomf thanks for your understanding! Yes: "Created" for all create commands, "Deleted" for all delete commands, and "Updated" for all update commands. Any other commands or subcommands that are more specific than that, we can leave as is!

arnaldomf:
Hi @nikkiki! I just changed the messages, should I open another PR or can I use this one?
Also, I couldn't find a way to test the update commands in case of success.

nikkiki:
@arnaldomf This PR or a new one.. whichever is easiest for you! Yes, the update command is only available in interactive mode, which we ourselves have not found a good way to write a stable, automated test for. Thanks for your hard work, I'll have another look at this today!

nikkiki:
I changed a setting in our repository so that these mandatory checks will run for PRs from forked repos.. looks like CI is running now 😄

echlebek:
I've merged master into this branch in the interests of keeping it up to date.

We will need to target a minor release for this changeset; since it changes the output of `sensuctl` commands, it could potentially break downstream tools that expect certain output from `sensuctl`.

portertech:
yesjacknicholson

Issue #2550, opened 2018-12-18
Updating a check breaks contract with schedule


nikkiki:
When an interval check is created (POST), it executes immediately after a calculated splay, and then on every interval thereafter.

When a cron check is created (POST), it does not calculate a splay, and is scheduled according to the cron contract.

https://github.com/sensu/sensu-go/blob/33fbc584feb7bdddfe3a3df3b7fe46213a289c93/backend/schedulerd/check_watcher.go#L115-L119

When an interval check is updated (PUT), it executes immediately, and then on every interval thereafter. If the interval time did not change, this breaks the interval contract and an extra check execution occurs.

When a cron check is updated (PUT), it executes immediately, and then scheduled according to the cron contract. Regardless if the cron string changed, this breaks the cron contract and an extra check execution occurs.

https://github.com/sensu/sensu-go/blob/33fbc584feb7bdddfe3a3df3b7fe46213a289c93/backend/schedulerd/check_watcher.go#L120-L125

In order to have parity with Sensu Classic, the schedule - if unchanged by the POST action - should persist, and the check should not fire until it's next scheduled time.

Rather than the interrupt immediately firing a check execution, we should special case it for different schedule switch states.
https://github.com/sensu/sensu-go/blob/33fbc584feb7bdddfe3a3df3b7fe46213a289c93/backend/schedulerd/check_scheduler.go#L142-L162

nikkiki:
Also, props to whoever wrote these comments. This code is super easy to follow!

PR #2587, opened 2019-01-04
Fix a bug in the check scheduler where a check would immediately fires if updated

nikkiki:

## What is this change?

Consolidates the `CheckScheduler` struct. Cleans up the check scheduler `schedule` function. Persists the check's timer if the schedule has not been updated (per 1.x parity). Special cases the interrupt issued by the check water to determine if the timer must be reset.

Closes https://github.com/sensu/sensu-go/issues/2550.

Yas.

Nah.

Nah.

We good.

### DIFF
--- a/backend/schedulerd/check_scheduler.go
+++ b/backend/schedulerd/check_scheduler.go
@@ -2,6 +2,7 @@ package schedulerd
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/sensu/sensu-go/backend/messaging"
 	"github.com/sensu/sensu-go/backend/store"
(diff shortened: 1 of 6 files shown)


backend/schedulerd/check_scheduler.go
```diff
@@ -145,18 +105,25 @@ func (s *CheckScheduler) start() {
...
 			return
 		case <-s.interrupt:
+			if err := s.refreshCheckFromStore(); err != nil {
+				s.logger.WithError(err).Error("unable to retrieve check in check scheduler")
+				return
+			}
+			// if a schedule change is detected, restart the timer
+			if s.toggleSchedule() {
+				timer.Stop()
+				defer s.Start()
```
golangcibot: Error return value of `s.Start` is not checked (from `errcheck`)

nikkiki: Why does `s.Start()` return an error to begin with? Unless I'm missing something, couldn't we simply not return anything..
https://github.com/sensu/sensu-go/blob/33fbc584feb7bdddfe3a3df3b7fe46213a289c93/backend/schedulerd/check_scheduler.go#L111-L115

nikkiki: Option 2 would be logging the error, although due to the nature of `Start()` that would be unreachable.
```
defer func() {
  if err := s.Start(); err != nil {
    log.Error(err)
  }
}()
```

echlebek: I think we should just change Start() to not return an error. I don't think CheckScheduler is implementing any meaningful interfaces.

portertech:
yesjacknicholson

PR #2588, opened 2019-01-04
Use package liveness in keepalived

### LINKED ISSUE #2530: Keepalive events will stop occurring on backend restart
Although PR #2502 fixes the problems with Check TTL and Keepalives in the common case, the solution is in-memory only, and will fail when the backend's state is refreshed (i.e., restarted).

While monitors can transition between backends via etcd, the failure loop that causes backends to emit events relies on in-memory state.

This is related to #2135 and users who rely on durable keepalive and check TTL failure behaviour should be aware of this.

echlebek:
## What is this change?

This commit makes use of package liveness in keepalived, in order
to achieve durable keepalives. With this change applied, failing
keepalives will continue firing even if one or more backends is
lost. (Assuming etcd quorum is intact)

Closes #2530 

Yes

## Testing Performed
Manual QA performed. Both agents and backends were removed, to ensure that keepalive failure events would be issued as expected. Keepalives continue to propagate even when the backend the agent was previously connected to is terminated. Also, keepalive events will continue to fire even after a cluster restart.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 ### Fixed
 - Fixed a bug where `sensuctl edit` was not removing the temp file it created.
 - Fixed a bug where adhoc checks were not retrieving asset dependencies.
+- Fixed a bug where check updates would cause the check to immediately fire.
 
 ## [5.1.0] - 2018-12-18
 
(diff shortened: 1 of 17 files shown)


echlebek:
Yeah, I need to retool how the concurrency works here. It's a bug in the test.

echlebek:
Opening a new PR instead.

Issue #2511, opened 2018-12-08
Metric extraction: a bad line should not abort extraction


portertech:
We ran into trouble extracting metrics from Kube State Metrics, using the [sensu-prometheus-collector plugin](https://github.com/sensu/sensu-prometheus-collector) and InfluxDB line output.

## Expected Behavior
The Sensu Agent will do its best to extract metrics from check output, logging invalid lines.

## Current Behavior
The Sensu Agent will abort check output metric extraction if it encounters an invalid line.

https://github.com/sensu/sensu-go/blob/a9f7289efe6107a8ba7c84fc571d2ae6dadf280c/agent/transformers/influx_db.go#L50

echlebek:
Half enhancement, half bug IMO :)

PR #2591, opened 2019-01-07
Fixes a bug where a bad line in check output would abort metric extraction

nikkiki:

## What is this change?

- Parse functions accept an Event as a parameter.
- Parse functions log errors rather than returning them.
- Transformers do not abort metric extraction when a bad line is encountered.
- Fixes bugs associated with https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices.

Closes https://github.com/sensu/sensu-go/issues/2511

Yas.

Nah.

Nah.

We good.


nikkiki:
@portertech 

giphy

PR #2593, opened 2019-01-09
Fix benchmark, and update for clustered testing

echlebek:

## What is this change?

Fixes the benchmark - it was broken, because using the same cachedir for all agents was causing a deadlock.

I've also updated the tool to support multiple backends for agent connection.

IMO no.

### DIFF
--- a/benchmark/cmd/main.go
+++ b/benchmark/cmd/main.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"os"
 	"os/signal"
+	"strings"
 	"sync"
 	"syscall"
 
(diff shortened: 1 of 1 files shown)


benchmark/cmd/main.go
```diff
@@ -26,16 +27,19 @@ func main() {
...
 
 	for i < *count {
 		name := uuid.New().String()
+		backend := backends[i%len(backends)]
 
 		cfg := agent.NewConfig()
 		cfg.API.Host = agent.DefaultAPIHost
 		cfg.API.Port = agent.DefaultAPIPort
-		cfg.CacheDir = path.SystemCacheDir("sensu-agent")
+		cfg.CacheDir = path.SystemCacheDir(fmt.Sprintf("sensu-agent-%s", name))
```
nikkiki: Good catch 👍

PR #2595, opened 2019-01-10
Add keepalive check output

### LINKED ISSUE #2585: Request to have check.output populated for keepalive events
#2081 added the check.output field.  However, I would like to see this field populated such that handlers that use the output (e.g. sensu-email-handler) can make use of it.  For example, in the case of using sensu-email-handler, the body of the generated email is made up of the check output, meaning it is empty for keepalive events.  I would like to see it populated with some simple text such as "Keeaplive failure for entity_name".
## Current Behavior

check.output is empty for keepalive events
## Possible Solution

To capture it during a failure event add something like this to backend/keepalived/keepalived.go:
```
        // this is a real keepalive event, emit it.
        event := createKeepaliveEvent(e)
        event.Check.Status = 1
        event.Check.Output = "Keepalive failed for " + entity.Name
```
However, I'm not sure how/where to populate similar for passing event.

1.  Add sensu-email-handler defined handler as a handler for keepalive handler
2.  Generate a keepalive event
3.  Body of email is empty

## Context

Would be a nice to have as other handlers may wish to use it as well.
## Your Environment

sensu-backend version 5.1.0#b2ea9fc, build b2ea9fcdb21e236e6e9a7de12225a6d90c786c57, built '2018-12-18T21:31:11+0000'
* Installation method (packages, binaries, docker etc.):  packages
* Operating System and version (e.g. Ubuntu 14.04):  Ubuntu 16.04

nixwiz:
## What is this change?

Adds check.output for keepalive events.

Closes #2585 
Yes

One thing that might stand out is that for passing keepalives, I use the absolute (clock on the wall) time as opposed to using the relative time for failures.  This is because when viewing keepalive events (e.g. in the dashboard) if I used relative time it would report the last seen as zero seconds all the time, which is not useful information.  I use the relative time (including the configured timeout) in failures to match 1.x functionality.

No.

I didn't see any references to keepalive events in the documentation where this would need to be updated.  The only visible change would be to events output when surfaced through a handler or seen in the dashboard.

### DIFF
--- a/backend/keepalived/keepalived.go
+++ b/backend/keepalived/keepalived.go
@@ -2,6 +2,7 @@ package keepalived
 
 import (
 	"context"
+	"fmt"
 	"sync"
 	"time"
 
@@ -316,6 +317,7 @@ func (k *Keepalived) handleUpdate(e *types.Event) error {
 	}
 	event := createKeepaliveEvent(e)
 	event.Check.Status = 0
+	event.Check.Output = fmt.Sprintf("Keepalive last sent from %s at %s", entity.Name, time.Unix(entity.LastSeen, 0).String())
 
 	return k.bus.Publish(messaging.TopicEventRaw, event)
 }
@@ -340,6 +342,7 @@ func (k *Keepalived) HandleFailure(e *types.Event) error {
 	// this is a real keepalive event, emit it.
 	event := createKeepaliveEvent(e)
 	event.Check.Status = 1
+	event.Check.Output = fmt.Sprintf("No keepalive sent from %s for %v seconds (>= %v)", entity.Name, time.Now().Unix()-entity.LastSeen, event.Check.Timeout)
 
 	if err := k.bus.Publish(messaging.TopicEventRaw, event); err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


nikkiki:
I changed a setting in our repository so that these mandatory checks will run for PRs from forked repos.. looks like CI is running now 😄

nixwiz:
> I changed a setting in our repository so that these mandatory checks will run for PRs from forked repos.. looks like CI is running now 😄

Looks like CI failed, but doesn't appear to be due to my code.

backend/keepalived/keepalived.go
```diff
@@ -329,6 +330,9 @@ func (k *Keepalived) dead(key string, prev liveness.State, leader bool) {
 		lager.Error(err)
 		return
 	}
+	event := createKeepaliveEvent(e)
```
nikkiki: I believe this might cause a compile time error (see L382).

nixwiz: But, those are outside of the changes I made as part of this commit.

nikkiki: Yeah, I totally understand that. I believe we had a long lived branch for some of those keepalive improvements (that's on us). We just need to make sure master will compile before we merge. We can take care of this for you if you'd like 😄

nixwiz: Can you, please?

PR #2601, opened 2019-01-11
Use package liveness in keepalived

### LINKED ISSUE #2530: Keepalive events will stop occurring on backend restart

### LINKED ISSUE #2622: Keepalives conflict with entity deregistration
## Expected Behavior
Deregisterer entities should not experience keepalive failure events, and should not be listed as entities after they have deregistered.

## Current Behavior
Entities can appear after deregistration.

## Context
Reported by pfak in the community slack.

echlebek:
## What is this change?

This commit makes use of package liveness in keepalived, in order
to achieve durable keepalives. With this change applied, failing
keepalives will continue firing even if one or more backends is
lost. (Assuming etcd quorum is intact)

Closes #2530 
Closes #2622 

Yes


agent/agent.go
```diff
@@ -173,6 +173,9 @@ func (a *Agent) Run() error {
 	if err := v2.ValidateName(a.config.AgentName); err != nil {
 		return fmt.Errorf("invalid agent name: %v", err)
 	}
+	if timeout := a.config.KeepaliveTimeout; timeout > 0 && timeout < 5 {
```
nikkiki: Can a KeepaliveTimeout be 0 seconds? That would essentially disable failing keepalives right?

echlebek: The minimum value is 5 seconds, and we don't support disabling keepalives currently.

backend/keepalived/keepalived.go
```diff
@@ -298,54 +304,114 @@ func createRegistrationEvent(entity *types.Entity) *types.Event {
...
-	entity.LastSeen = e.Timestamp
+	lager = lager.WithFields(logrus.Fields{"entity": name, "namespace": namespace})
+	lager.Info("entity is alive")
+}
 
-	if err := k.store.UpdateEntity(ctx, entity); err != nil {
-		logger.WithError(err).Error("error updating entity in store")
-		return err
+func (k *Keepalived) dead(key string, prev liveness.State, leader bool) {
+	lager := logger.WithFields(logrus.Fields{
```
nikkiki: I propose we change _all_ instances of `logger` to `lager` 🍻

agent/agent.go
echlebek: After chatting with Nikki I realized that the condition should only check for `timeout < 5`

backend/keepalived/keepalived.go
echlebek: But then my strategy for not overriding the logger variable would break :)

nikkiki: giphy

backend/keepalived/integration_test.go
```diff
@@ -55,23 +60,36 @@ func TestKeepaliveMonitor(t *testing.T) {
...
+		},
 		Entity:    entity,
 		Timestamp: time.Now().Unix(),
 	}
 
+	if err := store.UpdateEvent(ctx, keepalive); err != nil {
+		t.Fatal(err)
+	}
+
+	factory := liveness.EtcdFactory(context.Background(), client)
```
nikkiki: Can we use the same context that the store uses on L52, or is there another reason we have independent contexts?

echlebek: That context carries extra information, and since it's being created by a function call in another package, I'd prefer to keep them separate.

nikkiki: That makes sense, was just curious!

PR #2610, opened 2019-01-17
Add Authentication Providers

palourde:
## What is this change?

This change introduces the notion of authentication providers and move all the current authentication logic into a `basic` authentication provider.

In details:
- Authentication providers have two main methods; `Authenticate` & `Refresh`. The former is used to authenticate a user with a username/password combinaison and the latter is used to renew the user claims (mainly groups) without the password.
- The JWT claims now contain a `ProviderClaims` struct, which informs us of the provider used to authenticate the user and its username (which might differs from the username used by Sensu).
	- This is useful for renewing the user claims during the renewal of an access token
- Added the notion of an `Authenticator` to apid.
	- It has the knowledge of all enabled authentication providers and it contains the business logic to determine which providers should be used for authentication (e.g. use them in order or only use a specific one)
- Refactored the `Token` store and renamed some of its methods, so it can now accept a list of tokens/claims to allow or revoke.
- The authentication providers now return the user claims instead of a user struct; this struct is specific to the `basic` authentication provider, therefore we shouldn't have to use this "limiting" struct between the providers and the tokens.
- Refactored the error messages and log entries of the `AuthenticationRouter` so we don't leak information via the API about the reasons behind an unauthorized request, and instead provide the logs with all the required information.

Required by https://github.com/sensu/sensu-enterprise-go/issues/21

Added!

Nope!

Nope!

Nope!

### DIFF
--- a/backend/apid/routers/authentication.go
+++ b/backend/apid/routers/authentication.go
@@ -44,7 +44,8 @@ func (a *AuthenticationRouter) login(w http.ResponseWriter, r *http.Request) {
 	// Authenticate against the provider
 	claims, err := a.authenticator.Authenticate(r.Context(), username, password)
 	if err != nil {
-		logger.WithField("user", username).Error("invalid username and/or password")
+		logger.WithError(err).WithField("user", username).
+			Error("invalid username and/or password")
 		http.Error(w, http.StatusText(http.StatusUnauthorized), http.StatusUnauthorized)
 		return
 	}
--- a/backend/apid/routers/authentication.go
+++ b/backend/apid/routers/authentication.go
@@ -118,7 +118,7 @@ func (a *AuthenticationRouter) logout(w http.ResponseWriter, r *http.Request) {
 
 	// Remove the access & refresh tokens from the access list
 	if err := a.store.RevokeTokens(accessClaims, refreshClaims); err != nil {
-		logger.WithField("user", refreshClaims.Subject).Errorf(
+		logger.WithError(err).WithField("user", refreshClaims.Subject).Errorf(
 			"could not revoke tokens IDs %q & %q", accessClaims.Id, refreshClaims.Id,
 		)
 		http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
--- a/backend/apid/routers/authentication.go
+++ b/backend/apid/routers/authentication.go
@@ -2,6 +2,7 @@ package routers
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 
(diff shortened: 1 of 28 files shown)


palourde:
Note to myself: rebase master once https://github.com/sensu/sensu-go/pull/2278 is merged.

backend/apid/routers/authentication.go
```diff
@@ -33,69 +37,44 @@ func (a *AuthenticationRouter) login(w http.ResponseWriter, r *http.Request) {
...
 
 	// Authenticate against the provider
-	user, err := a.store.AuthenticateUser(r.Context(), username, password)
+	claims, err := a.authenticator.Authenticate(r.Context(), username, password)
 	if err != nil {
-		logger.WithField(
-			"user", username,
-		).WithError(err).Error("invalid username and/or password")
-		http.Error(w, "Request unauthorized", http.StatusUnauthorized)
+		logger.WithField("user", username).Error("invalid username and/or password")
```
nikkiki: Does the error in this case contain sensitive information?

backend/apid/routers/authentication.go
```diff
@@ -124,128 +102,118 @@ func (a *AuthenticationRouter) logout(w http.ResponseWriter, r *http.Request) {
...
 	}
 
 	// Remove the access & refresh tokens from the access list
-	tokensToRemove := []string{accessClaims.Id, refreshClaims.Id}
-	if err := a.store.DeleteTokens(refreshClaims.Subject, tokensToRemove); err != nil {
-		err = fmt.Errorf("could not remove the access and refresh tokens from the access list: %s", err)
-		logger.WithField("user", refreshClaims.Subject).Error(err)
-		http.Error(w, err.Error(), http.StatusInternalServerError)
+	if err := a.store.RevokeTokens(accessClaims, refreshClaims); err != nil {
+		logger.WithField("user", refreshClaims.Subject).Errorf(
```
nikkiki: Same thing here, can we log the error `WithError(err)` or does that have sensitive info as well?

backend/apid/routers/authentication.go
```diff
@@ -124,128 +102,118 @@ func (a *AuthenticationRouter) logout(w http.ResponseWriter, r *http.Request) {
...
-	// Make sure to the 'system:users' group is present
-	user.Groups = append(user.Groups, "system:users")
-
-	// Remove the old access token from the access list
-	if err := a.store.DeleteTokens(accessClaims.Subject, []string{accessClaims.Id}); err != nil {
-		err = fmt.Errorf("could not remove the access token from the access list: %s", err)
-		logger.WithField("user", refreshClaims.Subject).Error(err)
-		http.Error(w, err.Error(), http.StatusInternalServerError)
-		return
+	// Ensure backward compatibility by filling the provider claims if missing
```
nikkiki: 👍

backend/authentication/providers/LICENSE
```diff
@@ -0,0 +1,20 @@
+Copyright (c) 2017 Sensu Inc.
```
nikkiki: Do we need to update the copyright year since we've made changes to this code in 2019?

backend/authentication/providers/basic/LICENSE
```diff
@@ -0,0 +1,20 @@
+Copyright (c) 2017 Sensu Inc.
```
nikkiki: Same thing here, 2017->2019?

backend/authentication/providers/providers.go
```diff
@@ -0,0 +1,93 @@
...
+// Provider represents an abstracted authentication provider
+type Provider interface {
+	Authenticate(ctx context.Context, username, password string) (*v2.Claims, error)
+	GetName() string
+	Refresh(ctx context.Context, providerClaims v2.ProviderClaims) (*v2.Claims, error)
+	Type() string
+}
+
+// ID returns a unique identifier for a given provider
+func ID(p Provider) string {
```
nikkiki: Would it make more sense to define this function in `backend/authentication/providers/basic/basic.go` as `func (p *Provider) ID() string`?

backend/apid/routers/authentication.go
nikkiki: Was just curious why we weren't logging the specific error. If you don't think its necessary, that's fine too!

palourde: Oh my bad, I misread the code; we _should_ be logging the error, my bad!

backend/apid/routers/authentication.go
```diff
@@ -128,15 +108,25 @@ func (a *AuthenticationRouter) test(w http.ResponseWriter, r *http.Request) {
...
+		}
+	} else {
+		err = errors.New("basic provider is disabled")
 	}
+
+	logger.WithField(
+		"user", username,
+	).WithError(err).Info("invalid username and/or password")
+	http.Error(w, "Request unauthorized", http.StatusUnauthorized)
+	return
```
golangcibot: redundant return statement (from `megacheck`)

PR #2615, opened 2019-01-17
[Web UI] Updated toolbar and hover buttons for dark mode

### LINKED ISSUE #2301: "Live" button in dashboard difficult to see in dark mode
## Expected Behavior
The "Live" button should probably be in white

## Current Behavior
The button/text is black:

dashboard-dark

## Possible Solution
Change to white in dark mode

dabria:
## What is this change?
Closes #2301 
Changes the parent of the buttons to use `color: theme.palette.text.secondary`. This allows the colour of the text to switch when dark mode is enabled. In addition I updated the hovering buttons in the list view to have consistent colours as well.

From this:
screen shot 2019-01-15 at 2 07 53 pm screen shot 2019-01-15 at 2 08 08 pm

To this:
screen shot 2019-01-15 at 2 07 34 pm screen shot 2019-01-15 at 2 07 20 pm

The updated hover buttons:
screen shot 2019-01-17 at 2 55 32 pm

These two colours look the same, but they can actually be configured. Setting the colour to primary instead of secondary gives a pure white and black depending if dark mode is enabled. However, it seemed a bit harsh, so I went with secondary as it seems to match the side-nav links in dark mode nicely. 

The buttons were hard to read before.

Added one.

### DIFF
--- a/dashboard/src/components/partials/EventsList/EventsListItem.js
+++ b/dashboard/src/components/partials/EventsList/EventsListItem.js
@@ -133,14 +133,13 @@ class EventListItem extends React.Component {
                   <ResolveMenuItem
                     iconOnly
                     disabled={event.status === 0}
-                    description="Resolve event."
                     onClick={this.props.onClickResolve}
                   />
                 </ToolbarMenu.Item>
                 <ToolbarMenu.Item id="re-run" visible="never">
                   <QueueMenuItem
+                    iconOnly
                     disabled={event.check.name === "keepalive"}
-                    title="Re-run Check"
                     onClick={this.props.onClickRerun}
                   />
                 </ToolbarMenu.Item>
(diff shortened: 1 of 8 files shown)


dashboard/src/components/partials/EventsList/EventsListItem.js
```diff
@@ -133,14 +133,12 @@ class EventListItem extends React.Component {
...
                     iconOnly
                     disabled={event.status === 0}
-                    description="Resolve event."
                     onClick={this.props.onClickResolve}
                   />
                 </ToolbarMenu.Item>
                 <ToolbarMenu.Item id="re-run" visible="never">
                   <QueueMenuItem
                     disabled={event.check.name === "keepalive"}
-                    title="Re-run Check"
```
dabria: These were removed because they are already defined as "description" inside their respective ToolbarMenuItems

PR #2653, opened 2019-01-30
Add support for check output truncation

### LINKED ISSUE #2641: Support check output truncation
## Expected Behavior
I can push large amounts of metrics through Sensu in check output.

## Current Behavior
All check output is stored in etcd, limiting the amount of metrics per event to ~1 MB.

## Possible Solution
1.x Sensu supports check output truncation, that is enabled by default when checks contain metrics to be extracted.

Docs: https://docs.sensu.io/sensu-core/1.6/reference/checks/#check-attributes (search "truncate" on this page)

Output truncation can also be useful even if there are no metrics extracted.

## Context
Investigating an issue from a potential client, who are having difficulty coping with the amount of output their checks produce.

echlebek:
## What is this change?
    
This commit adds support for check output truncation. Truncation
can occur in either the backend or the agent, according to the
max_output_size and discard_output check configurations.

Closes #2641 

Included

Documentation will require an update for these features.

Unit tests have been expertly added.


echlebek:
It's well covered by the unit tests, but I'm happy to add some to the crucible as well!

PR #2676, opened 2019-02-04
Add support for wrapped resources storage

palourde:
## What is this change?

This PR adds support for resources stored using the wrapper.

Specifically:
- The API authenticator now identifies providers using their name only
- `sensuctl create` & `sensuctl edit` have been modified so they send resources as `types.Wrapper` to the API for all API groups except `core/v2`, where it still sends `corev2.Resource`.
- It exports `KeyBuilderFn` so it can be re-used in the enterprise store

Required by https://github.com/sensu/sensu-enterprise-go/pull/201

Nope!

Nope!

Not required for now.

Not yet.


echlebek:
@nikkiki no, it only works with wrapped resources. The difference is that now we send the wrapper to the HTTP server for non-core APIs.

PR #2684, opened 2019-02-04
[Web UI] Bump global polling interval

jamesdphillips:
## What is this change?

Bump global polling interval from 2.5s to 5s.

Queries can be expensive with large sets of data.

sure.

nada.

nada.

nada.

nada.


jamesdphillips:
> Can I ask the reason for this change? In addition what are the implications of it?

@dabria my apologies, I didn't really provide enough context in the description of the ticket.

This ticket does two things:

1. Consolidates polling values throughout all pages into one place; this makes it easier to tweak in the future and opens an avenue for making it configurable in the future. 
1. Settles on five seconds as the default value.
    * This is "bumping" the value only in so much as it is making the value we already use on the events page the standard everywhere.
    * It does mean that a few pages will be a bit "slower" but no more than the events page is already.
    * 5s still feels pretty responsive and avoids punishing the server.

PR #2685, opened 2019-02-05
Ensure dead entities are buried when appropriate

### LINKED ISSUE #2682: Keepalive and check TTL switches not removed on entity deletion
## Expected Behavior
After entities are deleted, switches no longer fire.

## Current Behavior
Switches continue to fire. This results in garbage occupying the DB, and log spam from removed entities.

## Possible Solution
Bury switches if it is discovered that entities do not exist in `dead` callbacks.

echlebek:
## What is this change?

This commit fixes bugs in keepalived and eventd, where entities
that have been removed by the user will continue to be partially
registered.

This had the effect of using up extra space in etcd, and writing
out extra logs, but was otherwise inconsequential.

Closes #2682 

Added

Test case added.

### DIFF
--- a/backend/eventd/eventd.go
+++ b/backend/eventd/eventd.go
@@ -219,6 +219,7 @@ func (e *Eventd) handleMessage(msg interface{}) error {
 		}
 		return e.handleUpdate(event)
 	} else {
+		// The check TTL has been disabled, there is no longer a need to track it
 		if err := switches.Bury(context.TODO(), switchKey); err != nil {
 			// It's better to publish the event even if this fails, so
 			// don't return the error here.
@@ -229,7 +230,7 @@ func (e *Eventd) handleMessage(msg interface{}) error {
 	return e.bus.Publish(messaging.TopicEvent, event)
 }
 
-func (e *Eventd) alive(key string, prev liveness.State, leader bool) {
+func (e *Eventd) alive(key string, prev liveness.State, leader bool) (bury bool) {
 	lager := logger.WithFields(logrus.Fields{
 		"status":          liveness.Alive.String(),
 		"previous_status": prev.String()})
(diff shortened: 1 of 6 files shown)


backend/eventd/eventd.go
```diff
@@ -270,26 +273,37 @@ func (e *Eventd) dead(key string, prev liveness.State, leader bool) {
...
-		return
+		return false
+	}
+
+	ctx := store.NamespaceContext(context.Background(), namespace)
+
+	// The entity has been deleted, and so there is no reason to track check
+	// TTL for it anymore.
+	if ent, err := e.store.GetEntityByName(ctx, entity); err == nil && ent == nil {
+		fmt.Println("LOL NO ENTITY:", entity)
```
ccressent: I'm assuming this `Println()` was only meant for testing? :)

backend/eventd/integration_test.go
```diff
@@ -75,6 +76,12 @@ func TestEventdMonitor(t *testing.T) {
 	event.Check.Interval = 1
 	event.Check.Ttl = 5
 
+	ctx := otherTestutil.ContextWithNamespace("default")(context.Background())
```
ccressent: How many ways do we have to create a context with a namespace key? :(
https://github.com/sensu/sensu-go/blob/master/backend/store/namespace.go#L25-L28
https://github.com/sensu/sensu-go/blob/master/testing/testutil/context.go#L28-L34

backend/eventd/eventd.go
nikkiki: It’s a better debug print statement than I usually go for 😂

backend/eventd/integration_test.go
echlebek: About as many `package testutil`s I think

PR #2688, opened 2019-02-05
[Web UI] Include proxy request configuration

### LINKED ISSUE #2668: [Web UI] Add missing fields to check details page
Part of: #2594
We need to add the following to the check details page:

Check Summary card:
- [ ] proxy_requests 
- [ ] output_metric_format
- [ ] output_metric_handlers

We've omitted a few from this list as more discussion is needed around them:
#2663 #2664 

Subdue is dependent on the redesign of said feature. #1955 

## Expected Behavior
Displays all important data

## Current Behavior
Doesn't display all important data

## Possible Solution
Add the data

jamesdphillips:
## What is this change?

Include proxy request configuration details on the check details page.

Closes #2668 

yes.

no.

no.

nada.

nada.

### DIFF
--- a/dashboard/src/components/partials/CheckDetailsContainer/CheckDetailsConfiguration.js
+++ b/dashboard/src/components/partials/CheckDetailsContainer/CheckDetailsConfiguration.js
@@ -238,7 +238,7 @@ class CheckDetailsConfiguration extends React.PureComponent {
 
                 <DictionaryEntry fullWidth={!!check.proxyRequests}>
                   <DictionaryKey>Proxy Requests</DictionaryKey>
-                  <DictionaryValue>
+                  <DictionaryValue explicitRightMargin>
                     <Maybe value={check.proxyRequests} fallback="None">
                       {val => (
                         <CodeBlock>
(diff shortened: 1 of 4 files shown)


dashboard/src/components/partials/CheckDetailsContainer/CheckDetailsConfiguration.js
```diff
@@ -348,25 +364,6 @@ class CheckDetailsConfiguration extends React.PureComponent {
             </Grid>
           </Grid>
         </CardContent>
-
-        {Object.keys(check.extendedAttributes).length > 0 && (
```
dabria: 🎉Throw 👏 it 👏 out 👏

dashboard/src/components/partials/CheckDetailsContainer/CheckDetailsConfiguration.js
```diff
@@ -264,9 +280,9 @@ class CheckDetailsConfiguration extends React.PureComponent {
                   </DictionaryValue>
                 </DictionaryEntry>
 
-                <DictionaryEntry>
+                <DictionaryEntry fullWidth={check.envVars.length > 0}>
                   <DictionaryKey>ENV Vars</DictionaryKey>
-                  <DictionaryValue scrollableContent>
```
dabria: I see that we've sort of replaced scrollableContent with fullWidth, but isn't having the values scroll still desirable in this context?

jamesdphillips: as per IRL discussion; would require a larger refactor which doesn't make sense at this juncture.

PR #2627, opened 2019-01-22
Better TLS support for sensuctl

ccressent:
## What is this change?

Add support for the following TLS related options to `sensuctl`:
- `--trusted-ca-file`
- `--insecure-skip-tls-verify`
- `--cert-file`
- `--key-file`

These have the same meaning as their counterparts in `sensu-backend` and `sensu-agent`.

## Why is this change necessary?

Closes #2541.
Closes #2575.

## Does your change need a Changelog entry?

Yes, added.   

## Have you reviewed and updated the documentation for this change? Is new documentation required?

The documentation needs to be updated to reflect the new options.
**I haven't done that yet.**

## Does this change require a new test case?

A few tests for various TLS configurations can be added to the QA checklist.
**I haven't done that yet.**

PR #2694, opened 2019-02-06
Use TLS client config when configuring GraphQL service

jamesdphillips:
Use TLS client config when configuring GraphQL service; otherwise skip verify option is not copied.

### DIFF
--- a/backend/apid/apid.go
+++ b/backend/apid/apid.go
@@ -73,12 +73,12 @@ func New(c Config, opts ...Option) (*APId, error) {
 		Authenticator:       c.Authenticator,
 	}
 
-	var tlsConfig *tls.Config
+	var tlsClientConfig *tls.Config
 	var err error
 	if c.TLS != nil {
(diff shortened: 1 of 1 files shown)


backend/apid/apid.go
```diff
@@ -73,12 +73,12 @@ func New(c Config, opts ...Option) (*APId, error) {
 		Authenticator:       c.Authenticator,
 	}
 
-	var tlsConfig *tls.Config
+	var tlsClientConfig *tls.Config
 	var err error
 	if c.TLS != nil {
 		// TODO(palourde): We should avoid using the loopback interface
 		c.TLS.InsecureSkipVerify = true
-		tlsConfig, err = c.TLS.ToTLSConfig()
```
jamesdphillips: NOTE: As of #2627 the `InsecureSkipVerify` option is no longer copied when fabricating a `TLS.Config`. See: [api/core/tls.go](https://github.com/sensu/sensu-go/blob/master/api/core/v2/tls.go#L31-L54).

Issue #2548, opened 2018-12-18
TLS is inflexible and limited


ccressent:
The `TLSOptions` type is too simple and "server centric". It needs to be more
granular and closer to the standard library type to allow for things like client
certificate request and verification.

The `tls.Config` type from the `crypto/tls` standard library is what is
actually underneath `TLSOptions`, and that type is much richer and granular.

At the moment, I see the following issues with sensu-go:

#### No support for client certificates

The backend never requests client certificates and cannot be configured to do
so, be it for API clients (sensuctl) or websocket clients (sensu-agent). The consequence 
is that both `--cert-file` and `--key-file` cannot have any real effect when given 
to sensu-agent and sensuctl.

See `tls.Config.ClientAuth`. Untils client authentication can be configured,
there is no point in having the `--cert-file` and `--key-file` flags for
sensu-agent and sensuctl.

#### `--trusted-ca-file` is currently useless when given to sensu-backend

This bundle of CA certificates is ultimately put in `tls.Config.RootCAs`, which
is only used by clients to verify the certificate presented by the server end of
the TLS connection. To verify the certificate presented by the client end, that
certificate bundle needs to be put in `tls.Config.ClientCAs` instead.

See the `crypto/tls` package documentation:
`tls.Config.RootCAs` is "the set of root certificate authorities that clients
use when verifying server certificates".
`tls.Config.ClientCAs` is "the set of root certificate authorities that servers
use if required to verify a client certificate".

#### `--insecure-skip-tls-verify` is currently useless when given to the backend

Since the backend doesn't check client certificates, `--trusted-ca-file` or even
the system's certificate bundle (used by default) is never needed.

#### No reason to couple the various TLS flags

I don't see a reason for `--trusted-ca-file` to need both `--cert-file` and
`--key-file` to also be specified (see [here](https://github.com/sensu/sensu-go/blob/c4373a6aece87679d225829160c6fd19fb14c20a/backend/cmd/start.go#L178-L198)).

Not specifying `--trusted-ca-file` will make `tls.Config.RootCAs` nil, which in
turn will make the library use the system's bundle of root CAs (on non windows
OSes).

Presenting a certificate (`--cert-key` and `cert-file`) is orthogonal to
verifying a certificate presented to you (`--trusted-ca-file`).

## Possible Solution

* Rework the various TLS flags and options to fully take advantage of what the standard library gives us for free.
* Decouple the flags as appropriate

## Context

This was discovered while working on #2300.
[...]

PR #2700, opened 2019-02-07
closes #2657 - strengthen our TLS security posture

### LINKED ISSUE #2657: TLS: don't accept version < 1.2
## Expected Behavior
Insecure TLS versions should not be allowed to negotiate a connection.

## Current Behavior
TLS 1.0 clients can negotiate a connection.

## Possible Solution
Tell the TLS server to use 1.2 or greater.

## Context
Security audit

gbolo:
## What is this change?

Enforces stronger TLS settings

To mitigate known TLS vulnerabilities. The existing TLS settings were affected by:
```
LUCKY13 (CVE-2013-0169)  - uses cipher block chaining (CBC) ciphers with TLS.
SWEET32 (CVE-2016-2183, CVE-2016-6329)  - VULNERABLE, uses 64 bit block ciphers
BEAST (CVE-2011-3389) - TLS1: ECDHE-RSA-AES128-SHA
                          ECDHE-RSA-AES256-SHA
                          AES128-SHA AES256-SHA
                          ECDHE-RSA-DES-CBC3-SHA
                          DES-CBC3-SHA 
```

it's transparent (app functionality is no changed), but perhaps it should be mentioned?

no, but if a reviewer is not super familiar with TLS, I offer the explanation(s) below:
- min version of TLS 1.2 is being offered in order to ONLY support the best possible ciphers
- `DefaultCipherSuites` was modified to remove CBC ciphers and non-TLS_1.2 ciphers. This list now ONLY includes *(perfect) forward secrecy* ciphers
- `tlsCurvePreferences` was added to prefer EC curve X25519. this is an opensource curve with no NSA involvement.
- `+build go1.8` was added to prevent older go versions from attempting to compile this package (since some curves and ciphers are not available in previous versions)

no

no changes to doc required

probably not

### DIFF
--- a/api/core/v2/tls.go
+++ b/api/core/v2/tls.go
@@ -1,6 +1,3 @@
-// +build go1.8
-// enforce go 1.8+ just so we can support stronger TLS settings (X25519 curve and chacha_poly suites)
-
 package v2
 
 import (
@@ -29,46 +26,42 @@ var (
 	// curve reference: http://safecurves.cr.yp.to/
 	tlsCurvePreferences = []tls.CurveID{
 		// this curve is a non-NIST curve with no NSA influence. Prefer this over all others!
-		// this curve required go 1.8+
 		tls.X25519,
-		// These curves are provided by NIST; prefer in descending order
-		tls.CurveP521,
+		// These curves are provided by NIST; optimal order
 		tls.CurveP384,
 		tls.CurveP256,
+		tls.CurveP521,
 	}
 )
 
-// ToTLSConfig outputs a tls.Config from TLSOptions
-//
-// ToTLSConfig should only be used for server TLS configuration.
-func (t *TLSOptions) ToTLSConfig() (*tls.Config, error) {
+// ToServerTLSConfig should only be used for server TLS configuration. outputs a tls.Config from TLSOptions
+func (t *TLSOptions) ToServerTLSConfig() (*tls.Config, error) {
 	cfg := tls.Config{}
 
 	if t.CertFile != "" || t.KeyFile != "" {
 		cert, err := tls.LoadX509KeyPair(t.CertFile, t.KeyFile)
 		if err != nil {
-			return nil, fmt.Errorf("Error loading tls client certificate: %s", err)
+			return nil, fmt.Errorf("Error loading tls server certificate: %s", err)
 		}
 
 		cfg.Certificates = []tls.Certificate{cert}
 	}
 
+	// useful when we present multiple certificates
 	cfg.BuildNameToCertificate()
 
 	// apply hardened TLS settings
 	cfg.MinVersion = tlsMinVersion
 	cfg.CurvePreferences = tlsCurvePreferences
 	cfg.CipherSuites = DefaultCipherSuites
-
-	// Tell the server to prefer its cipher suite ordering over what the client
-	// prefers.
+	// Tell the server to prefer it's own cipher suite ordering over the client's preferred ordering
 	cfg.PreferServerCipherSuites = true
 
 	return &cfg, nil
 }
 
-// ToTLSClientConfig is like ToTLSConfig but intended for TLS client config.
-func (t *TLSOptions) ToTLSClientConfig() (*tls.Config, error) {
+// ToClientTLSConfig is like ToServerTLSConfig but intended for TLS client config.
+func (t *TLSOptions) ToClientTLSConfig() (*tls.Config, error) {
 	cfg := tls.Config{}
 	cfg.InsecureSkipVerify = t.InsecureSkipVerify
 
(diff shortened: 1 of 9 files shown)


gbolo:
I updated description with technical details of the change to aid in the code review

gbolo:
With these TLS strong settings we will **LOSE** support for the following clients (when TLS is enabled ofcourse):

 - Internet Explorer 6 XP
 - Internet Explorer 7 Vista
 - Internet Explorer 8 Win 7 and XP
 - Opera 17 Win 7
 - Android 4.2.2 and lower
 - Java 6u45 and lower
 - Java 7u25 and lower

gbolo:
also, we need to avoid hard-coding things like this into the code: `c.TLS.InsecureSkipVerify = true`. This defeats the whole authentication process of an x509 certificate

echlebek:
Thanks for the pull request! I ended up making comments on your branch because I wasn't paying attention, oops!

I don't think we mind losing support for the clients you list, but thanks for mentioning that.

The portion of the code that disables TLS verification is only for an internal reverse proxy, where the dashboard server uses an HTTP client to talk to the API. This is obviously not ideal and we have an issue on the backlog to fix it.

gbolo:
Hi @echlebek 

Thank you for your review, good catch on the 521 cipher. I have applied the changes. 

Also I have noticed that the same keypair is being used for both client and server TLS config (client being `registerGraphQLService`). It does not hurt right now, but should I correct this anyway?

echlebek:
Ah, yes please!

api/core/v2/tls.go
```diff
@@ -63,6 +70,10 @@ func (t *TLSOptions) ToTLSClientConfig() (*tls.Config, error) {
 		if err != nil {
 			return nil, err
 		}
+		// TODO(gbolo): we need an option to first load the system root CAs, then append the user specified CA(s).
+		// In this current form, our client will ONLY trust the specified CA(s).
+		// This may be desired by the user, but if this client config is used for a variety of endpoints,
+		// then perhaps the user should decide if they want system roots enabled or not.
```
ccressent: I think the rationale for not trying to combine the system's default CA certificate bundle and the user provided bundle is that the user can then completely override the system's default if they want or need to.

If users wants to combine the system's default bundle and a custom bundle, they can just create a new bundle containing the two? Or am I missing a pretty common use case that would really benefit from adding more flags? I'm in favor of simplicity and reducing the number of available flags! :)

backend/cmd/start.go
```diff
@@ -172,8 +172,11 @@ func newStartCommand() *cobra.Command {
 			// Sensu APIs TLS config
 			certFile := viper.GetString(flagCertFile)
 			keyFile := viper.GetString(flagKeyFile)
-			trustedCAFile := viper.GetString(flagTrustedCAFile)
 			insecureSkipTLSVerify := viper.GetBool(flagInsecureSkipTLSVerify)
+			// TODO(gbolo): trustedCAFile will be useful in the future when TLS mutual auth is supported.
+			// Not used for backend right now, but leaving it here for now since it appears
+			// that it is being used for etcd client config
+			trustedCAFile := viper.GetString(flagTrustedCAFile)
```
ccressent: I'd favor separating the two, as explained here: https://github.com/sensu/sensu-go/issues/2548#issuecomment-462521459. That can probably be done in a separate PR though.

backend/cmd/start.go
```diff
@@ -182,18 +185,6 @@ func newStartCommand() *cobra.Command {
...
-				if certFile == "" {
-					emptyFlags = append(emptyFlags, flagCertFile)
-				}
-				if keyFile == "" {
-					emptyFlags = append(emptyFlags, flagKeyFile)
-				}
-
-				if len(emptyFlags) > 0 {
-					return fmt.Errorf("missing the following cert flags: %s", emptyFlags)
-				}
```
ccressent: I think we should keep a form of that branch that checks that _both_ `certFile` and `keyFile` are present at the same time because, unless I'm mistaken, providing one without the other doesn't make sense.

ie: keep that branch, drop the `trustedCAFile != ""` condition, reword the error message to say "you used X without Y" (or the opposite) accordingly.

api/core/v2/tls.go
echlebek: I agree. Using the operating system's certificate store should be the preferred approach, with the command line flag being a way to short-circuit that.

jamesdphillips:
The clients mentioned are not supported by the web UI as it is. 👍

api/core/v2/tls.go
gbolo: Yea that makes sense. I will remove this comment then. Thanks guys

backend/cmd/start.go
gbolo: well technically according to RFC, one is allowed to include both a certificate and private key in the same pem file. And Go supports loading both from same pem file as well. However, in practice, I have never seen people do this lol. OK I will add back the friendly error if one is missing :)

backend/cmd/start.go
gbolo: I will change comment to better explain that

gbolo:
Hi @ccressent @echlebek, thanks for the reviews. I have updated the PR to fix issues mentioned in the review, as well as fix issues that I have **previously missed.** Please review again

Issue #2671, opened 2019-01-31
[Web UI] Spike - Entity page redesign


dabria:
Part of: #2594
Implement missing fields: #2669
This is for a general discussion on redesigning the entity page. We would like to add the ability to collapse a section of the card which would allow us to hide some less important details. This makes it so the user isn't overwhelmed with data. We may already have an efficient and sufficient solution in place.

Whichever redesign ticket is grabbed first will likely create the collapsing cards, or perhaps that should be another ticket as well.
Other redesign tickets:
#2667 #2670

Issue #2670, opened 2019-01-31
[Web UI] Spike - Check Details page redesign


jamesdphillips:
Beyond simply adding fields that are currently missing, we would like to take a step back and take a more wholistic view of the page.

* which data should we push to have "above the fold".
* which data should be hidden by default.
* what data (if any) could be omitted.
* what (if any) helpful auxiliary data can we expose to the user; maybe entities where the check is failing, graphs, etc. 
* better use of space when viewport is very wide.

### Notes

* Initial implementation of missing fields #2668.
* Document: https://docs.google.com/document/d/1blo8zcCKUgQzQvJnxeIo5x3hYD-Uwg0S-Vot6vMfbZE/edit#heading=h.j0wh2i930xae.

PR #2701, opened 2019-02-07
[Web UI] Labels and Annotations

### LINKED ISSUE #2687: [Web UI] Add annotations and labels
We would like to add annotations and labels to the web ui. We want to give our users access to these fields quickly, as to see the many ways they're being used. Following that, we would like to redesign the details pages and display them in a more cohesive and better fitting manner.

* Child of #2594
* Redesign tickets #2670 #2671

## Expected Behavior
Display these fields

## Current Behavior
Doesn't display these fields

## Possible Solution
Display these fields

dabria:
## What is this change?
Closes #2687 
Adds support for annotations and labels in the web ui on the entity and check details pages. This is an initial implementation and we will likely change how it's displayed as needs change. We just wanted to get this in the hands of our users.
Relevant design tickets: #2670 #2671

screen shot 2019-02-07 at 11 19 55 am

Users should see this information

Yes


dabria:
Realized that some people have rather large annotations and keys, this will need to be revisited.

dabria:
screen shot 2019-02-12 at 3 41 33 pm
I've updated the annotations section

dashboard/src/components/partials/LabelsAnnotationsCell/LabelsAnnotationsCell.js
```diff
@@ -0,0 +1,140 @@
...
+  };
+
+  static defaultProps = {
+    entity: null,
+    check: null,
+  };
+
+  static fragments = {
+    check: gql`
+      fragment LabelsAnnotationsCell_check on CheckConfig {
```
jamesdphillips: it would make the component simpler if we made the fragment only query the Metadata component, eg.

```graphql
fragment LabelsAnnotationsCell_metadata on Metadata {
  labels { key, val }
  annotations { key, val }
}
```

it would reduce the complexity and allow us to reuse it with any object that has annotations and labels.

PR #2712, opened 2019-02-11
Add entity labels to graphql

roganartu:
## What is this change?

Entity labels are not currently exposed via graphql. This change exposes them under the `labels` field.

To allow users to access labels via the graphql endpoint. I imagine this would also be useful for the Sensu dashboard to allow showing more detailed entity information on check result pages etc.

I can add one if needed.

Not really. Seems like a minor change, and I built and ran it locally and it works as expected.

Nope.

I haven't seen any documentation for thee graphql endpoint that isn't just reading the source, so I don't think there's anything to update.

I didn't see any existing tests anywhere I made changes, so not sure where I'd put it or if it's necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Fixed
+- Fixed a regression in the agent that would not allow proxy checks to be
+run for subsequent executions.
+
 ## [5.2.0] - 2019-02-06
 
 ### Added
--- a/agent/check_handler.go
+++ b/agent/check_handler.go
@@ -10,7 +10,7 @@ import (
 	"time"
 
 	"github.com/sensu/sensu-go/agent/transformers"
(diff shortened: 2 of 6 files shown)


jamesdphillips:
Hey @roganartu, thanks for the contribution! However, we recently added the field in the (WIP) PR [2701]*https://github.com/sensu/sensu-go/pull/2701/files#diff-2d1e04067ce60772acf4590bf32a982eR16). The implementation does differ somewhat, so please give it a look and tell me if the approach will work out for your use case.

---

As an aside: I don't want to discourage contribution or usage but I just want to make clear that the GraphQL service is still unstable and interface is bound to continue changing (maybe even significantly) over the next little bit.

roganartu:
Implementation difference is fine, I'll close this in favour of that one thanks. I searched the issues but not the existing PRs (and I'm not sure github issue search would've helped here anyway!).

I'm aware that the GraphQL endpoint is unstable and that's ok, I just need to pay attention to the changelogs and version diffs for each release.

PR #2722, opened 2019-02-14
closes #2642 separate TLS config for dashboard

### LINKED ISSUE #2642: Create separate TLS config for dashboardd
## Expected Behavior
The dashboard can use a separate TLS config from the API

## Current Behavior
The dashboard TLS config is aliased to the API TLS config

## Rationale
The dashboard should be allowed to present its own TLS certificate. Users may wish to host the dashboard on a different hostname than the API.

gbolo:
Signed-off-by: gbolo <george.bolo@gmail.com>

## What is this change?

See #2642

In the chance that someone would like to present a different certificate on the dashboard port

yes

there are two new flags. The behaviour of omitting the flags is the same as the previous release (backwards compatible). Is this desired?

no

I have not updated the docs yet, this change will require that.

would be nice, I did not include one yet.

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -250,10 +250,6 @@ func Initialize(config *Config) (*Backend, error) {
 			dashboardTLSConfig.CertFile = config.DashboardTLSCertFile
 			dashboardTLSConfig.KeyFile = config.DashboardTLSKeyFile
 		}
-	} else {
-		// important for this to be nil, since there's
-		// a nil check on dashboardd Start()
-		dashboardTLSConfig = nil
 	}
 	dashboard, err := dashboardd.New(dashboardd.Config{
 		APIURL: config.APIURL,
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -32,6 +32,7 @@ import (
 	etcdstore "github.com/sensu/sensu-go/backend/store/etcd"
 	"github.com/sensu/sensu-go/rpc"
 	"github.com/sensu/sensu-go/system"
+	"github.com/sensu/sensu-go/types"
 )
 
 // Backend represents the backend server, which is used to hold the datastore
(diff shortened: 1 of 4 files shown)


ccressent:
I'm OK with the idea of this PR; is this something you've needed @gbolo, or is this just a "nice to have"? I'm afraid of a flag explosion and the potential it may have to confuse operators.

Can anyone else give their 2 cents?

gbolo:
> I'm OK with the idea of this PR; is this something you've needed @gbolo, or is this just a "nice to have"? I'm afraid of a flag explosion and the potential it may have to confuse operators.
> 
> Can anyone else give their 2 cents?

Hey @ccressent 
This issue was not opened by me, typically I use a reverse proxy for TLS across my containers/services. I was simply going through TLS issues and trying to close them with PRs. I thought that would help me get more familiar with the code base.

If you ask me though, I would separate all components from backend (apid, agentd, and dashboardd) for more flexibility, instead of this one monolithic backend. That's my 2 cents :)

ccressent:
Sorry I somehow missed the linked issue (#2642)!

Thanks again for your work @gbolo! The change looks good to me, but we are currently focusing on what we have agreed to include in our upcoming release (5.3). In the interest of not moving goalposts, I'm marking this PR as 5.3.next, meaning we'll approve, test and merge it _after_ the 5.3 release.

Long story short, don't worry if this stays open for a week or two!

gbolo:
Thanks @ccressent I have no objection to that

backend/backend.go
```diff
@@ -234,11 +235,31 @@ func Initialize(config *Config) (*Backend, error) {
...
+		// override the TLS cert/key if user specified them
+		if config.DashboardTLSCertFile != "" && config.DashboardTLSKeyFile != "" {
+			dashboardTLSConfig.CertFile = config.DashboardTLSCertFile
+			dashboardTLSConfig.KeyFile = config.DashboardTLSKeyFile
+		}
+	} else {
+		// important for this to be nil, since there's
+		// a nil check on dashboardd Start()
+		dashboardTLSConfig = nil
+	}
```
ccressent: I think this `else` branch is not necessary, since the declaration of `dashboardTLSConfig` on line 238 would make its value `nil` already (Go's zero value for pointers). What do you think @gbolo?

echlebek:
It was my intent that the TLS configurations for the dashboard and the API not be coupled. I would expect them to work as you were expecting.

This is a breaking change; users who were previously using TLS will need to configure TLS in the dashboard.

Making this change will require good messaging to inform users, and possibly some changes to the default configuration files for the agent.

ccressent:
What about the following "meet in the middle" approach:
- with only `cert-file` and `key-file` used, that certificate is used for both API and Dashboard (current behaviour)
- with only `dashboard-cert-file` and `dashboard-key-file` used, that certificate is used only by the Dashboard; API is insecure
- with all 4 options used, both API and Dashboard use TLS, with different certificates

That way it's not a breaking change for existing users but it still allows separate API and Dashboard certificates if needed.

backend/backend.go
gbolo: Hi @ccressent , I would love to remove that, however I believe I didn't have much of a choice due to an error when attempting to start dashabordd. I will revisit it later to see why i explicitly set it to nil.

gbolo:
> What about the following "meet in the middle" approach:
> 
> * with only `cert-file` and `key-file` used, that certificate is used for both API and Dashboard (current behaviour)
> * with only `dashboard-cert-file` and `dashboard-key-file` used, that certificate is used only by the Dashboard; API is insecure
> * with all 4 options used, both API and Dashboard use TLS, with different certificates
> 
> That way it's not a breaking change for existing users but it still allows separate API and Dashboard certificates if needed.

I think these are good suggestions. If there is consensus with that, i can make the required adjustments.

I would like to state again, that it would be nice to completely separate all the components to their own binaries (similar to original sensu approach) since i believe this monolithic binary provides little advantage over the separation approach, and that "advantage" seems to diminish with each passing commit :)

backend/backend.go
gbolo: Hi @ccressent , I removed the `nil` and tested the options, which seemed to work just fine. Sorry I can't remember why I was so explicit with that `nil`. Thanks for catching that

ccressent: No worries, that's what reviews are for! :)

ccressent:
Let's go with the aforementioned "meet in the middle" approach then; it gives the most flexibility while being backwards compatible.

gbolo:
> Let's go with the aforementioned "meet in the middle" approach then; it gives the most flexibility while being backwards compatible.

Sounds good, I will try and make the necessary adjustments on the weekend or sooner.

gbolo:
@ccressent should be working as desired now.

PR #2733, opened 2019-02-19
Add persist

### LINKED ISSUE #2732: Feature Request: Support persisting mutations
Allow mutators to persist their changes to the store, instead of just forwarding them down the pipeline.

## Expected Behavior
- As a user I can specify an option in the spec of my mutator to instruct Sensu to save the changes it makes (if any) to the etcd store.
- The mutator runs as normal
- On receiving the mutated event, Sensu checks if the event was modified (optional, might have performance implications) and persists the changes to the store before forwarding the mutated event to the next step in the pipeline as usual.

## Current Behavior
Mutations are wholly contained within the pipeline with which they run.

## Possible Solution
https://github.com/sensu/sensu-go/pull/2733

## Context
Sometimes events need some dynamic context added to them, which is why mutators already exist. Allowing mutators to persist this dynamic data allows tools querying Sensu data (eg: sensutctl, the dashboard, and any other tooling users may create or use that leverages the API) to provide this extra  context without having to implement the ability to query these external dependencies themselves.

Some examples that come to mind where this feature might be useful:
- A handler that raises tickets (eg JIRA) for events might want to persist the ticket key or link in the events
- A mutator that looks up some dynamic state for the alerting host from an external service

roganartu:
## What is this change?

Add the ability for mutators to persist their mutations to the etcd store.

Closes #2732

Yes, will add shortly.

I don't think so.

It was not exactly clear how to generate the proto file changes.

I ended up using the following command from `api/core/v2` and then only checking in the relevant changes:
```
protoc --proto_path=${GOPATH}/src:${GOPATH}/src/github.com/gogo/protobuf/protobuf:. --gogo_out=. mutator.proto
```

The generated proto files in general seem to be inconsistent in their usage of golang vs gogoproto. gogoproto is used extensively, but not all the generated proto files import and use it. Perhaps I was simply generating them wrong, but I would expect the correct generation invocation to result in very few changes. I excluded all new generated changes because it builds fine from this branch, I didn't want to break things, and figured the CI has this configured correctly.

New documentation will be required. Where can I make that change?

Built and ran a fork in a test cluster over the weekend that processed thousands of generated and real events with this new mutator option enabled. Persisted as expected with no errors in the logs.


backend/pipelined/mutate.go
```diff
@@ -82,6 +82,16 @@ func (p *Pipelined) mutateEvent(handler *types.Handler, event *types.Event) ([]b
...
 	}
 
+	if mutator.Persist {
+		modifiedEvent := &types.Event{}
+		err := json.Unmarshal(eventData, modifiedEvent)
+		if err != nil {
+			logger.WithFields(fields).WithError(err).Error("failed to unmarshal the mutated event")
+			return nil, err
+		}
+		p.store.UpdateEvent(ctx, modifiedEvent)
```
golangcibot: Error return value of `p.store.UpdateEvent` is not checked (from `errcheck`)

roganartu: Fixed to make this fail the mutator with an error if persisting fails. Unsure whether this is the most desirable behaviour or if we should log an error and let the rest of the pipeline execute.

portertech:
What if there are multiple mutators with persist for a particular event?

What if there are multiple pipelines for a particular event and their event data no longer what's in etcd?

Is losing the original event data acceptable?

echlebek:
One issue with storing mutated events is that it's legal for mutators to result in any sequence of bytes. As implemented, the PR would allow people to construct pipelines that fail to update events, if the mutator does not produce JSON that fits the event spec.

roganartu:
> What if there are multiple mutators with persist for a particular event?

I considered this, and I think the persist option should be clearly documented that it is an eventual consistency feature and that enforcing a particular order of execution is not supported. I couldn't think of a nice workaround for that without serialising all event pipelines which seems quite undesirable.

> What if there are multiple pipelines for a particular event and their event data no longer what's in etcd?

I'm not sure what you mean here. Trying to think what you might mean has made me think of a question of my own though: what if multiple events fire for the same check/host combination? I think there's a race condition for check history where history items may be lost.

> Is losing the original event data acceptable?

Under what circumstances do you think this would result in original event data being lost? A (buggy) mutator that returns an empty JSON object might present a problem that I'm not sure I considered here. On further consideration, perhaps this should be limited to _adding_ data instead of fully mutating, since that's the use case driving the original feature request. Could potentially:
1) compare mutated event with event that was passed to mutator. If no changes, continue in pipeline as normal
2) if changes, re-fetch the event from the store (with a mutex? Unsure if there's precedent or appetite for locking these ops) and add/update only supported fields (labels and annotations?). Mutex protects against the store PUT call making accidental changes.
3) save and release mutex

This doesn't quite fix the race condition of multiple persisting mutators, but it restricts it to only affecting mutators that attempt to modify the same field, which seems more acceptable (with clear documentation). This also avoids the history loss problem I raised above.

roganartu:
> One issue with storing mutated events is that it's legal for mutators to result in any sequence of bytes. As implemented, the PR would allow people to construct pipelines that fail to update events, if the mutator does not produce JSON that fits the event spec.

I don't think this will be a problem in practice as this feature as proposed is opt-in. Existing mutators will be unaffected, and the documentation on the `persist` option can emphasise that mutators should ensure they return a valid event when using `"persist": true`. Perhaps `persist` should be an enum instead of a bool:
- `0` - don't persist
- `1` - attempt to persist, but don't stop pipeline on failure just log the error
- `2` - must persist (this would be the same as `"persist": true` with this PR as currently implemented)

This would allow users slightly more control. I'm not attached to any particular implementation here.

annaplotkin:
Per comments in #2732, closing this PR.

PR #2737, opened 2019-02-19
Fix yaml and wrapped-json output for sensuctl user list

### LINKED ISSUE #2569: sensuctl user list: no output with yaml and wrapped-json format
The `sensuctl user list` command doesn't produce any output if it is configured to output yaml or wrapped-json. Requesting json or the tabular format gives the expected output.

## Current Behavior

Tabular (and json, not shown here) output works:
```
$ sensuctl user list --format tabular
  Username       Groups       Enabled  
 ────────── ──────────────── ───────── 
  admin      cluster-admins   true     
  agent      system:agents    true
```

But yaml and wrapped-json doesn't:
```
$ sensuctl user list --format yaml
$ sensuctl user list --format wrapped-json
```

## Possible Solution

yaml and wrapped-json being output formats that include metadata information while tabular and json do not, I suggest there may be a problem somehow related to metadata handling for users in `sensuctl`.

I've tested most (not all) of the other object types, and all 4 output formats appear to work just fine. 

1. Start a backend
2. No need to manually create users to test: `admin` and `agent` are already seeded
3. Observe the output of `sensuctl user list --format {yaml,wrapped-json,tabular,json}`

## Context
Reported by Yakhatape on the community Slack for `--format yaml`.
Reproduced by me with both `yaml` and `wrapped-json`.

## Your Environment

* Sensu version : 5.1.0
* Installation method: git
* Operating System: Linux 4.19

ccressent:

Closes #2569.

Yes, added.

It isn't fully clear to me why `types.User` is not a `types.Resource`. That is the real root cause as of why this bug appeared in the first place. This led me to add "snowflake" code for `types.User` in `PrintYAML()`.

No.

- Wrote unit tests for the json, wrapped-json and yaml outputs for `sensuctl user list`. 
- With varied users created, checked the output of `sensuctl user list --format {json,wrapped-json,yaml}`

I don't believe we need to add RSpecs for this as this is covered by the unit tests.


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 - Check results sent via the agent socket now support handlers.
+- `sensuctl user list` can now output yaml and wrapped-json
```
palourde: Should this go into `CHANGELOGS/unreleased`?

ccressent: Yes. If that's OK with you, I'll leave it for this PR as I don't think there is any harm in just merging in the Changelog directly this time around, but I'll use the new CHANGELOGS folder in the future.

nikkiki: @ccressent don't speak too soon 😆 
screen shot 2019-03-04 at 11 21 01 am

PR #2740, opened 2019-02-20
Add package ringv2

### LINKED ISSUE #2735: Add v2 ring package
Add a new ring package that works more reliably and simply than the old ring package.

Specs: https://docs.google.com/document/d/1lxgTouMOCOBDJG5ri0Jp2G97banHFRDrRbxA6sj_ZEs

echlebek:
## What is this change?

This commit adds the ringv2 package. It replaces most of the
functionality in the ring package.

The ringv2 package is much simpler than the original ring package
and operates on the basis of watchers.

A caveat of the new package is that the smallest interval supported
is 5s, a constraint imposed by etcd.

Closes #2735 

No, this is merely plumbing.

No doc changes required.

An extensive integration test suite.


backend/ringv2/ringv2.go
```diff
@@ -0,0 +1,433 @@
...
+		client:        client,
+		itemPrefix:    path.Join(storePath, "items"),
+		keySeqKey:     path.Join(storePath, "seq"),
+		intervalKey:   path.Join(storePath, "interval"),
+		triggerPrefix: path.Join(storePath, "triggers"),
+	}
+}
+
+// dump is for debugging
+func (r *Ring) dump(ctx context.Context, w io.Writer) {
```
golangcibot: U1000: func `(*Ring).dump` is unused (from `unused`)

echlebek: This is fine

backend/ringv2/ringv2.go
```diff
@@ -0,0 +1,433 @@
...
+	case EventRemove:
+		return "EventRemove"
+	case EventTrigger:
+		return "EventTrigger"
+	case EventError:
+		return "EventError"
+	case EventClosing:
+		return "EventClosing"
+	default:
+		return "INVALID"
```
nikkiki: giphy-4

backend/ringv2/ringv2.go
```diff
@@ -0,0 +1,433 @@
...
+
+	getresp, err := r.client.Get(ctx, itemKey)
+	if err != nil {
+		return fmt.Errorf("couldn't add %q to ring: %s", value, err)
+	}
+
+	if len(getresp.Kvs) > 0 && len(getresp.Kvs[0].Value) > 0 {
+		// Item already exists
+		return nil
+	}
```
ccressent: > If the value already exists, it will not be disturbed.

So this ring implementation is not a true container in the sense that it "inspects" values that are added to it and can take actions based on those values; in that case not add any "duplicates" to the ring.

This is fine for the original use case of the ring (round robin checks), but we also discussed using the new ring package for other things (to be determined!), implying it might need to be a bit more general and not assume that adding the same value several times is nonsensical.

What do you think?

backend/ringv2/ringv2.go
```diff
@@ -0,0 +1,499 @@
...
+
+func (r *Ring) nextInRing(ctx context.Context, n int) ([]*mvccpb.KeyValue, error) {
+	resp, err := r.client.Get(ctx, r.itemPrefix, nextInRingOps(n)...)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't get next item in ring: %s", err)
+	}
+	return resp.Kvs, nil
+}
+
+func (r *Ring) bumpSequence(ctx context.Context, value string) error {
```
golangcibot: U1000: func `(*Ring).bumpSequence` is unused (from `unused`)

backend/ringv2/LICENSE
```diff
@@ -0,0 +1,20 @@
+Copyright (c) 2017 Sensu Inc.
```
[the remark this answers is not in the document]
echlebek: Apparently it has to be in every directory... I suppose we should update it and others with the current year.

backend/ringv2/ringv2.go
```diff
@@ -0,0 +1,433 @@
...
+	"fmt"
+	"io"
+	"path"
+
+	"github.com/coreos/etcd/clientv3"
+	"github.com/coreos/etcd/mvcc/mvccpb"
+	"github.com/sensu/sensu-go/backend/etcd"
+)
+
+var _ Interface = &Ring{}
```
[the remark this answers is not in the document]
echlebek: It's a build-time assertion that `*Ring` implements `Interface`

backend/ringv2/ringv2.go
echlebek: My view of the data structure is that it is a set - adding values should be idempotent. Additionally, the ordering should not be disturbed when an existing value is added.

PR #2746, opened 2019-02-22
Don't set a timeout for asset downloads

### LINKED ISSUE #2745: Assets will fail to download if they do not download within 30s
## Expected Behavior
Assets should be allowed to be retrieved at any speed, even on slow networks. For large assets on slow networks, much more than 30s of download time may be both required and tolerable.

## Current Behavior
Assets will stop downloading after 30s, and the process will be restarted. This can result in assets downloading forever, but never finishing.

## Possible Solution
The asset fetcher's HTTP client should not specify a timeout.

Use assets that your system cannot download within 30s. You should observe that the asset never finishes downloading.

## Context
Reported by @jspaleta

echlebek:
## What is this change?

This commit instructs the asset fetcher to not apply any HTTP
client timeout when fetching assets. Previously, assets that took
longer than 30s to download would retry until they took less than
30s to download.

Buffering has also been added, which should improve the performance
of asset downloads.

Closes #2745 

Yes

Verified manually by Jef.

### DIFF
--- a/CHANGELOGS/unreleased/2745-bugfix_asset_fetcher.md
+++ b/CHANGELOGS/unreleased/2745-bugfix_asset_fetcher.md
@@ -1,0 +1,5 @@
+### Fixed
+- Assets downloads no longer specify a client timeout.
+
+### Changed
+- Asset downloading now uses buffered I/O.
--- a/asset/fetcher.go
+++ b/asset/fetcher.go
@@ -1,6 +1,7 @@
 package asset
 
 import (
+	"bufio"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -25,9 +26,7 @@ type urlGetter func(string) (io.ReadCloser, error)
 
 // Get the target URL and return an io.ReadCloser
 func httpGet(url string) (io.ReadCloser, error) {
-	client := &http.Client{Timeout: defaultHTTPGetTimeout}
-
-	resp, err := client.Get(url)
+	resp, err := http.Get(url)
 	if err != nil {
 		return nil, fmt.Errorf("error fetching asset: %s", err)
 	}
@@ -58,8 +57,12 @@ func (h *httpFetcher) Fetch(url string) (*os.File, error) {
 		return nil, fmt.Errorf("can't open tmp file for asset")
 	}
 
-	if _, err = io.Copy(tmpFile, resp); err != nil {
+	buffered := bufio.NewWriter(tmpFile)
+	if _, err = io.Copy(buffered, resp); err != nil {
 		return nil, fmt.Errorf("error downloading asset")
+	}
+	if err := buffered.Flush(); err != nil {
+		return nil, fmt.Errorf("error downloading asset: %s", err)
 	}
 
 	if err := tmpFile.Sync(); err != nil {


jspaleta:
Just tried the PR branch on Fedora 29 on a slow adsl link.
The asset download doesn't timeout any longer...
but sensu-agent from this PR appears to hang when trying to stop it with CTRL-C signal.  I think the download thread is preventing sensu-agent from stopping until the download is complete. But because the download is being performed over a slow link... the hang to exit can end up being very long.

is there a way to have the asset manager see the terminate signal and stop the download?

echlebek:
Hmm, how many CPU cores (virtual or otherwise) on the agent where you observed the slow response to CTRL+C?

jspaleta:
Running sensu-agent in Fedora 29 on T480 laptop with 8 cores... no virtualization layer in this latest test.

For my test using a 12MB asset, the agent finally exited when the download was completed and the asset moved from /tmp/ in to the correct /var/cache location. 

No truncated partial download left in /tmp/ when sensu-agent finally exits.

palourde:
Just tested that the `CTRL+C` issue already existed before this PR.

That being said, I'm kind of uncomfortable with removing any timeout value; let's say sensu-agent is configured to download an asset that does timeout for reasons outside of control (you could simulate a timeout of 230 seconds with this URL: `http://httpstat.us/200?sleep=230000`), you effectively end up with a zombie sensu-agent process, that can only be stopped with `TERM` signal, and will produce failure keepalives events.

While the default 30 seconds timeout might not be a sensitive default value for everyone, it did prevent zombie processes in this particular case. My feeling is that we should probably make that timeout value configurable instead and keep a sensible default values for most cases, so users who increase it are aware of the potential consequences.

Just my two cents.

echlebek:
That's entirely within our control, as we trap the signal and then wait for things to finish up before we terminate. We could just not trap the signal.

palourde:
Indeed, but that would still require users to go kill each sensu-agent process affected, which is not optimal in term of user experience in my opinion.

echlebek:
I see, the concern is that child processes will be orphaned and left in a zombie state?

We can cancel both those child processes and any asset downloads when we handle the signal, just need to thread a context through to the various things that need canceling.

HTTP client requests can be cancelled by making use of https://golang.org/pkg/net/http/#Request.WithContext

palourde:
It's not the child process, it's sensu-agent itself. For example, if I configure an asset with `http://httpstat.us/200?sleep=230000`, while it tries to download that, everything else is paused; the keepalives are no longer being responded by the agent, nor other check requests.

jspaleta:
okay filed a new issue for the SIGINT handling: #2756

gbolo:
Hey, I think that the timeout value was misplaced and meant to be placed somewhere in the `http.Transport` and/or the `net.Dialer` rather than directly to the `http.Client.Timeout`. The goal here is to have a reasonable timeout before the first packet is sent (time to first byte), not to timeout the entire length of the connection (while a transfer is still in progress). These settings have implications for Denial Of Service based attacks (although probably not much of a risk here TBH). The default http client that go provides has fairly long timeouts set in its transport struct:

```
var DefaultTransport RoundTripper = &Transport{
        Proxy: ProxyFromEnvironment,
        DialContext: (&net.Dialer{
                Timeout:   30 * time.Second, <-- this seems high to me
                KeepAlive: 30 * time.Second,
                DualStack: true,
        }).DialContext,
        MaxIdleConns:          100,
        IdleConnTimeout:       90 * time.Second,
        TLSHandshakeTimeout:   10 * time.Second, <-- this also seems high to me
        ExpectContinueTimeout: 1 * time.Second,
}
```

It also respects the usual proxy environment variables (`$HTTP_PROXY` and `$NO_PROXY`)  which may or may not be your intention? Furthermore, in the future you may want to support custom TLS configuration for the asset download client (to support TLS custom CAs or even possibly mutual auth?)

echlebek:
I think the default should suffice here for now, until we have evidence otherwise. 30s is a long time, yes, but the function that creates the client gets retried until it succeeds, anyhow.

echlebek:
As for TLS, you raise a good point! Outside the scope of this PR though. :)

gbolo:
> I think the default should suffice here for now, until we have evidence otherwise. 30s is a long time, yes, but the function that creates the client gets retried until it succeeds, anyhow.

sounds good to me

Issue #2750, opened 2019-02-22
Feature Request: Allow more complex event grouping


roganartu:
Sensu (and most other monitoring platforms) group events by two attributes:
- hostname
- check name

This does not work well with checks that may apply multiple times to the same host but for a different parameter. Consider a simple `ping` check on a machine with multiple network interfaces. One may want one ping check per interface, which can be achieved by duplicating the config and changing the interface. However then users are faced with the following choice:
- keep the check names identical, or
- append some unique identifier (like the interface name) to the check name

The former creates the problem that the checks are now confused with each other, where a down event from one interface will be cleared by an up event from the other interface. It also prevents users from silencing alerts from only one of the interfaces (kind of. I think Sensu would actually allow this by silencing based on a label value or similar but it's not very user-friendly), so the only real option is to append a unique identifier to the end of the check and duplicating a bunch of config.

This is how many Nagios configs live in the wild and it's pretty unwieldy.

Instead, Sensu should allow users to specify some attributes to group checks by in the `CheckConfig`.

This will require significant work, but it is not fundamentally incompatible with the underlying Sensu Go architecture (in many ways it actually Just Works™) and I think if it's going to be done it would be better to do it sooner than later before more and more assumptions about the uniqueness of hostname + checkname are made.

## Expected Behavior
Consider the a proxy check result as follows:
```
{
  "event": {
    "check": {
      "proxy_entity_name": "host1",
      "publish": false,
      "metadata": {
        "name": "check-ping",
        "namespace": "default"
      }
    }
    "group_by": {
      "name": true, # Optional, defaults to true
      "hostname": true, # Optional, defaults to true
      "check_labels": [],
      "entity_labels": [],
      "event_labels": ["interface"]
    }
  },
  "metadata": {
    "labels": {
      "interface": "eth1"
    }
  }
}
```
Sensu will create a event, let's call it `event1` for simplicity.

If Sensu then receives another proxy check result but with the following event metadata:
```
  "metadata": {
    "labels": {
      "interface": "eth0"
    }
  }
```
It will create a second event, `event2`, that is considered unique to `event1` (ie: neither appears in the other's history, resolving one will not resolve the other, etc).

The etcd store actually already supports this and required very minimal work. Because everything is stored under a namespaced path, it was as simple as appending some more fields to the event path.
```
/namespace/default/events/host1/check-ping
```
simply became
```
/namespace/default/events/host1/check-ping/event:interface:eth1
```

[...]

PR #2751, opened 2019-02-22
Proof of concept for grouping events by labels

roganartu:
Add the ability to group events by labels instead of just hostname and checkname combination.

This is not a PR that I expect to be merged in its current state. It is a proof-of-concept for #2750

Yes.

See detailed description in #2750

New docs would definitely be required.

I have not done any verification.


palourde:
Hi @roganartu, I just wanted to thank you once again for opening this PR! I believe we now better understand your needs, and I just raised the discussion to our product team.

Nesting events within an additional (optional) level in their path would however cause many problems around the codebase so if we will need to rethink this solution if we decide to tackle that issue. So for now, I'll just close this PR and move back the discussion to https://github.com/sensu/sensu-go/issues/2750.

Thanks!

PR #2754, opened 2019-02-25
Split agent subscriptions properly in CLI flags

### LINKED ISSUE #2753: Entity subscriptions are stored incorrectly
## Expected Behavior
Entity subscriptions are an array of subscriptions that are specified on the agent, plus an extra entity subscription added by the server.

i.e.
```
["eenie", "meenie", "miney", "moe", "entity:tiger"]
```

## Current Behavior
```
["eenie,meenie,miney,moe", "entity:tiger"]
```
The array of subscriptions is always of length 2.

```
sensuctl entity list --format json | jq .[0].subscriptions
```

## Context
Encountered while doing development work.

echlebek:
## What is this change?

This commit fixes a bug where agents were sending the wrong
subscriptions in entities, as they would not be split properly
before being sent to the config.

Due to a quirk in how subscriptions are handled, this doesn't
affect any current scheduling mechanism. Nevertheless,
subscriptions must be properly stored in entities, especially for
future work that depends on entity details being correct.

Closes #2753 

Yes

No doc changes required.

Manual verification.

### DIFF
--- a/agent/cmd/start.go
+++ b/agent/cmd/start.go
@@ -161,21 +161,8 @@ func newStartCommand() *cobra.Command {
 				cfg.BackendURLs = append(cfg.BackendURLs, newURL)
 			}
 
-			// Get a single or a list of redact fields
-			redact := viper.GetString(flagRedact)
-			if redact != "" {
-				cfg.Redact = splitAndTrim(redact)
-			} else {
-				cfg.Redact = viper.GetStringSlice(flagRedact)
-			}
-
-			// Get a single or a list of subscriptions
-			subscriptions := viper.GetString(flagSubscriptions)
-			if subscriptions != "" {
-				cfg.Subscriptions = splitAndTrim(subscriptions)
-			} else {
-				cfg.Subscriptions = viper.GetStringSlice(flagSubscriptions)
-			}
+			cfg.Redact = viper.GetStringSlice(flagRedact)
+			cfg.Subscriptions = viper.GetStringSlice(flagSubscriptions)
 
 			sensuAgent := agent.NewAgent(cfg)
 			if err := sensuAgent.Run(); err != nil {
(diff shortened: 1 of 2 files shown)


agent/cmd/start.go
```diff
@@ -170,8 +170,12 @@ func newStartCommand() *cobra.Command {
 			}
 
 			// Get a single or a list of subscriptions
-			subscriptions := viper.GetStringSlice(flagSubscriptions)
-			cfg.Subscriptions = subscriptions
+			subscriptions := viper.GetString(flagSubscriptions)
+			if subscriptions != "" {
+				cfg.Subscriptions = splitAndTrim(subscriptions)
+			} else {
+				cfg.Subscriptions = viper.GetStringSlice(flagSubscriptions)
```
palourde: Maybe I'm missing something but I don't see how we could reach this condition without the `flagSubscriptions` flag being empty, and therefore subscriptions value at all.

Both `--subscriptions foo,bar` & `--subscriptions foo --subscriptions bar` will cause the _if_ statement to be true, and in the latter case the first value (foo) will be ignored. 

For reference, https://github.com/spf13/viper/issues/380.

echlebek: I was just doing what you did :grin: 
https://github.com/sensu/sensu-go/blame/master/agent/cmd/start.go#L164

I'm not sure why it would drop the first value foo, but I think you're right about the if statement never being false.

palourde: "Do what I say, don't do what I do" 🤣

palourde: @echlebek Actually what I did worked because I set a slice as the default value, which will cause `viper.GetString` to return an empty value:

https://github.com/sensu/sensu-go/blob/master/agent/cmd/start.go#L238

Definitely not a clean solution though

echlebek: OK. The subscriptions flag has a default value of an empty slice as well. I don't think we can support the canonical way to represent slices, (`--foo a --foo b`) as it would break the CLI... any suggestions?

palourde: @echlebek Oh, we simply need to use `StringSlice` in pflag, I missed that. So we could simply do:

```
cfg.Subscriptions = viper.GetStringSlice(flagSubscriptions)

[...]

cmd.Flags().StringSlice(flagSubscriptions, viper.GetStringSlice(flagSubscriptions), "comma-delimited list of agent subscriptions")
```

Which would result in the following:

```
$ sensu-agent start --subscriptions foo
slice: [foo]
len: 1

$ sensu-agent start --subscriptions foo,bar
slice: [foo bar]
len: 2

$ sensu-agent start --subscriptions foo,bar --subscriptions qux
slice: [foo bar qux]
len: 3
```

echlebek: Sounds like a plan, thanks!

echlebek:
@palourde I've incorporated your suggestion and verified that it works. Thanks again!

PR #2758, opened 2019-02-27
[Web UI] Offline state detection

### LINKED ISSUE #2103: [Web UI] Offline network error state

10xjs:
## What is this change?

This change introduces several features to the Web UI driven by the ability to detect disconnection from the backend API.

If the Web UI loses ability to communicate with the backend API while running, it enters a persistent offline state. While offline, automatic live updates are disabled and a prominent banner notification is displayed. The persistent offline state ends when a new user-initiated request to the backend API succeeds or when the user clicks the "reconnect" button in the alert banner (doing so triggers any live-updating components to attempt to resume updating).

The offline detection heuristic is currently quite simple (and is limited by the amount of detail provided by the browser on certain network errors).  More complicated logic could be implemented, potentially by detecting a threshold quantity of failures within a set time range or by using the browser `NetworkInformation` API. This is something to revisit after this gets some use.

One possible concern that I can foresee is with the current requirement for user input for the app to resume communication. A timer-based auto-reconnect behavior could be beneficial for certain use-cases (e.g. a status dashboard).

Closes #2103 

With the ability to navigate between "pages" and display cached data without an active network connection, a single-page browser app can obscure the fact that the network has failed.

Displaying a blocking error modal would be an option in this case, however we want to allow the app remain as usable as possible while offline while letting the user know that communication with the backend API has stopped.

The global alert approach implemented here used instead of one relying each API-connected component to implement its own detection and reporting. This saves a lot of repeated effort. The implementation is intentionally designed to be transparent to API consumers and will naturally support new different API requests added in the future. In addition, since the offline status is universal to all components, this approach enables components to effectively communicate the offline event with no additional effort (e.g. a component which posts a user-filled form to the API could automatically disable submission if a live update request fails elsewhere in the app).

Yes

No

No complications per se, but this change does affect mostly foundational parts of the code. There is potential for effects to manifest on all views of the Web UI. (I do trust that our standard QA process will sufficiently catch any potential regressions.)

Through much manual testing in various browsers.

Offline state can be simulated locally in some browsers (but not Safari 🙁) see https://developers.google.com/web/ilt/pwa/tools-for-pwa-developers#simulate_offline_behavior. (Firefox does not block connections to localhost when in offline mode, a locally hosted web app must be accessed over a different host for this to work)


10xjs:
The DCO check is needlessly complaining 🙄- ready for review otherwise
> Expected "Melissa Page 73hl10n@gmail.com", but got "Melissa P 73hl10n@gmail.com".

jamesdphillips:
@portertech probably would know best, but, considering all the DCO "issues" are related to merge commits I think this is safe to be merged. The PRs associated with said merge comments all passed DCO checks.

dabria:
We confirmed with @portertech previously it is fine to merge 👍

Issue #2583, opened 2019-01-03
sensuctl won't delete hook (or other resources) without specifying namespace


nixwiz:
When attempting to delete a hook in the current namespace, it shouldn't be required to specify namespace via the --namespace argument to sensuctl.  Is this wrong?  Or is it working as designed?  If it is working as designed then documentation needs to be updated to reflect this and the code for argument checking should require it and usage and error output should list it is such.

## Expected Behavior

'sensuctl hook delete hook_name' should delete a hook in the current namespace.

Or it should error out better about requiring namespace.
```
# sensuctl hook delete linux-process-list-hook --skip-confirm
delete hooks given name

Usage:  sensuctl hook delete [NAME] [flags] --namespace string

Flags:
  -h, --help           help for delete
      --skip-confirm   skip interactive confirmation prompt

Global Flags:
      --api-url string      host URL of Sensu installation
      --cache-dir string    path to directory containing cache & temporary files (default "/root/.cache/sensu/sensuctl")
      --config-dir string   path to directory containing configuration files (default "/root/.config/sensu/sensuctl")
      --namespace string    namespace in which we perform actions (default "default"), required for delete

Error: --namespace required for delete
```

## Current Behavior

If you don't specify a namespace via the --namespace argument, running 'sensuctl hook delete hook_name' fails with "Error: not found".
## Steps to Reproduce (for bugs)

1. Create a hook in a non-default namespace
```
# cat hook.json
{
  "type": "HookConfig",
  "api_version": "core/v2",
  "spec": {
    "command": "/usr/bin/top -b -n 1 | head -20",
    "timeout": 30,
    "stdin": false
  },
  "metadata": {
    "name": "linux-process-list-hook",
    "namespace": "AWS"
  }
}

# sensuctl create -f hook.json
```
2. Confirm that your sensuctl is configured for that namespace
```
# sensuctl config view                          
=== Active Configuration
API URL:       http://127.0.0.1:8080
Sensu Edition: enterprise
Namespace:     AWS
Format:        tabular
```
3. Confirm the hook exists
```
# sensuctl hook info linux-process-list-hook
=== linux-process-list-hook
Name:      linux-process-list-hook
Command:   /usr/bin/top -b -n 1 | head -20
Timeout:   30
Stdin?:    false
Namespace: AWS
```
4. Attempt to delete the hook without specifying the current namespace
```
# sensuctl hook delete linux-process-list-hook --skip-confirm
Error: not found
```
[...]

PR #2762, opened 2019-03-01
Sensuctl delete client should use generic functions, commands should load namespace

nikkiki:

## What is this change?

Fixes a bug in `sensuctl` where handlers and filters would only be deleted from the default namespace, unless a `--namespace` flag was specified.

Closes https://github.com/sensu/sensu-go/issues/2583

Yas.

Nah.

Yes, it seemed pretty simple at first, but refactoring ended up being required. I also had to touch some graphql stuff so I added @jamesdphillips in on that (deleting resources on the web ui still works).

We good.

https://github.com/sensu/sensu-go-qa-crucible/pull/23
```
sensuctl namespace create devops
sensuctl config set-namespace devops
sensuctl hook create hook1 --command "echo hook1"
sensuctl hook delete hook1 --skip-confirm
```


cli/client/filter.go
```diff
@@ -32,7 +32,7 @@ func (client *RestClient) CreateFilter(filter *types.EventFilter) (err error) {
 
 // DeleteFilter deletes a filter from configured Sensu instance
 func (client *RestClient) DeleteFilter(filter *types.EventFilter) error {
-	path := filtersPath(filter.Namespace, filter.Name)
+	path := filtersPath(client.config.Namespace(), filter.Name)
```
ccressent: I think `filter.Namespace` is correct here!

In fact this breaks the `--namespace` flag. The following works before this patch but returns `Error: not found` after applying it :(
```
sensuctl config set-namespace default
sensuctl hook delete hook1 --skip-confirm --namespace devops
```

My understanding of the problem is that since the caller of `DeleteFilter()` (here the cobra command returned by `DeleteCommand()`) doesn't set `filter.Namespace`, it gets treated as the default namespace. I think what should be done instead is set `filter.Namespace = client.config.Namespace()` in the caller of `DeleteFilter()`, and let `filter.Namespace` get overridden if the `--namespace` flag is set to a non-empty value (as it is done now).

cli/client/hook.go
```diff
@@ -50,7 +50,7 @@ func (client *RestClient) UpdateHook(hook *types.HookConfig) (err error) {
 
 // DeleteHook deletes hook from configured Sensu instance
 func (client *RestClient) DeleteHook(hook *types.HookConfig) error {
-	path := hooksPath(hook.Namespace, hook.Name)
+	path := hooksPath(client.config.Namespace(), hook.Name)
```
ccressent: ditto, but for hooks instead of filters.

cli/client/filter.go
nikkiki: https://github.com/sensu/sensu-go/blob/master/cli/client/check.go#L54

ccressent:
Regarding the conflict in `CHANGELOG.md`, if you want to avoid dealing with them, we've started putting change logs in the `CHANGELOGS/unreleased` folder, which gets concatenated and cleaned during a release.

cli/client/filter.go
nikkiki: That's a good call out.. I'll have to dive a little deeper here.

backend/apid/graphql/mutations.go
```diff
@@ -86,12 +86,12 @@ func (r *mutationsImpl) DeleteCheck(p schema.MutationDeleteCheckFieldResolverPar
 	ctx := setContextFromComponents(p.Context, components)
 
 	client := r.factory.NewWithContext(ctx)
-	check, err := client.FetchCheck(components.UniqueComponent())
+	_, err := client.FetchCheck(components.UniqueComponent())
```
ccressent: So, since we don't need the check returned by `FetchCheck()` anymore, we are purely making a call to that function for the side effect it has of checking if a check actually exists?

cli/commands/event/delete.go
```diff
@@ -32,7 +32,12 @@ func DeleteCommand(cli *cli.SensuCli) *cobra.Command {
 				}
 			}
 
-			err := cli.Client.DeleteEvent(entity, check)
+			namespace := cli.Config.Namespace()
+			if namespaceFlag, _ := cmd.Flags().GetString("namespace"); namespaceFlag != "" {
+				namespace = namespaceFlag
+			}
```
ccressent: After #2770 is merged, I'm fairly sure those calls to `cmd.Flags().Get*()` will not be needed anymore. They were a crutch to make up for an expected parse error earlier in the program's execution, which #2770 fixes through the error whitelisting facility newer versions of Cobra provide!

`namespace := cli.Config.Namespace()` should now correctly give you the right value: the default namespace, overridden by what's in the config file, in turn overridden by a `--namespace` flag anywhere on the command line.

Not requesting a change, just sharing that fact for future changes to those commands!

cli/commands/root/root.go
```diff
@@ -31,7 +30,7 @@ func Command() *cobra.Command {
 	cmd.PersistentFlags().Bool("insecure-skip-tls-verify", false, "skip TLS certificate verification (not recommended!)")
 	cmd.PersistentFlags().String("config-dir", path.UserConfigDir("sensuctl"), "path to directory containing configuration files")
 	cmd.PersistentFlags().String("cache-dir", path.UserCacheDir("sensuctl"), "path to directory containing cache & temporary files")
-	cmd.PersistentFlags().String("namespace", config.DefaultNamespace, "namespace in which we perform actions")
+	cmd.PersistentFlags().String("namespace", "", "namespace in which we perform actions")
```
ccressent: I'm not sure this is necessary: if namespace is equal to `""`, [it is changed to `config.DefaultNamespace` anyway](https://github.com/sensu/sensu-go/blob/bugfix/resource-delete/cli/client/config/basic/reader.go#L31-L33).

And in a sense, `config.DefaultNamespace` here correctly shows that should the namespace never be explicitly given anywhere (config file or command line), it will default to `config.DefaultNamespace`.

What do you think? Maybe I'm missing something?

cli/client/check.go
```diff
@@ -50,19 +50,8 @@ func (client *RestClient) UpdateCheck(check *types.CheckConfig) (err error) {
...
-func (client *RestClient) DeleteCheck(check *types.CheckConfig) error {
-	path := checksPath(client.config.Namespace(), check.Name)
-	res, err := client.R().Delete(path)
-
-	if err != nil {
-		return err
-	}
-
-	if res.StatusCode() >= 400 {
-		return UnmarshalError(res)
```
ccressent: I don't think `UnmarshalError()` is in any potential code path anymore after these changes. Could that change the CLI's output in the face of some errors?

cli/commands/event/delete.go
nikkiki: Yeah, I'll probably rebase after your PR lands, update the necessary things and retest.

backend/ringv2/ringv2.go
```diff
@@ -0,0 +1,544 @@
...
+	interval int
+
+	// triggerPrefix is the prefix that contains ring triggers
+	triggerPrefix string
+
+	// watchCtr counts the number of open watchers
+	watchCtr int64
+
+	// cron is a cron schedule
+	cron cron.Schedule
```
golangcibot: `cron` is unused (from `structcheck`)

backend/agentd/session.go
```diff
@@ -66,7 +71,7 @@ type SessionConfig struct {
 // encounters a receive error.
 func NewSession(cfg SessionConfig, conn transport.Transport, bus messaging.MessageBus, store Store) (*Session, error) {
 	// Validate the agent namespace
-	ctx := context.TODO()
+	ctx, cancel := context.WithCancel(context.Background())
```
golangcibot: the cancel function is not used on all paths (possible context leak) (from `govet`)

cli/client/check.go
nikkiki: I'm not sure what you mean.. the generic `Delete` function could still reach the `UnmarshalError`: https://github.com/sensu/sensu-go/blob/6780d12f05bca9c296e95ca99688a17b9b1e7eb3/cli/client/generic.go#L18

cli/commands/root/root.go
nikkiki: This is a global and persistent flag, so the default value will populate the namespace field wherever the flag applies. The user in this case is in a non-default namespace, and wants to delete a resource in that namespace without specifying the `--namespace` flag. In order for that to happen, this value cannot have a default. The default should assume the current namespace.

ex. in namespace `devops`
```
$ sensuctl check delete check-in-devops
$ Error: not found
```

nikkiki: Alright, I've had to work around some other changes after your global namespace PR. It requires another refactor here, but I think it'll get us the solution we're looking for.

cli/client/check.go
ccressent: Indeed! Sorry that's my bad.

PR #2766, opened 2019-03-04
Rework round robin

### LINKED ISSUE #1973: Round robin checks execute more often than expected when multiple subscriptions are specified
## Expected Behavior
Round robin checks will only execute 1/N times on a given agent, even in the presence of multiple subscriptions.

## Current Behavior
When a set of agents matches multiple round robin subscriptions, the round robin check will execute (1/N)*NUM_SUBSCRIPTIONS per agent.

## Possible Solution
Deduplicate subscriptions? We need to do some design work around this, about what should happen. What if the sets of agents are disjoint in terms of their subscriptions, etc. Difficult use case.

1. Create a round robin check that targets multiple subscriptions.
2. Create a set of agents that have all of those subscriptions.
3. Observe greater than expected round robin check execution.

## Context
Staging

### LINKED ISSUE #2699: Epic - Re-enabling round robin scheduling.
Epic to house the work required to re-enable round robin scheduling in Sensu Go. 

Proposal document: https://docs.google.com/document/d/1lxgTouMOCOBDJG5ri0Jp2G97banHFRDrRbxA6sj_ZEs/edit

Parent of #2431 
Parent of #2005 
Parent of #2735

echlebek:
## What is this change?

This PR reworks the round robin scheduling feature. Round robin scheduling now uses the new ringv2 package, and supports proxy check execution, interval and cron scheduling.

The `schedulerd` package has been given a substantial rework. I've chosen to favour code duplication over re-use, so that different schedulers can exist as separate datatypes. This makes switching scheduling modes easier, since schedulers can just be replaced.

The amount of I/O required for scheduling checks is now much smaller. Before, check scheduling would require a scan of all the entities in the store, for every check. Now, an entity cache is loaded when schedulerd starts up, and is maintained via watchers.

The ring package has been altered in the course of this work, and now supports TTLs for ring items, and also multiple items per ring advancement. Additionally, the storage layout of the ring is now simpler and requires fewer I/O operations.

Keepalived has been modified to add and remove items in scheduling rings - this allows the rings to contain the "right" elements for scheduling.

A future effort should rework ad-hoc check scheduling mechanism as a scheduler, instead of an executor, to better fit the scheduler/executor paradigm.

Closes #2699 
Closes #1973 

Added

* Several iterations of code/test/debug were required
* Ring interface needed altering as requirements became more clear
* Schedulerd needed refactoring to properly express round robin scheduling in the context of the other scheduling modes.

We need to update the documentation to tell people round robin scheduling is enabled again.

Substantial testing was performed, with a small Sensu cluster running in docker-compose. Proxy checks with round robin interval and cron scheduling were tested and found to be working as expected. New QA tests are also needed but not added yet.

I also took a lot of time to test that entities could be added and removed, and scheduling would keep working as expected.


backend/schedulerd/entity_cache.go
```diff
@@ -0,0 +1,206 @@
...
+	stop := sort.Search(len(cache), func(i int) bool {
+		return cache[i].Namespace >= endNS
+	})
+	if stop > len(cache) {
+		stop = len(cache)
+	}
+	return cache[start:stop]
+}
+
+func entityEQ(x, y *corev2.Entity) bool {
```
golangcibot: `entityEQ` is unused (from `deadcode`)

backend/schedulerd/check_watcher.go
```diff
@@ -117,9 +142,23 @@ func (c *CheckWatcher) handleWatchEvent(watchEvent store.WatchEventCheckConfig)
...
+			if err := c.startScheduler(check); err != nil {
+				logger.WithError(err).Error("unable to start check scheduler")
+			}
+		}
+		if sched.Type() == GetSchedulerType(check) {
+			logger.Info("restarting scheduler")
+			sched.Interrupt(check)
+		} else {
+			logger.Info("stopping existing scheduler, starting new scheduler")
+			sched.Stop()
```
golangcibot: Error return value of `sched.Stop` is not checked (from `errcheck`)

backend/agentd/session.go
```diff
@@ -66,7 +71,7 @@ type SessionConfig struct {
 // encounters a receive error.
 func NewSession(cfg SessionConfig, conn transport.Transport, bus messaging.MessageBus, store Store) (*Session, error) {
 	// Validate the agent namespace
-	ctx := context.TODO()
+	ctx, cancel := context.WithCancel(context.Background())
```
golangcibot: the cancel function is not used on all paths (possible context leak) (from `govet`)

backend/ringv2/ringv2.go
```diff
@@ -361,41 +409,20 @@ func (r *Ring) advanceRing(ctx context.Context, prevKv *mvccpb.KeyValue, numValu
...
-
-	if prevKv != nil && path.Base(string(prevKv.Key)) != path.Base(string(items[0].Key)) {
-		// Another client has already advanced the ring, there is no need to
-		// continue.
-		return repeatItems, nil
+	if len(items) < numValues+1 {
+		// There are fewer items than requested values
+		nextItem = items[0]
+	} else {
+		items = items[:len(items)-1]
```
golangcibot: SA4006: this value of `items` is never used (from `staticcheck`)

backend/agentd/session.go
echlebek: False positive; the cancel function is called in Stop()

echlebek:
Reviewers, please ignore the failing status from GolangCI as it is reporting a false positive.

backend/schedulerd/interface.go
```diff
@@ -0,0 +1,56 @@
+package schedulerd
+
+import corev2 "github.com/sensu/sensu-go/api/core/v2"
+
+// Scheduler is the abstract interface of a scheduler.
+type Scheduler interface {
```
nikkiki: This pleases me.

echlebek: Ah that reminds me I have to improve the docs for that type. "Scheduler is the abstract interface of a scheduler" is not great documentation. :)

backend/keepalived/keepalived.go
```diff
@@ -423,5 +441,27 @@ func (k *Keepalived) handleUpdate(e *types.Event) error {
...
+	if entity.EntityClass == corev2.EntityAgentClass {
+		// Refresh the rings that the entity is involved in
+		for _, sub := range entity.Subscriptions {
+			if strings.HasPrefix(sub, "entity:") {
+				// Entity subscriptions don't get rings
+				continue
+			}
+			ring := k.ringPool.Get(ringv2.Path(entity.Namespace, sub))
+			if e.Check.Timeout == 0 {
+				panic("0 timeout")
```
echlebek: I need to remove this panic.

backend/schedulerd/cron_scheduler.go
```diff
@@ -0,0 +1,125 @@
...
+
+func (s *CronScheduler) toggleSchedule() (stateChanged bool) {
+	defer s.setLastState()
+
+	if s.lastCronState != s.check.Cron {
+		s.logger.Info("cron schedule has changed")
+		return true
+	}
+
+	s.logger.Debug("cron schedule unchanged")
```
nikkiki: Since you already log the `schedule_type` field (awesome btw), perhaps this message can remain the same (regardless of the scheduler type), ex: `check schedule has not changed`. This limits a very minor amount of redundancy.

PR #2769, opened 2019-03-04
[WIP] Rough POC of handler annotations -> handler env vars

portertech:
This pull-request is intended to kick off a discussion about Handler configuration through annotations and Check/Entity configuration overrides. These changes translate Handler annotations to valid environment variables for the Handler to optionally read while executing for an Event. I will add to the pro/con list as we discuss/comment.

Pros:
- Consistent way of setting configuration values across object types

Cons:
- Already a way to set ENV variables for handlers (this adds a second way)

### DIFF
--- a/api/core/v2/util.go
+++ b/api/core/v2/util.go
@@ -16,6 +16,12 @@ func FakeHandlerCommand(command string, args ...string) *Handler {
 	env := []string{"GO_WANT_HELPER_HANDLER_PROCESS=1"}
 
 	handler := &Handler{
+		ObjectMeta: ObjectMeta{
+			Namespace:   "default",
+			Name:        "fake",
+			Labels:      make(map[string]string),
+			Annotations: make(map[string]string),
+		},
 		Command: trimmedCmd,
 		EnvVars: env,
 	}
--- a/backend/pipelined/handle.go
+++ b/backend/pipelined/handle.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/sensu/sensu-go/asset"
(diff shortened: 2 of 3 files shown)


portertech:
it is literally highway to the danger zone something _62fd3e946bde805b2df319f4fd2d0d59

calebhailey:
image

portertech:
Alternatively, we set a standard practice of setting Handler environment variables, and then override with Check/Entity annotations within the Event, example:

`env_vars["PAGERDUTY_API_TOKEN=abc"]`

`check.Annotations["sensu.io/plugins/pagerduty/api_token"] = "def"`

echlebek:
I like the idea of exporting our annotations as environment variables, but I'm concerned that the existing method of setting environment variables will cause confusion. This change makes annotations shadow names in the "env" field - potentially a breaking change for some users.

I think it's worth exploring the option of deprecating the existing env var mechanisms, in favour of one enabled by annotations. This would be a good candidate for inclusion in our next API version. We could expose it as a beta for the time being in order to garner feedback.

echlebek:
To elaborate on my previous comment:

We could expose this feature in a beta API, and use conditional logic in pipelined to make a breaking change without affecting core/v2 users.

Assumptions:
* Types know their API version (need to implement this, OR use wrapper type more pervasively)

Example: in pipelined, where the handler environment behaviour lives,
```
switch handler.GetAPIVersion() {
  case "v2":
    oldWay()
  case "v3beta1":
    newWay()
  default:
    return errors.New("unknown API version")
}
```

portertech:
@calebhailey @echlebek I remember the other driver, eliminate the need for template files on disk, e.g. 
https://github.com/sensu/sensu-email-handler/pull/15/files

portertech:
Closing this PR. I suspect discussion will start up again w/ further integration work.

PR #2770, opened 2019-03-05
Fix issue causing global flags to be ignored

### LINKED ISSUE #2345: Global/Persistent Flags are ignored if they come after command local flags
## Expected Behavior
`sensuctl entity list --format tabular --namespace devops` should return all entities within the devops namespace

## Current Behavior
The namespace is ignored so entities of the default namespace are returned.

## Possible Solution
Strangely if the namespace flag is passed before the format flag, it works (e.g. `sensuctl entity list --namespace devops --format tabular`).

I assume there's something weird with the format flag.

1. Run the command
2. Observe the result
3. ???
4. Profit

## Context

Found during QA

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): beta.8

### LINKED ISSUE #2731: sensuctl configure non-interactive: config-dir flag is ignored
## Expected Behavior
We should be able to use sensuctl configure with --non-interactive and --config-dir flags together.

## Current Behavior
When configuring sensuctl with sensuctl configure --non-interactive, --config-dir flag is ignored and config is always generated in ~/.config/sensu/sensuctl.
This doesn't happen when using interactive mode where --config-dir flag is correctly used.

1. Execute
```
sensuctl configure --non-interactive \
    --config-dir /tmp/sensuctl \
    --url "api_url" \
    --username "admin" \
    --password "P@ssw0rd!" \
    --namespace default \
    --format yaml
```
2. You can see that configuration has been generated in $HOME/.config/sensu/sensuctl instead of  /tmp/sensuctl

## Context
We need to be able to generate multiple sensuctl configurations because you can have more than one sensu cluster to manage.

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): 5.2.1
* Installation method (packages, binaries, docker etc.): Ubuntu packages
* Operating System and version (e.g. Ubuntu 14.04): Ubuntu 18.04

ccressent:
## What is this change?

When Load() is called, some sub-command local flags, such as
--format, are not registered yet and this leads to "unknown flags"
errors being returned by cobra. Such an error can throw off the cobra
parser, leading to all the flags appearing after the offending,
supposedly unknown flag to be ignored.

For now, we just ignore such "unknown flags" errors in order not to
potentially ignore other flags as a side effect. cobra will still store
the name/value of all the flags, even if they are registered later.

Unfortunately, a (rather big) refactor of the CLI is probably the
only way to get around that. Such a refactor would involve completely
reordering the way we define commands, load the configuration files
and override properties with command line flags.

Fixes #2345.
Fixes #2731.

Yes, it has been added to the `CHANGELOGS/unreleased` directory.

No.

Many! There were many red herrings during this investigation, and I fell for most of them, wasting a lot of time.

Until I came across the ability to whitelist certain parse errors in latest cobra releases, this lead me to trying many different approaches that were all ultimately rather unsatisfactory in the end: they would introduce other breakage, bad coupling, ...

Should we rework `sensuctl` in the future, we should probably embrace a cobra/viper combo to manage the fact that the configuration is coming from several sources (configuration files, command line flags, ... maybe environment variables?) that need to be merged.

No changes required.

Given the lack of tests for the CLI, I've manually tested various flag combinations, in various orders, to verify that the output is what we expect every time.

I mostly tested the `--namespace`, `--trusted-ca-file` and `config-dir` flags with the `entity {list,info}`, `hook list` and `filter list` sub-commands.


cli/commands/asset/create.go
```diff
@@ -32,7 +28,14 @@ func CreateCommand(cli *cli.SensuCli) *cobra.Command {
 				_ = cmd.MarkFlagRequired("url")
 			}
 		},
-		RunE: exec.Run,
+		RunE: func(cmd *cobra.Command, args []string) error {
```
palourde: Out of curiosity, what's the difference here between these two formats?

ccressent: I'm assuming you're asking about the difference here between assigning `exec.Run` vs assigning an anonymous function. No difference!

The important change here is that I moved the creation of `executor` and the namespace it holds from command creation time to command run time.

I made that change before converging on my actual, final solution of using the parse error whitelisting facilities of newer versions of Cobra. I suspect that this specific change in `asset/create.go` might be unnecessary now: since the commands are created after the flags are parsed, `cli.Config.Namespace()` will probably have the right value.

That being said, I think keeping this change makes sense because it makes it so the command uses the namespace that's defined when it _runs_ rather than when it's _defined_. An important difference, especially if there is a refactor that creates all the commands before parsing the flags!

cli/commands/configure/configure.go
```diff
@@ -115,11 +115,11 @@ func Command(cli *cli.SensuCli) *cobra.Command {
 	}
 
 	_ = cmd.Flags().BoolP("non-interactive", "n", false, "do not administer interactive questionnaire")
-	_ = cmd.Flags().StringP("url", "", cli.Config.APIUrl(), "the sensu backend url")
```
nikkiki: What if a user runs `sensuctl configure` after they have initially configured? If they leave these flags blank, wouldn't their current configuration be overwritten by the defaults?

cli/commands/configure/configure.go
```diff
@@ -115,11 +115,11 @@ func Command(cli *cli.SensuCli) *cobra.Command {
...
+	_ = cmd.Flags().StringP("url", "", config.DefaultAPIURL, "the sensu backend url")
```
nikkiki: What if a user runs `sensuctl configure` after they have initially configured? If they leave these flags blank, wouldn't their current configuration be overwritten by the defaults?

ccressent: Great catch @nikkiki!
I'm going to revert this commit: https://github.com/sensu/sensu-go/pull/2770/commits/06e975d86342bf810ddbdba206b55fa238e0e067; it was a mistake on my part.

PR #2783, opened 2019-03-07
Don't store events with a zero timestamp

### LINKED ISSUE #2636: POST /events do not create a timestamp (must be provided)
From the Classic to Go gap analysis:

> API Endpoint that allows users to inject events into the event pipeline on an ad-hoc basis, similar to the /request POST endpoint in Sensu 1.x. Currently in Sensu Go, we have POST and PUT endpoints for the Events API, but they don’t generate their own timestamps. (I haven’t tested whether events created by these endpoints are processed by handlers.)

echlebek:
## What is this change?

Events that are sent to the store with a zero value as a timestamp
will now be given the current time instead.

Closes #2636 

Integration test

### DIFF
--- a/CHANGELOGS/unreleased/2636-event-timestamp.md
+++ b/CHANGELOGS/unreleased/2636-event-timestamp.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Fixed a bug where events could be stored without a timestamp.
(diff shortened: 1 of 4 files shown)


portertech:
Are we doing Event validation at the edge that could be rejecting Events without a timestamp?

echlebek:
I've added a unit test for event validation, and ensured we aren't doing any ad-hoc validation of events.

portertech:
dog-dog

PR #2785, opened 2019-03-11
Don't allow metrics to be stored in any case

### LINKED ISSUE #2743: Avoid writing event.metrics to etcd (only store record).
## Expected Behavior
As a user of the sensu-prometheus-collector check, I would like the check to only result in metrics events, so that the metrics are not stored in Sensu.

## Current Behavior
All checks are stored along with metrics data in Sensu.

## Possible Solution 1
Add a flag called "metrics_only" to check configurations, so that successful check executions are only transmitted as metrics events, and thus not stored. Unsuccessful checks are still transmitted as regular check events, so that users can easily detect failure.

## Possible Solution 2
Add a flag called "no_store_metrics" to check configurations, so that extracted metrics are not stored along with the check event.

## Context
A user is attempting to push a large amount of metrics through Sensu, but the default etcd request size limit is hampering that effort.

echlebek:
## What is this change?

This commit explicitly nils out metrics when events arrive at the
store, in order to ensure that metrics are never stored in etcd.

Care is taken to not write to the pointer that is given to the
store. To this end, some unrelated existing code in the same
UpdateEvent method has also been modified to be safer.

Closes #2743 

Included

This change brings us in line with the documentation.

A new integration test.

### DIFF
--- a/backend/store/etcd/event_store.go
+++ b/backend/store/etcd/event_store.go
@@ -160,7 +160,7 @@ func (s *Store) GetEventByEntityCheck(ctx context.Context, entityName, checkName
 
 // UpdateEvent updates an event.
 func (s *Store) UpdateEvent(ctx context.Context, event *types.Event) error {
-	if event.Check == nil {
+	if event == nil || event.Check == nil {
 		return errors.New("event has no check")
 	}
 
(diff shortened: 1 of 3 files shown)


backend/store/etcd/event_store_test.go
```diff
@@ -112,3 +112,24 @@ func TestEventStorage(t *testing.T) {
 		assert.Error(t, err)
 	})
 }
+
+func TestDinnaeStoreMetrics(t *testing.T) {
```
nikkiki: I don't think I'm smart enough to know what `Dinnae` means..

echlebek: https://en.wikipedia.org/wiki/Scotticism

ccressent: I had to consult that wiki page too :(
Should it just read `TestDoNotStoreMetrics()`?

backend/store/etcd/event_store.go
```diff
@@ -161,7 +161,7 @@ func (s *Store) GetEventByEntityCheck(ctx context.Context, entityName, checkName
 
 // UpdateEvent updates an event.
 func (s *Store) UpdateEvent(ctx context.Context, event *types.Event) error {
-	if event.Check == nil {
+	if event == nil || event.Check == nil {
 		return errors.New("event has no check")
```
ccressent: Is that the correct error for the `event == nil` case? I think it should be something along the lines of "cannot update nil event" or something.

echlebek: The == nil case is really just a guard against a panic, not sure it deserves its own error case.

PR #2786, opened 2019-03-12
Add Dashboard Asset Collections

jamesdphillips:
Required by sensu/sensu-enterprise-go#274.

## What is this change?

Introduces notion of asset collections to the dashboard daemon.

Further,

* collections allow one or more packs of assets to be served by the daemon.
  * If items with duplicate paths exist in the packs, the last pack applied takes precedence. This should allow a pack to override previous a module/graphic/etc.
* Splits existing asset HTTP handler into two.
  * The handler mounted at `/static` is responsible for all static assets; it assumes that asset pack(s) follow a convention that all items in the static directory can be considered immutable. It then ensures that appropriate HTTP headers are set so that the files are cached indefinitely.
  * The other handler is responsible for all other files. The HTTP headers used ensure that the files are never cached and will fallback to serving `index.html` if the given path does not exist in any of the asset packs. 
* Adds endpoint exposing all accessible files,
  * among other things this will be useful for implementing updates with service workers.

Required for sensu/sensu-enterprise-go#229

Yup!

Not at this time!

Nada.

Yes, and no additional documentation is required.

Unit tests and running the backend independently.


backend/dashboardd/asset/collection.go
```diff
@@ -0,0 +1,143 @@
...
+	atom := &atomic.Value{}
+	atom.Store([]http.FileSystem{})
+	return &Collection{
+		bs: atom,
+		mu: &sync.Mutex{},
+	}
+}
+
+// Extend collection with given set of assets
+func (a *Collection) Extend(b http.FileSystem) {
```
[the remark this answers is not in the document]
jamesdphillips: I was mulling over the possibility, that the collection could be extended by the asset manager, in which case we could see concurrent writes. Albeit, this is unlikely to be a feature we introduce in the near term.

backend/dashboardd/dashboardd.go
```diff
@@ -240,3 +228,65 @@ func newBackendProxy(APIURL string, TLS *types.TLSOptions) (*httputil.ReversePro
...
+	w.Header().Set("Content-Type", "application/json")
+
+	// Marshal
+	bytes, err := json.Marshal(d)
+	if err != nil {
+		return err
+	}
+
+	// Write
+	if _, err = w.Write(bytes); err != nil {
```
golangcibot: SA9003: empty branch (from `staticcheck`)

backend/dashboardd/asset/collection.go
```diff
@@ -0,0 +1,155 @@
...
+			return dir, err
+		}
+		stat, err := f.Stat()
+		if err != nil || !stat.IsDir() {
+			return dir, err
+		}
+		if dir.name == "" {
+			dir.name = stat.Name()
+		}
+		if es, _ := f.Readdir(-1); es != nil {
```
golangcibot: S1031: unnecessary nil check around range (from `gosimple`)

Issue #2795, opened 2019-03-13
[Tessen] API core


nikkiki:
Implement versioned API endpoints such as `/api/core/v2/tessen` to GET and PUT the Tessen configuration.

Depends on https://github.com/sensu/sensu-go/issues/2789

Issue #2789, opened 2019-03-13
[Tessen] Store


nikkiki:
Implement an etcd store for Tessen configuration. The store path should be `sensu.io/tessen` with functions to create and update the stored configuration. These values can be set by the user through the API (see other tickets) or through backend configuration options. Implement these configuration options as well.

```
type TessenConfig struct {
    // OptOut is the opt-out status of Tessen data collection
    OptOut bool `json:"opt_out"`
    // Upper is the upper bound frequency, in minutes, that Tessen data will be collected (default 360)
    Upper uint32 `json:"upper"`
    // Lower is the lower bound frequency, in minutes, that Tessen data will be collected (default 1)
    Lower uint32 `json:"lower"`
}
```

PR #2818, opened 2019-03-19
Implements tessen store and api

nikkiki:

## What is this change?

- Implements a Tessen store at /sensu.io/tessen
- Implements /api/core/v2/tessen GET
- Implements /api/core/v2/tessen PUT

Closes https://github.com/sensu/sensu-go/issues/2789
Closes https://github.com/sensu/sensu-go/issues/2795

Yas.

Nah.

Nah.

Tracking docs issue: https://github.com/sensu/sensu-docs/issues/1320

All covered in unit and integration tests. You can GET and PUT /tessen to verify from an e2e perspective.


backend/tessen/config.go
```diff
@@ -0,0 +1,34 @@
...
+
+	// UpperBound is the maximum frequency that tessen will phone home
+	UpperBound = 360
+
+	// DefaultTessenFrequency is the default frequency at which tessen will phone home
+	DefaultTessenFrequency = 1
+)
+
+// Config is the tessen configuration
+type Config struct {
```
palourde: I might have missed some discussions around that so I apologize in advance but my understanding was that we would create a new API group for Tessen, so I expected this to live in a folder like `api/tessen/v2` and see the Tessen API routes under the same path, which isn't the case.

backend/store/etcd/tessen_store.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/sensu/sensu-go/backend/tessen"
+)
+
+func getTessenPath() string {
+	return fmt.Sprintf("%s/tessen", EtcdRoot)
+}
+
+// CreateOrUpdateTessenConfig creates or updates the tessen configuration
+func (s *Store) CreateOrUpdateTessenConfig(ctx context.Context, config *tessen.Config) error {
+	return CreateOrUpdate(ctx, s.client, getTessenPath(), "", config)
```
palourde: 👍👍👍 for using the generic store functions haha

backend/tessen/config.go
nikkiki: You're totally right, we initially discussed and spec'd a versioned API. We had another small conversation regarding leader election and some other tessen details. We concluded versioning might be a little overkill, since it's not necessarily a consumer resource, and right now we only let the user determine opt-out status. Very similar to the `/health` endpoint. Do you have any strong feelings for versioning it, because we can def discuss it further!

backend/store/etcd/tessen_store.go
nikkiki: The less code, the better! Thanks for providing them 😉

backend/tessen/config.go
palourde: I'm kind of on the fence. I agree that if we only let users determine opt-out status, it's definitively overkill. I guess it ultimately depends on what the users can control; are we going to allow customization of the lower and upper bound frequencies? What about the HTTP proxy stuff mentioned in the specs? If so, we are limiting configuration to the `sensuctl tessen` subcommand (in contrast to `sensuctl create`).

Also, it might not be important but it prevents targeting the Tessen config in RBAC; you'll need access to everything to manage non-resources endpoints like `/tessen`, which is probably fine, but just something to keep in mind.

nikkiki: I feel that for sure. Right now, the only thing we're letting the user determine is opt-out status, but if in the future if we do need to add fields, it might better if we already have versioning concepts in place. I assumed tessen would not be gated by RBAC, but we never explicitly discussed that so I could be wrong. @portertech @echlebek do you have any thoughts to add on all of this?

echlebek: I think only admins should be able to enable and disable tessen.

backend/tessen/config.go
```diff
@@ -0,0 +1,34 @@
...
+func ValidateFrequency(freq uint32) error {
+	if !(freq >= LowerBound && freq <= UpperBound) {
+		return fmt.Errorf("tessen frequency must be in between the lower and upper bound limits (%d-%d)", LowerBound, UpperBound)
+	}
+	return nil
+}
+
+// DefaultConfig returns the default tessen configuration
+func DefaultConfig() *Config {
+	return &Config{}
```
echlebek: Should this be populated with default values?

nikkiki: The default value of the `OptOut` field is already false.. unless the empty struct is not programmatically equivalent?

echlebek: I was thinking about the frequency values

nikkiki: Frequency is hardcoded at once per minute, and a user cannot modify it. The frequency can only be modified or overwritten if the Tessen service returns a new frequency within the header. I figured the current frequency would be a field belonging to the Tessen daemon. Does that make sense?

backend/tessen/config.go
nikkiki: This has all been resolved by versioning the tessen api 😄

PR #2819, opened 2019-03-20
Add queueing to the agent /events API

### LINKED ISSUE #2801: Implement queueing for the agent /events API
As a user, I can send a Sensu Go Event to the agent /events API and expect the event to be accepted for processing even if the agent is not connected to the backend.

Acceptance criteria: 

1. The /events API responds with a '202' status code when valid events are POSTed.
1. Events POSTed to the /events API are persistently queued before being sent to the backend.
1. If the agent is not connected a backend, it will retry sending queued events in FIFO order until the queue is exhausted.
1. If an event fails to be sent to the backend, it will be placed back in the queue in its original position.
1. All received events will be logged.

### LINKED ISSUE #2805: Add Quality-of-Service features to agent event relaying
As an agent, when I connect to a backend and send it queued events, I will not overwhelm it with requests. (rate limiting)

As an agent, when I am trying to connect to a backend that is unavailable, I will not retry constantly. (Avoid the thundering herd problem.) Instead, I will try again with exponential backoff.

Acceptance Criteria: 
1. Provide a way of limiting on events per second (must be configurable). 
1. Include a reasonable default configuration for rate limiting.
1. Include configuration for exponential backoff.

echlebek:
## What is this change?

This PR adds queueing to the /events API. Agents will persist all messages received from /events to disk, before sending those messages to their backend.

The queueing is based on a library I wrote a few years ago called lasr, designed to be a durable, transactional, embedded queue. The library has been forked to depend on `coreos/bbolt` instead of `boltdb/bolt`, as that avoids an additional dependency for sensu-go.

Closes #2801 
Closes #2805 

We want agents to be able to buffer messages sent to /events when they are disconnected from backends. Additionally, the events need to persist when agents are restarted.

Yes

The lasr queue library is based on bolt DB, which implements an ever-growing database - deleting from the database will never make the file on disk shrink. To deal with this, the lasr library supplies a Compact method.

This method works by copying the entire database to a new file. It's a bit tricky to know when a good time to do that is.

On-line queue compaction is hard to get right. For now, I've set up the agent to compact the queue on startup, so that users can easily trigger compaction by restarting the agent.

New documentation is required.

Integration and unit testing, and a good amount of manual testing. New QA materials will be added.


agent/agent.go
```diff
@@ -60,10 +62,13 @@ type Agent struct {
 	systemInfo      *v2.System
 	systemInfoMu    sync.RWMutex
 	wg              sync.WaitGroup
+	db              *bolt.DB
```
golangcibot: `db` is unused (from `structcheck`)

agent/agent.go
```diff
@@ -14,7 +14,9 @@ import (
 	"strings"
 	"sync"
 
+	bolt "github.com/coreos/bbolt"
```
palourde: I just realized the asset manager uses `bolt "go.etcd.io/bbolt"`. Should we create an issue to migrate to only one of these repo?

agent/queue.go
```diff
@@ -0,0 +1,51 @@
...
+	return buf.Bytes()
+}
+
+func decompressMessage(message []byte) []byte {
+	dst := new(bytes.Buffer)
+	src, _ := gzip.NewReader(bytes.NewReader(message))
+	defer src.Close()
+	_, err := io.Copy(dst, src)
+	if err != nil {
+		panic(err)
```
palourde: Could we avoid panicking here and just return empty bytes?

agent/agent.go
echlebek: Oh, I meant to use that one. I'll change it for this PR.

agent/queue.go
echlebek: Yes, the panic was actually just an assertion that I wanted while testing. Good catch!

agent/api.go
```diff
@@ -50,6 +54,57 @@ func healthz(connected func() bool) http.HandlerFunc {
...
+			if !ok {
+				return
+			}
+			msg := &transport.Message{
+				Type:    transport.MessageTypeEvent,
+				Payload: decompressMessage(message.Body),
+				SendCallback: func(err error) {
+					if err != nil {
+						logger.WithError(err).Error("couldn't send queued message, retrying")
+						_ = message.Nack(true)
```
nikkiki: I'm not sure what this function returns without looking at the dependency, but is it anything that would be useful to log?

echlebek: If there is an error writing to disk, the message will be left in the unacked state, and the error will be returned. It will eventually make its way back into the ready state on subsequent queue operations.

There's not really much point in logging the error as subsequent queue operations will just return the same error if there is a problem with the disk.

PR #2822, opened 2019-03-21
[Web UI]  Create separate vendor and lib DLL bundles

10xjs:
## What is this change?

This change breaks the single Web UI client bundle into multiple bundles.

Core components and shared dependencies are provided by "lib" and "vendor" bundles that can be referenced by downstream webpack builds allowing them to add additional functionality to the core Web UI app.

This changes enables the creation of the Sensu Enterprise Web UI. (https://github.com/sensu/sensu-enterprise-go/issues/229)

Yes.

No.

This feature required major changes of the Web UI codebase and build directory structures.
The dashboard assets go generate logic does not support the changes to the build directory structure - regenerated assets are intentionally omitted from this PR to avoid breaking things.

The `yarn start` script no longer runs a build and does not reload changes while the server is running. It is necessary to `yarn build` and re-run `yarn start` to see any changes in the browser. Depending on the scope of the changes, it may be possible to run `yarn build:app` or `yarn build:lib && yarn build:app` to speed things up.

No,

By running `yarn test` and by manual Web UI testing.


dashboard/src/app/static/index.html
```diff
@@ -4,17 +4,30 @@
...
     <meta name="format-detection" content="telephone=no">
 
     <!-- https://developers.google.com/web/fundamentals/engage-and-retain/web-app-manifest/ -->
     <meta name="theme-color" content="#2D3555">
-    <link rel="manifest" href="/manifest.json">
-    <link rel="shortcut icon" href="/favicon.ico">
+    <link rel="manifest" href="${require('./manifest.macro.js')}">
+
+    <!-- Windows 8.1 + IE11 and above -->
+    <meta name="msapplication-config" content="${require('./browserconfig.macro.js')}" />
```
jamesdphillips: neato

dashboard/src/app/component/partial/SilencesList/SilencesListItem.js
```diff
@@ -115,20 +116,18 @@ class SilencesListItem extends React.Component {
                 paddingTop: 8, // one spacing unit
               }}
             >
-              <Maybe value={silence.creator}>
```
jamesdphillips: I think this may be a regression that was introduced in the rebase. Was introduced in #2817.

10xjs: good catch! This is one of two conflicts I had to resolve

PR #2823, opened 2019-03-22
[Web] Update vulnerable dev packages

jamesdphillips:
## What is this change?

**NOTE:** Before any (understandable) excitement arises, none of the vulnerable packages had any attack surface that would have affected to the dashboard itself; only the dev tooling was "vulnerable."

**Removes** unused package `react-dev-tools`

**Updates**

* lodash to latest
* js-yaml to latest
* jest to ^23.5

**New**

Adds `yarn audit-deps` which prints out any vulnerable packages in a clean format. Would be nice to hook this up to CI in the future.

Frankly the vulnerable banner on Github is really poor optics.

Ab-so-lutely.

nada

nada

Not required.

`yarn audit-deps`


dashboard/package.json
```diff
@@ -110,20 +111,19 @@
     "http-proxy-middleware": "^0.18.0",
     "intl": "^1.2.5",
     "intl-relativeformat": "^2.1.0",
-    "jest": "^22.4.3",
-    "jest-enzyme": "^6.0.0",
-    "jest-fetch-mock": "^1.5.0",
-    "jest-localstorage-mock": "^2.2.0",
+    "jest": "^23.6.0",
```
jamesdphillips: Updating to `^25.4` would have required updating `babel` as well, which, I felt was out of the scope of this PR.

10xjs: Upgrading to babel 7 is something I would like to do ASAP tho. Doing so shouldn't cause us any problems.

PR #2824, opened 2019-03-22
[Web] [Chore] Update babel to ^7

jamesdphillips:
**TODO** switch base branch to `master` after #2823 is merged

## What is this change?

Updates babel to ^7.4 [[CHANGELOG](https://github.com/babel/babel/blob/master/CHANGELOG.md#v742-2019-03-21)]

Also updated to work w/ babel v7

* `jest` ^24.5 [[CHANGELOG](https://github.com/facebook/jest/blob/master/CHANGELOG.md)]
* `eslint-import-resolver-babel-module` ^5.0.1 [[CHANGELOG](https://github.com/tleunen/eslint-import-resolver-babel-module/blob/master/CHANGELOG.md)]
* `eslint` ^5.15.0 [[CHANGELOG](https://github.com/eslint/eslint/blob/master/CHANGELOG.md)]
* `css-loader` ^2.0 [[CHANGELOG](https://github.com/webpack-contrib/css-loader/blob/master/CHANGELOG.md)]

Prereq for sensu/sensu-enterprise-go#76.

Yep!

Never.

Not in particular.

None required.

* `yarn test`
* `yarn lint`
* manual testing


dashboard/src/lib/component/base/Duration/Duration.js
```diff
@@ -18,14 +18,16 @@ class Duration extends React.PureComponent {
...
+    if (duration < 10000) {
       // ex. 8.02s
       const formatter = this.numberFormatter({
         minimumFractionDigits: 2,
         maximumFractionDigits: 2,
       });
       return `${formatter.format(duration / 1000)}s`;
-    } else if (duration < 60000) {
+    }
+    if (duration < 60000) {
```
dabria: Should this remain an if/else if? Otherwise we will run both of these blocks if `duration is < 6000` - and it looks to me as though it overwrites what occurs in the first block. It still gives the correct results, but it's just extra cycles.

dashboard/package.json
```diff
@@ -37,6 +37,17 @@
...
     "@10xjs/date-input-controller": "0.1.6",
     "@10xjs/form": "^0.1.7",
+    "@babel/core": "^7.0.0",
+    "@babel/plugin-proposal-class-properties": "^7.0.0",
+    "@babel/plugin-proposal-object-rest-spread": "^7.0.0",
+    "@babel/plugin-syntax-dynamic-import": "^7.0.0",
+    "@babel/plugin-transform-destructuring": "^7.0.0",
+    "@babel/plugin-transform-react-jsx": "^7.0.0",
+    "@babel/plugin-transform-react-jsx-self": "^7.0.0",
+    "@babel/plugin-transform-react-jsx-source": "^7.0.0",
```
jamesdphillips: I believe `-self` and `-source` can be removed but I didn't want to until I was sure that the error modal was not using it(?)

dashboard/src/lib/util/checkStatus.js
```diff
@@ -12,9 +12,11 @@ const unknown = "unknown";
 export function statusCodeToId(st) {
   if (st === 0) {
     return success;
-  } else if (st === 1) {
+  }
+  if (st === 1) {
     return warning;
-  } else if (st === 2) {
+  }
+  if (st === 2) {
```
dabria: Could probably make this a switch, but that's my personal preference.

dashboard/src/lib/component/base/Duration/Duration.js
dabria: Oh wait, sorry I had my signs backwards

dabria: Wait, it still stands. Sorry, still working through my tea this morning!

jamesdphillips: I believe it's the [`no-else-return`](https://eslint.org/docs/rules/no-else-return#allowelseif-false) rule that made this change. It's a bit odd at first glance but in the interest of consistency I kept it.

dabria: This feels wrong to me, but if consistency is that important...

dabria: I didn't see there was a return in the statement, retract my comment.

dashboard/package.json
10xjs: These are now implicitly added by https://babeljs.io/docs/en/babel-preset-react and aren't necessary here

PR #2827, opened 2019-03-22
Use contexts for cancellation

### LINKED ISSUE #2533: agent retry logic (first connection) on multiple backends
Looks like we might want to return `err` instead of `nil` on `L381`, so that it can be logged in `L390`. That code block appears unreachable.
https://github.com/sensu/sensu-go/blob/e3421f7ed050bfb3310f83a28aed5b9d85b4be64/agent/agent.go#L377-L391

I also noticed that on an agents first connection, it only attempts retry logic on one backend (provided by the `RandomBackendSelector`). I'm wondering if it would be beneficial for that agent to try connecting to other backends until its successful.. unless this isn't the actual/intended behavior already. https://github.com/sensu/sensu-go/blob/e3421f7ed050bfb3310f83a28aed5b9d85b4be64/agent/agent.go#L201

### LINKED ISSUE #2756: Feature Request: Cancel Asset download and running check commands when agent recieves SIGINT
sensu-agent run from cmdline will wait for asset download and check commands to complete when using CTRL-C.  The wait can be several minutes long (depending on external factors like asset size and network bandwidth), and this makes it appear as if the agent is "stuck".  

 
## Expected Behavior
CTRL-C results in clean agent shutdown within reasonable amount of time. say 10 seconds

## Current Behavior
CTRL-C while asset is downloading causes agent to appear to "hang" until asset download is complete without any indication as to what is going on.

## Possible Solution
Refactor signal handling so that client processes such as asset download and check commands can be passed SIGINT  and do a graceful aboirt/clean up as necessary so agent can exit more quickly.

1. setup check using "large" asset
2. run sensu-agent from cmdline with subscription necessary to run check
3. wait for asset to start download ... download cached under /tmp/
4. attempt to quit agent using Ctrl-C
5. wait for asset download to complete

## Context
I originally came across this issue when testing the asset download timeout fix in PR #2746.  The timeout fix is unrelated but makes agent "hang" on CTRL-C more obvious.

echlebek:
## What is this change?

Use contexts for cancellation through the agent application. This
means threading a context through components that are also used in
the backend, so the backend has been nominally modified as well.

Note that cancellation problems still exist in the backend and will
require a larger, focused effort to resolve.

The Agent Stop() method has been removed, and Agent Run() now
blocks until the agent context is cancelled.

The asset manager now accepts a context, and will stop trying to
retrieve assets when the context is cancelled. This prevents
problems with asset managers never stopping, when assets cannot be
retrieved.

The agent's connection manager now stops trying to reconnect when
a context cancellation is observed.

Agents now shut down cleanly when SIGTERM and SIGINT are received.
All in-progress HTTP connections are given 1s to shut down and are
then terminated forcibly.

Agents still block on SIGTERM and SIGINT, but the longer waits for
shutdown should now be resolved.

Closes #2756 
Closes #2533 

Yes, added.

Several components required refactoring to accomplish this.

No documentation changes are required.

Automated testing covers running the agent quite well. I also did manual testing to ensure that agents terminate properly when asset download is in-flight.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -9,6 +9,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"log"
 	"net/http"
 	"os"
 	"strings"
@@ -48,7 +49,6 @@ type Agent struct {
 	config          *Config
 	connected       bool
 	connectedMu     sync.RWMutex
-	context         context.Context
 	entity          *v2.Entity
 	executor        command.Executor
 	handler         *handler.MessageHandler
@@ -57,7 +57,6 @@ type Agent struct {
 	inProgressMu    *sync.Mutex
 	statsdServer    *statsd.Server
 	sendq           chan *transport.Message
-	stopping        chan struct{}
 	systemInfo      *v2.System
 	systemInfoMu    sync.RWMutex
 	wg              sync.WaitGroup
@@ -67,19 +66,14 @@ type Agent struct {
 // NewAgent creates a new Agent. It returns non-nil error if there is any error
 // when creating the config.CacheDir.
 func NewAgent(config *Config) (*Agent, error) {
-	ctx := context.TODO()
-	ctx, cancel := context.WithCancel(ctx)
 	agent := &Agent{
 		backendSelector: &RandomBackendSelector{Backends: config.BackendURLs},
-		cancel:          cancel,
-		context:         ctx,
 		connected:       false,
 		config:          config,
 		executor:        command.NewExecutor(),
 		handler:         handler.NewMessageHandler(),
 		inProgress:      make(map[string]*v2.CheckConfig),
 		inProgressMu:    &sync.Mutex{},
-		stopping:        make(chan struct{}),
 		sendq:           make(chan *transport.Message, 10),
 		systemInfo:      &v2.System{},
 	}
@@ -119,7 +113,7 @@ func (a *Agent) refreshSystemInfo() error {
 	return nil
 }
 
-func (a *Agent) refreshSystemInfoPeriodically() {
+func (a *Agent) refreshSystemInfoPeriodically(ctx context.Context) {
 	defer logger.Debug("shutting down system info collector")
 	ticker := time.NewTicker(time.Duration(DefaultSystemInfoRefreshInterval) * time.Second)
 	defer ticker.Stop()
@@ -130,7 +124,7 @@ func (a *Agent) refreshSystemInfoPeriodically() {
 			if err := a.refreshSystemInfo(); err != nil {
 				logger.WithError(err).Error("failed to refresh system info")
 			}
-		case <-a.stopping:
+		case <-ctx.Done():
 			return
 		}
 	}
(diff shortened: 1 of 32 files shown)


agent/agent.go
```diff
@@ -171,54 +170,53 @@ func (a *Agent) Run() error {
...
 		a.connected = true
 		a.connectedMu.Unlock()
 
-		done := make(chan struct{})
-
-		go a.receiveLoop(conn, done)
-		a.sendLoop(conn, done)
+		ctx, cancel := context.WithCancel(ctx)
+		go a.receiveLoop(ctx, cancel, conn)
+		a.sendLoop(ctx, cancel, conn)
```
golangcibot: Error return value of `a.sendLoop` is not checked (from `errcheck`)

agent/agent.go
```diff
@@ -48,7 +49,6 @@ type Agent struct {
 	config          *Config
 	connected       bool
 	connectedMu     sync.RWMutex
-	context         context.Context
```
ccressent: Is the `cancel` field above still needed?

agent/check_handler_env_test_windows.go
```diff
@@ -29,7 +30,7 @@ func TestEnvVars(t *testing.T) {
 	agent.sendq = ch
 
 	entity := agent.getAgentEntity()
-	agent.executeCheck(request, entity)
+	agent.executeCheck(context.TODO(), request, entity)
```
ccressent: Why `context.TODO()` here and `context.Background()` in the unix version?

agent/check_handler_internal_test.go
```diff
@@ -43,13 +44,13 @@ func TestHandleCheck(t *testing.T) {
 	agent.inProgressMu.Lock()
 	agent.inProgress[checkKey(request)] = request.Config
 	agent.inProgressMu.Unlock()
-	assert.Error(agent.handleCheck(payload))
+	assert.Error(agent.handleCheck(context.TODO(), payload))
```
ccressent: Just curious about the choice of `context.TODO()` over `context.Background()`. They boil down to the same context, but I would have gone with `Background()` here.

nikkiki: Yeah my interpretation of the docs was that `context.Background()` is typically intended for tests (or the main function). They both return the same empty context, so its definitely a small thing!

agent/agent.go
```diff
@@ -340,51 +340,40 @@ func (a *Agent) StartAPI() {
...
 		Multiplier:           10,
+		Ctx:                  ctx,
 	}
 
-	if err := backoff.Retry(func(retry int) (bool, error) {
+	err := backoff.Retry(func(retry int) (bool, error) {
+		url := selector.Select()
 		c, err := transport.Connect(url, tlsOpts, header)
 		if err != nil {
 			logger.WithError(err).Error("reconnection attempt failed")
```
nikkiki: Is returning `false, nil` intentional here, or would we rather return `false, err`?

agent/agent.go
echlebek: Nope, I'll remove it

agent/check_handler_env_test_windows.go
echlebek: No particular reason

agent/check_handler_internal_test.go
echlebek: My thinking at the time was that we might want to cancel the context when the test terminates one day.

agent/agent.go
echlebek: A quirk in our retry library means that we cannot return non-nil error without terminating the retry loop.

https://github.com/sensu/sensu-go/blob/master/util/retry/retry.go#L74

nikkiki: Oh interesting.. thanks for clarifying!

PR #2830, opened 2019-03-25
Add pagination support to the API

### LINKED ISSUE #2604: Add pagination to the HTTP API
At the moment, I can't find clear guidance on pagination.  Will the Sensu Go API simply not use pagination?  Will it be added at some point with breaking changes to code written with the assumption of no paging?

I didn't see an issue on the API side, figured it would be worth having something others can search and find

## Expected Behavior

* Pagination behavior is documented
* There is a roadmap of sorts for API basics, if there are plans in place (paging, auth in particular)

## Current Behavior

* No docs on paging, presumably no support for paging?

## Possible Solution

* Add paging, or let us know if it's not something you plan to implement (eek!)?

Thanks!

ccressent:
## What is this change?

This change adds pagination support to the REST API. All the endpoints that can
be listed now accept 2 optional query parameters: `limit` and `continue`.
`limit` is the maximum number of objects to return per page (the page size)
while continue is an opaque token that allows resuming the pagination from a
previous point.

To paginate, one first issues a request with only the `limit` parameter set to a
non-zero value. If the `Sensu-Continue` header is set in the response, that
means that there are more objects to be fetched and the value held by that
continue header should be passed as the `continue` value in subsequent API calls.
By doing so until no `Sensu-Contiue` header is returned, one can paginate
through sets of objects.

Closes #2604.

Yes.

No.

Instead of passing the page size and continue token through the context all the
way down to the store, we decided to have the store take in and return
pagination parameter directly. This makes the store API much nicer to work with,
but it lead to an awful lot of code changes!

The stores could really benefit from using the "generic" object store we now
have: that would greatly reduce the amount of duplicated logic and make for a
more consistent handling of all resources.

No yet; this is tracked by https://github.com/sensu/sensu-docs/issues/1297.

I've added unit tests for all the controllers and pagination specific
integration tests for all the stores.

Additionally, I've tested paginating through most object types with the API
manually with `curl`.


palourde:
We also need a patch to `sensu-enterprise-go` for the new list argument

PR #2832, opened 2019-03-25
[Web UI] Set up consumable npm package

10xjs:
## What is this change?

Update built npm package (currently the local `dashboard.tgz`) to support downstream web-ui variant builds.

Required for https://github.com/sensu/sensu-enterprise-go/pull/290.

No.

No.

Nothing worth noting.

No.

The app builds and runs as it did before. `yarn test` passes


dashboard/config/base.webpack.config.js
```diff
@@ -117,16 +120,23 @@ export default ({
...
+            test: /\.(mjs|js|jsx)$/,
+            include: [
+              path.join(root, "src"),
+              path.join(root, "node_modules/@sensu/web-ui/src"),
+            ],
             loader: require.resolve("babel-loader"),
             options: {
+              babelrcRoots: [
+                root,
+                path.join(root, "node_modules/@sensu/web-ui"),
```
jamesdphillips: I can't find where this is consumed? Is this so that can be consumed by babel on the other side?

jamesdphillips: How would you feel about `@sensuapp/web`? We own the `@sensuapp` npm org. Eventually we could have `@sensuapp/sdk`.

10xjs: Yeah, this is for when the config is reused elsewhere

PR #2845, opened 2019-03-29
Refactoring of the list handlers

palourde:

## What is this change?

This is a required refactoring to support enterprise API filtering.

I apologize in advance for the size of this PR. Most changes simply consist of renaming/removing/moving stuff around. It can be summarized as follow:

- All _controllers_ were renamed from `query` to `list` for consistency
- These controllers now only deal with fetching and returning results from the store as `[]corev2.Resource`; we now deal with pagination in the generic lister _router_ (more about that below)
	- Therefore I removed all unit tests focusing on pagination for these controllers
- We now use a `*store.SelectionPredicate{}` to pass information about the request (pagination etc.) from apid to the store, which reduces the number of arguments and parameters required.
- The `store` are no longer exported in controllers (no idea why they were in the first place)
- `listerHandler` and `List` functions were introduced (see `backend/apid/routers/lister.go`)
	- `listerHandler()` is a workaround for mux.Router; the original idea was to override it from sensu-enterprise-go but turns out it isn't possible. Once the router is started, the actual handlers are immutable, so what I did is have it call the function stored in the `Lister` variable instead, which refers to the `List()` function in core. Then, from sensu-enterprise-go, we can assign a new function to this variable during runtime.
- The `Lister` function is responsible for handling pagination and fetching results from the store (via the controller function passed as a parameter). It also deals with a `suffix` route variable, in case we want to list a subresource (e.g. `/events/:entity`).
- All specific routers for listing were removed since they are no longer required
- All listing methods in the store are now using the generic `List` function, so  the pagination logic for etcd is now located in a single place.
	- Therefore all unit tests specific to pagination were removed except for the generic function

N.B. It seems like silenced entries do not support pagination at this time so I had to figure out a workaround for now

Required for https://github.com/sensu/sensu-enterprise-go/issues/282.

Yep.

Nope.

It would have been faster to do if we had generic handlers and stores for our core resources!

This is only a refactoring, no changes in functionality are expected at this point.

I manually verified every route.

### DIFF
--- a/backend/apid/actions/events.go
+++ b/backend/apid/actions/events.go
@@ -2,7 +2,6 @@ package actions
 
 import (
 	"context"
-	"encoding/base64"
 
 	"github.com/sensu/sensu-go/backend/messaging"
 	"github.com/sensu/sensu-go/backend/store"
@@ -12,49 +11,72 @@ import (
 
 // EventController expose actions in which a viewer can perform.
 type EventController struct {
-	Store store.EventStore
-	Bus   messaging.MessageBus
+	store store.EventStore
+	bus   messaging.MessageBus
 }
 
 // NewEventController returns new EventController
 func NewEventController(store store.EventStore, bus messaging.MessageBus) EventController {
 	return EventController{
-		Store: store,
-		Bus:   bus,
+		store: store,
+		bus:   bus,
 	}
 }
 
 // Query returns resources available to the viewer filter by given params.
-func (a EventController) Query(ctx context.Context, entityName, checkName string) ([]*corev2.Event, string, error) {
+func (a EventController) List(ctx context.Context, pred *store.SelectionPredicate) ([]corev2.Resource, error) {
 	var results []*corev2.Event
-	var nextContinueToken string
-
-	pageSize := corev2.PageSizeFromContext(ctx)
-	continueToken := corev2.PageContinueFromContext(ctx)
+	var err error
 
 	// Fetch from store
-	var serr error
-	if entityName != "" && checkName != "" {
-		var result *corev2.Event
-		result, serr = a.Store.GetEventByEntityCheck(ctx, entityName, checkName)
-		if result != nil {
-			results = append(results, result)
-		}
-	} else if entityName != "" {
-		results, nextContinueToken, serr = a.Store.GetEventsByEntity(ctx, entityName, int64(pageSize), continueToken)
+	if pred.Suffix != "" {
+		results, err = a.store.GetEventsByEntity(ctx, pred.Suffix, pred)
 	} else {
-		results, nextContinueToken, serr = a.Store.GetEvents(ctx, int64(pageSize), continueToken)
+		results, err = a.store.GetEvents(ctx, pred)
 	}
 
-	if serr != nil {
-		return nil, "", NewError(InternalErr, serr)
+	if err != nil {
+		return nil, NewError(InternalErr, err)
 	}
 
-	// Encode the continue token with base64url (RFC 4648), without padding
-	encodedNextContinueToken := base64.RawURLEncoding.EncodeToString([]byte(nextContinueToken))
+	resources := make([]corev2.Resource, len(results))
+	for i, v := range results {
+		resources[i] = corev2.Resource(v)
+	}
 
-	return results, encodedNextContinueToken, nil
+	return resources, nil
 }
+
+// func (a EventController) Query(ctx context.Context, entityName, checkName string) ([]*corev2.Event, string, error) {
+// 	var results []*corev2.Event
+// 	var nextContinueToken string
+
+// 	pageSize := corev2.PageSizeFromContext(ctx)
+// 	continueToken := corev2.PageContinueFromContext(ctx)
+
+// 	// Fetch from store
+// 	var serr error
+// 	if entityName != "" && checkName != "" {
+// 		var result *corev2.Event
+// 		result, serr = a.store.GetEventByEntityCheck(ctx, entityName, checkName)
+// 		if result != nil {
+// 			results = append(results, result)
+// 		}
+// 	} else if entityName != "" {
+// 		results, nextContinueToken, serr = a.store.GetEventsByEntity(ctx, entityName, int64(pageSize), continueToken)
+// 	} else {
+// 		results, nextContinueToken, serr = a.store.GetEvents(ctx, int64(pageSize), continueToken)
+// 	}
+
+// 	if serr != nil {
+// 		return nil, "", NewError(InternalErr, serr)
+// 	}
+
+// 	// Encode the continue token with base64url (RFC 4648), without padding
+// 	encodedNextContinueToken := base64.RawURLEncoding.EncodeToString([]byte(nextContinueToken))
+
+// 	return results, encodedNextContinueToken, nil
+// }
 
 // Find returns resource associated with given parameters if available to the
 // viewer.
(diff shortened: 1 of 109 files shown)


backend/apid/actions/events_test.go
```diff
@@ -20,24 +21,22 @@ func TestNewEventController(t *testing.T) {
...
-		continueToken         string
-		storeErr              error
-		expectedLen           int
-		expectedContinueToken string
-		expectedErr           error
+		name        string
+		ctx         context.Context
+		events      []*types.Event
+		entity      string
+		check       string
```
golangcibot: U1000: field `check` is unused (from `unused`)

backend/apid/actions/roles_test.go
```diff
@@ -276,84 +276,53 @@ func TestRoleGet(t *testing.T) {
...
-		storeErr              error
-		continueToken         string
-		expectedResult        []*types.Role
-		expectedContinueToken string
-		expectedErr           bool
-		expectedErrCode       ErrCode
+		name          string
+		ctx           context.Context
+		storeErr      error
+		continueToken string
```
golangcibot: U1000: field `continueToken` is unused (from `unused`)

backend/apid/routers/users.go
```diff
@@ -59,21 +59,6 @@ func (r *UsersRouter) Mount(parent *mux.Router) {
 	routes.Path("{id}/{subresource:password}", r.updatePassword).Methods(http.MethodPut)
 }
 
-func (r *UsersRouter) list(w http.ResponseWriter, req *http.Request) (interface{}, error) {
-	records, continueToken, err := r.controller.Query(req.Context())
-
-	// Obfustace users password
-	for i := range records {
-		records[i].Password = ""
```
ccressent: The generic List doesn't do that for Users. Is that a problem?

backend/store/etcd/check_store_test.go
```diff
@@ -39,121 +39,18 @@ func TestCheckConfigStorage(t *testing.T) {
...
-		checks, continueToken, err = store.GetCheckConfigs(ctx, 0, "")
+		checks, err = s.GetCheckConfigs(ctx, pred)
 		assert.NoError(t, err)
 		assert.NotEmpty(t, checks)
 		assert.Equal(t, 1, len(checks))
-		assert.Empty(t, continueToken)
+		assert.Empty(t, pred.Continue)
+		for _, check := range checks {
+			fmt.Printf("TestCheckConfigStorage: %#v\n", check)
+		}
```
ccressent: I'm assuming that's a debug `for` loop that can be removed?

backend/store/etcd/entity_store.go
```diff
@@ -5,10 +5,10 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"path"
 
 	"github.com/coreos/etcd/clientv3"
 	"github.com/sensu/sensu-go/backend/store"
+	"github.com/sensu/sensu-go/types"
```
ccressent: It's OK not to change now, but as a reminder for future changes/refactors: we have type aliases to move away from the `types` package. We want to use the `corev2 "github.com/sensu/sensu-go/api/core/v2` package instead.

backend/apid/actions/roles.go
```diff
@@ -82,23 +81,17 @@ func (a RoleController) Get(ctx context.Context, name string) (*types.Role, erro
...
-			return nil, "", NewErrorf(NotFound)
-		default:
-			return nil, "", NewError(InternalErr, err)
-		}
+		return nil, NewError(InternalErr, err)
 	}
 
-	// Encode the continue token with base64url (RFC 4648), without padding
-	encodedNewContinueToken := base64.RawURLEncoding.EncodeToString([]byte(newContinueToken))
+	resources := make([]corev2.Resource, len(results))
```
nikkiki: I see this code block is repeated often in the refactor. Is this something we could maybe make a generic function?

backend/store/etcd/check_store_test.go
nikkiki: I'm impressed your debug print statements contain no vulgarity 😆

backend/apid/routers/lister_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+// router := mux.NewRouter()
+// router.PathPrefix("/foo/{suffix}").HandlerFunc(List(controller.List))
+// router.PathPrefix("/foo").HandlerFunc(List(controller.List))
+
+// fixtures := []corev2.Resource{types.FixtureCheckConfig("check1")}
+// controller.On("List", mock.Anything, mock.AnythingOfType("*store.SelectionPredicate")).Return(fixtures, nil)
+
+// router.ServeHTTP(w, r)
+// }
```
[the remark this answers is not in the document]
palourde: 😬

backend/store/etcd/check_store_test.go
palourde: 😬

backend/apid/actions/events.go
```diff
@@ -12,48 +11,40 @@ import (
...
-	if entityName != "" && checkName != "" {
-		var result *corev2.Event
-		result, serr = a.Store.GetEventByEntityCheck(ctx, entityName, checkName)
-		if result != nil {
-			results = append(results, result)
-		}
-	} else if entityName != "" {
-		results, nextContinueToken, serr = a.Store.GetEventsByEntity(ctx, entityName, int64(pageSize), continueToken)
+	if pred.Suffix != "" {
+		results, err = a.store.GetEventsByEntity(ctx, pred.Suffix, pred)
```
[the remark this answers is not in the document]
palourde: As discussed, we will go ahead with `subcollection` instead, which is way more clearer!

backend/apid/routers/users.go
palourde: Oh god yes, good catch, that's actually a real problem haha. I'll move that to the controller actually!

backend/apid/actions/roles.go
palourde: At first I tried to create a function like this, but obviously it won't work because an `interface{}` doesn't implement the corev2.Resource interface:

```go
func returnResources(results []corev2.Resource) []corev2.Resource {
	resources := make([]corev2.Resource, len(results))
	for i, v := range results {
		resources[i] = corev2.Resource(v)
	}
	return resources
}
```

If someone has an idea, I'm all ears!

ccressent: Unfortunately, in Go, even though `[]corev2.Role` would hand-wavely qualify as a `[]corev2.Resource` because `Role` meets the `Resource` interface, the compiler won't accept it because it would have to construct a potentially expensive operation behind the scene to do what you want: convert each `corev2.Role` into a `Resource` interface (they have different memory representations), which is O(n). This is what you explicitly do with that `for` loop. If the compiler did that "magic", you could directly use `results` (type `[]corev2.Role`) as the function's return value (which needs to be `[]corev2.Resource`); no need for that for loop!

So in Go, boilerplate it is! If you want to abstract it away you can still write that `[]interface{} -> []corev2.Resource` function provided you're willing to use reflection. Something like that should do the trick:

```go
func returnResources(input interface{}) []corev2.Resource {
	var resources []corev2.Resource

	v := reflect.ValueOf(input)

	if v.Kind() != reflect.Slice {
		panic("input needs to be a slice")
	}

	for i := 0; i < v.Len(); i++ {
		resource := v.Index(i).Interface().(corev2.Resource)
		resources = append(resources, resource)
	}

	return resources
}
```

Note that the type assertion will panic if the elements of the input slice do not meet the `corev2.Resource` interface.

nikkiki: This is purely aspirational, but I appreciate that you tried to simplify it in your approach @palourde! While @ccressent approach would likely work, it may not necessary to introduce reflection for a small win over code duplication.

PR #2868, opened 2019-04-08
Audit and fix all etcd watchers

### LINKED ISSUE #2635: 100% CPU usage after a random period of time
## Expected Behavior

Sensu backend should run normally.

## Current Behavior

After a random period of time (can be lass than 2 hours or more than 12 hours), the sensu-backend process start to consume 100% of the CPU available on the virtual machines.

A restart of the process solves the problem until the next time. CPU is normally around 5% usage (there is only a few agents registered and only 3/4 checks per agents for now).

This bug is totally random.

There is nothing abnormal in the logs. The backend still works as expected (scheduling checks, handling events, etc) but the sensu-backend process is taking 100% of the CPU.

## Context
Backend configuration:
```
# general configuration
state-dir: "/var/lib/sensu"

# etcd cluster configuration
etcd-advertise-client-urls: "http://sensu-etcd-cluster.staging.monitoring.xxx.com:2379"
no-embed-etcd: true

# deregistration
deregistration-handler: "deregister"

# logging
log-level: "debug"
```
## Your Environment
* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): 5.1.0
* Installation method (packages, binaries, docker etc.): Ubuntu packages
* Operating System and version (e.g. Ubuntu 14.04): Ubuntu 18.04

Backends are on AWS EC2 instances behind an ALB. etcd cluster is accessed through a NLB also on EC2 instances.

echlebek:
## What is this change?

During a recent QA session, it was discovered that watchers could
fail permanently, but their consumers were using them as if they
could not. This was causing issues such as 100% CPU usage in some
components, and other components would simply fail to update until
service restart.

The problem mostly presents on backends that are using the
--no-embed-etcd mode, but can theoretically also occur on backends
that do embed etcd.

All watchers now get reinstated on a retry loop when they fail, and
all watchers now require that a leader is present, so that watchers
do not get stuck on partitions.

Closes #2635 

Added

This change is still unverified beyond automated testing, and needs to be put through a full QA cycle.

### DIFF
--- a/backend/liveness/liveness.go
+++ b/backend/liveness/liveness.go
@@ -7,10 +7,12 @@ import (
 	"path"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/coreos/etcd/clientv3"
 	"github.com/coreos/etcd/mvcc/mvccpb"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/time/rate"
 )
 
 var (
@@ -234,7 +236,6 @@ func (t *SwitchSet) getTTLFromEvent(event *clientv3.Event) (int64, State) {
 // monitor starts a goroutine that monitors the SwitchSet prefix for key PUT
 // and DELETE events.
 func (t *SwitchSet) monitor(ctx context.Context) {
-	wc := t.client.Watch(ctx, t.prefix, clientv3.WithPrefix(), clientv3.WithPrevKV())
 	go func() {
 		for event := range t.events {
 			if key, bury := event(); bury {
@@ -245,15 +246,23 @@ func (t *SwitchSet) monitor(ctx context.Context) {
 		}
 	}()
 	go func() {
+		var wc clientv3.WatchChan
+		ctx := clientv3.WithRequireLeader(ctx)
+		limiter := rate.NewLimiter(rate.Every(time.Second), 1)
+		_ = limiter.Wait(ctx)
+	OUTER:
+		wc = t.client.Watch(ctx, t.prefix, clientv3.WithPrefix(), clientv3.WithPrevKV())
 		for {
 			select {
 			case <-ctx.Done():
 				close(t.events)
 				return
-			case resp := <-wc:
-				if err := resp.Err(); err != nil {
+			case resp, ok := <-wc:
+				if err := resp.Err(); err != nil && err != context.Canceled {
 					t.logger.WithError(err).Error("error monitoring toggles")
-					continue
+				}
+				if resp.Canceled || !ok {
+					goto OUTER
 				}
 				for _, event := range resp.Events {
 					t.handleEvent(ctx, event)
--- a/backend/store/etcd/watchers.go
+++ b/backend/store/etcd/watchers.go
@@ -91,7 +91,7 @@ func (s *Store) GetEntityWatcher(ctx context.Context) <-chan store.WatchEventEnt
 				entity.Name = meta.ResourceName
 			} else {
 				if err := json.Unmarshal(response.Object, &entity); err != nil {
-					logger.WithField("key", response.Key).WithError(err).Error("unable to unmarshal check config from key")
+					logger.WithField("key", response.Key).WithError(err).Error("unable to unmarshal entity from key")
 					continue
 				}
 			}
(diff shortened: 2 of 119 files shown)


palourde:
@echlebek Quick note, the watcher wrapper was already used by sensu-enterprise-go so once this PR is merged, it will probably require a patch too!

backend/store/etcd/watchers.go
```diff
@@ -130,37 +112,34 @@ func (s *Store) GetEntityWatcher(ctx context.Context) <-chan store.WatchEventEnt
 // restart the watcher, if needed.
```
nikkiki:
```suggestion
// The watcher does its best to recover from errors.
```

backend/store/etcd/watchers.go
```diff
@@ -130,37 +112,34 @@ func (s *Store) GetEntityWatcher(ctx context.Context) <-chan store.WatchEventEnt
...
+			var tessen corev2.TessenConfig
 
-				watchEvent := store.WatchEventTessenConfig{
-					Action:       action,
-					TessenConfig: tessen,
+			if response.Type == store.WatchDelete {
+				tessen = *corev2.DefaultTessenConfig()
+			} else {
+				if err := json.Unmarshal(response.Object, &tessen); err != nil {
+					logger.WithField("key", response.Key).WithError(err).Error("unable to unmarshal check config from key")
```
nikkiki:
```suggestion
					logger.WithField("key", response.Key).WithError(err).Error("unable to unmarshal tessen config from key")
```

backend/store/etcd/watchers.go
```diff
@@ -81,46 +68,41 @@ func (s *Store) GetCheckConfigWatcher(ctx context.Context) <-chan store.WatchEve
...
-				watchEvent := store.WatchEventEntity{
-					Action: action,
-					Entity: &entity,
+			if response.Type == store.WatchDelete {
+				meta := store.ParseResourceKey(response.Key)
+				entity.Namespace = meta.Namespace
+				entity.Name = meta.ResourceName
+			} else {
+				if err := json.Unmarshal(response.Object, &entity); err != nil {
+					logger.WithField("key", response.Key).WithError(err).Error("unable to unmarshal check config from key")
```
nikkiki:
```suggestion
					logger.WithField("key", response.Key).WithError(err).Error("unable to unmarshal entity from key")
```

backend/liveness/liveness.go
```diff
@@ -245,15 +246,23 @@ func (t *SwitchSet) monitor(ctx context.Context) {
 		}
 	}()
 	go func() {
+		var wc clientv3.WatchChan
+		ctx := clientv3.WithRequireLeader(ctx)
+		limiter := rate.NewLimiter(rate.Every(time.Second), 1)
+		_ = limiter.Wait(ctx)
+	OUTER:
+		wc = t.client.Watch(ctx, t.prefix, clientv3.WithPrefix(), clientv3.WithPrevKV())
```
palourde: Stupid question: do we need to move the rate limiter `Wait` method within the `OUTER` label? Maybe I'm misreading it because what I understand is that it would only be rate limiting once the goroutine starts.

echlebek: Ah good question. The intent of the limiter is that no more than one `Watch` request will be made per second. That is, if the system tries to create a `Watch` over and over again, but fails consistently, it will not do so faster than once per second.

backend/store/etcd/watchers.go
```diff
@@ -81,86 +68,78 @@ func (s *Store) GetCheckConfigWatcher(ctx context.Context) <-chan store.WatchEve
...
 			}
 		}
 	}()
+
 	return ch
 }
 
 // GetTessenConfigWatcher returns a channel that emits WatchEventTessenConfig structs notifying
 // the caller that a TessenConfig was updated. If the watcher runs into a terminal error
 // or the context passed is cancelled, then the channel will be closed. The caller must
```
nikkiki: I wish you could do multi-line suggestions

```suggestion
// or the context passed is cancelled, then the channel will be closed.
```

echlebek: Thanks, I'll look into it

Issue #2720, opened 2019-02-13
Results submitted to agent socket do not follow 1.x behavior wrt source attribute


cwjohnston:
## Expected Behavior

Results received via agent socket treat the value of `source` attribute as the entity with which the check result should be associated.

## Current Behavior

When results are received via the agent socket, the value of the `source` attribute is not considered when associating the event with an entity.

## Possible Solution

* honor the value of `source` attribute as specifying the entity with which the check result (event) should be associated
* also honor [`proxy_entity_name` as a synonym for `source`](https://github.com/sensu/sensu-go/issues/2605)
* discard the `client` attribute (under 1.x this is treated as a custom check result attribute and does not influence which entity the result is attributed to)

## Context

In preparing [updates for agent socket documentation](https://github.com/sensu/sensu-docs/pull/1200), we've found that the handling of source and client attributes differs between Sensu 1.x and Sensu Go. As we believe the agent socket should be backward compatible with the 1.x check result format, we also expect the treatment of `source` attributes to be the same under Go as it is under Sensu 1.x.

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): 5.2.1
* Installation method (packages, binaries, docker etc.): packages
* Operating System and version (e.g. Ubuntu 14.04): Centos 7

echlebek:
* The agent should fill in the `client` attribute, and the user should not be required to. Currently, the user is required to fill in `client`, this is a bug.

* Add the `source` attribute  and use it to set the proxy entity name.

Issue #2873, opened 2019-04-10
Agent socket should support 1.x custom attributes as labels


nikkixdev:
Related: #2720, #417

## Expected Behavior
Custom fields passed to the agent socket are added as 1.x custom attributes, or 2.x labels.

## Current Behavior
Custom fields passed to the agent socket are ignored, and the event is still created without the custom fields.

## Possible Solution
Use some reflection to populate the fields as labels.

## Steps to Reproduce (for bugs)
1. Send this payload to the agent socket: `{"name": "check-mysql-status", "output": "error!", "status": 1, "foo": "bar"}`
2. Observe an event is created without the field `"foo":"bar"`

## Context
1.x parity

## Your Environment
* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): https://github.com/sensu/sensu-go/commit/34315c36d36ce21cd87ebf5652584fb698867ce1
* Installation method (packages, binaries, docker etc.): binaries
* Operating System and version (e.g. Ubuntu 14.04): MacOS

nikkiki:
If the agent socket is deprecated, further development/support is probably no longer necessary. Leaving this open until we can discuss in Inbox cleanup.

portertech:
@nikkixdev can you please confirm the current behaviour. If the socket accepts the data and sends an event sans custom attributes, we can close this.

nikkixdev:
The current _behavior_ is still the same. The socket accepts the data and sends an event without custom attributes or labels. Closing the issue per this comment: https://github.com/sensu/sensu-go/issues/2873#issuecomment-534691014.

PR #2874, opened 2019-04-10
Deprecates client in favor of source in the agent socket

nikkiki:

## What is this change?

Changes the 1.x `client` field to `source` in the 1.x compatible agent socket. The `client` field is now deprecated.

Closes https://github.com/sensu/sensu-go/issues/2720.

Yas.

Nah.

Filed https://github.com/sensu/sensu-go/issues/2873 after discussing the agent socket with @cwjohnston.

https://github.com/sensu/sensu-docs/issues/1356

1. Run `echo '{"name": "check", "status": 1, "output": "error!", "client":"client"}' | nc localhost 3030`
- An event `client/check` is created
- `event.check.proxy_entity_name=client`
- A proxy entity `client` is created

2. Run `echo '{"name": "check", "status": 1, "output": "error!", "source":"source"}' | nc localhost 3030`
- An event `source/check` is created
- `event.check.proxy_entity_name=source`
- A proxy entity `source` is created

3. Run `echo '{"name": "check", "status": 1, "output": "error!", "source":"agent"}' | nc localhost 3030` where the agent running is named `agent`
- An event `agent/check` is created
- `event.check.proxy_entity_name` is empty
- A proxy entity is not created

4. Run `echo '{"name": "check", "status": 1, "output": "error!"}' | nc localhost 3030` where the agent running is named `agent`
- An event `agent/check` is created
- `event.check.proxy_entity_name` is empty
- A proxy entity is not created

### DIFF
--- a/CHANGELOGS/unreleased/2720-add-source-agent-socket.md
+++ b/CHANGELOGS/unreleased/2720-add-source-agent-socket.md
@@ -1,0 +1,2 @@
+### Added
+- Added the 1.x `source` field as a synonym for `proxy_entity_name` in the 1.x compatible agent socket.
--- a/CHANGELOGS/unreleased/2720-add-source-agent-socket.md
+++ b/CHANGELOGS/unreleased/2720-add-source-agent-socket.md
@@ -1,2 +1,0 @@
-### Added
-- Added the 1.x `source` field as a synonym for `proxy_entity_name` in the 1.x compatible agent socket.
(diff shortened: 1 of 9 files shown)


nikkiki:
We haven’t removed the `client` field and currently still support it with the same functionality as `source`. Unless I’m misunderstanding the comment @echlebek?

portertech:
"client" was never part of the 1.x spec, it should just be removed.

portertech:
"client" was never part of the 1.x spec, it should just be removed.

echlebek:
@nikkiki oh my bad, I saw a deletion around the 'Client' attribute but now I see that it was simply moved towards the bottom of the struct!

portertech:
We are choosing to not remove `client`, as we do not want to make a breaking change. We need to deprecate the whole socket.

Issue #2882, opened 2019-04-16
[Tessen] Use etcd WithCountOnly option when collecting entity/backend counts 


nikkiki:
Optimize data collection for entity/backend counts by using etcd's `WithCountOnly()`. A short benchmarking exercise proved that this method was more efficient, especially at scale.

## Expected Behavior
```
resp, err := client.Get(ctx, entitiesKey, clientv2.WithPrefix(), clientv3.WithCountOnly())
```

## Current Behavior
Iterates over the entire entity list.
```
entities, err := t.store.GetEntities(t.ctx, &store.SelectionPredicate{})
```

## Possible Solution
Update the functions used to collect the backend and entity count.

## Context
This is a performance optimization.

## Your Environment
* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): 5.5.0
* Installation method (packages, binaries, docker etc.): binaries
* Operating System and version (e.g. Ubuntu 14.04): MacOS

nikkiki:
Closed by https://github.com/sensu/sensu-go/pull/2886.

Issue #2880, opened 2019-04-16
[Tessen] Opt-out notification payload


nikkiki:
Implement the data payload specified in the spike, https://github.com/sensu/sensu-go/issues/2862 so that TessenD sends an opt-out notification event when the configuration is updated. Although InfluxDB discards duplicate data points, we should ensure only 1 backend sends this event, rather than _every_ backend in the cluster.

nikkiki:
Closed by https://github.com/sensu/sensu-go/pull/2886.

PR #2886, opened 2019-04-18
Add message bus to tessend

nikkiki:

## What is this change?

Adds the message bus to Tessend in order to track Tessen configuration changes from the API.

Closes https://github.com/sensu/sensu-go/issues/2880.
Closes https://github.com/sensu/sensu-go/issues/2882.

Yas.

Nah.

Filed https://github.com/sensu/sensu-go/issues/2891.

We good.

[TestRail test case](https://sensuinc.testrail.io/index.php?/cases/view/5514&group_by=cases:section_id&group_order=asc&group_id=1074)
```
$ sensuctl tessen opt-in
$ sensuctl tessen opt-out --skip-confirm
```
**Tessen Go dashboard:**
Screen Shot 2019-04-18 at 3 03 53 PM

### DIFF
--- a/backend/tessend/tessend.go
+++ b/backend/tessend/tessend.go
@@ -4,8 +4,8 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"strconv"
@@ -161,10 +161,7 @@ func (t *Tessend) startMessageHandler() {
 	for {
 		msg, ok := <-t.messageChan
 		if !ok {
-			select {
-			case t.errChan <- errors.New("tessen message channel closed"):
-			default:
-			}
+			logger.Debug("tessen message channel closed")
 			return
 		}
 
@@ -401,13 +398,11 @@ func (t *Tessend) getTessenMetrics(now int64, data *Data) {
 	var entityCount, backendCount float64
 
 	// collect entity count
-	entities, err := t.store.GetEntities(t.ctx, &store.SelectionPredicate{})
+	entities, err := etcd.Count(t.ctx, t.client, etcd.GetEntitiesPath(t.ctx, ""))
 	if err != nil {
 		logger.WithError(err).Error("unable to retrieve entity count")
 	}
-	if entities != nil {
-		entityCount = float64(len(entities))
-	}
+	entityCount = float64(entities)
 
 	// collect backend count
 	cluster, err := t.client.Cluster.MemberList(t.ctx)
@@ -463,8 +458,9 @@ func (t *Tessend) send(data *Data) string {
 		logger.WithError(err).Error("tessen phone-home service failed")
 		return ""
 	}
+	defer resp.Body.Close()
 	if resp.StatusCode >= 400 {
-		body, _ := ioutil.ReadAll(resp.Body)
+		body, _ := ioutil.ReadAll(io.LimitReader(resp.Body, 4096))
 		logger.Errorf("bad status: %d (%q)", resp.StatusCode, string(body))
 		return ""
 	}
--- a/backend/tessend/tessend.go
+++ b/backend/tessend/tessend.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -14,6 +15,7 @@ import (
 	"github.com/coreos/etcd/clientv3"
 	"github.com/google/uuid"
 	corev2 "github.com/sensu/sensu-go/api/core/v2"
+	"github.com/sensu/sensu-go/backend/messaging"
 	"github.com/sensu/sensu-go/backend/ringv2"
 	"github.com/sensu/sensu-go/backend/store"
 	"github.com/sensu/sensu-go/backend/store/etcd"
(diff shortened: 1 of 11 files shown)


backend/tessend/tessend.go
```diff
@@ -123,7 +149,43 @@ func (t *Tessend) Err() <-chan error {
...
+	return componentName
+}
+
+// Receiver returns the tessen receiver channel.
+func (t *Tessend) Receiver() chan<- interface{} {
+	return t.messageChan
+}
+
+func (t *Tessend) startMessageHandler() {
+	for {
```
golangcibot: S1000: should use for range instead of for { select {} } (from `gosimple`)

backend/tessend/tessend.go
```diff
@@ -123,7 +149,44 @@ func (t *Tessend) Err() <-chan error {
...
+			data := t.getDataPayload()
+			t.getTessenConfigMetrics(time.Now().UTC().Unix(), tessen, data)
+			logger.WithFields(logrus.Fields{
+				"url":                       t.url,
+				"id":                        data.Cluster.ID,
+				"opt-out":                   tessen.OptOut,
+				data.Metrics.Points[0].Name: data.Metrics.Points[0].Value,
+			}).Info("sending opt-out status event to tessen")
+			_ = t.send(data)
+		default:
```
golangcibot: SA5004: should not have an empty default case in a for+select loop. The loop will spin. (from `staticcheck`)

backend/tessend/tessend.go
```diff
@@ -123,7 +149,41 @@ func (t *Tessend) Err() <-chan error {
...
+
+// Receiver returns the tessen receiver channel.
+func (t *Tessend) Receiver() chan<- interface{} {
+	return t.messageChan
+}
+
+func (t *Tessend) startMessageHandler() {
+	for {
+		msg, ok := <-t.messageChan
+		if !ok {
```
echlebek: This will only happen when the daemon is stopped, and since the closing happens in the same package I don't really think this is an error condition. I think it's probably fine to just return here.

backend/tessend/tessend.go
```diff
@@ -123,7 +149,41 @@ func (t *Tessend) Err() <-chan error {
...
+
+		data := t.getDataPayload()
+		t.getTessenConfigMetrics(time.Now().UTC().Unix(), tessen, data)
+		logger.WithFields(logrus.Fields{
+			"url":                       t.url,
+			"id":                        data.Cluster.ID,
+			"opt-out":                   tessen.OptOut,
+			data.Metrics.Points[0].Name: data.Metrics.Points[0].Value,
+		}).Info("sending opt-out status event to tessen")
+		_ = t.send(data)
```
echlebek: Should this error be handled and logged?

backend/tessend/tessend.go
```diff
@@ -345,27 +391,83 @@ func (t *Tessend) collect(now int64) *Data {
...
+
+	return data
+}
+
+// getTessenMetrics populates the data payload with # backends/entities.
+func (t *Tessend) getTessenMetrics(now int64, data *Data) {
+	var entityCount, backendCount float64
+
+	// collect entity count
+	entities, err := t.store.GetEntities(t.ctx, &store.SelectionPredicate{})
```
echlebek: It would be best to avoid getting the entire list of entities here. Can you add an `EntityCount` method to the store using the `clientv3.WithCountOnly` option?

We'll need this for another component also, so we can get two birds stoned at once. :bird: :evergreen_tree:

backend/tessend/tessend.go
```diff
@@ -345,27 +391,83 @@ func (t *Tessend) collect(now int64) *Data {
...
+	}
 }
 
-// send the data payload to the tessen url
-func (t *Tessend) send(data *Data) (*http.Response, error) {
+// send sends the data payload to the tessen url and retrieves the interval response header.
+func (t *Tessend) send(data *Data) string {
 	b, _ := json.Marshal(data)
-	return http.Post(t.url, "application/json", bytes.NewBuffer(b))
+	resp, err := http.Post(t.url, "application/json", bytes.NewBuffer(b))
```
echlebek: Need to do:
```
defer resp.Body.Close()
```
to avoid leaking resources here.

backend/tessend/tessend.go
```diff
@@ -345,27 +391,83 @@ func (t *Tessend) collect(now int64) *Data {
...
+	// TODO(nikki): special case logs on a per error basis
+	if err != nil {
+		logger.WithError(err).Error("tessen phone-home service failed")
+		return ""
+	}
+	if resp.StatusCode >= 400 {
+		body, _ := ioutil.ReadAll(resp.Body)
```
echlebek: We should use `io.LimitReader` here. An attacker could theoretically MITM the tessen service and send an infinite stream of bytes, causing a crash. For instance,

```
body, _ := ioutil.ReadAll(io.LimitReader(resp.Body, 4096))
```

backend/tessend/tessend.go
nikkiki: This function doesn't return an error, but errors are logged within in.

backend/tessend/tessend.go
nikkiki: Yeah, I filed https://github.com/sensu/sensu-go/issues/2882 back when we discovered that improvement. I suppose I can add that to this PR.

nikkiki: Alright, added a `Count()` function to the generic store.. utilizing it here.

PR #2892, opened 2019-04-23
Add support for field selectors

palourde:
## What is this change?

This PR adds functions to retrieve fields for resources that can be listed via the REST API.

It also adds an argument to the lister HTTP handler that correspond to these functions signature.

Required by https://github.com/sensu/sensu-enterprise-go/issues/283.

Added!

Nope!

Nope! I hesitated to add a method to the `Resource` interface but I decided against it because it could be considered as a breaking change; third-party code relying on it would break with this change.

Not yet.

Manual testing.

### DIFF
--- a/api/core/v2/asset.go
+++ b/api/core/v2/asset.go
@@ -106,9 +106,9 @@ func NewAsset(meta ObjectMeta) *Asset {
 // AssetFields returns a set of fields that represent that resource
 func AssetFields(r Resource) map[string]string {
 	resource := r.(*Asset)
-	fields := make(map[string]string, 3)
-	fields["asset.name"] = resource.ObjectMeta.Name
-	fields["asset.namespace"] = resource.ObjectMeta.Namespace
-	fields["asset.filters"] = strings.Join(resource.Filters, ",")
-	return fields
+	return map[string]string{
+		"asset.name":      resource.ObjectMeta.Name,
+		"asset.namespace": resource.ObjectMeta.Namespace,
+		"asset.filters":   strings.Join(resource.Filters, ","),
+	}
 }
--- a/api/core/v2/asset.go
+++ b/api/core/v2/asset.go
@@ -8,6 +8,7 @@ import (
 	"net/url"
 	"path"
 	"regexp"
+	"strings"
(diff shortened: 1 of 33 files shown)


palourde:
CircleCI builds are just getting OOM...

api/core/v2/namespace.go
```diff
@@ -27,7 +27,7 @@ func FixtureNamespace(name string) *Namespace {
 	}
 }
 
-// URIPath returns the path component of a Namespace URI.
+// URIPath returns the path component of a Namespace 	URI.
```
ccressent: Minor nitpick: that whitespace diff doesn't look right.

backend/apid/routers/lister_test.go
```diff
@@ -92,8 +92,8 @@ func TestList(t *testing.T) {
 			w := httptest.NewRecorder()
 
 			router := mux.NewRouter()
-			router.PathPrefix("/foo/{subcollection}").HandlerFunc(List(controller.List))
-			router.PathPrefix("/foo").HandlerFunc(List(controller.List))
+			router.PathPrefix("/foo/{subcollection}").HandlerFunc(List(controller.List, corev2.CheckConfigFields))
+			router.PathPrefix("/foo").HandlerFunc(List(controller.List, corev2.CheckConfigFields))
```
ccressent: I would suggest passing in `func(r corev2.Resource) map[string]string { return map[string]string{} }`, a "no-op" `FieldsFunc`, if you just need one for the sake of passing one in.

api/core/v2/asset.go
```diff
@@ -101,3 +102,13 @@ func (a *Asset) URIPath() string {
 func NewAsset(meta ObjectMeta) *Asset {
 	return &Asset{ObjectMeta: meta}
 }
+
+// AssetFields returns a set of fields that represent that resource
+func AssetFields(r Resource) map[string]string {
+	resource := r.(*Asset)
+	fields := make(map[string]string, 3)
```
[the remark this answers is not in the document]
palourde: Even better, we will just use composite literals

PR #2895, opened 2019-04-24
Filtering support for sensuctl

ccressent:
## What is this change?

This change adds filtering support to `sensuctl`. There are now 2 additional flags, `--field-selector` and `--label-selector`, for all the resource types that support the `list` subcommand. `--field-selector` is a string representing a filter that acts on resource fields; `--label-selector` is a string representing a filter that acts on resource labels. This is very similar to the `kubectl` approach.

Filtering support is "enterprise only": while `sensuctl` is able to construct queries that include field and label selectors no matter what, the backend might answer with an error if it doesn't have a valid enterprise license.

Closes https://github.com/sensu/sensu-enterprise-go/issues/286.

Yes, added.

No.

I had to refactor/duplicate a mock, `MockConfig`, to avoid a circular dependency between the `cli/client` and `cli/client/testing` packages. Ultimately, we should probably clean up and unify the various mock objects we have but that will entails quite a bit of editing across the code base.

New documentation is required. This is tracked by https://github.com/sensu/sensu-docs/issues/1395.

*THIS IS STILL BEING TESTED AGAINST THE FILTERING FEATURE BRANCH!*


cli/commands/asset/list.go
```diff
@@ -31,8 +33,22 @@ func ListCommand(cli *cli.SensuCli) *cobra.Command {
...
+			}
+
+			labelSelector, err := cmd.Flags().GetString(flags.LabelSelector)
+			if err != nil {
+				return err
+			}
+
+			opts := client.ListOptions{}
+			opts.FieldSelector = fieldSelector
+			opts.LabelSelector = labelSelector
```
palourde: I feel like we could create a nice wrapper for the 12 lines above since it's basically the same across all commands!

ccressent: You're right, I'll refactor that!
I got a bit carried away with my vim macros ;)

PR #2896, opened 2019-04-24
Add Windows service for sensu-agent

echlebek:
This changeset adds a service wrapper for Windows. The service can be installed or uninstalled with:

```
sensu-agent service install
```
and
```
sensu-agent service uninstall
```

When the service is installed, it will be in the "started" state. Users can use the Windows service manager to stop and start the service after this.

When the service is uninstalled, it will be implicitly stopped.

The service logs the start/stop state to the event log, as well as any fatal errors encountered during service execution. The service will attempt to restart when there is a fatal error.

The application logs are redirected to a file, by default, `C:\ProgramData\sensu\log\sensu-agent.log`.


cmd/sensuctl/main_windows.go
```diff
@@ -0,0 +1,26 @@
+package main
```
palourde: Could you add a comment similar to what we find in `cmd/sensu-agent/main_windows.go` that basically explains why we have this file (that's essentially a copy of `main.go`)? Took me a bit to understand that 😅

PR #2900, opened 2019-04-26
Create a generic function for continue token generation

palourde:

## What is this change?

This PR introduces exports the ComputeContinueToken function and introduce events handling.

Required for enterprise filtering.

Added

Nope

Nope

Nope

Already covered by integration tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 `sensuctl cluster member-list` commands in tabular format.
 - Added a `Header` field to the `HealthResponse` type returned by `/health`.
 
+### Changed
+- The continue token of events is now computed by the same function as the other
+resources.
+
 ### Fixed
 - Fixed the agent `--annotations` and `--labels` flags.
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 `sensuctl cluster member-list` commands in tabular format.
 - Added a `Header` field to the `HealthResponse` type returned by `/health`.
 
+### Changed
+- The continue token of events is now computed by the same function as the other
+resources.
+
```
ccressent: I'm not sure we even need a changelog entry for that; in my opinion it's very much internal stuff that doesn't (or shouldn't!) affect anything or anyone.

backend/store/etcd/store.go
```diff
@@ -294,10 +294,31 @@ func namespaceFound(namespace string) clientv3.Cmp {
...
+			return "/" + eventNamespace + "/" + resource.Entity.Name + "/" + resource.Check.Name + "\x00"
+		}
+		return resource.Entity.Name + "/" + resource.Check.Name + "\x00"
+	}
+
+	objMeta := r.GetObjectMeta()
+
 	if queriedNamespace == "" {
 		token = path.Join(objMeta.Namespace, objMeta.Name) + "\x00"
 	} else {
```
ccressent: I would put the default, "regular" case in the switch as part of `default`.

To me that makes it clearer that there are distinct cases, and if a code path in `case *corev2.Event` is missing a `return`, it won't lead to some nonsensical operation being executed.

PR #2924, opened 2019-05-07
Create generic PrintTitle function, properly check for formats

### LINKED ISSUE #2917: sensu-go 5.6 cluster member-list is not valid JSON
## Expected Behavior

I'd expect `--format json` for `sensuctl cluster member-list` to be valid JSON that can be parsed by standard JSON parsers.

## Current Behavior

Currently the output of `sensuctl cluster member-list --format json` is two separate JSON hashes separated by a new line which isn't valid JSON.

```
[root@sensu-backend-peer1 ~]# sensuctl cluster member-list --format json | python -m json.tool
Extra data: line 7 column 1 - line 37 column 1 (char 99 - 649)
```

This was not an issue with sensu-go 5.5.

## Context

The Puppet module for sensu-go expects `--format json` to be valid JSON for Ruby JSON parser.  Neither python nor Ruby can parse the output of `sensuctl cluster member-list --format json`.

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): 5.6
* Installation method (packages, binaries, docker etc.): RPM
* Operating System and version (e.g. Ubuntu 14.04): CentOS 7

### LINKED ISSUE #2918: sensuctl cluster commands and wrapped-json raise errors
## Expected Behavior

I'd expect `sensuctl cluster` commands with `--format wrapped-json` to not raise errors.

```
[root@sensu-backend-peer1 ~]# sensuctl cluster member-list --format wrapped-json
Error: &{%!t(uint64=18198607357687572901) %!t(uint64=17055393785846869528) %!t(int64=0) %!t(uint64=185)} is not a Resource
[root@sensu-backend-peer1 ~]# sensuctl cluster health --format wrapped-json
Error: &{%!t(uint64=18198607357687572901) %!t(uint64=17055393785846869528) %!t(int64=0) %!t(uint64=185)} is not a Resource
```

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): 5.6
* Installation method (packages, binaries, docker etc.): RPM
* Operating System and version (e.g. Ubuntu 14.04): CentOS 7

nikkiki:

## What is this change?

Implements a generic `PrintTitle` function and properly checks the configuration format. The function will print a styled title for tabular format only.

Closes #2917 
Closes #2918 

Yas.

Nope, this was totally my fault and I knew exactly what I did wrong.

Nah.

We good.

```
$ sensuctl cluster health --format tabular
$ sensuctl cluster health --format json
$ sensuctl cluster health --format wrapped-json
$ sensuctl cluster health --format yaml
$ sensuctl cluster health --format foo
$ sensuctl cluster member-list --format tabular
$ sensuctl cluster member-list --format json
$ sensuctl cluster member-list --format wrapped-json
$ sensuctl cluster member-list --format yaml
$ sensuctl cluster member-list --format foo
```

### DIFF
--- a/cli/commands/helpers/print.go
+++ b/cli/commands/helpers/print.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/sensu/sensu-go/cli/client/config"
 	"github.com/sensu/sensu-go/cli/commands/flags"
+	"github.com/sensu/sensu-go/cli/elements/list"
 	"github.com/sensu/sensu-go/types"
 	"github.com/spf13/cobra"
 )
@@ -57,4 +58,21 @@ func PrintFormatted(flag string, format string, v interface{}, w io.Writer, prin
 	default:
 		return printToList(v, w)
 	}
+}
+
+// PrintTitle prints a title for tabular format only.
+// Flag overrides the cli config format, if set.
+func PrintTitle(flag string, format string, title string, w io.Writer) error {
+	if flag != "" {
+		format = flag
+	}
+	// checking the formats exclusively to cover invalid formats
+	// that get defaulted to tabular
+	if format != config.FormatJSON && format != config.FormatWrappedJSON && format != config.FormatYAML {
+		cfg := &list.Config{
+			Title: title,
+		}
+		return list.Print(w, cfg)
+	}
+	return nil
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 - Fixed `sensuctl` color output on Windows.
+- Fixed a regression in `sensuctl cluster` json/wrapped-json output.
 
 ## [5.6.0] - 2019-04-30
 
--- a/cli/commands/cluster/health.go
+++ b/cli/commands/cluster/health.go
@@ -7,7 +7,6 @@ import (
 	"github.com/coreos/etcd/etcdserver/etcdserverpb"
 	"github.com/sensu/sensu-go/cli"
 	"github.com/sensu/sensu-go/cli/commands/helpers"
-	"github.com/sensu/sensu-go/cli/elements/list"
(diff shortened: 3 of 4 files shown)


cli/commands/helpers/print.go
```diff
@@ -58,3 +59,20 @@ func PrintFormatted(flag string, format string, v interface{}, w io.Writer, prin
...
+
+// PrintTitle prints a title for tabular format only.
+// Flag overrides the cli config format, if set.
+func PrintTitle(flag string, format string, title string, w io.Writer) error {
+	if flag != "" {
+		format = flag
+	}
+	// checking the formats exclusively to cover invalid formats
+	// that get defaulted to tabular
+	if format != config.FormatJSON && format != config.FormatWrappedJSON && format != config.FormatYAML {
```
ccressent: I'm not sure I understand 100% why this can't be `if format == config.FormatTabular` instead of `!=` all the other types. Given your comment on line 70 I'm assuming I'm missing something!

I think invalid formats get defaulted to `config.DefaultFormat`, which may or may not be equal to `config.FormatTabular`.

nikkiki: Invalid formats do get formatted to tabular, but the flag would still represent something else. If we did `if format == config.FormatTabular`, then `--format foo` for example would override the config, and the title would not get printed. See `PrintFormatted` above where the `default` block covers both cases.

nikkiki: L66-L67 have the flag override the config format, which is the intended behavior. So it's possible that `format` doesn't yet abide to a defined format.

ccressent: I see, so that `config.DefaultFormat` is only the default value used when the flag is not mentioned at all on the command line. I thought we were rewriting unknown formats to become `config.DefaultFormat` in order to always know all the values that the format variable can take. Maybe we should. Else that branch with all the `!=` will break when/if we add additional formats and forget to update it.

Issue #2905, opened 2019-05-01
Tessen Phase 2


nikkiki:
To develop a [magic number](https://trello.com/c/VboOtI3I/235-develop-a-magic-number-for-sensu-go) for Sensu Go, we need to leverage the existing Tessen/TessenD framework to collect additional metrics. The additional metrics that we’re interested in at this point are the number of resources by type, and the number of events being processed.

Spec: https://docs.google.com/document/d/1oofPosaGakUyD5joYkseO8s_GOLi5CesnovmodvqSc0

nikkiki:
RE: https://github.com/sensu/sensu-go/pull/2936#discussion_r282889058 and collecting metrics for enterprise resources.. we will reconvene for discussion next week.

PR #2936, opened 2019-05-10
Add per resource counts to tessen data collection

### LINKED ISSUE #2909: Collect per resource counts for tessen
Update [getTessenMetrics()](https://github.com/sensu/sensu-go/blob/9dfb8d29ace7ac94d4c89a62a09196a957855393/backend/tessend/tessend.go#L397) so that TessenD counts the total number of each of the following resources:
- entity (already implemented)
- backends (already implemented)
- asset
- check
- cluster-role
- cluster-role-binding
- event
- filter
- handler
- hook
- mutator
- namespace
- role
- role-binding
- silenced
- user

To distribute the load on etcd, allow about 5-10 seconds before proceeding to the next resource.

nikkiki:

## What is this change?

Adds per resource counts to tessen data collection.

Closes #2909

Yas.

Nah.

We didn't estimate this ticket, so I understand if there are questions regarding the implementation.

https://github.com/sensu/sensu-docs/issues/1445

Verified the new data points showed up in the Tessen dashboard.
https://github.com/sensu/sensu-go-qa-crucible/issues/42

### DIFF
--- a/backend/store/etcd/silenced_store.go
+++ b/backend/store/etcd/silenced_store.go
@@ -25,7 +25,6 @@ var (
 // splat as a checkName may return multiple values.
 
 // GetSilencedPath gets the path of the silenced store.
-// populates type keyBuilder with org and env info, returns a prefix
 func GetSilencedPath(ctx context.Context, name string) string {
 	return silencedKeyBuilder.WithContext(ctx).Build(name)
 }
--- a/backend/store/etcd/silenced_store.go
+++ b/backend/store/etcd/silenced_store.go
@@ -24,8 +24,9 @@ var (
 // stored with a splat in string form. Key lookups for subscriptions with a
 // splat as a checkName may return multiple values.
 
+// GetSilencedPath gets the path of the silenced store.
 // populates type keyBuilder with org and env info, returns a prefix
-func getSilencedPath(ctx context.Context, name string) string {
+func GetSilencedPath(ctx context.Context, name string) string {
 	return silencedKeyBuilder.WithContext(ctx).Build(name)
 }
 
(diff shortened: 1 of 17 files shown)


backend/tessend/tessend.go
```diff
@@ -42,6 +42,32 @@ const (
...
+
+	// perResourceDuration is the duration of time, in seconds, that TessenD will
+	// wait in between resources when collecting its respective count.
+	perResourceDuration = 5 * time.Second
+)
+
+var (
+	// resourceMetrics maps the metric name to the etcd function
+	// responsible for retrieving the resources store path.
+	resourceMetrics = map[string]func(context.Context, string) string{
```
palourde: A bit off topic but reading through this list just made me realize it would be cool to have some metrics about enterprise resources. Do you know if that's something planned?

backend/store/etcd/silenced_store.go
```diff
@@ -24,8 +24,9 @@ var (
 // stored with a splat in string form. Key lookups for subscriptions with a
 // splat as a checkName may return multiple values.
 
+// GetSilencedPath gets the path of the silenced store.
```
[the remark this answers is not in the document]
nikkiki: Not a nitpick.. that's a good catch!

backend/tessend/tessend.go
nikkiki: That's a really good question. I don't think we explicitly talked about this. @annaplotkin @portertech thoughts?

I suppose if we did want enterprise resource metrics, we would have to export this field and append any enterprise store paths of interest.

annaplotkin: @nikkiki We haven't discussed the Enterprise portion specifically but since folks with licenses are required to keep Tessen on, it would make sense 👍 Would love to hear @portertech thoughts as well.

nikkiki: https://github.com/sensu/sensu-go/issues/2905#issuecomment-491410646

nikkiki:
[GolangCI](http://golangci.com/) is currently returning a 522. I ran `golangci-lint run` manually and no _new_ issues were found. I will however be filing an issue for existing linter complaints. Merging this!

Issue #2760, opened 2019-02-28
Leases aren't being cleaned up after a server is deregistered


gattie:
## Expected Behavior
After an agent is removed or deregisters from sensu-go the keepalive leases should disappear and not return after their TTL is met.

## Current Behavior
TL;DR - When an agent is removed from sensu-go, either by auto deregistration or manually deleting it from the UI or API, once all of it's keepalive leases reach their TTL they all get recreated and leased over and over at the TTL interval forever.

I connect an agent to my sensu-go cluster.

If I run `etcdctl least list` I see 7 total leases. 6 of them represent each backend, and the 7th is the agent's keepalive.

Every 20s (the keepalive value) a new lease is created and the old lease no longer has an associated key.

Since the keepalive value is 20s and the TTL is 120s we end up with 5 leases with no attached keys and 1 lease for the agent's keepalive key.

Running something like `for i in $(etcdctl lease list | tail -n +2); do etcdctl lease timetolive --keys $i; done` I can see each lease and the keys in that lease.

```
lease 5f816935c078c817 granted with TTL(120s), remaining(5s), attached keys([])
lease 5f816935c078c882 granted with TTL(120s), remaining(25s), attached keys([])
lease 5f816935c078c8ed granted with TTL(120s), remaining(45s), attached keys([])
lease 5f816935c078c019 granted with TTL(60s), remaining(45s), attached keys([/sensu.io/backends/5f816935c078c019])
lease 491f6935c078f90c granted with TTL(60s), remaining(47s), attached keys([/sensu.io/backends/491f6935c078f90c])
lease 61956935c0789c0c granted with TTL(60s), remaining(49s), attached keys([/sensu.io/backends/61956935c0789c0c])
lease 5f816935c078c027 granted with TTL(60s), remaining(54s), attached keys([/sensu.io/backends/5f816935c078c027])
lease 5f816935c078c032 granted with TTL(60s), remaining(54s), attached keys([/sensu.io/backends/5f816935c078c032])
lease 5f816935c078c043 granted with TTL(60s), remaining(54s), attached keys([/sensu.io/backends/5f816935c078c043])
lease 5f816935c078c958 granted with TTL(120s), remaining(65s), attached keys([])
lease 5f816935c078c9c3 granted with TTL(120s), remaining(85s), attached keys([])
lease 5f816935c078ca2e granted with TTL(120s), remaining(105s), attached keys([/sensu.io/switchsets/keepalived/default/sensu-go-agent])
```

[...]

PR #2940, opened 2019-05-10
Always use the namespace in entities switchsets

palourde:

## What is this change?

It ensures we always use the entity's namespace within the keepalives switchets. Previously, we would bury `/sensu.io/switchsets/keepalived/whisky.local` while creating a lease for `/sensu.io/switchsets/keepalived/default/whisky.local` (notice the `default` namespace), so the switchsets would never actually be deleted.

Closes https://github.com/sensu/sensu-go/issues/2760

Added!

I've tested and re-tested this change with multiple scenarios, I believe it's 100% working but I'm not totally familiar with this codebase.

Nope!

Nope!

Manually tested with a 3-nodes cluster, probably a good idea to test against sensu-staging once the benchmarking is done.


backend/liveness/liveness.go
```diff
@@ -139,7 +139,8 @@ func NewSwitchSet(client *clientv3.Client, name string, dead, alive EventFunc, l
 // Bury buries a live or dead switch. The switch will no longer
 // or callbacks.
 func (t *SwitchSet) Bury(ctx context.Context, key string) error {
-	key = path.Join(t.prefix, key)
+	t.logger.WithFields(logrus.Fields{"key": key}).Debug("buring key")
```
echlebek: s/buring/burying/

palourde:
TODO @palourde : Figure out how to delete the swithsets keys that do not include a namespace, so they don't spam etcd with PUT operations!

palourde:
@echlebek I've realized my change would probably have broke the check TTL feature, so I've done a bit of refactoring so all 3 public methods of the liveness interface accept an _ID_ instead of the full key path. Also, I modified the keepalived logic so if it receives a "bad key", it buries the key directly instead of letting it trigger a new event in the watcher.

I've tested keepalives & check TTLs, everything seems to be working now!

backend/keepalived/keepalived.go
```diff
@@ -328,23 +329,29 @@ func (k *Keepalived) alive(key string, prev liveness.State, leader bool) bool {
...
+		// events to the watcher.
 		lager.Error(err)
-		return false
+		return true
 	}
 
 	lager = lager.WithFields(logrus.Fields{"entity": name, "namespace": namespace})
 	lager.Warn("keepalive timed out")
 
-	if !leader {
```
echlebek: Why has this been removed? This prevents the non-leader nodes from making unnecessary requests.

palourde: The diff is confusing but it has only been moved up, on line 336, so we don't return `true`, and therefore bury the key if it's invalid and we are not the leader. Makes sense?

echlebek: I see! That makes sense, but wondering if it should happen after the logger gets its additional context (namespace, etc)

palourde: Ah yes, I can move things around so we add the entity name & namespace before logging everything and then check the error.

PR #2951, opened 2019-05-14
[GraphQL] Implement simpler extensible filtering

jamesdphillips:
## What is this change?

This pull request replaces the existing JS based approach of filtering with a simpler extensible one. While this approach makes filtering slightly less expressive, by no longer being burdened by the `dynamic` and JS VM packages it is now significantly more efficient and simpler to reason about.

image

### Interface

On fields where a `filter` argument was previously accepted, the argument is now deprecated and replaced with a `filters` argument that takes a list.

```graphql
# for example
events(
  # ...
  "deprecated; please use filters"
  filter: String
  "filters accepts a list of arbitrary statements."
  filters: [String!] = []
): MyConnection!
```

Filters take the form of a identifier and a value separated but a colon (`KEY:VAL`). The identifier is used to match the statement with a particular matcher and the value is used by the matcher to determine if the resource should appear in the filter list of results. For example, the events field has among others `status` & `silenced` filters configured, allowing me to express `["status:critical","silenced:false"]`.

image

* Improved efficiency should improve friction as user's events & entities grow.
* Extensible interface will allow for label & field selectors to be added. (sensu/sensu-enterprise-go#285)

Yup.

No.

Not particularly.

Not as of yet. I will do this after the web app has been updated.

Unit tests & manual testing.


backend/apid/graphql/handler_filters.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	v2 "github.com/sensu/sensu-go/api/core/v2"
+	"github.com/sensu/sensu-go/backend/apid/graphql/filter"
+)
+
+// HandlerFilters returns collection of filters used for matching resources.
+func HandlerFilters() map[string]filter.Filter {
+	filters := map[string]filter.Filter{
+		// type:pipe | type:socket | type:set
```
wesleycyu: I'm not sure `type:socket` is a correct example. In the docs, I think we distinguish between `udp` and `tcp` connections ([src](https://docs.sensu.io/sensu-go/5.7/reference/handlers/#spec-attributes)).

I might be missing something, but is there logic that takes the `type:socket` filter and returns `udp` and `tcp` handlers?

PR #2956, opened 2019-05-14
Add --etcd-cipher-suites backend config option

echlebek:
## What is this change?

This changeset adds the `--etcd-cipher-suites` config option to `sensu-backend`. As a necessity, etcd is upgraded to 3.3.13, since the current vendored version does not support overriding cipher suites.

Operators need to be able to defend their TLS servers from attacks on weak TLS ciphers.

Yes

The change will be tested on sensu-staging with a variety of cipher suites and the testssl.sh tool.


echlebek:
I tested this on sensu-staging and can confirm that we successfully hook the etcd `cipher-suites` option. However, this actually doesn't provide much benefit, as the TLS 1.3 cipher suites aren't working yet.

A better option for users that wish to use TLS 1.3 would be adding an environment variable:
```
GODEBUG="tls13=1"
```

PR #2966, opened 2019-05-16
Add chunk-size option to sensuctl (pagination)

### LINKED ISSUE #2904: [Pagination] Add sensuctl support
Now that the REST API supports pagination, we need to add it to sensuctl.

TODO: Determine if a default limit is desired (e.g. 500).

ccressent:
## What is this change?

This change is somewhat of a pale version of what I had originally envisioned.
It is nonetheless a step toward having a better behaving `sensuctl` on
installations with large number of objects, greatly reducing if not completely
removing potential timeouts during API calls.

The idea is to introduce a new `--chunk-size` flag on `list` sub-commands to
tell the client to internally use the API's pagination capability to retrieve
chunks of at most `chunk-size` objects. The sum of these chunks corresponds to
the same result set that would have been retrieved in one go, without the
`--chunk-size` flag.

Alleviates problems when trying to list large number of resources.
Closes #2904.

Yes, added.

The original intention was to stream the chunks to the output, as to be able to
start displaying results as soon as the first chunk arrives. This would have lead
to a pretty constant memory usage and more "responsiveness" in the output.

As it turns out, receiving a resource set as a collection of independent chunks
raises a few questions:

-   With `tabular` output, the header may need to be repeated
    This is in order to keep proper table alignment in the output. This might be
    indesirable for users trying to parse that output (they should use one of the
    structured outputs for that though). `kubectl` repeats the header, though that
    sometimes confuse people. See <https://github.com/kubernetes/kubernetes/issues/65727>.

-   With `json` or `yaml` output, we need to "stream" to the output
    Each chunk is an array of objects. Simply printing them as they come changes
    the output to become an array of array of objects with the current printing
    methods we have. This is not desirable because it makes the output of `list`
    and `list --chunk-size x` different and breaks operations like dumping the
    checks in a json file (e.g. `sensuctl check list --format json --chunk-size
      500 > my-checks.json`) and loading them elsewhere (`sensuctl create -f
      my-checks.json`).

As a consequence, right now, we reassemble all the chunks before printing
everything in one go. That still fixes potential timeouts when querying the API
for large object sets, but it doesn't change the ballooning memory usage of
`sensuctl` or the delay before showing some output.

Ideally, we would rework or add to the `Print*()` helpers in order for them to
support a more streaming way of printing objects. With that out of the way, we
would be able to either:

-   change `cli.Client` so that it can stream its response object set to its caller
-   move the pagination logic out of `cli.Client` and into `cli.Commands`, in
    order for the commands to "print as they query"

The `sensuctl` reference needs to be updated. This is tracked here: https://github.com/sensu/sensu-docs/issues/1467.

- manual testing on local sensu instance
- checked against the `cli.commands` tests

Testing this feature uncovered a bug in the Namespace and User pagination: https://github.com/sensu/sensu-go/pull/2960.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 - Added event processing counts to tessen data collection.
 - Added ability to query for `Handlers` (individual and collections) from the GraphQL query endpoint.
 - --etcd-cipher-suites option is now available for sensu-backend.
+- Added the `--chunk-size` flag to `sensuctl * list` sub-commands
 
 ### Changed
 - eventd and keepalived now use 1000 handlers for events.
--- a/backend/apid/graphql/dataloader.go
+++ b/backend/apid/graphql/dataloader.go
@@ -35,7 +35,7 @@ func loadAssetsBatchFn(c client.APIClient) dataloader.BatchFunc {
 	return func(ctx context.Context, keys dataloader.Keys) []*dataloader.Result {
 		results := make([]*dataloader.Result, 0, len(keys))
 		for _, key := range keys {
-			records, err := c.ListAssets(key.String(), client.ListOptions{})
+			records, err := c.ListAssets(key.String(), &client.ListOptions{})
 			result := &dataloader.Result{Data: records, Error: handleListErr(err)}
 			results = append(results, result)
 		}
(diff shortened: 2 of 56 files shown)


ccressent:
Yes, sorry I wasn't clear: using `--chunk-size` on Namespace and User _requires_ the backend to have #2960!

I probably should have just kept everything in 1 PR to make that clear.

Issue #2644, opened 2019-01-29
Tuning embedded etcd


nikkiki:
An operator _may_ find it beneficial to configure embedded etcd's storage size `QuotaBackendBytes` and max request size `MaxRequestBytes`.
https://github.com/sensu/sensu-go/blob/66b0502402384f7a5de669513f2e1862473a0e74/backend/etcd/etcd.go#L161-L162 These values can be loaded as backend flags and config file arguments.

Related to https://github.com/sensu/sensu-go/issues/2641

## Expected Behavior
The following command would configure embedded etcd's variables accordingly.
```
sensu-backend start --etcd-max-request-bytes $((1.5 * 1024 * 1024)) --etcd-quota-backend-bytes $((4 * 1024 * 1024 * 1024))
```

## Current Behavior
There is no way to configure embedded etcd's variables with sensu, however using external etcd would probably allow you to do so.

## Possible Solution
Provide config arguments. https://github.com/sensu/sensu-go/issues/2641 solves a similar issue.

## Steps to Reproduce (for bugs)
N/A

## Context
This would allow us to empower operators to tune embedded etcd. Developers may need to better understand the implications of allowing operators to tune embedded etcd rather than using default (and suggested) values.

## Your Environment
N/A

ninjasloth22:
I would also like to be able to tweak things like the heartbeat interval and election timeouts. 
Some way to add additional configuration options to etcd would be great.

palourde:
Closed by https://github.com/sensu/sensu-go/pull/2968

PR #2968, opened 2019-05-16
Make etcd database size configurable

echlebek:
## What is this change?

Etcd database size is now configurable, along with etcd maximum request size.

Some users may wish to tune these parameters when their workflows demand it.

Affirmative

Negative

Yes, not salient to the PR.

Docs changes not required, these are advanced options and best exposed through the CLI help.

I ran the backend with a maximum database size of 1024 bytes, and observed that it failed to start. Similar for max request size.


nikkiki:
Does this PR close https://github.com/sensu/sensu-go/issues/2644?

echlebek:
@nikkiki indeed

backend/etcd/etcd.go
```diff
@@ -31,6 +31,9 @@ const (
 	// EtcdStartupTimeout is the amount of time we give the embedded Etcd Server
 	// to start.
 	EtcdStartupTimeout = 60 // seconds
+
+	DefaultMaxRequestBytes   = 1.5 * (1 << 20)
```
nikkiki: Would you oppose comments for these constants?

backend/etcd/etcd.go
```diff
@@ -162,8 +170,8 @@ func NewEtcd(config *Config) (*Etcd, error) {
 	// revision.
 	cfg.AutoCompactionMode = "revision"
 	cfg.AutoCompactionRetention = "2"
-	// Default to 4G etcd size. TODO: make this configurable.
-	cfg.QuotaBackendBytes = int64(4 * 1024 * 1024 * 1024)
+	cfg.QuotaBackendBytes = config.QuotaBackendBytes
```
nikkiki: Do we need to do any validation for these values? Curious what the implications are if an arbitrary/random number is selected.

palourde: Here's a good explanation about this limit here: https://github.com/etcd-io/etcd/issues/7045#issuecomment-268586740

If you configure it to use 12GB, etcd will print the following warning in the logs:
```
{"component":"etcd","level":"warning","msg":"backend quota 12884901888 exceeds maximum recommended quota 8589934592","pkg":"etcdserver","time":"2019-05-17T09:13:41-04:00"}
```

If you use a much lower value, sensu-backend will reach the quota real fast and errors will show up in the logs about `database space exceeded` or something like this.

I don't feel like we need to validate values passed directly to etcd but I'm okay if you'd prefer to!

backend/etcd/etcd.go
echlebek: Done.

nikkiki: giphy

Issue #2957, opened 2019-05-14
Use protobuf serialization when storing resources in etcd


echlebek:
Rationale
-------------
A recent battery of load tests with some experimental patches showed that protobuf serialization dramatically improved event throughput and database size, by reducing the size of KVs stored in etcd.

With protobuf serialization enabled for events and entities, we achieved twice the event throughput, while using much less database storage. Since etcd databases cannot grow to a very large size, this proved to be a very useful improvement, allowing far more events to be stored.

Spec
-------
Implement protobuf serialization across all data types, allowing for backwards compatibility with databases that already contain JSON. The store should transparently work with either serialization format for reads, but only write to the store with protobuf format going forward.

Backwards Compatibility
---------------------------------
All methods that read from the store should try deserializing with `proto.Unmarshal` first, and try `json.Unmarshal` on error. If both methods produce an error, return the result of `proto.Unmarshal`.

Proof of concept
----------------------
A PoC exists here: https://github.com/sensu/sensu-go/commit/f5cf3c2527ec6acc500a525e9ed186e5873e0f80

PR #2973, opened 2019-05-17
Protobuf serialization

palourde:
## What is this change?

This PR introduces Protobuf serialization for most of our resources.

Closes https://github.com/sensu/sensu-go/issues/2957

Done

Nope

Some resources were not generated using protobuf and others stored using `types.Wrapper` will have to wait!

Nope

I added a specific regression test in `TestList`. I also tested deployed directly this new version with sensu-staging, and tested an upgrade scenario.


backend/store/etcd/generic_object.proto
```diff
@@ -0,0 +1,26 @@
...
+option go_package = "etcd";
+option (gogoproto.populate_all) = true;
+option (gogoproto.equal_all) = true;
+option (gogoproto.marshaler_all) = true;
+option (gogoproto.unmarshaler_all) = true;
+option (gogoproto.sizer_all) = true;
+option (gogoproto.testgen_all) = true;
+
+// Asset defines an asset agents install as a dependency for a check.
+message GenericObject {
```
palourde: FYI we use this resource to test the generic store so obviously, to have it working with protobuf I had to create this file.

backend/store/etcd/store.go
```diff
@@ -37,7 +39,7 @@ func NewStore(client *clientv3.Client, name string) *Store {
 
 // Create the given key with the serialized object.
 func Create(ctx context.Context, client *clientv3.Client, key, namespace string, object interface{}) error {
-	bytes, err := json.Marshal(object)
+	bytes, err := proto.Marshal(object.(proto.Message))
```
echlebek: Hmm this seems a bit unsafe. Could we change the function signature?

```
func Create(ctx context.Context, client *clientv3.Client, key, namespace string, object proto.Message) error
```

backend/store/etcd/store_test.go
```diff
@@ -213,45 +191,54 @@ func TestList(t *testing.T) {
...
 		ctx := context.WithValue(context.Background(), types.NamespaceKey, "default")
 		require.NoError(t, Create(ctx, s.client, "/sensu.io/generic/default/obj1", "default", obj1))
 
-		obj2 := &genericObject{ObjectMeta: corev2.ObjectMeta{Name: "obj2", Namespace: "acme"}}
+		// Let's encode a resources with JSON (instead of Protobuf) to ensure the
+		// List function supports both encoding format
+		obj2 := &GenericObject{ObjectMeta: corev2.ObjectMeta{Name: "obj2", Namespace: "acme"}}
 		ctx = context.WithValue(context.Background(), types.NamespaceKey, "acme")
-		require.NoError(t, Create(ctx, s.client, "/sensu.io/generic/acme/obj2", "acme", obj2))
+		// require.NoError(t, Create(ctx, s.client, "/sensu.io/generic/acme/obj2", "acme", obj2))
```
ccressent: Should this line be removed altogether?

palourde: Haha yes, good catch

PR #2976, opened 2019-05-21
Fix the number version with the build.sh script

realPy:
Compile with a linux build environment not specify the Version , SHA and date build.
This settings is correctly set with the windows build environnement.

This pull request fix this missing settings

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -71,6 +71,7 @@ build_binary () {
     unset GOOS
     unset GOARCH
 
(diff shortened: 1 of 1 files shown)


nikkiki:
@amdprophet I think you might be the best person to look at this. Seems OK on my end.

amdprophet:
We're no longer using this build script to produce our packages but I'll merge this as you're likely trying to build your own packages.

PR #2987, opened 2019-05-23
[GraphQL] Add toJSON field to core resources

jamesdphillips:
## What is this change?

Add `toJSON` field to core GraphQL types to allow front-end to retrieve the REST API representation of the type.

Required for sensu/sensu-enterprise-go#443

Sure!

No.

No.

None required.

Unit & manual.


backend/apid/graphql/check.go
```diff
@@ -95,8 +96,7 @@ func (r *checkCfgImpl) Silences(p graphql.ResolveParams) (interface{}, error) {
 
 // ToJSON implements response to request for 'toJSON' field.
 func (r *checkCfgImpl) ToJSON(p graphql.ResolveParams) (interface{}, error) {
-	check := p.Source.(*types.CheckConfig)
-	return types.WrapResource(check), nil
+	return types.WrapResource(p.Source.(v2.Resource)), nil
```
ccressent: You can probably change `types.WrapResource()` to `v2.WrapResource()` while you're at it, maybe leading to the removal of the `github.com/sensu/sensu-go/types` import altogether.

backend/apid/graphql/check.go
```diff
@@ -221,6 +221,11 @@ func (r *checkImpl) RuntimeAssets(p graphql.ResolveParams) (interface{}, error)
 	return results, err
 }
 
+// ToJSON implements response to request for 'toJSON' field.
+func (r *checkImpl) ToJSON(p graphql.ResolveParams) (interface{}, error) {
+	return types.WrapResource(p.Source.(v2.Resource)), nil
```
ccressent: Use `v2` instead of `types` in new code :)

backend/apid/graphql/schema/asset.graphql
```diff
@@ -26,4 +26,10 @@ type Asset implements Node, Namespaced, HasMetadata {
   queries are joined by the "AND" operator.
   """
   filters: [String]
+
+  """
+  toJSON returns a REST API compatible representation of the resource. Handy for
+  sharing snippets that can then be imported with `sensuctl import`.
```
ccressent: What's `sensuctl import`? Do you mean `sensuctl create`?

backend/apid/graphql/check.go
jamesdphillips: Unless I'm blind, I didn't see the wrapper wasn't in the `v2` package, is this intentional? Otherwise, I can move it over and add aliases in the `types` package.

backend/apid/graphql/schema/asset.graphql
jamesdphillips: That's what I get for doing this quickly and copy and pasting it all over. 😅

backend/apid/graphql/check.go
ccressent: I don't think that's intentional, because `v2` is an alias for `types` and we're supposed to gradually move from `types` to `v2`!

jamesdphillips: Hmm, after looking at this more closely, I noticed that the Wrapper implements it's own map of packages. https://github.com/sensu/sensu-go/blob/master/types/wrapper.go#L32-L37. This leads me to believe that the Wrapper was intended to stay outside of specific api package(?)

/cc @echlebek

jamesdphillips: Resolving as per discussion in Slack.

> Eric [3:28 PM]
> I deliberately left it in the `types` package, since it's not a part of the corev2 API
> 
> Eric [3:29 PM]
> It's meant to be re-usable across API packages

https://sensu.slack.com/archives/CCHNSF7RD/p1558736744034800

PR #2988, opened 2019-05-23
Fix/arm overflow int

realPy:
The last tag and master failed to compiled on ARM architecture with 
`backend/cmd/start.go:290:18: constant 4294967296 overflows int` error


echlebek:
Thanks!

PR #2992, opened 2019-05-23
Revert "Fix the number version with the build.sh script"

amdprophet:
Reverts sensu/sensu-go#2976 as it the date flags differ between GNU & BSD date.


amdprophet:
Closing this as it doesn't actually break builds; the date will not be included when building on BSD or macOS though.

PR #2996, opened 2019-05-24
Update cli/client package so errors are consistent

jamesdphillips:
## What is this change?

Update cli/client package so that all API errors are handled consistently across all resources.

* Generally improves the user experience of using `sensuctl`.
* A predictable interface allows clients (`sensuctl` or for the foreseeable future `graphql`) to handle errors consistently.

Yes, probably.

No.

A reflex to shave yaks.

None required.

Manual.


ccressent:
Thank you for this clean up!

PR #3011, opened 2019-05-30
Refactor eventd

### LINKED ISSUE #3002: Move event update logic into event store
Refactor the `store.UpdateEvent` method so that the merging logic for event updates is contained within. Store consumers like `eventd` would then only need to call `UpdateEvent`, and not concern themselves with merging logic directly.

This pattern is better, because databases with richer transactions can update events in a single transaction, or even in a single write.

echlebek:
## What is this change?

This commit moves a significant portion of event update logic to
the core/v2 package, and another portion to the store package.

Now, the store/etcd.UpdateEvent method does all of the necessary
logic to update an event in-place, and returns the updated event
and also the previous event.

Flap detection logic has been moved into the core/v2 package, and
has been incorporated into the Check.MergeWith method.

This enables a better data access pattern for stores that can
handle event updates in a single transaction.

As an incidental bug fix, check state and check total state change
have been corrected. The previous algorithm used the oldest history
item when doing a comparison with the newest event, rather than the
most recent history item. Additionally, the algorithm looped over
the history in the wrong order before. It now loops over the
history in descending order.

Closes #3002 

Yes

Only integration tests so far. More QA is required before merging this.


backend/store/etcd/event_store.go
```diff
@@ -203,34 +203,58 @@ func (s *Store) GetEventByEntityCheck(ctx context.Context, entityName, checkName
...
+	// Maintain check history.
+	if prevEvent != nil {
+		if !prevEvent.HasCheck() {
+			return nil, nil, errors.New("invalid previous event")
+		}
+
+		event.Check.MergeWith(prevEvent.Check)
 	}
 
+	updateOccurrences(event.Check)
```
palourde: Would it make sense to have this business logic in the `api/core/v2` package too? Or it's only relevant for etcd?

echlebek: It's something that's going to be different for each store unfortunately.

echlebek:
I've tested this with docker-compose, seems to work fine. :)

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Fixed
+- The check state and check total_state_change properties are now more correct.
```
ccressent: :)

PR #3013, opened 2019-05-30
Optimize silenced data access patterns for events

### LINKED ISSUE #3008: Allow retrieving multiple 'Silenced' entries in a single transaction
When events are processed, silenced entries are retrieved for every subscription associated with the check. This is expensive, as it requires that we issue multiple read requests for every event processed.

Instead, we could modify the Silenced store to accept multiple subscription parameters, and return all the information in a single transaction.

This should significantly reduce network and disk I/O for event processing, when checks have multiple subscriptions.

echlebek:
## What is this change?

This change modifies how silenced entries are accessed by eventd, so that the number of requests made to etcd remains constant.

Closes #3008 

Yes

Integration tests, which I believe are sufficient.


backend/eventd/silenced.go
```diff
@@ -76,25 +68,21 @@ func handleExpireOnResolveEntries(ctx context.Context, event *types.Event, store
...
-			}
+	entries, err := store.GetSilencedEntriesByName(ctx, event.Check.Silenced...)
+	if err != nil {
+		return fmt.Errorf("couldn't resolve silences: %s", err)
+	}
+	toDelete := entries[:0]
+	toRetain := []string{}
+	for _, entry := range entries {
+		if entry.ExpireOnResolve {
+			toDelete = append(toDelete, entry)
```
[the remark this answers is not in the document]
echlebek: Ah you're right. This is simply a mistake.

backend/store/etcd/silenced_store.go
```diff
@@ -29,48 +29,56 @@ func GetSilencedPath(ctx context.Context, name string) string {
...
 		return nil, err
 	}
 	return silencedArray, nil
 }
 
-// GetSilencedEntriesBySubscription gets all silenced entries that match a subscription.
-func (s *Store) GetSilencedEntriesBySubscription(ctx context.Context, subscription string) ([]*types.Silenced, error) {
-	if subscription == "" {
-		return nil, errors.New("must specify subscription")
+// GetSilencedEntriesBySubscription gets all silenced entries that match set of subscriptions.
```
nikkiki: Nitpick.. totally don't need to do this lol
```suggestion
// GetSilencedEntriesBySubscription gets all silenced entries that match a set of subscriptions.
```

backend/store/etcd/silenced_store.go
```diff
@@ -190,3 +214,23 @@ func (s *Store) arraySilencedEntries(resp *clientv3.GetResponse) ([]*types.Silen
...
+	for _, resp := range resp.Responses {
+		for _, kv := range resp.GetResponseRange().Kvs {
+			leaseID := clientv3.LeaseID(kv.Lease)
+			ttl, err := s.client.TimeToLive(ctx, leaseID)
+			if err != nil {
+				return nil, fmt.Errorf("couldn't get silenced entries: %s", err)
+			}
+			var silenced corev2.Silenced
+			if err := unmarshal(kv.Value, &silenced); err != nil {
+				return nil, fmt.Errorf("couldn't get silenced entries: %s", err)
```
nikkiki: Another nitpick, we reuse the error statement `couldn't get silenced entries` about 5 times. We _could_ consolidate.

echlebek: Is the suggestion:

```
func silencedErr(err error) error {
    return fmt.Errorf("couldn't get silenced entries: %s", err)
}
```
and use that function when errors are returned?

nikkiki: A function, or a constant. Or do nothing at all. Just something I noticed.

echlebek: I've done this in the past, but have come to conclude that the indirection isn't worth it when reading the code.

PR #3016, opened 2019-06-01
Optimize proxy check scheduling (proxy entity filtering)

### LINKED ISSUE #2979: Round robin proxy checks perform poorly with high proxy entity counts
CPU can spike above 100% under certain conditions. These conditions were produced in the staging environment, and reproduced locally. A clustered Sensu installation can contain 1k entities at normal CPU consumption. When round robin proxy checks are added to that environment, we observed abnormal CPU consumption.

From @echlebek:
So here is what we know:
- there are a reasonable number of goroutines running
- the CPU time on backend-2 is being spent on rings and watchers
- there are far too many auth tokens in etcd
- tokens and rings do not directly interact
- there seems to be too much memory allocated for the system load
- the issue has been present for at least one other release
- etcd requests are taking too long to execute, >100 ms
- the key structure of rings looks OK
- the key structure of switchsets looks OK
_after restarting and the CPU drops_
What does this imply?
- the CPU load on backend-2 was due to the program state, not the database state
- so therefore the CPU load was due to an infinite recursion or a resource leak
Steps to reproduce:
- After I created a round robin proxy check, CPU usage increased to ~140%
- So it seems that the CPU usage increases as a function of proxy entities when proxy checks are scheduled
- The 1000 entities do not share a subscription with the proxy check

## Your Environment
* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): 5.7.0, 5.8.0
* Operating System and version (e.g. Ubuntu 14.04): Ubuntu 16

echlebek:
## What is this change?

    This commit uses a batching strategy for filtering proxy entities,
    so that all of the proxy entities that are to be filtered for a
    given check are done so with a single Javascript invocation,
    instead of M*N invocations. This should dramatically reduce CPU
    usage for scheduling proxy checks.
    
    Below is a benchmark of the previous filtering mechanism compared
    with the new one.
    
    BenchmarkEvaluate1000-4                10 183369379 ns/op 174297743 B/op  1581652 allocs/op
    BenchmarkEvaluateEntityFilters1000-4  100  14916854 ns/op 6278989 B/op    166571  allocs/op
    
    The new mechanism is over 10 times faster at filtering 1000 proxy
    entities, uses 10 times fewer allocations, and 6 MB instead of
    174 MB of memory.
    
    The only downside to this approach is that it's no longer possible
    to detect syntax errors versus semantic errors in javascript
    filters, so all errors are logged at debug level.

Close #2979 

Yes

No doc changes required.

Integration tests and unit tests. I have also added benchmarks that demonstrate the performance impact.

EDIT: I realized there were some issues with this commit, in that we'd get duplicate results, and that performance could be improved even further.

------------------------

 Further improve proxy entity filtering

This commit improves upon the last one in several ways.

* It compiles the expressions, since they are re-used between
entity evaluations. This saves a considerable amount of time.

* It avoids allocations inside the javascript VM.

* The function stops evaluating subsequent filters for entities
once they match true.

* Results reported are 1:1 with the number of entities provided,
leading to a more natural API that doesn't require additional
mapping.

The benchmark results:

```
BenchmarkMatchEntities1000-4  1000  1739455 ns/op   1002147 B/op   22663 allocs/op
BenchmarkEvaluate1000-4       10    180055940 ns/op 174295188 B/op 1581635 allocs/op
```

MatchEntities() is over 100 times faster at matching entities than
the naive Evaluate(). Additionally, it only uses 1 MB of memory
instead of 174 MB to evaluate 1000 entities, and 22k allocs instead
of 1.5M.


echlebek:
I tried to use an LRU for caching synthesized entities, but unfortunately entities can contain unhashables.

Issue #1804, opened 2018-07-10
Add unit tests for event filter routes


nikkiki:
Follow the template in https://github.com/sensu/sensu-go/commit/674880f7ab5cdd7ef7908cc246e1c2ee43c3ae5c to make the event filter controller abstract and mock the controller in unit tests. Ensure these tests improve code coverage.

Issue #1037, opened 2018-02-16
Unit test all routers


echlebek:
Most if not all of the routers are lacking unit tests. We should write unit tests for them, and refactor the routers if necessary to facilitate this.

When writing tests, we should ensure that:

- required mux variables are supplied
- path-escaped mux variables and url-escaped query parameters are handled correctly

PR #3018, opened 2019-06-03
APId Refactoring

palourde:
## What is this change?

This PR introduces a significant refactoring to the `backend/apid` package. It started as a bugfix for https://github.com/sensu/sensu-enterprise-go/issues/481 but I quickly got frustrated by how 90%+ of our routers and controllers are almost identical, which results in a lot of code duplication and different behaviors across routers (e.g. a PUT sometimes returns 200 with a response body, sometimes a 204 without a body). 

I apologize for the size of this PR, there's a lot of lines changes but it's mostly caused by code moved around and deleted. It basically boils down to:

### `api/core/v2`
- A `StorePrefix()` method was added to the `Resource` interface. It allows us to handle `corev2.Resource` in apid & the store and determine the key path in the store without having to assert the concrete type.
- I moved code around so most methods that satisfies the `Resource` interface are now grouped together.
- The (huge) `api/core/v2/checks.go` file was split so we can easily make a distinction between a check, a check config & a check request. 

### `backend/apid/actions`
- Most of the controllers of "standard" resources were removed and replaced with handlers (see below).
- I made small adjustments to the methods name so they are standardized & we pass pointers to resources in arguments.

### `backend/apid/handlers`
- In place of a controller, we now have a `Handlers` struct, which deals with `corev2.Resource` and uses a new `store.ResourceStore`.
- This `Handlers` struct stores a `corev2.Resource` which represents the concrete type handled. With this resource, we are able to decode the request payloads and the store keys.
- There's also a new `CheckMeta` function that ensures the request payload matches the request URI (in order to actually fix https://github.com/sensu/sensu-enterprise-go/issues/481)

### `backend/apid/routers`
- Most of the standard resources are now exclusively using handlers (instead of controllers) and now have 100% test coverage.
- Some resources are partially or not using handlers:
	- `ChecksRouter` is using handlers for the common operations (CRUD) but require some custom routes and a controller for adhoc requests and check hooks
	- `EventsRouter` wasn't touched, excepted to add full test coverage and standardize some stuff. Considering we are about to have an enterprise store for this resource, I preferred to wait and it's a bit of a snowflake because it doesn't use a `id` path variable.
	- `SilencedRouter` won't be easy to migrate, because the store needs to deal with leases so using a generic store would be a really hard task.
	- `UsersStore` is also a snowflake, because we need to deal with users passwords.
	- There's actually two routers for Tessen, I would have to take a step back and explore a bit more but I believe `TessenController` could be replaced.
- I added some manual test coverage for routers that do not use handlers so we are now at 75.8% test coverage for this package.
	- I made sure to protect against https://github.com/sensu/sensu-enterprise-go/issues/481 for these resources.

### `backend/store/etcd/resource.go`
- This store is only a wrapper around the generic store, and its only job is to deal with the resource keys in etcd.

I didn't tried to refactor everything, maybe just the easiest 80%. From that point, I believe the next big win would be to refactor most of the store methods so deal with concrete types (and therefore are easier to use in the codebase) but uses the generic store, a bit like RBAC resources do (e.g. https://github.com/sensu/sensu-go/blob/master/backend/store/etcd/role_store.go)

Closes https://github.com/sensu/sensu-enterprise-go/issues/481
Closes https://github.com/sensu/sensu-go/issues/1037
Closes https://github.com/sensu/sensu-go/issues/1801
Closes https://github.com/sensu/sensu-go/issues/1802
Closes https://github.com/sensu/sensu-go/issues/1803
Closes https://github.com/sensu/sensu-go/issues/1804
Closes https://github.com/sensu/sensu-go/issues/1805
Closes https://github.com/sensu/sensu-go/issues/1806
Closes https://github.com/sensu/sensu-go/issues/1807
Closes https://github.com/sensu/sensu-go/issues/1808
Closes https://github.com/sensu/sensu-go/issues/1809
Closes https://github.com/sensu/sensu-go/issues/1810
Closes https://github.com/sensu/sensu-go/issues/1811
Closes https://github.com/sensu/sensu-go/issues/1812

Yes!

I _really_ hesitated with the new `handlers` package, wasn't sure if it would have made more sense to put that stuff into the actions package but I liked the idea to have a clear distinction between the two. Anyway, I'm open to suggestions!

Yep :) 

I deployed several times new clusters with sensu-staging, and even did an upgrade test. I also added a ton of unit tests to test some specific stuff like resources metadata.


backend/apid/routers/helpers_test.go
```diff
@@ -0,0 +1,343 @@
...
+		storeFunc: func(s *mockstore.MockStore) {
+			s.On("DeleteResource", mock.Anything, mock.AnythingOfType("string"), mock.AnythingOfType("string")).
+				Return(&store.ErrInternal{}).
+				Once()
+		},
+		wantStatusCode: http.StatusInternalServerError,
+	}
+}
+
+var deleteResourceSuccessTestCase = func(pathPrefix, kind string) routerTestCase {
```
golangcibot: `deleteResourceSuccessTestCase` is unused (from `varcheck`)

api/core/v2/adhoc.proto
```diff
@@ -14,6 +14,9 @@ option (gogoproto.sizer_all) = true;
 option (gogoproto.testgen_all) = true;
 
 message AdhocRequest {
+  option (gogoproto.face) = true;
+  option (gogoproto.goproto_getters) = false;
+
```
ccressent: Do you have a 2 sentences rationale behind the need set these options?
I haven't read the gogoproto manual so it's not obvious what that is to me ;)

api/core/v2/adhoc.proto
```diff
@@ -24,5 +27,5 @@ message AdhocRequest {
   string reason = 4 [(gogoproto.nullable) = true];
 
   // Metadata contains the name, namespace, labels and annotations of the AdhocCheck
-  ObjectMeta metadata = 5 [(gogoproto.jsontag) = "metadata", (gogoproto.embed) = true, (gogoproto.nullable) = false];
+  ObjectMeta metadata = 5 [(gogoproto.jsontag) = "metadata,omitempty", (gogoproto.embed) = true, (gogoproto.nullable) = false];
```
ccressent: Are we sure that `omitempty` is not going to cause subtle breakage somewhere else?

api/core/v2/asset.go
```diff
@@ -13,11 +12,28 @@ import (
...
-// AssetNameRegex used to validate name of asset
-var AssetNameRegex = regexp.MustCompile("^" + AssetNameRegexStr + "$")
+// StorePath returns the path prefix to assets in the store
+func (a *Asset) StorePath() string {
+	return AssetsResource
+}
+
+// URIPath returns the path component of an asset URI.
+func (a *Asset) URIPath() string {
+	return path.Join(URLPrefix, "namespaces", url.PathEscape(a.Namespace), AssetsResource, url.PathEscape(a.Name))
```
ccressent: I _think_ it's unlikely to be a problem, but the old `Sprintf()` implementation would return a string with a trailing `/` if `a.Namespace` is `""`. This implementation with `path.Join()` doesn't.

ccressent:
So, in summary, we want to have more general code that works on `Resource` rather than duplicated code for each resource type. I like the idea and I'm all for handling everything in a uniform fashion, enforced by the fact that all types go through the same functions.

My concern is about potentially creating subtle breakages. I haven't seen any obvious ones while having a look at the change, but we never know! Hopefully we have enough relevant test cases... yes, I'm a bit paranoid ;)

Is the plan to extend what you did to `Check` to all the other resource types?
Likewise, will all the various store methods be replaced by the new `{Get, List, ...}Resource()`?

api/core/v2/adhoc.proto
palourde: Long story short, I these attributes missing were causing a bug but turns out it didn't. I reverted these changes to prevent any weird bug!

api/core/v2/adhoc.proto
palourde: See my comment above!

api/core/v2/asset.go
palourde: The only place in our codebase where we use the `URIPath` method (for actual logic, and not printing something) is: https://github.com/sensu/sensu-go/blob/5ffdd077e8df24d76bafa769577fa78f5cb8e3cd/cli/client/generic.go#L109

We can safely assume that we have a resource name at this place since we are doing a PUT request. So basically yeah, it won't cause any problem!

PR #3026, opened 2019-06-05
Implement POST /api/core/v2/tessen/metrics

### LINKED ISSUE #2982: [Tessen] Create an API endpoint for tessen metrics
Create an unauthorized API endpoint, POST `/api/core/v2/tessen/metrics`. This route should accept a payload of `[]corev2.MetricPoint`:
https://github.com/sensu/sensu-go/blob/a786ceeb603601f89a39276824de346b098ad5e4/api/core/v2/metrics.pb.go#L86-L99

Create a message bus topic called `TopicTessenMetrics` (similar to `TopicTessen`):
https://github.com/sensu/sensu-go/blob/9dfb8d29ace7ac94d4c89a62a09196a957855393/backend/messaging/message_bus.go#L26-L27

The endpoint should bypass the store, and send the metrics off to a specific message bus channel (similar to the tessen config controller):
https://github.com/sensu/sensu-go/blob/9dfb8d29ace7ac94d4c89a62a09196a957855393/backend/apid/actions/tessen.go#L35-L41

nikkiki:

## What is this change?

Adds the POST `/api/core/v2/tessen/metrics` API endpoint.

Closes #2982 

Yas.

Nah.

Nah.

New docs are not required.

Unit/integration coverage. E2E tested with POST the following []corev2.MetricPoint on `/api/core/v2/tessen/metrics`.
```
[
    {
        "name": "cpu.idle_percentage",
        "value": 61,
        "timestamp": 1525462242,
        "tags": [
            {
                "name": "name0",
                "value": "value0"
            },
            {
                "name": "name1",
                "value": "value1"
            }
        ]
    },
    {
        "name": "mem.sys",
        "value": 104448,
        "timestamp": 1525462242,
        "tags": []
    }
]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Added
 - [Web] Adds ability to delete entities
+- Added POST `/api/core/v2/tessen/metrics`.
 
 ### Changed
 - [Web] Updated embedded web assets from `275386a` ... `b0c1138`
--- a/backend/apid/actions/tessen.go
+++ b/backend/apid/actions/tessen.go
@@ -54,3 +54,24 @@ func (c TessenController) Get(ctx context.Context) (*corev2.TessenConfig, error)
 
 	return config, nil
 }
+
+// TessenMetricController exposes actions which a viewer can perform
+type TessenMetricController struct {
+	bus messaging.MessageBus
+}
+
+// NewTessenMetricController returns a new TessenMetricController
+func NewTessenMetricController(bus messaging.MessageBus) TessenMetricController {
+	return TessenMetricController{
+		bus: bus,
+	}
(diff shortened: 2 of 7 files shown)


nikkiki:
@ccressent I can definitely understand your concern about the endpoint being unauthorized. This was apart of our collaborative spec with our front end folks, as we do plan in the future to track metrics on unauthorized actions such as failed login attempts.

ccressent:
I figured that must have been the reason... I'm not asking for the spec (or this implementation) to change now, but now that I think about it, maybe we should have planned for several endpoints, like `metrics/privileged` and `metrics/unprivileged`, or for the API to check what type of metric is being posted and allow/disallow depending on authentication?

10xjs:
The ability to call this from the FE is working great.

However, is this missing the timestamp and cluster information injection?

PR #3028, opened 2019-06-06
[Graphql] Simple autosuggestion

jamesdphillips:
## What is this change?

Adds a field to the GraphQL service that returns a set of values associated with a given field, type and namespace. Enabling add initial rudimentary autosuggestion features to our clients.

**Ref**
    
The 'ref' is an identifier used to uniquely reference a resource in the system as well as a field on said resource. Refs take the form: `:group/:version/:type/:field`. (Within reason,) `:group/:version` should always match the package, `:type` should match a name registered in the typemap, and the `:field` components should be snake_case. Further, the field segment may be a path in and of it's own, eg. `metadata/name` would refer to the name field nested inside a resource's metadata.

The following are valid examples:
 
  `core/v2/asset/metadata/name`
        `core/v2/asset/metadata/labels`
        `core/v2/asset/metadata/labels/region`
        `core/v2/check_config/subscriptions`
        `core/v2/check_config/command`
        `core/v2/check_config/timeout`
        `core/v2/entity/system/os`
        `core/v2/entity/system/platform`
        `core/v2/filter/metadata/name`
        `core/v2/handler/command`
        `core/v2/hook_config/command`
        `core/v2/mutator/command`
        `core/v2/mutator/timeout`
        `core/v2/silenced/creator`

**Schema**

```graphql
extend Query {
  """
  Given a ref, field and a namespace returns a set of suggested values.

  As an example if you would like a list of check names you might use:
  suggest(ref: "core/v2/check_config/metadata/name", namespace: "default")

  Or, if you would like a list of subscriptions...
  suggest(ref: "core/v2/entity/subscriptions", namespace: "default")

  You may filter the results with the q argument, for example:
  suggest(ref: "core/v2/check_config/metadata/name", namespace: "default", q: "disk")

  By default the results are ordered by the frequency in which the result occurs in the set. The order argument allow you to tweak this behaviour, for example:
  suggest(ref: "core/v2/check_config/metadata/name", namespace: "default", order: ALPHA_DESC)
  """
  suggest(
    """
    If the value of a field does not contain the value of this argument it will
    be omitted from the response. Operation is case-insensitive.
    """
    q: String = "",
    """
    Ref is used to uniquely identify a resource in the system as well as a field
    on said resource. Refs take the form: :group/:version/:type/:field. The
    field segment may be a path in and of it's own, eg. metadata/name would
    refer to the name field nested inside a resource's metadata.
    The following are valid example values for this argument:
    
        `core/v2/asset/metadata/name`
        `core/v2/asset/metadata/labels`
        `core/v2/asset/metadata/labels/region`
        `core/v2/check_config/subscriptions`
        `core/v2/check_config/command`
        `core/v2/check_config/timeout`
        `core/v2/entity/system/os`
        `core/v2/entity/system/platform`
        `core/v2/filter/metadata/name`
        `core/v2/handler/command`
        `core/v2/hook_config/command`
        `core/v2/mutator/command`
        `core/v2/mutator/timeout`
        `core/v2/silenced/creator`
    """
    ref: String!,
    namespace: String!,
    limit: Int = 10,
    order: SuggestionOrder = FREQUENCY,
  ): SuggestionResultSet
}
```

**Usage**

```graphql
query HandlerSuggestions {
  # fetch the first five handler names; values are sorted alphanumerically
  suggest(ref: "core/v2/handler/metadata/name", namespace: "default", limit: 5, order: ALPHA_DESC) {
    values
  }
}

query LabelSuggestions {
  # fetch the ten most popular label names among checks
  suggest(ref: "core/v2/check_config/metadata/labels", namespace: "default", order: FREQUENCY) {
    values
  }
}

query LabelValueSuggestions {
  # among checks, fetch the first ten values under the label "region". Sorted alphanumerically.
  suggest(ref: "core/v2/check_config/metadata/labels/region", namespace: "default", order: ALPHA_ASC) {
    values
  }
}

query CheckCommandSuggestions {
  # among checks, fetch the first ten commands that contain the string "echo".
  suggest(ref: "core/v2/check_config/command", namespace: "default", q: "echo", order: ALPHA_DESC) {
    values
  }
}

query MutatorTimeoutSuggestions {
  # among mutators, fetch the five most frequently used values for timeout.
  # note that all items in the result set are strings regardless of the original type.
  suggest(ref: "core/v2/mutator/timeout", namespace: "default", order: FREQUENCY, limit: 5) {
    values
  }
}
```

Auto-suggestion.

Yes!

Not really.

Reflection is painful.

None should be required.

Many unit tests and some manual testing as well.


backend/apid/graphql/query.go
```diff
@@ -61,6 +71,81 @@ func (r *queryImpl) Handler(p schema.QueryHandlerFieldResolverParams) (interface
...
+
+	if len(values) > p.Args.Limit {
+		values = values[:p.Args.Limit]
+	}
+
+	results["values"] = values
+	return results, nil
+}
+
+func mustBeSlice(objs interface{}) {
```
golangcibot: `mustBeSlice` is unused (from `deadcode`)

PR #3031, opened 2019-06-06
Add EventStoreProxy

echlebek:
## What is this change?

EventStoreProxy is designed to allow patching the EventStore used
by the backend.

This is a strut for enterprise, so that we can support alternative event stores.

Nein

### DIFF
--- a/backend/store/proxy.go
+++ b/backend/store/proxy.go
@@ -1,0 +1,67 @@
+package store
+
+import (
+	"context"
+	"sync/atomic"
+	"unsafe"
+
+	corev2 "github.com/sensu/sensu-go/api/core/v2"
+	"github.com/sensu/sensu-go/types"
+)
+
+// EventStoreProxy is a mechanism for providing an EventStore with a replaceable
+// underlying implementation. It uses an atomic so that calls are not impeded by
+// mutex overhead.
+type EventStoreProxy struct {
+	impl unsafe.Pointer
+
+	// guard against the store being garbage collected,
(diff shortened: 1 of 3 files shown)


backend/store/proxy.go
```diff
@@ -0,0 +1,67 @@
...
+
+func NewEventStoreProxy(s EventStore) *EventStoreProxy {
+	return &EventStoreProxy{
+		impl:    unsafe.Pointer(&s),
+		gcGuard: s,
+	}
+}
+
+func (e *EventStoreProxy) do() EventStore {
+	return *((*EventStore)(atomic.LoadPointer(&e.impl)))
```
nikkiki: Do you have any literature that might help me understand what this means?

nikkiki: Do you have any literature that might help me understand what this means?

nikkiki: Okay cool GitHub I already said that. Resolving this dupe.

echlebek: https://golang.org/ref/spec#Package_unsafe

Basically, what's happening here is the pointer to the store is being stored as a raw pointer in the `EventStoreProxy`, and is being loaded with the atomic package so that when it's replaced, there is no danger of a race condition.

The expression you are asking about is saying: convert this raw unsafe pointer to a pointer to an `EventStore` interface, and then dereference it, yielding an `EventStore`.

nikkiki: giphy

Issue #2983, opened 2019-05-22
[Tessen] Message handler for metrics topic


nikkiki:
Create a component in TessenD to handle all incoming messages on the channel, populate the cluster/license information, populate the timestamps, and send the metrics off to Tessen (similar to the message handler):
https://github.com/sensu/sensu-go/blob/9dfb8d29ace7ac94d4c89a62a09196a957855393/backend/tessend/tessend.go#L160-L184

PR #3033, opened 2019-06-06
Tessend metrics message bus

nikkiki:
## What is this change?

Adds the ability in TessenD to listen for metric points on the message bus, populate, and send them to the Tessen service.

Closes https://github.com/sensu/sensu-go/issues/2983

Yas.

Nah.

Nah.

Will be updated at completion of feature.

Integration test, logs, looking at chronograph. TestRail case will be added at completion of feature.


backend/tessend/tessend.go
```diff
@@ -200,20 +217,44 @@ func (t *Tessend) startMessageHandler() {
...
-			"id":                        data.Cluster.ID,
-			"opt-out":                   tessen.OptOut,
-			data.Metrics.Points[0].Name: data.Metrics.Points[0].Value,
-		}).Info("sending opt-out status event to tessen")
-		_ = t.send(data)
+		metrics, ok := msg.([]corev2.MetricPoint)
+		if ok {
+			if t.enabled() {
+				data := t.getDataPayload()
+				now := time.Now().UTC().Unix()
```
echlebek: Just so you're aware, this doesn't compute the offset to UTC from the local TZ - `time.Now().Unix()` delivers the same result as `time.Now().UTC().Unix()`.

https://golang.org/pkg/time/#Time.UTC

nikkiki: giphy

backend/tessend/tessend.go
```diff
@@ -200,20 +217,44 @@ func (t *Tessend) startMessageHandler() {
 		}
 
 		tessen, ok := msg.(*corev2.TessenConfig)
-		if !ok {
-			logger.WithField("msg", msg).Errorf("received non-TessenConfig on tessen message channel")
-			return
+		if ok {
+			data := t.getDataPayload()
+			t.getTessenConfigMetrics(time.Now().UTC().Unix(), tessen, data)
```
ccressent: Unix time has no timezone; I'm not 100% sure how `UTC()` and `Unix()` interact in the standard library. I'm assuming that `UTC()` has no effect in this case. To be certain, I would just drop `UTC()` altogether!

backend/tessend/tessend.go
ccressent: Ah well, that answers my comment a few lines above!

PR #3049, opened 2019-06-14
Add an `upsert` parameter to the `PutWrapped` mutation

wesleycyu:
## What is this change?

This change updates the `PutWrapped` mutation resolver in order to conditionally call the resource POST endpoint if the user of the mutation does not want the mutation to upsert the resource they are passing as the `raw` argument of the mutation. 

Previously, the mutation would always call the resource POST endpoint, which overwrites (upserts) the resource even if the resource already exists. Since users of the mutation don't always want this behaviour, we're updating the implementation of the `PutWrapped` mutation resolver to conditionally call the POST endpoint if the `upsert` parameter is `false`. By default, the `upsert` parameter is `true` in order to maintain the original behaviour of the mutation (for backwards compatibility).

To support this change, this PR also updates the mutations schema to reflect the addition of the `upsert` argument in the `PutWrapped` mutation.

This change supports a bug reported in Sensu Enterprise
> https://github.com/sensu/sensu-enterprise-go/issues/496

Yes, 

No.

Yes. We implemented a minor hack in order to correctly call the `POST` end point for a resource

If the `upsert` parameter on this mutation is `false`, we want to return an error if the resource already exists, instead of updating the existing resource. To do this, we make a request to the POST endpoint instead of the PUT endpoint. However, since the URI path resolved from `ret.Value.URIPath()` includes the name of the resource in the path string, we need to remove the name in order to call the correct endpoint. 

Ex. 
```
POST `/api/core/v2/namespaces/default/checks` 
PUT `/api/core/v2/namespaces/default/checks/check-cpu` 
```

Here we use `strings.TrimSuffix` to remove the resource name in order to call the correct endpoint. This works for the resource we currently create from the Web UI (Checks and Handlers), but it's possible that this _may_ not work for future resources.

No documentation is required.

Manually and with automated tests.

### DIFF
--- a/backend/apid/graphql/mutations.go
+++ b/backend/apid/graphql/mutations.go
@@ -3,6 +3,7 @@ package graphql
 import (
 	"encoding/json"
 	"errors"
+	"net/url"
 	"strings"
 	"time"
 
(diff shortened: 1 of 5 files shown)


backend/apid/graphql/mutations.go
```diff
@@ -33,20 +36,66 @@ type deleteRecordPayload struct {
...
+		// POST `/api/core/v2/namespaces/default/checks`
+		// PUT `/api/core/v2/namespaces/default/checks/check-cpu`
+		//
+		// Here we use Regex to replace the resource name with an empty
+		// string in order to call the correct endpoint.
+		// This works for the resource we currently create from the Web UI,
+		// but it's possible that this MAY not work for future resources.
+		var re *regexp.Regexp
+		exp := fmt.Sprintf(`\/%s$`, url.PathEscape(ret.Value.GetObjectMeta().Name))
+		re, err = regexp.Compile(exp)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

wesleycyu: thx golangcibot

backend/apid/graphql/mutations.go
```diff
@@ -33,20 +34,67 @@ type deleteRecordPayload struct {
...
+		// Ex.
+		// POST `/api/core/v2/namespaces/default/checks`
+		// PUT `/api/core/v2/namespaces/default/checks/check-cpu`
+		//
+		// Here we slice off the resource name in order to call the correct
+		// endpoint. This works for the resource we currently create from the
+		// Web UI (Checks and Handlers), but it's possible that this MAY not
+		// work for future resources.
+		suffix := "/" + url.PathEscape(ret.Value.GetObjectMeta().Name)
+		if strings.HasSuffix(path, suffix) {
```
golangcibot: S1017: should replace this if statement with an unconditional strings.TrimSuffix (from `gosimple`)

Issue #3039, opened 2019-06-11
Store a unique sensu cluster id


annaplotkin:
We need to ensure that cluster IDs are unique and accurately represented as such.

nikkiki:
Due to https://github.com/etcd-io/etcd/issues/10814, the ClusterID provided by etcd is not necessarily unique. This is problematic primarily for Tessen, where we rely on cluster id uniqueness to differentiate clusters.

**Proposed solution/workaround:**
Store a [randomly generated UUID](https://godoc.org/github.com/google/uuid#New), `SensuClusterID`, at `/sensu.io/cluster_id`. This value should be populated at start up IFF an entry does not already exist at that store (first write wins in a multi-node cluster). The `SensuClusterID` must be exposed by the API and by sensuctl, via new/existing endpoints and commands.

PR #3052, opened 2019-06-14
Add a sensu cluster id (uuid) and use this value in tessen cluster identification

nikkiki:

## What is this change?

Adds a store and GET `/api/core/v2/cluster/id` endpoint for the unique sensu cluster id (uuid). Updates TessenD to use this value rather than the etcd cluster id (which is not unique).

Closes https://github.com/sensu/sensu-go/issues/3039.

Yas.

Nah.

Yeah. I'm not crazy how I had to rewrite some logic in the store and API (even though it was simple). We have many generalizations based off the assumption that a struct satisfies the `Resource` interface. However, all resources are prone to be updated with `sensuctl create` / POST. The sensu cluster id cannot and should not be modified by the user, so for that reason, it is not a `Resource` and exists as a special snowflake. 

https://github.com/sensu/sensu-docs/issues/1548

Unit, integration, and E2E. Crucible tests will be updated with the new command to retrieve the ID.


palourde:
I'm a bit surprised to see the `id` command as a root-level command; the first thing I did was to run `sensuctl cluster id` without realizing it wasn't a sub-command.

Considering it's the _id_ of the cluster and I don't expect users to query that information often, what do you think of moving it to the `sensuctl cluster` command and `/cluster` endpoint? Or maybe there's a good reason for not doing that and I missed it!

nikkiki:
@palourde I'm worried that moving it to the `cluster` subcommands that users will confuse the etcd cluster id and the sensu cluster id, since all `sensuctl cluster` commands are specifically related to etcd cluster information. If you don't think that will be an issue, I can definitely move the subcommand there, let me know your thoughts!

ccressent:
I too think it makes more sense to use `sensuctl cluster id` and a `/cluster/id` endpoint; I'll let you resolve with whatever consensus was reach in Slack this morning.

Regarding `Resource` not being granular enough; maybe we should split the interface as follow:

``` go
interface CRUDResource {
    Creatable
    Readable
    Updatable
    Deletable
}

interface Creatable {
    Create(...)
}

interface Readable {
    Get/List(...)
}
...
```

I'm not sure it's very Go-y though?

backend/backend.go
```diff
@@ -135,6 +136,18 @@ func Initialize(config *Config) (*Backend, error) {
 	logger.Debug("Done initializing store")
 	b.Store = stor
 
+	_, err = stor.GetClusterID(b.ctx)
+	if err != nil {
+		switch err := err.(type) {
+		case *store.ErrNotFound:
+			if storeErr := stor.CreateClusterID(b.ctx, uuid.New().String()); storeErr != nil {
```
ccressent: A quick note to mention that `uuid.New()` can _panic_ if it's not able to read enough bytes from the source of randomness it ultimately uses (OS dependent). It's probably fine since it would panic early (during the backend initialization), but I thought it may be worth mentioning...

nikkiki: Definitely good to mention. I'm not sure that we can do anything to avoid the panic in this case, but during initialization is probably the best place to do so if there was an issue.

PR #3053, opened 2019-06-14
[GraphQL] Add ability to extend object types

jamesdphillips:
## What is this change?

Adds support for GraphQL object extensions. This will allow consumers / product
variants to extend upstream types with additional fields, interfaces and values.

```graphql
#
# file_a.graphql
#

type Query {
  widgets(namespace: String!): [Widget!]!
}



#
# file_b.graphql
#

# Note the @named directive is used to avoid conflicts when generating interfaces
# in the resulting code.
extend type Query @named(suffix: "Gizmos") {
  gizmos(namespace: String!): [Gizmo!]!
}

extend type Query @named(suffix: "FooBars") {
  foo(namespace: String!): Foo
  bars: [Bar!]!
}
```

```go
//
// service.go
//

schema.RegisterQueryExtensionGizmos(service, &gizmoFieldResolvers{})
schema.RegisterQueryExtensionFooBars(service, &fooBarFieldResolvers{})
```

Allow enterprise to be extended with additional types.

Probably.

No.

When I originally build the code generator- instead of simply using Go templates-- I chose to use `github.com/dave/jennifer/jen` for generating the interfaces. While jen is very powerful and allows us to do things that might be tricky in a template, it can also be a bit more difficult to grok the code what's going on. As a result took me a little bit of time to get re-acquainted with the package.

None required.

Updated the integration tests and manual testing in enterprise repo.


jamesdphillips:
I apologize as I realize having moved things around this must have been a hairy review. Thanks for pushing through.

Issue #3042, opened 2019-06-12
Silenced Entries Cache


palourde:
While performing benchmarks, we realized that [queries](https://github.com/sensu/sensu-go/blob/613f31c430c7d2b8d00de82e8815756f68f118c9/backend/eventd/silenced.go#L24) to determine if an event is silenced create a bottleneck in etcd.

We should implement a caching mechanism for silenced entries, similar to what we have for entities. I've wrote a [PoC](https://github.com/sensu/sensu-go/commit/9aef5954d332aeff44b683d58ce1ed0c2295d545) 1 month ago, but the entities cacher changed since then so it will require a bit more work.

echlebek:
In benchmarking, disabling silenced lookups resulted in a dramatic reduction in etcd reads, especially as events scaled. With silenced entries being looked up for each event, reads scaled linearly, resulting in thousands of reads under load. Without these lookups, etcd reads become a function of agent inventory size.

Eliminating this lookup can save a large amount of disk I/O, network traffic, and CPU usage.

Posted below are some screenshots from testing the enterprise edition of Sensu Go with a PostgresQL event store, introduced in order to achieve a higher event throughput. Removing silenced reads enabled us to achieve a much higher throughput overall. Reducing these reads should help the OSS/Free version's performance as well.

![Screenshot from 2019-06-11 14-01-36](https://user-images.githubusercontent.com/1633772/59366725-99633980-8cef-11e9-99c0-70a17f818aec.png)
_above, a screenshot showing linear growth in etcd reads with silenced entry lookups enabled_

![Screenshot from 2019-06-11 17-46-33](https://user-images.githubusercontent.com/1633772/59366738-9d8f5700-8cef-11e9-86fb-f0490b1e65ec.png)
_above, a screenshot showing a constant number of etcd reads with silenced entry lookups disabled_

PR #3064, opened 2019-06-18
Silenced entries cache

palourde:
## What is this change?

This PR introduces a generic caching framework for corev2.Resource and implements it for the entities and silenced entries caches.

Closes https://github.com/sensu/sensu-go/issues/3042

Added!

Nope!

Nope!

This PR does not change the UX.

I've adapted the existing unit tests, did plenty of local testing & deployed on sensu-staging and tested with automated tests.


backend/store/cache/cache.go
```diff
@@ -0,0 +1,258 @@
...
+	v := Value{Resource: resource}
+	if synthesize {
+		v.Synth = dynamic.Synthesize(resource)
+	}
+	return v
+}
+
+// MakeSliceCache ...
+func MakeSliceCache(resources []corev2.Resource, synthesize bool) []Value {
+	cache := make([]Value, len(resources))
```
ccressent: Should you replace `[]Value` with the new `resourceSlice` type?

echlebek: Originally, the `sliceCache` was introduced in order to be able to simply return a slice from the EntityCache. This was because the scheduler needed to access all entities, and it would have been inefficient to constantly create a slice from a map.

I don't know if, in the context of a generic cache, it's worth removing that optimization.

backend/schedulerd/check_watcher.go
```diff
@@ -50,10 +51,9 @@ func (c *CheckWatcher) startScheduler(check *types.CheckConfig) error {
 		if existing.Type() == GetSchedulerType(check) {
 			logger.Error("scheduler already exists")
 			return nil
-		} else {
```
nikkiki: Good catch

backend/schedulerd/cron_scheduler.go
```diff
@@ -5,12 +5,13 @@ import (
...
 	"github.com/sensu/sensu-go/backend/messaging"
 	"github.com/sensu/sensu-go/backend/store"
+	"github.com/sensu/sensu-go/backend/store/cache"
 	"github.com/sirupsen/logrus"
 
 	corev2 "github.com/sensu/sensu-go/api/core/v2"
 )
 
-// IntervalScheduler schedules checks to be executed on a cron schedule.
+// CronScheduler schedules checks to be executed on a cron schedule.
```
nikkiki: This is hilarious

backend/schedulerd/proxy_check.go
```diff
@@ -34,7 +35,7 @@ func matchEntities(entities []EntityCacheValue, proxyRequest *types.ProxyRequest
 
 	for i, result := range results {
 		if result {
-			matched = append(matched, entities[i].Entity)
+			matched = append(matched, entities[i].Resource.(*corev2.Entity))
```
nikkiki: I suppose `len(results) == len(entities)` is checked above, but this definitely threw me off using `entities[i]` while looping through `results`. Not sure if there is a more fluid approach, just thought I'd mention it.

backend/schedulerd/schedulerd.go
```diff
@@ -45,7 +49,8 @@ func New(c Config, opts ...Option) (*Schedulerd, error) {
 		ringPool:    c.RingPool,
 	}
 	s.ctx, s.cancel = context.WithCancel(context.Background())
-	cache, err := NewEntityCache(s.ctx, s.store)
+	// cache, err := NewEntityCache(s.ctx, s.store)
```
nikkiki:
```suggestion
```

backend/store/cache/cache.go
```diff
@@ -0,0 +1,258 @@
...
+		select {
+		case watcher.ch <- struct{}{}:
+		default:
+			// if there is already a notification in the buffer, don't send
+			// another.
+		}
+	}
+	newWatchers := make([]cacheWatcher, 0, len(r.watchers))
+	for i, w := range r.watchers {
+		if _, ok := deletes[i]; !ok {
```
nikkiki: Can you explain what `deletes` is tracking here. Unless I'm misunderstanding something, `deletes` is only populated with `struct{}{}` on L175.

backend/store/etcd/watchers.go
```diff
@@ -65,14 +66,14 @@ func (s *Store) GetCheckConfigWatcher(ctx context.Context) <-chan store.WatchEve
 	return ch
 }
 
-// GetEntityWatcher returns a channel that emits WatchEventEntity structs notifying
-// the caller that an Entity was updated. If the watcher runs into a terminal error
-// or the context passed is cancelled, then the channel will be closed.
+// GetTessenConfigWatcher returns a channel that emits WatchEventTessenConfig structs notifying
+// the caller that a TessenConfig was updated. If the watcher runs into a terminal error
+// or the context passed is cancelled, then the channel will be closed. The caller must
```
nikkiki:
```suggestion
// or the context passed is cancelled, then the channel will be closed.
```

backend/store/cache/cache.go
```diff
@@ -0,0 +1,259 @@
...
+	"github.com/coreos/etcd/clientv3"
+	corev2 "github.com/sensu/sensu-go/api/core/v2"
+	"github.com/sensu/sensu-go/backend/store"
+	"github.com/sensu/sensu-go/backend/store/etcd"
+	"github.com/sensu/sensu-go/types/dynamic"
+)
+
+// key is used to uniquely identify cache entries
+type key struct {
+	name      string
```
golangcibot: `name` is unused (from `structcheck`)

backend/store/cache/cache.go
```diff
@@ -0,0 +1,259 @@
...
+	namespace string
```
golangcibot: `namespace` is unused (from `structcheck`)

backend/store/cache/cache.go
```diff
@@ -0,0 +1,259 @@
...
+
+	"github.com/coreos/etcd/clientv3"
+	corev2 "github.com/sensu/sensu-go/api/core/v2"
+	"github.com/sensu/sensu-go/backend/store"
+	"github.com/sensu/sensu-go/backend/store/etcd"
+	"github.com/sensu/sensu-go/types/dynamic"
+)
+
+// key is used to uniquely identify cache entries
+type key struct {
```
golangcibot: `key` is unused (from `deadcode`)

backend/store/cache/cache.go
palourde: I believe Eric originally wrote that bit but my understanding is that we track the watchers in `r.watchers`.

Before notifying them, we make sure they were are still active here: https://github.com/sensu/sensu-go/blob/a1458e05a2df19d6e11befbf6a72aa2716055f93/backend/store/cache/cache.go#L168

If they were stopped, we add the specific watcher to the `deletes` slice: https://github.com/sensu/sensu-go/blob/a1458e05a2df19d6e11befbf6a72aa2716055f93/backend/store/cache/cache.go#L169

Which is later used here, to build a new list of active watchers: https://github.com/sensu/sensu-go/blob/a1458e05a2df19d6e11befbf6a72aa2716055f93/backend/store/cache/cache.go#L181

Makes sense?

backend/schedulerd/proxy_check.go
palourde: To be honest I'm not sure how we could do that in a more simplistic way without changing the signature of `js.MatchEntities` so it returns the matched entities instead, which I'd like to avoid!

PR #3066, opened 2019-06-18
API: Delete events associated with deleted entity

### LINKED ISSUE #2846: Remove events when corresponding entities are removed
## Expected Behavior
An entity is removed. Afterwards, no events associated with the entity exist.

## Current Behavior
An entity is removed. Afterwards, the events associated with the entity still exist.

## Issues
Right now, the event lifecycle is somewhat decoupled from that of entities. Changing how this works will probably need careful analysis.

Schedule a discussion to identify any issues around deleting events when entities are deleted. Look for corner cases that could create bugs.

echlebek:
## What is this change?

This commit fixes a bug where events were not deleted when their
associated entity was deleted via the API, the dashboard, or
sensuctl.

Closes #2846 

Yes

Tested manually

### DIFF
--- a/backend/apid/routers/entities.go
+++ b/backend/apid/routers/entities.go
@@ -3,22 +3,27 @@ package routers
 import (
 	"github.com/gorilla/mux"
 	corev2 "github.com/sensu/sensu-go/api/core/v2"
+	"github.com/sensu/sensu-go/backend/apid/actions"
 	"github.com/sensu/sensu-go/backend/apid/handlers"
 	"github.com/sensu/sensu-go/backend/store"
 )
 
 // EntitiesRouter handles requests for /entities
 type EntitiesRouter struct {
-	handlers handlers.Handlers
+	handlers   handlers.Handlers
+	store      store.Store
+	eventStore store.EventStore
 }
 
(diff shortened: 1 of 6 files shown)


backend/apid/routers/entities.go
```diff
@@ -3,22 +3,27 @@ package routers
...
 	"github.com/sensu/sensu-go/backend/apid/handlers"
 	"github.com/sensu/sensu-go/backend/store"
 )
 
 // EntitiesRouter handles requests for /entities
 type EntitiesRouter struct {
-	handlers handlers.Handlers
+	handlers   handlers.Handlers
+	store      store.Store
+	eventStore store.EventStore
```
palourde: Since `store.Store` adheres to the `store.EventStore` interface, could we just keep `store.Store` in the router and pass it as the EntityStore & EventStore below, instead of having to deal with both stores in the router?

echlebek: No, it needs to be its own thing due to pluggable event stores.

nikkiki: I was also curious about the 2 stores

PR #3085, opened 2019-06-24
[GraphQL] Queries for 'Event Filter' Resources

wesleycyu:
## What is this change?

This PR adds the ability to query for Event Filter resources from the GraphQL schema to support https://github.com/sensu/sensu-enterprise-go/issues/501. There are two ways users can retrieve handlers.

Users can retrieve a single Event Filter given a namespace and a name from the `eventFilter` field on the root query type,
Or, users can retrieve a collection of Event Filters (EventFiltersCollection) from the `eventFilters` field on the `namespace` type.

Screen Shot 2019-06-25 at 1 13 51 PM

This change supports a future update in the in the `sensu/web` repo (https://github.com/sensu/sensu-enterprise-go/issues/507) where Sensu users will have the ability to view a list of event filters and see the details of a particular selected event filter. This change allows the Web UI to query for event filters in order to display this data to the use

Yes,

To manually verify this change you can run a sensu-backend and query the GraphQL endpoint via GraphiQL.

1. Run a sensu-backend locally
2. Create a number of event filters. I did this using `sensuctl -create --file example_event_filters.json` 
3. Run the sensu/web application locally with the apollo dev tools Chrome extension installed. (Don't forge to update the graphQLSchemaRef in package.json to the latest commit on this PR `76e06c0`)
4. Run queries to ensure that handlers are being returned correctly.
example query:
```gql
query {
  eventFilter(namespace: "sensu-devel", name: "filter-production-only") {
    name
    namespace
    expressions
    runtimeAssets {
      name
    }
  }
  
  namespace(name: "sensu-devel") {
    eventFilters {
      nodes {
        name
      }
    }
  }
}
```


backend/apid/graphql/schema/event_filter.graphql
```diff
@@ -0,0 +1,60 @@
...
+  node: EventFilter
+  cursor: String!
+}
+
+"""
+Describes ways in which a list of mutators can be ordered.
+"""
+enum EventFilterListOrder {
+  NAME
+  NAME_DESC
```
jamesdphillips: I noticed in your ticket you had wanted to sort by action type as well, is this something you still want to do?

backend/apid/graphql/schema/event_filter.graphql
```diff
@@ -0,0 +1,60 @@
...
+  name: String!
+
+  "metadata contains name, namespace, labels and annotations of the record"
+  metadata: ObjectMeta!
+
+  "Command is the command to be executed."
+  action: EventFilterAction!
+
+  "Env is a list of environment variables to use with command execution"
+  expressions: [String!]
```
jamesdphillips: I think it's safe to say this won't ever return `null`. 

`expressions: [String!]!`

backend/apid/graphql/schema/event_filter.graphql
```diff
@@ -0,0 +1,60 @@
...
+
+  "RuntimeAssets are a list of assets required to execute the event filter"
+  runtimeAssets: [Asset!]
```
jamesdphillips: Same here.

backend/apid/graphql/event_filter.go
```diff
@@ -0,0 +1,40 @@
...
+
+var _ schema.EventFilterFieldResolvers = (*eventFilterImpl)(nil)
+
+//
+// Implement EventFilterFieldResolvers
+//
+
+type eventFilterImpl struct {
+	schema.EventFilterAliases
+}
```
jamesdphillips: As I recently did with mutators... don't forget the Global ID.

backend/apid/graphql/schema/event_filter.graphql
wesleycyu: Sorry -- that line in the ticket is a bit mis-leading
> The list should be sortable by name (asc/desc) and action (allow/deny)

The `action` attribute will be filter-able using the `namespace.eventFilters(filters: ["action:allow"]` query syntax. 

I don't _think_ this list needs to be sort-able against actions, right? I don't think that would be useful, would it?

backend/apid/graphql/event_filter.go
```diff
@@ -0,0 +1,46 @@
...
```
jamesdphillips: Can you double check if the the `type` field is working? it might conflict with the enum.

backend/apid/graphql/globalid/event_filters.go
```diff
@@ -0,0 +1,23 @@
+package globalid
+
+import "github.com/sensu/sensu-go/types"
+
+//
+// Event Filters
+//
+
+var eventFilterName = "eventFilters"
```
jamesdphillips: Eventually I think I want to do away with this package, however, until then I think it might be nice to have this match the resource name.. https://github.com/sensu/sensu-go/blob/master/api/core/v2/filter.go#L16

backend/apid/graphql/event_filter.go
```diff
@@ -0,0 +1,46 @@
...
+
+// ID implements response to request for 'id' field
+func (*eventFilterImpl) ID(p graphql.ResolveParams) (string, error) {
+	return globalid.EventFilterTranslator.EncodeToString(p.Source), nil
+}
+
+// IsTypeOf is used to determine if a given value is associated with the type
+func (*eventFilterImpl) IsTypeOf(s interface{}, p graphql.IsTypeOfParams) bool {
+	_, ok := s.(*types.EventFilter)
```
jamesdphillips: It would be preferable to use `*v2.EventFilter` instead. That way we can avoid adding another alias to the `types` package.

backend/apid/graphql/event_filter.go
wesleycyu: Yeah -- you mean the `action` field, and yes -- we'll need to write a custom resolver for this.

wesleycyu:
@jamesdphillips -- Updated! And, confirmed the two queries are working in the Apollo GraphiQL Client 🎉

Issue #3012, opened 2019-05-30
Cancelled contexts in watchers cause high CPU usage


palourde:
We discovered during 5.7.0 QA where upon removing 1000 proxy entities, the CPU usage would remain extremely high on at least one backend node. Here's the graph of the profile the reliability team took against a 5.7.0 cluster: [caviar_pprof.pdf](https://github.com/sensu/sensu-go/files/3237850/caviar_pprof.pdf)

From @ccressent:

> Basically, it boils down to this loop going haywire: https://github.com/sensu/sensu-go/blob/0d61cbddf963c2c62d04ccb56b23d9605ac4d770/backend/store/etcd/watcher.go#L83 Checking `ctx.Err()` very rapidly causes all this time spent locking/unlocking. There are things I want to try to finishing understanding what's going on, but like you, I've been unable to _reproduce_ the problem in the first place...

Here's some things we could try to reproduce the problem:

- [x] Try to reproduce this problem against 5.9.0
- [x] Try to reproduce this problem against 5.7.0
- [x] Try to reproduce this problem with a cluster first deployed with 5.6.0 and then upgraded to 5.7.0 to simulate the conditions against which QA was done.

palourde:
I was unable to reproduce this issue against a 5.9.0 cluster. Is it possible this issue was fixed in 5.8.0 or 5.9.0? Could it be related to the serialization changes we made in 5.7.0?

ccressent:
@echlebek made a good catch: the [limiter used in that function](https://github.com/sensu/sensu-go/blob/master/backend/store/etcd/watcher.go#L82) should be _inside_ the `for ctx.Err() == nil { ... }` loop! I'd still like to find a way to reproduce the issue so that we can "prove" that reworking the limiter fixes it.

palourde:
@ccressent I've deployed a 5.6.0 cluster, upgraded to 5.7.0, then created/destroyed 1000 proxy entities without any success 😓. The load always quickly come back to almost nothing.

ccressent:
Based on my testing and profiling, this issue is caused by a combination of 2
things: a failure of the `etcd` client to properly report that a watcher is
effectively dead, and our code assuming that it can always detect when a watcher
dies.


<a id="orgc7f99c1"></a>

### `etcd`'s watcher can fail to properly report their death

As reported in [this issue (still open)](https://github.com/etcd-io/etcd/issues/8980), watchers can fail to properly report
[...]

PR #3091, opened 2019-06-26
Rework watchers to avoid tight loops

ccressent:
## What is this change?

This small change is the result of many hours of fighting with etcd watchers in a staging environment.

In a nutshell, it alters 2 abstractions we've built on top of etcd watchers, `etcd.Watcher` (in `/backend/store`, yes the name is unfortunate) and `ringv2.Ring`, so that they can better detect when an etcd watcher dies.

We also don't `range` over those etcd watcher channels anymore; we `for { select { } }` instead, which allows us to avoid cases where we would end up ranging over a closed channel.

Mostly addresses #3012 and #2635.
~There are still improvements that could be made, see #3092.~
Closes https://github.com/sensu/sensu-go/issues/3092
Closes https://github.com/sensu/sensu-go/issues/3012

Yes, added.

Even after spending that much time with etcd watchers, I'm still not 100% sure of the correct, foolproof way to manage their life cycle.

Also, I tried to refactor `Ring.startWatchers()` and `Watcher.start()` so that they share a common abstraction for etcd watchers, but that didn't work out. I'm wondering if I'm missing subtleties between the 2 that justifies the 2 separate implementations?

- we started with no way to reproduce the problem

  That was unfortunate and that considerably slowed down progress at the beginning. Eventually found a way to reproduce by overloading the backend by creating a lot of proxy entities (took up to an hour to do) and then ultimately found a way to reproduce by simply disconnecting backends from each other.

- hard to iterate fast with staging
  
  I ended up writing shell scripts to help with the task; they might make a good addition to `sensu-staging` after a cleanup.

- couldn't refactor watcher/ring to use 1 implementation
  
  I just kept breaking everything and I'm wondering if there is actually a subtle reason why they each have their abstraction over etcd watchers.

- CPU usage can still be higher than normal 
  
  While a backend is disconnected from the cluster, its CPU usage still goes up, though not nearly has high as before, and it comes back down to normal once it reconnects. This is 100% coming from the goroutine spawned by `Watcher.start()`, but I've been unable
to address that without breaking `Watcher` altogether... #3092 tracks this specific case.

This is an internal change that requires no documentation addition or change.
[This comment I made during my investigation](https://github.com/sensu/sensu-go/issues/3012#issuecomment-501851111) may be helpful in the future.

- Pass all unit tests for watcher and ring
- Pass all integration tests for watcher and ring
- Pass automated QA tests in a staging environment
- Many hours overloading or disconnecting cluster members and observing/profiling their CPU usage

### DIFF
--- a/backend/store/cache/cache.go
+++ b/backend/store/cache/cache.go
@@ -195,7 +195,6 @@ func (r *Resource) start(ctx context.Context) {
 		case <-ctx.Done():
 			return
 		case event := <-r.watcher:
-			fmt.Println("received event")
 			r.updates = append(r.updates, event)
 		case <-ticker.C:
 			if len(r.updates) > 0 {
--- a/backend/store/cache/cache.go
+++ b/backend/store/cache/cache.go
@@ -195,6 +195,7 @@ func (r *Resource) start(ctx context.Context) {
 		case <-ctx.Done():
 			return
 		case event := <-r.watcher:
+			fmt.Println("received event")
 			r.updates = append(r.updates, event)
 		case <-ticker.C:
 			if len(r.updates) > 0 {
--- a/backend/store/cache/cache_integration_test.go
+++ b/backend/store/cache/cache_integration_test.go
@@ -94,7 +94,6 @@ func TestEntityCacheIntegration(t *testing.T) {
 
 	newEntity := corev2.FixtureEntity("new")
 	newEntity.EntityClass = corev2.EntityProxyClass
-	fmt.Println("update...")
 	if err := store.UpdateEntity(ctx, newEntity); err != nil {
 		t.Fatal(err)
 	}
--- a/backend/store/cache/cache_integration_test.go
+++ b/backend/store/cache/cache_integration_test.go
@@ -97,7 +97,6 @@ func TestEntityCacheIntegration(t *testing.T) {
 	if err := store.UpdateEntity(ctx, newEntity); err != nil {
 		t.Fatal(err)
 	}
-	fmt.Println("updated")
 	<-watcher
 
 	got := cache.Get("default")
--- a/backend/store/cache/cache_integration_test.go
+++ b/backend/store/cache/cache_integration_test.go
@@ -94,10 +94,11 @@ func TestEntityCacheIntegration(t *testing.T) {
 
 	newEntity := corev2.FixtureEntity("new")
 	newEntity.EntityClass = corev2.EntityProxyClass
+	fmt.Println("update...")
 	if err := store.UpdateEntity(ctx, newEntity); err != nil {
 		t.Fatal(err)
 	}
-
+	fmt.Println("updated")
 	<-watcher
 
 	got := cache.Get("default")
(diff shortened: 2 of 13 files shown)


backend/store/etcd/watcher.go
```diff
@@ -87,35 +86,30 @@ func (w *Watcher) start(ctx context.Context) {
 		defer close(w.resultChan)
 		_ = limiter.Wait(ctx)
 		for ctx.Err() == nil {
-			for watchResponse := range watcherChan {
+			select {
+			case watchResponse, closed := <-watcherChan:
```
echlebek: It's unclear to me why this variable is named `closed` - a true value here indicates that the value was received from an open channel.

backend/store/etcd/watcher.go
```diff
@@ -87,35 +86,30 @@ func (w *Watcher) start(ctx context.Context) {
...
-					}
 					logger.WithError(err).Info("error from watch response")
 					w.resultChan <- store.WatchEvent{
 						Type: store.WatchError,
 						Err:  err,
 					}
-					if watchResponse.Canceled {
-						// Reinstate the watcher and break to the outer loop
+
+					if watchResponse.Canceled || closed {
```
echlebek: My reading of this is that this will result in the watcher being created for every watch event.

backend/store/etcd/watcher.go
ccressent: It shouldn't be named closed, you're 100% correct. It should be named `ok`.

backend/store/etcd/watcher.go
ccressent: This branch is inside the `if err := watchResponse.Err(); err != nil { }` branch; so the watcher is reinstated only if that `err != nil`.

Phrased another way, and assuming I rename that `closed` to `ok`, it means: if _any_ error was reported (`watchResponse.Err() != nil`) AND if the watch failed (`watchResponse.Canceled == true` according to etcd docs) OR if the channel is still open (`ok == true`) THEN just recreate the watcher.

Since the watcher can fail to close its channel or set `Canceled` to true, that's the only thing I tried that ended up working to not spin indefinitely...

I have a hard time explaining it even after spending that much time on it. Not a good sign...

echlebek: I see! This is indeed unintuitive. Thanks.

palourde:
@ccressent I've merged and pushed my branch to yours, I believe this is ready to be 🚢

backend/store/cache/cache.go
```diff
@@ -195,6 +195,7 @@ func (r *Resource) start(ctx context.Context) {
 		case <-ctx.Done():
 			return
 		case event := <-r.watcher:
+			fmt.Println("received event")
```
echlebek: Should kill this line

backend/store/cache/cache_integration_test.go
```diff
@@ -94,10 +94,11 @@ func TestEntityCacheIntegration(t *testing.T) {
 
 	newEntity := corev2.FixtureEntity("new")
 	newEntity.EntityClass = corev2.EntityProxyClass
+	fmt.Println("update...")
 	if err := store.UpdateEntity(ctx, newEntity); err != nil {
 		t.Fatal(err)
 	}
-
+	fmt.Println("updated")
```
echlebek: And this

Issue #2811, opened 2019-03-18
Should Sensu-agent be able to register an entity in any namespace?


funkypenguin:
I'm using RBAC to build a multi-tenanted Sensu platform. I create a per-department namespace, create users, assign them to groups, and apply the cluster-role "admin" to the group with a role-binding in their namespace. The intended result is that the users in each department are isolated from named resources in any other department.

## Expected Behavior
I misconfigure agent.yml, and allow a user in my "preprod" department (who's inherited cluster-role admin with a role binding to the "preprod" namespace) to register an entity in the "default" namespace. The entity is not permitted to be registered to the default namespace, since the username used by the agent has no permissions to "default" namespace (or any other besides their own).

## Current Behavior
Provided I specify a valid user/pass in agent.yml, I'm able to register entities into any namespace I please. This presents a problem because having misconfigured my agent, I can't later _delete_ the entity from the incorrect namespace, since my user doesn't have permission to even _list_ entities in that namespace.

## Possible Solution
It feels as if this is to do with the fact that on receiving a Keepalive from an agent, the API will register an entity for that agent if one is not already seen. I wonder whether the user/permissions check is not applied at this point, allowing the entity to be erroneously registered.

## Steps to Reproduce (for bugs)
1. Create namespace "test-rbac-issue" (```sensuctl namespace create test-rbac-issue```)
2. Create user "test-rbac-issue" (```sensuctl user create test-rbac-issue -p "test-rbac-issue```
3. Add user to group (```sensuctl user add-group test-rbac-issue group-test-rbac-issue```)
4. Add role-binding (in namespace) to group (```sensuctl role-binding create role-test-rbac-issue-admin --cluster-role admin -g group-test-rbac-issue --namespace test-rbac-issue```
5. Attempt to setup a new agent, leaving "namespace" value in agent.yml comment out (per default), but populate the user/pass correctly. When you start the agent, the entity is registered in the "default" namespace.
6. Create _another_ namespace (to prove it's not specific to default) (```sensuctl namespace create super-secret```)
7. Change the namespace in agent.yml to "super-secret", and restart sensu-agent
8. The entity is now registered in the "super-secret" namespace.

## Context
The risk here is that a misconfigured agent would pollute other namespaces, and the operators responsible for that agent would be unable to remove the entity from the erroneous namespaces

## Your Environment
Running sensu/sensu Docker container under Kubernetes on GKE

Server:
```
/ # sensu-backend version
[...]

PR #3094, opened 2019-06-26
Bugfix/websocket authorization

nikkiki:
## What is this change?

Adds basic authorization middleware to the agent web socket connection.

Closes https://github.com/sensu/sensu-go/issues/2811

Yas.

Nah.

Yes, needed to pair with @palourde and @ccressent for a bit, but they were able to clarify the missing piece for me! (it was injecting the namespace into the context)

It's a bugfix, docs should be good.

Integration and e2e testing. New test case is probably not required.

### DIFF
--- a/backend/agentd/agentd.go
+++ b/backend/agentd/agentd.go
@@ -43,7 +43,7 @@ type Agentd struct {
 	wg         *sync.WaitGroup
 	errChan    chan error
 	httpServer *http.Server
-	store      Store
+	store      store.Store
 	bus        messaging.MessageBus
 	tls        *types.TLSOptions
 	ringPool   *ringv2.Pool
@@ -83,7 +83,7 @@ func New(c Config, opts ...Option) (*Agentd, error) {
 		return nil, err
 	}
 
-	handler := middlewares.BasicAuthentication(http.HandlerFunc(a.webSocketHandler), a.store)
+	handler := middlewares.BasicAuthorization(middlewares.BasicAuthentication(http.HandlerFunc(a.webSocketHandler), a.store), a.store)
 	a.httpServer = &http.Server{
 		Addr:         fmt.Sprintf("%s:%d", a.Host, a.Port),
 		Handler:      handler,
--- a/backend/agentd/agentd.go
+++ b/backend/agentd/agentd.go
@@ -24,12 +24,6 @@ var (
 	upgrader = &websocket.Upgrader{}
 )
 
-// Store specifies storage requirements for Agentd.
-type Store interface {
-	middlewares.AuthStore
-	SessionStore
-}
-
 // Agentd is the backend HTTP API.
 type Agentd struct {
 	// Host is the hostname Agentd is running on.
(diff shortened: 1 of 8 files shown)


backend/agentd/agentd.go
```diff
@@ -43,7 +43,7 @@ type Agentd struct {
 	wg         *sync.WaitGroup
 	errChan    chan error
 	httpServer *http.Server
-	store      Store
+	store      store.Store
```
palourde: Can we remove the Store interface at line 28?

backend/apid/middlewares/authorization.go
```diff
@@ -1,10 +1,15 @@
...
 import (
+	"context"
 	"net/http"
 
 	"github.com/sensu/sensu-go/backend/apid/actions"
 	"github.com/sensu/sensu-go/backend/authorization"
+	"github.com/sensu/sensu-go/backend/authorization/rbac"
+	"github.com/sensu/sensu-go/backend/store"
+	"github.com/sensu/sensu-go/transport"
+	corev2 "github.com/sensu/sensu-go/types"
```
palourde: Could we replace `github.com/sensu/sensu-go/types` with `github.com/sensu/sensu-go/api/core/v2`?

backend/apid/middlewares/authorization.go
```diff
@@ -44,3 +49,41 @@ func (a Authorization) Then(next http.Handler) http.Handler {
 		next.ServeHTTP(w, r.WithContext(ctx))
 	})
 }
+
+// BasicAuthorization performs basic authorization for entity creation via the agent websocket.
```
palourde: s/entity/event/

backend/apid/middlewares/authorization.go
```diff
@@ -44,3 +49,41 @@ func (a Authorization) Then(next http.Handler) http.Handler {
...
+func BasicAuthorization(next http.Handler, store store.Store) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		namespace := r.Header.Get(transport.HeaderKeyNamespace)
+		ctx := r.Context()
+		ctx = context.WithValue(ctx, corev2.NamespaceKey, namespace)
+
+		user, err := store.GetUser(ctx, r.Header.Get(transport.HeaderKeyUser))
```
palourde: I don't believe we should do this, considering the agent user could be within an authentication provider. Instead, we should rely on the claims injected into the context by [`middlewares.BasicAuthentication`](https://github.com/sensu/sensu-go/blob/c80ddead34075326a45ae9d3f878304c6c43ce53/backend/apid/middlewares/authentication.go#L84) and fetch those like this: https://github.com/sensu/sensu-go/blob/a82a17edaf9f7455c5d86cc2a63bbe5bfc152835/backend/apid/middlewares/authorization_attributes.go#L120

nikkiki: Good point! Fixing this also made me realize I had the authorization and authentication middlewares in the wrong order. 2 🐦1 stone!

Issue #3098, opened 2019-06-27
Configurable buffers & workers


palourde:
We recently increased the default number of workers that handle incoming events & keepalives to `1000`, after realizing they constituted our main bottleneck. The number was chosen arbitrarily.

We also realized that the buffered channels between agentd, eventd & pipelined were only able to buffer a couple of messages, which can provoke some backpressure on the agent side.

Since not all installations are the same, tweaking these numbers can be useful in small or large clusters.

PR #3100, opened 2019-06-27
Configurable buffer size & worker count

palourde:
## What is this change?

This PR introduces configurable buffer size & worker count for keepalived, eventd & pipelined.

More specifically, it allows an operator to configure the `sensu:event-raw`, `sensu:event` & `sensu:keepalive` topics of Wizard Bus in order to determine the buffer size and the number of workers that consume these topics.

Closes https://github.com/sensu/sensu-go/issues/3098

Added!

Nope!

Nope!

https://github.com/sensu/sensu-docs/issues/1575

I manually tested this change and we successfully used this change on the performance testing environment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ is silenced.
 - Added --disable-assets flag to sensu-agent.
 - Added ability to query mutators to the GraphQL service
 - Added ability to query event filters to the GraphQL service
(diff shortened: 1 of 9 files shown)


palourde:
@echlebek do you have any suggestions for sensible default values?

echlebek:
@palourde I'm thinking a buffer size of 100, and a goroutine pool size of 10 would be appropriate. WDYT?

Issue #3097, opened 2019-06-27
Validate that an entity exists before deleting its events


palourde:
When deleting an entity, the API no longer returns an error (I'd assume 404) if the entity does not exist.

We should check if the entity does exist first, then delete its events and then finally the entity itself.

**Edit** Maybe we should delete events first, whether or not the entity exist, and then check if it does exist.

PR #3103, opened 2019-06-28
Return an error when deleting an entity that does not exist

palourde:

## What is this change?

It fixes the behaviour of the DELETE endpoint for entities, which would not return an error if the entity didn't exist since 5.10.1.

Closes https://github.com/sensu/sensu-go/issues/3097

Added!

Nope!

Nope!

Nope!

Manually tested, we have a failing test case in sensu-staging that will be fixed with this patch.


ccressent:
Nice, thanks for the quick turnaround!

Issue #3106, opened 2019-07-01
sensu-agent disconnected from master and never attempted to reconnect


roganartu:
Our cluster got into a strange state, where the master removed an agent from the pool because it disconnected, but the agent didn't seem to notice the disconnect and hence never attempted to reconnect.

## Expected Behavior
A disconnected agent should continuously try to reconnect, with some kind of capped exponential backoff.

## Current Behavior
It is possible for an agent to be disconnected from the master and never attempt to reconnect.

## Steps to Reproduce (for bugs)
Unsure how exactly to reproduce, though given the logs I saw perhaps manually breaking the socket between the agent and the master would successfully reproduce.

## Context
Backend logs at time of disconnect (debug level):
```
╰ sudo journalctl -u sensu_backend --since 12:58 | rg "agentd"          
Jul 01 12:58:28 shanetmon1 sensu-backend[22273]: {"addr":"10.16.80.141:36860","component":"agentd","id":"shanetmon1-agent","level":"warning","msg":"stopping session","recv error":"Connection error: read tcp 10.16.80.141:8081-\u003e10.16.80.141:36860: read: connection reset by peer","time":"2019-07-01T12:58:28-04:00"}
Jul 01 12:58:28 shanetmon1 sensu-backend[22273]: {"component":"agentd","level":"info","msg":"session disconnected - stopping recvPump","time":"2019-07-01T12:58:28-04:00"}
Jul 01 12:58:28 shanetmon1 sensu-backend[22273]: {"component":"agentd","level":"info","msg":"shutting down - stopping subPump","time":"2019-07-01T12:58:28-04:00"}
Jul 01 12:58:28 shanetmon1 sensu-backend[22273]: {"component":"agentd","level":"info","msg":"shutting down - stopping sendPump","time":"2019-07-01T12:58:28-04:00"}
Jul 01 12:58:28 shanetmon1 sensu-backend[22273]: {"component":"agentd","level":"info","msg":"removing agent from ring","time":"2019-07-01T12:58:28-04:00"}
Jul 01 12:58:28 shanetmon1 sensu-backend[22273]: {"component":"agentd","level":"info","msg":"removing agent from ring","time":"2019-07-01T12:58:28-04:00"}
Jul 01 12:58:28 shanetmon1 sensu-backend[22273]: {"component":"agentd","level":"info","msg":"removing agent from ring","time":"2019-07-01T12:58:28-04:00"}
[...]

PR #3111, opened 2019-07-03
Properly identify the agent in agentd session

palourde:

## What is this change?

It makes sure we identify the agent in log entries of a session, using its ID and its namespace.

Related to https://github.com/sensu/sensu-go/issues/3106

I don't believe it's worth mentioning it.

Nope

Nope

Nope

Manually tested.

### DIFF
--- a/backend/agentd/session.go
+++ b/backend/agentd/session.go
@@ -81,7 +81,7 @@ func NewSession(cfg SessionConfig, conn transport.Transport, bus messaging.Messa
 	logger.WithFields(logrus.Fields{
 		"addr":          cfg.AgentAddr,
 		"namespace":     cfg.Namespace,
-		"id":            cfg.AgentName,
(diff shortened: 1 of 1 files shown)


backend/agentd/session.go
```diff
@@ -261,7 +262,10 @@ func (s *Session) Stop() {
 	close(s.checkChannel)
 	for _, sub := range s.cfg.Subscriptions {
 		ring := s.ringPool.Get(ringv2.Path(s.cfg.Namespace, sub))
-		logger.Info("removing agent from ring")
+		logger.WithFields(logrus.Fields{
+			"namespace": s.cfg.Namespace,
+			"id":        s.cfg.AgentName,
```
echlebek: Maybe use the key "agent" instead of "id", just to remove any ambiguity about what the name is?

PR #3120, opened 2019-07-06
Fix a bug in round robin check scheduling

### LINKED ISSUE #3119: Round robin proxy check schedulers restart too often
Both cron and interval round robin proxy check schedulers frequently shut down and restart whenever entities are updated. Since entities are updated whenever keepalives occur, this happens way too often.

This can cause check executions to misfire, leading to round robin checks not executing frequently enough.

The check schedulers should only restart when a material change to entities occurs, and the cardinality of the agents that are required to service the request changes.

echlebek:
This commit fixes a bug where check schedulers can be shut
down and recreated too often. In some cases this can lead to checks
not being executed frequently enough.

The problem is caused by entity updates being treated as cause for
shutting down schedulers, when in fact, it is only necessary to do
so in the rare circumstance that entity attributes change, leading
to a different cardinality of round robin ring.

To avoid this happening, before cancelling the existing scheduler,
the cardinality of the request is inspected, and if unchanged, the
existing scheduler is left alone.

Closes #3119 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ not exist.
 were not getting properly populated.
 - Fixed a bug where multiple nested set handlers could be incorrectly flagged as
 deeply nested.
+- Fixed a bug where round robin proxy checks could fail to execute.
 
 ## [5.10.1] - 2019-06-25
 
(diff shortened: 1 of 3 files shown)


echlebek:
We could expose the problem in our QA suite by shortening the keepalive interval, but otherwise, I think it would be quite difficult to write a test for this.

Issue #1507, opened 2018-05-11
Agent Check command allow list


portertech:
## Background

A Sensu Backend can instruct an Agent to run Checks, which execute commands on the Agent's host. In Sensu 1.x, the Sensu Client has a "safe_mode" feature, which stops the Client from executing Checks when it does not have local configuration for them (including the commands to execute). As all configuration in Sensu 2.x is done centrally via the Backend API (w/ RBAC), a "safe_mode" feature has not yet been implemented. This functionality is desired by some organizations, e.g. financial.

## Feature Request

Support optional Agent configuration to specify a list of allowed Check commands. If configured, the Agent compares received Check commands with its allow list. If a Check command is not in the allow list, the Agent does not execute the command and responds with an Event declaring this fact (with Check status of 3).

**Spec/Implementation details:** [Agent 'Allow' List spec](https://docs.google.com/document/d/1ZaOhpG0J2SHhX-MZmDgRrFDWFME3FV6LTScv_10ssYE)

portertech:
@jspaleta we can probably start w/ full command string matching.

echlebek:
I see the motivations for using the command or the command executable name in an allow-list, but I have to disagree about implementing the feature this way.

*Alternate proposal*:
Allow a list of checks by name, with optional Etag.

*Rationale*
The check name is the only unique identifier for a check, and it is ultimately the only key we want to provide in check management. With the addition of the Etag, the assertion becomes even stronger, verifying that the check is the same as it was at the time it was added to the allow list.

jspaleta:
i think check name plus etag covers the underlying issue.  if someone changes the check definition (either by accident or malicious intent) the etag prevents running the check until agent undates the etag.

Could this also spawn a sensu event if a list check is out of check-etag compliance? 
Maybe make it possible to define a handler for check-etag compliance violation?  Similar to how statsd metric event has a defined handler without being a check result.

portertech:
@echlebek @jspaleta etag would need to be consistent across installations. I should be able to rebuild an installation and my allow list should still be correct. Using only the command was intentional, allowing me to change handlers etc w/o having to update the 10k+ allow lists 😬

echlebek:
@portertech consistent etags are achieved easily enough with hashing, but maybe I'm missing something?

Due to the existence of check environment variables, the check command alone is probably not enough to achieve safety. (Suppose `PATH=/path/to/evil` or any other number of tricks you can perform with environment variables)

[...]

PR #3130, opened 2019-07-10
Implement agent allow list for checks and hooks

nikkiki:

## What is this change?

Adds functionality for the agent `--allow-list` configuration, which whitelists check and check hook executables in a yaml/json file.

Closes https://github.com/sensu/sensu-go/issues/1507

Yas.

Nah.

Matching the SHA in the appropriate env was difficult. See `helpers.go` and `LEGAL`.

https://github.com/sensu/sensu-docs/issues/1608

100% unit test coverage.

example `allow-list.yaml`:
```
- exec: /Users/nikkiki/go/src/github.com/sensu/sensu-go/hello.sh
  args:
  - ""
  sha512: 736ac120323772543fd3a08ee54afdd54d214e58c280707b63ce652424313ef9084ca5b247d226aa09be8f831034ff4991bfb95553291c8b3dc32cad034b4706
  enable_env: true
  foo: bar
- exec: /Users/nikkiki/go/src/github.com/sensu/sensu-go/hook.sh
  args:
  - ""
  sha512: 28d61f303136b16d20742268a896bde194cc99342e02cdffc1c2186f81c5adc53f8550635156bebeed7d87a0c19a7d4b7a690f1a337cc4737e240b62b827f78a
- exec: echo-asset.sh
  args:
  - "foo"
  sha512: cce3d16e5881ba829f271df778f9014f7c3659917f7acfd7a60a91bfcabb472eea72f9781194d310388ba046c21790364ad0308a5a897cde50022195ba90924b
```
example `checks.yaml`:
_see `allow_list_on_deny_status` for configurable status_
```
type: CheckConfig
api_version: core/v2
metadata:
  name: check-allow-list
  namespace: default
spec:
  check_hooks:
  - "0":
    - hook-allow
  command: /Users/nikkiki/go/src/github.com/sensu/sensu-go/hello.sh
  env_vars: null
  handlers: []
  high_flap_threshold: 0
  interval: 10
  low_flap_threshold: 0
  output_metric_format: ""
  output_metric_handlers: null
  proxy_entity_name: ""
  publish: true
  round_robin: false
  runtime_assets: null
  stdin: false
  subdue: null
  subscriptions:
  - allow
  timeout: 0
  ttl: 0
---
type: CheckConfig
api_version: core/v2
metadata:
  annotations:
    allow_list_on_deny_status: "5"
  name: check-deny-list
  namespace: default
spec:
  check_hooks: null
  command: echo deny
  env_vars: null
  handlers: []
  high_flap_threshold: 0
  interval: 10
  low_flap_threshold: 0
  output_metric_format: ""
  output_metric_handlers: null
  proxy_entity_name: ""
  publish: true
  round_robin: false
  runtime_assets: null
  stdin: false
  subdue: null
  subscriptions:
  - allow
  timeout: 0
  ttl: 0
---
type: CheckConfig
api_version: core/v2
metadata:
  name: check-with-asset
  namespace: default
spec:
  check_hooks: null
  command: echo-asset.sh
  env_vars: null
  handlers: []
  high_flap_threshold: 0
  interval: 10
  low_flap_threshold: 0
  output_metric_format: ""
  output_metric_handlers: null
  proxy_entity_name: ""
  publish: true
  round_robin: false
  runtime_assets:
  - staging-asset
  stdin: false
  subdue: null
  subscriptions:
  - misc
  timeout: 0
  ttl: 0
```
expected output:
```
    Entity          Check                                   Output                              Status   Silenced             Timestamp            
 ──────────── ────────────────── ───────────────────────────────────────────────────────────── ──────── ────────── ─────────────────────────────── 
  Nikkis-MBP   check-allow-list   hi                                                                 0   false      2019-07-10 13:26:55 -0700 PDT  
                                                                                                                                                   
  Nikkis-MBP   check-deny-list    Check command denied by the Agent Allow list                       5   false      2019-07-10 13:27:23 -0700 PDT  
  Nikkis-MBP   check-with-asset   you have installed an asset, kudos!                                0   false      2019-07-10 13:27:23 -0700 PDT 
  Nikkis-MBP   check-deny-this    Check command denied by the Agent Allow list                       3   false      2019-07-10 13:27:23 -0700 PDT 
```


agent/check_handler.go
```diff
@@ -19,9 +20,12 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+const allowListOnDenyStatus = "allow_list_on_deny_status"
+const allowListOnDenyOutput = "Check command denied by the Agent Allow list"
```
palourde: Any reason why `Agent Allow` are capitalized but not list? Also, error strings should usually not be capitalized (https://github.com/golang/go/wiki/CodeReviewComments#error-strings).

nikkiki: I literally just copied and pasted the string from the spec 😆I've fixed this in the check and hook error messages!

agent/allow_list.go
```diff
@@ -0,0 +1,81 @@
...
+	Sha512    string   `yaml:"sha512" json:"sha512"`
+	EnableEnv bool     `yaml:"enable_env" json:"enable_env"`
+}
+
+func readAllowList(path string, readBytes func(string) ([]byte, error)) ([]allowList, error) {
+	var allowList []allowList
+	if path == "" {
+		return allowList, nil
+	}
+	unmarshalFuncs := map[string]func(in []byte, out interface{}) error{
```
echlebek: Can we just use yaml unmarshaling in all cases? JSON is a subset of YAML. Then we wouldn't need to care about the extension.

agent/allow_list.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	return nil
+}
+
+func (a *Agent) matchAllowList(command string) (allowList, bool) {
+	for _, al := range a.allowList {
+		remaining := command
+		if strings.Contains(command, al.Exec) {
+			remaining = strings.Trim(remaining, al.Exec)
```
echlebek: This is probably not what you want, and could lead to incorrect results. Unlike TrimPrefix and TrimSuffix, Trim uses a set of runes to be trimmed, not a whole string.

See the example here to see what I'm talking about: https://golang.org/pkg/strings/#Trim

agent/allow_list.go
nikkiki: yaml.Unmarshal works on JSON files?

agent/allow_list.go
nikkiki: The intention is to trim leading and trailing whitespace.

echlebek: Trim doesn't trim whitespace, that's https://golang.org/pkg/strings/#TrimSpace

nikkiki: Oops I misread this. The intention is to remove the substring `al.Exec` from the string `remaining`.

nikkiki: After reading through the example I see what you mean!

echlebek: I mean, obviously Trim will trim whitespace if you give it whitespace to trim, but it will trim the entire string of whitespace, not only trailing and leading :)

agent/allow_list.go
nikkiki: When using `yaml.Unmarshal` in all cases, I'm getting some errors: `yaml: line 2: found character that cannot start any token`. I believe this is due to the use of tabs in JSON rather than spaces. I think we may still want to support tabs, wdyt?

echlebek: I see. Well don't worry about it then :)

Issue #2566, opened 2018-12-20
Don't use an internal reverse proxy for dashboardd


echlebek:
The dashboardd HTTP server is reverse-proxying its requests, to the URL the apid HTTP server is serving on. The design generally works, but has some issues, including:

* Roundtripping, in the internal application, means that something which should be a function call is hitting the network stack. This has adverse performance effects, due to an amplification of system calls.
* Proxy semantics can yield 502 errors, obscuring other errors.
* Requires binding to localhost. Operators may wish to disallow services on localhost, and this would prevent them from using the product in the way they would like.
* Proxying TLS to localhost is more work than it is worth. We skip hostname verification for TLS requests.

We should be able to re-use the apid http.Handler in dashboardd, instead of using a reverse proxy.

Create a top level mux that includes the apid routes, and use that to serve requests with dashboardd. Serve non-api requests with the existing handlers.

echlebek:
Proposal doc: https://docs.google.com/document/d/1wdfAhajJsXCEmC0vquifkbdP5O33HSPH9PbjFLC46b4/edit#

PR #3133, opened 2019-07-11
Refactor authentication into its own API

echlebek:
## What is this change?

This change extracts business logic from the authentication router and places it in its own type. The intent is that this type will be reused by dashboardd, so that it doesn't need to reverse proxy requests. This is the first of many such changes.

These changes are in support of #2566

No

### DIFF
--- a/backend/apid/routers/authentication.go
+++ b/backend/apid/routers/authentication.go
@@ -2,18 +2,14 @@ package routers
 
 import (
 	"encoding/json"
-	"errors"
-	"fmt"
 	"net/http"
 
-	"github.com/sensu/sensu-go/api/core/v2"
-
 	"github.com/gorilla/mux"
+	"github.com/sensu/sensu-go/backend/api"
 	"github.com/sensu/sensu-go/backend/authentication"
-	"github.com/sensu/sensu-go/backend/authentication/jwt"
-	"github.com/sensu/sensu-go/backend/authentication/providers/basic"
 	"github.com/sensu/sensu-go/backend/store"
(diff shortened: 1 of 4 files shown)


backend/dashboardd/dashboardd.go
```diff
@@ -133,6 +138,18 @@ func (d *Dashboardd) Name() string {
 	return "dashboardd"
 }
 
+func registerGraphQLRouter(parent *mux.Router, url string, store store.Store, tls *tls.Config) {
```
golangcibot: `registerGraphQLRouter` is unused (from `deadcode`)

backend/apid/routers/authentication.go
```diff
@@ -131,138 +84,42 @@ func (a *AuthenticationRouter) test(w http.ResponseWriter, r *http.Request) {
...
-	// Remove the access & refresh tokens from the access list
-	if err := a.store.RevokeTokens(accessClaims, refreshClaims); err != nil {
-		logger.WithError(err).WithField("user", refreshClaims.Subject).Errorf(
-			"could not revoke tokens IDs %q & %q", accessClaims.Id, refreshClaims.Id,
-		)
-		http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
-		return
-	}
+	http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
+	return
```
golangcibot: S1023: redundant `return` statement (from `gosimple`)

Issue #929, opened 2018-01-24
Hook assets


grepory:
As a user, I want to be able to use assets for check hooks.

craigpfeiffer:
as a work around to check hooks not having direct asset configuration i did some testing to see if a check hook has the same $PATH as the check that is triggering the hook.  unfortunately it does not appear as a viable solution either.

created a check with an asset and attempted to run the script within the asset in the hook and it failed with a command not found error.  also updated the hook to output the $PATH as well as the check to output the $PATH to compare them.

check $PATH:
`output": "/var/cache/sensu/sensu-agent/eca51a0f4c9ad1d520d24025b0579dea9a56899a668475f8a0f4c319b40e2578278cd3109d2846a1ce1dccd09f27a1fd7c08ef7eef5897d1741ab070099d5446/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin\n",`

check hook $PATH:
`"output": "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin\n",`

annaplotkin:
This has come up as a higher priority due to a use case from teams at NCR. 

In the event that a service encounters an issue, the teams tail the logs. Sometimes the logs are quite large and it would be beneficial for them to be able to leverage hooks for powershell assets to achieve this rather than writing dubious checks. Raising higher in the Backlog.

PR #3134, opened 2019-07-11
Add assets to check hooks

nikkiki:

## What is this change?

Adds the `runtime_assets` field to `HookConfig`. Adds assets to check hook execution.

Closes https://github.com/sensu/sensu-go/issues/929

Yas.

Nah.

Nah.

https://github.com/sensu/sensu-docs/issues/1609

https://sensuinc.testrail.io/index.php?/cases/view/8783&group_by=cases:section_id&group_order=asc&group_id=175
Required by https://github.com/sensu/sensu-staging/pull/144
Required by https://github.com/sensu/sensu-go-qa-crucible/pull/61


agent/check_handler.go
```diff
@@ -227,7 +227,7 @@ func (a *Agent) executeCheck(ctx context.Context, request *corev2.CheckRequest,
 	event.Timestamp = time.Now().Unix()
 
 	if len(checkHooks) != 0 {
-		event.Check.Hooks = a.ExecuteHooks(request, checkExec.Status)
+		event.Check.Hooks = a.ExecuteHooks(request, checkExec.Status, assets)
```
ccressent: I'm not very familiar with this code, but at a glance it seems that `assets` here corresponds to the check's assets? Is it not possible for checks and hooks to use different assets?

agent/hook.go
```diff
@@ -37,7 +38,7 @@ func (a *Agent) ExecuteHooks(request *corev2.CheckRequest, status int) []*corev2
 				// code and severity (ex. 0, ok)
 				in := hookInList(hookConfig.Name, executedHooks)
 				if !in {
-					hook := a.executeHook(hookConfig, request.Config.Name)
+					hook := a.executeHook(hookConfig, request.Config.Name, assets)
```
ccressent: Can't different hooks use different assets?

backend/schedulerd/cron_scheduler.go
```diff
@@ -59,6 +59,7 @@ func (s *CronScheduler) schedule(timer *CronTimer, executor *CheckExecutor) {
 	}
 }
 
+// Start starts thr cron scheduler.
```
ccressent: s/thr/the

backend/schedulerd/executor.go
```diff
@@ -327,17 +326,17 @@ func buildRequest(check *types.CheckConfig, s store.Store) (*types.CheckRequest,
...
+	}
+
 	// Guard against iterating over assets if there are no assets associated with
 	// the check in the first place.
 	if len(check.RuntimeAssets) != 0 {
-		// Explode assets; get assets & filter out those that are irrelevant
-		assets, err := s.GetAssets(ctx, &store.SelectionPredicate{})
-		if err != nil {
-			return nil, err
-		}
```
ccressent: I'm not 100% sure why this is moved outside of the `if len(check.RuntimeAssets) != 0 { }` branch. I'm also not 100% sure what that branch is even for :)

agent/check_handler.go
nikkiki: Yes, it's completely possible for checks and hooks to use different assets. They're all [loaded in by schedulerd](https://github.com/sensu/sensu-go/blob/feature/hook-assets/backend/schedulerd/executor.go#L347-L366) and send to the agent. I should update the variable name here to distinguish it is not just the check's assets, rather the entire request.

agent/hook.go
nikkiki: Yes they can. Here `assets` is the entire list of assets that are relevant to the request.

backend/schedulerd/executor.go
nikkiki: This call `s.GetAssets(ctx, &store.SelectionPredicate{})` retrieves _all_ assets within the namespace. I moved it outside of the branch to avoid calling the function twice (for checks and for hooks). The `if len(check.RuntimeAssets) != 0` and `if len(hook.RuntimeAssets) != 0` blocks simply verify if the check or hook have any runtime assets associated with their configs. If not, we don't have to iterate through the list to load in the assets.

agent/check_handler.go
ccressent: I see, the scheduler sends the agent a list of all the assets that are needed for it to be able to run the check, its hooks, ... all in one go, as one big bundle? I assumed it was "compartmentalized" but never checked.

My question then is as follow: could there be situations in which the check's assets and the hook's assets are fine as separate sets, but incompatible when merged? I'm not expecting this PR to attempt to solve such issues, but I'm curious to know what you think about it!

agent/hook.go
ccressent: Similarly to my other comment, can we imagine hooks each with assets that are fine taken as individual sets but somehow incompatible (or worse silently producing a wrong result) when merged into a single set of assets?

agent/check_handler.go
nikkiki: For each check being scheduled, a request is built with the check, it's hooks, and it's assets (this includes all assets needed by the check and the potential hooks). You are correct, the check request is sent from the backend to the agent as one big bundle. I'm not sure how to compartmentalize this to be more efficient, since only the backend is aware of assets which exist.

I'm not sure what you mean by "incompatible when merged." Could you elaborate?

backend/schedulerd/executor.go
ccressent: > This call `s.GetAssets(ctx, &store.SelectionPredicate{})` retrieves _all_ assets within the namespace. I moved it outside of the branch to avoid calling the function twice (for checks and for hooks). 

Thanks, that makes perfect sense!

> The `if len(check.RuntimeAssets) != 0` and `if len(hook.RuntimeAssets) != 0` blocks simply verify if the check or hook have any runtime assets associated with their configs. If not, we don't have to iterate through the list to load in the assets.

That's my bad, I read the code too quickly; I now understand why we short-circuit... we're not ranging over the assets/hooks for the current check. We're ranging over _all_ the assets/hooks in the namespace! So the more assets/hooks there are in the namespace, the longer it takes to run through `{asset,hook}IsRelevant()`. I doubt this is a bottleneck for most setup, but ideally I think we should just fetch the needed assets from the store directly, rather than fetch everything and then compare one by one. This is of course completely out of the scope of this PR!

nikkiki: I totally agree, there is definitely some room for optimization here. However, at least in the case of hooks, we won't know which assets are actually needed until the check executes (check status influences the hooks which run, and the assets required to run the hook).

agent/check_handler.go
ccressent: The compartmentalization I mention is more about guaranteeing the environment checks and hooks expect more than anything else. I know our assets are pretty rudimentary, but let's take typical package manager problems into consideration. Suppose checkAsset is an asset used by a check and hookAsset is an asset used by a hook.

- checkAsset and hookAsset are not compatible

- checkAsset depends on A at version X; hookAsset depends on A at version Y
After all assets are installed, the check or the hook might not work

- checkAsset includes a dependency on A; hookAsset includes a dependency on B; A and B are incompatible
After all assets are installed, the check and the hook might not work.

In each of those cases, "running" {check + its assets} or {hook + its assets} would work, but "running" {check + all assets} or {hook + all assets} might not (or worse, produce wrong results).

I need to check how assets are used by the agent; either way, this is not an issue with your PR at all, but that's what I meant by "incompatible when merged"!

nikkiki: Hm, I see. If there were assets that were incompatible, they would cached in different directories (as all assets are), and only the environments of the relevant assets are loaded into the executable's environment. I suppose that is sufficient for now, but we can definitely revisit this if we run into an issue in the future!

Issue #3135, opened 2019-07-11
proxy checks don't get scheduled


sjthespian:
##  Expected Behavior
There seems to be a bug with template variable and proxy checks that keeps them from getting scheduled in some situations. In my case, I have a pair or proxy checks that work on some entities but never appear on others.

## Current Behavior
The proxy checks are appearing on some entities but not others. The error message that appears in the log is less than helpful, it doesn't specify the check or the entity that is causing the error:
```
Jul 11 15:29:06 morpheus sensu-backend[19233]: {"component":"schedulerd","error":"unmatched token: template: :1:46: executing \"\" at \u003c.system.hostname\u003e: map has no entry for key \"hostname\"","level":"error","msg":"error executing check","time":"2019-07-11T15:29:06-07:00"}
```

## Possible Solution
To start with, the error message needs at least the entity name and check that are casing the error included in the message. At the moment, it is virtually impossible to debug what is causing that message.

## Steps to Reproduce (for bugs)
1. I haven't found a reproducible scenario for this yet

## Context

For one of my proxy checks I found that the issue was caused by the entity attribute in the check existing on a non-proxy host. When I removed that the check suddenly started working. To be specific, here is a stripped-down check definition (removing nulls and 0 values):
```
type: CheckConfig
api_version: core/v2
metadata:
  annotations:
    sensu.io.json_attributes: '{"occurrences":3,"refresh":3600,"dependencies":["keepalive"]}'
  name: proxy-http
  namespace: default
spec:
  command: check-http.rb -u {{ .labels.url }}
  handlers:
  - mailer
  interval: 180
  proxy_entity_name: ""
  proxy_requests:
    entity_attributes:
    - entity.labels.has_http_server == 'true'
    splay: true
    splay_coverage: 90
  publish: true
  round_robin: true
  runtime_assets:
  - sensu-ruby-runtime
  - sensu-ruby-runtime-armv7l
  - sensu-plugins-http
  subscriptions:
  - standalone
```
When I have `has_http_server: 'true'` defined on a non-proxy host, this doesn't get applied everywhere. If I remove that from all of the agents it will suddenly appear on all of the proxy hosts that have `has_http_server` set as a label.

I have yet to figure out what is causing this check to not be scheduled anywhere:
```
type: CheckConfig
api_version: core/v2
metadata:
  annotations:
    sensu.io.json_attributes: '{"occurrences":3,"refresh":3600,"dependencies":["keepalive"]}'
  name: proxy-fping
[...]

PR #3136, opened 2019-07-12
Adding a new error message with entity and check name for debuging proxy entity issues

jspaleta:
## What is this change?
Better diagnostic message to help track down which proxy entity and check is throwing a token substitution error. 
This PR is a result of troubleshooting #3135 

Troubleshooting aid..without this context you have to do a full review of all your proxy entities looking for the problem. 

Maybe an Added statement about the addition of new error log message?  

May want to refactor this as an updated error message thrown by agent/TokenSubsitution to include entity and check name. Not sure which is better. I like this implementation because it makes it clear its happening in the context of a proxy entity.

Nope.

I dont think we have documentation that covers detailed error diagnostics.

local build of backend, created a bad proxy check substitution and watched the log for new error message.


backend/schedulerd/proxy_check.go
```diff
@@ -53,6 +53,7 @@ func substituteProxyEntityTokens(entity *corev2.Entity, check *corev2.CheckConfi
 	// entity
 	checkBytes, err := agent.TokenSubstitution(synthesizedEntity, check)
 	if err != nil {
+		logger.Error(fmt.Errorf("error in TokenSubstitution for proxy entity: %s check: %s err: %s", entity.Name, check.Name, err))
```
jamesdphillips: You could do...

```go
logger.Errorf("error in TokenSubstitution for proxy entity: %s check: %s err: %s", entity.Name, check.Name, err)
```

but, given that logrus gives us structured logging I think it would be even more preferable to do:

```go
logger.WithField("check", check.Name).WithField("entity", entity.Name).WithError(err).Error("unable to substitute tokens")
```

jspaleta: cool.. i'll refactor

jspaleta: refactored now

jspaleta:
merged in master and fixed dco signoff

palourde:
@jspaleta It appears your PR will need some love, there's a lot of conflicts & the diff seems a bit weird, it looks like it changes a lot of stuff unrelated to the original PR!

echlebek:
Closing in favour of #3220

Issue #3078, opened 2019-06-23
some checks are not being schedule by schedulerd


Alexc0007:
some of my checks are not being scheduled automatically by schedulerd.
some other checks run just fine.
check definition is basically almost the same for those checks so i cant figure out what's wrong.

here is an example of a check that stops scheduling automatically:
=== cpu_load
Name:              cpu_load
Interval:          180
Command:           /usr/lib64/nagios/plugins/check_load -w 15,10,5 -c 30,25,20
Cron:              
Timeout:           0
TTL:               0
Subscriptions:     all
Handlers:
Runtime Assets:    
Hooks:             
Publish?:          true
Stdin?:            false
Proxy Entity Name: 
Namespace:         default
Metric Format:     nagios_perfdata
Metric Handlers:   influx-db

has anyone encountered such behaviour?

## Expected Behavior
check interval is 180 seconds , so i expect the check to run every 3 minutes.

## Current Behavior
after restarting the backend it seems like the check start executing correctly every 3 minutes , but at some point it just stops.

if i try to re-run the check , it works fine , but then again - not being executed on its own.

here is an example of several checks that are being executed just fine , along with 
cpu_load and check_self_signed_cert_rootca that stopped executing 3 days ago during the weekend

![image](https://user-images.githubusercontent.com/15061757/59972669-fe464b80-959b-11e9-9da9-e97a3168e769.png)

## Possible Solution
i have no idea why it happens or what is causing it...
so i cant offer a solution

## Steps to Reproduce (for bugs)
the only thing in common between the checks that are not working properly
is that they all have low amount of subscriptions...
they are basic checks and should run on all servers , so they use the "all"
subscription.

## Your Environment
most of the checks run on amazon-linux or amazon-linux2
however the checks that are running on also run on same enviromnets

Alexc0007:
as i keep debugging this case , i can see constant errors in the backend log:
{"component":"eventd","error":"error setting silenced entries: couldn't get silenced entries: etcdserver: too many operations in txn request","level":"error","msg":"eventd - error handling event","time":"2019-06-24T09:04:58Z"}

when sensu stop executing some of the checks(those checks mentioned above) - about 10 minutes after , those errors stop.
when i restart the backend , the errors return

annaplotkin:
@Alexc0007 what version are you running?

Alexc0007:
Hi Anna , 
i tried it on 5.7 , 5.8 , 5.9 , 5.10 - it was the same behavior.
[...]

PR #3138, opened 2019-07-12
Websocket heartbeats

palourde:
## What is this change?

The agent now sends heartbeat to the backend over the WebSocket (via the ping/pong control frames), so it can detect network failures faster and reconnect to a different backend (if available).

It also reduces the handshake timeout from 45 seconds to 15 seconds when the agent negotiates its WebSocket connection, so it can failover to a an another backend (if available) faster.

Closes https://github.com/sensu/sensu-go/issues/3078
Closes https://github.com/sensu/sensu-go/issues/3106

Added

Nope

WebSockets are weird 😄 

Yes, we need to document the new config attributes.

I've done manual e2e testing. This change would be really hard to test via an integration test so I will attempt to add something to our QA crucible instead.


transport/transport.go
```diff
@@ -168,6 +174,49 @@ func (t *WebSocketTransport) Closed() bool {
...
+					return
+				}
+			case <-ctx.Done():
+				logger.Debug("websocket connection has been closed, stopping the heartbeat")
+				return
+			}
+		}
+	}()
+
+	t.Connection.SetReadDeadline(time.Now().Add(pongWait))
```
golangcibot: Error return value of `t.Connection.SetReadDeadline` is not checked (from `errcheck`)

transport/transport.go
```diff
@@ -168,6 +174,49 @@ func (t *WebSocketTransport) Closed() bool {
...
+	t.Connection.SetPongHandler(func(string) error {
+		logger.Debugf("pong received from the backend, setting the read deadline to %d", time.Now().Add(pongWait).Unix())
+		t.Connection.SetReadDeadline(time.Now().Add(pongWait))
```
golangcibot: Error return value of `t.Connection.SetReadDeadline` is not checked (from `errcheck`)

backend/agentd/session_test.go
```diff
@@ -30,6 +31,10 @@ func (t *testTransport) Close() error {
 	return nil
 }
 
+func (t *testTransport) Heartbeat(ctx context.Context, interval, timeout int) {
+	return
```
golangcibot: S1023: redundant `return` statement (from `gosimple`)

Issue #2796, opened 2019-03-13
Use go modules for vendoring instead of dep


echlebek:
Now that Go modules has matured a little, and is poised to become the default for the Go toolchain, we should move to a modules vendoring approach instead of using dep.

With modules, we can:

* Use only the go toolchain for dependency management. No dep versioning to worry about.
* Avoid committing dependencies for the project, while still enabling `go get` to work.
* Be certain that the vendoring solution will be supported long-term, and will be subject to the Go compatibility promise.

echlebek:
@portertech says we will do this in 5.12 :laughing:

nikkiki:
![giphy](https://user-images.githubusercontent.com/10634078/59627425-336e1c00-910c-11e9-91fb-48b8997b3b33.gif)

PR #3165, opened 2019-07-24
Convert project to go modules

echlebek:

## What is this change?

Remove dep, add go modules support.

It's better to use the official method of dependency management.

Yes

I ran `go build ./...`


nikkiki:
Here's the associated issue! https://github.com/sensu/sensu-go/issues/2796

PR #3171, opened 2019-07-27
Use module proxy

echlebek:

## What is this change?

Uses the go proxy for modules, which should result in a build speedup for modules that are not cached.


echlebek:
Appveyor got a nice speedup with this, and the build is under 3 mins now.

Issue #2742, opened 2019-02-21
Feature Request: Dump config 


asachs01:
## Expected Behavior
As an operator, I would like an easy way to dump my Sensu check, filter, mutator, handler, and asset definitions to a yaml file that could be reimported should I need to blow away etcd. 

## Current Behavior
There isn't an operator-centric way of exporting an existing config. I'd have to write up a script to iterate through the various API endpoints, or use sensuctl. This is fairly onerous and makes restoring my Sensu configuration a less-than-fun process. Even if I'm using something like config management, or have a repo of files, I'd still have to add those back one-by-one. 

## Possible Solution
```
sensuctl dump --help

sensuctl dump {checks,filters,handlers,assets,mutators,all} --file FILE --namespace

sensuctl dump all -F /etc/sensu/site-config.yaml
```
Could also do an API call
`curl -H "Authorization: Bearer TOKEN" \
http://127.0.0.1:8080/api/core/v2/namespaces/default/settings/{checks,filters,mutators,handlers,assets}`

or 

`curl -H "Authorization: Bearer TOKEN" \
http://127.0.0.1:8080/api/core/v2/settings`

The idea here being that I could either as the deployment owner dump the ENTIRE config, or I could say as a team member dump the config for my namespace.

rgeniesse:
Another use case here is for use in troubleshooting. Sensu Classic provides means to dump the entire configuration so support can look at everything as a whole easily. Users could also use this to get the entire picture in a single output/file.

johntdyer:
I second this, I want a simple way to backup my configs as JSON

SiqingYu:
https://docs.sensu.io/sensu-core/1.7/reference/configuration

portertech:
Would also like the ability to dump other API resource, e.g. Events.

aphix:
Here's a simple script I whipped up to tackle this for myself. I'm not sure how easily I can re-import it, but it's better than nothing.

`print-all-resources.sh`
```bash
#!/bin/bash

get_date_safe_for_filenames() {
	date +"%Y%m%d-%H%M%S"
}

# use -h or --help to get usage info
if [[ "$*" =~ -h ]]; then
    echo "Outputs all sensu resources to specified directory using \`sensuctl\`.
Default output folder is \"./sensu_backup_[CURRENT_DATE_TIME]/\".

    usage: $0 <output_dir, default: \"./sensu_backup_$(get_date_safe_for_filenames)/\">
    "
    exit
fi

[...]

PR #3176, opened 2019-07-31
Sensuctl dump

nikkixdev:
## What is this change?

Implements `sensuctl dump`.

Closes https://github.com/sensu/sensu-go/issues/2742

Yas.

Nah.

There was some nuance with error handling with regard to how unavailable enterprise resources might be handled.

Filing a docs issue now.

Filing QA crucible issues now.

examples:
```
$ sensuctl dump check
dumps all checks in the current namespace

$ sensuctl dump check --namespace dev
dumps all checks in the dev namespace

$ sensuctl dump check --all-namespaces
dumps all checks in all namespaces

$ sensuctl dump all --all-namespaces
dumps all resources in all namespaces
non-namespaced resources are prioritized first to mitigate errors when using sensuctl create (ex. namespaces must be created before resources within that namespace)

$ sensuctl dump check,foo
returns an error because foo is invalid

$ sensuctl dump check,check
returns an error because there are duplicate resource types

$ sensuctl dump all --all-namespaces --format wrapped-json
dumps all resources in all namespaces in wrapped-json format (only yaml and wrapped-json are supported due to sensuctl create)

$ sensuctl dump all --all-namespaces --file resources.yaml
dumps all resources in all namespaces to the designated file
```

### DIFF
--- a/cli/commands/dump/dump.go
+++ b/cli/commands/dump/dump.go
@@ -141,7 +141,6 @@ func execute(cli *cli.SensuCli) func(*cobra.Command, []string) error {
 			return err
 		}
 		if fp == "" {
-			_, err := fmt.Print(out)
 			return err
 		}
 		f, err := os.Create(fp)
--- a/cli/commands/dump/dump.go
+++ b/cli/commands/dump/dump.go
@@ -145,10 +145,10 @@ func execute(cli *cli.SensuCli) func(*cobra.Command, []string) error {
 			return err
 		}
 		f, err := os.Create(fp)
-		defer f.Close()
 		if err != nil {
 			return err
 		}
+		defer f.Close()
 		_, err = f.WriteString(out)
 		return err
 	}
--- a/cli/commands/dump/dump.go
+++ b/cli/commands/dump/dump.go
@@ -89,10 +89,14 @@ func execute(cli *cli.SensuCli) func(*cobra.Command, []string) error {
 		} else {
 			types := strings.Split(args[0], ",")
 			for _, t := range types {
+				length := len(actions)
 				for _, action := range All {
 					if t == action.Resource {
 						actions = append(actions, action)
 					}
+				}
+				if length == len(actions) {
+					return fmt.Errorf("invalid arg: %s", t)
 				}
 			}
 		}
@@ -120,11 +124,7 @@ func execute(cli *cli.SensuCli) func(*cobra.Command, []string) error {
 			}
 
 			// execute the command and build wrapped-json or yaml lists
-			originalBytes, err := exec.Command(os.Args[0], ctlArgs...).CombinedOutput()
-			if err != nil {
-				_, _ = cmd.OutOrStdout().Write(originalBytes)
-				return fmt.Errorf("couldn't get resource: %s", err)
-			}
+			originalBytes, _ := exec.Command(os.Args[0], ctlArgs...).CombinedOutput()
 			if len(originalBytes) == 0 {
 				continue
 			}
--- a/cli/commands/dump/dump.go
+++ b/cli/commands/dump/dump.go
@@ -141,6 +141,7 @@ func execute(cli *cli.SensuCli) func(*cobra.Command, []string) error {
 			return err
 		}
 		if fp == "" {
+			_, err := fmt.Print(out)
 			return err
 		}
 		f, err := os.Create(fp)
--- a/cli/commands/dump/dump.go
+++ b/cli/commands/dump/dump.go
@@ -96,7 +96,7 @@ func execute(cli *cli.SensuCli) func(*cobra.Command, []string) error {
 					}
 				}
 				if length == len(actions) {
-					return fmt.Errorf("invalid arg: %s", t)
+					return fmt.Errorf("invalid resource type: %s", t)
 				}
 			}
 		}
--- a/cli/commands/dump/dump.go
+++ b/cli/commands/dump/dump.go
@@ -103,7 +103,7 @@ func execute(cli *cli.SensuCli) func(*cobra.Command, []string) error {
 
 		// iterate the matched actions and start building a sensuctl command
 		var out string
-		for i, a := range actions {
+		for _, a := range actions {
 			ctlArgs := []string{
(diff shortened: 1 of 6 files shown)


cli/commands/dump/dump.go
```diff
@@ -0,0 +1,155 @@
...
+		fp, err := cmd.Flags().GetString("file")
+		if err != nil {
+			return err
+		}
+		if fp == "" {
+			_, err := fmt.Print(out)
+			return err
+		}
+		f, err := os.Create(fp)
+		defer f.Close()
```
golangcibot: SA5001: should check returned error before deferring f.Close() (from `staticcheck`)

cli/commands/dump/dump.go
```diff
@@ -0,0 +1,155 @@
...
+
+		// get the configured format or the flag override
+		format := cli.Config.Format()
+		if flag := helpers.GetChangedStringValueFlag("format", cmd.Flags()); flag != "" {
+			format = flag
+		}
+		switch format {
+		case "yaml", "wrapped-json":
+		default:
+			format = "yaml"
```
palourde: This bit is a bit counterintuitive for me. If you print the help usage, it tells you the following:

```
$ sensuctl dump -h
[...]
      --format string    format of data returned ("json"|"wrapped-json"|"tabular") (default "tabular")
```

Which is actually not true, since the default format is `yaml` and only `wrapped-json` & `yaml` are supported. I think we should remove the default format flag (added at the line 60 above) and create our own flag with the correct default & allowed values.

cli/commands/dump/dump.go
```diff
@@ -0,0 +1,155 @@
...
+		var actions []Action
+		if args[0] == "all" {
+			actions = All
+		} else {
+			types := strings.Split(args[0], ",")
+			for _, t := range types {
+				length := len(actions)
+				for _, action := range All {
+					if t == action.Resource {
+						actions = append(actions, action)
```
palourde: We should probably make sure we don't add duplicates into this slice, otherwise right now if I do `sensuctl dump check,check`, it duplicates the resources printed. I could see this being problematic when you want to dump all resources except one and the list of types become long and therefore easy to make a mistake and duplicate a type.

cli/commands/dump/dump.go
nikkixdev: Kinda stole this from https://github.com/sensu/sensu-go/blob/daa6748dfc5d808f65418b474a2fdb58b1ba40a4/cli/commands/edit/edit.go#L45-L48 but I suppose it would be a little confusing if someone did `--format tabular` and got yaml instead. I agree, I'll add a special flag rather than the global one.

cli/commands/dump/dump.go
nikkixdev: Good catch!

cli/commands/dump/dump.go
nikkixdev: Actually, here is another weird UX thing. Say I'm configured in default tabular format, and run `sensuctl dump all` without a format flag. In other cases of sensuctl we would want to use the current format, however this flag isn't consistent with that. I'm curious if a format flag which defaults to yaml rather than what's configured would be confusing to users..

palourde: > I'm curious if a format flag which defaults to yaml rather than what's configured would be confusing to users..

We don't have to change this behavior, I'm simply saying the help usage should be modified to reflect what's happening behind the scenes. We can simply avoid calling `helpers.AddFormatFlag(cmd.Flags())` and instead manually define the `format` flag so at least the help usage displays the right values as default and supported, while still trying the use the value the user has in its config.

Showing `tabular` or even `json` as a support format value for this command is more confusing to me, because it's not the reality.

nikkixdev: - Updated help usage in `dump` to read `--format string    format of data returned ("wrapped-json"|"yaml") (default "wrapped-json")` (where wrapped-json is the current config)
- Updated help usage in the global format flag to read `--format string    format of data returned ("json"|"wrapped-json"|"tabular"|"yaml") (default "tabular")` (yaml was missing before)

cli/commands/dump/dump.go
nikkixdev: In the event of duplicate resource types, the command returns the following error: `Error: duplicate resource type: check`

cli/commands/dump/dump.go
```diff
@@ -0,0 +1,165 @@
...
+				} else if ok {
+					ctlArgs = append(ctlArgs, "--all-namespaces")
+				} else {
+					ctlArgs = append(ctlArgs, "--namespace")
+					ctlArgs = append(ctlArgs, cli.Config.Namespace())
+				}
+			}
+
+			// execute the command and build wrapped-json or yaml lists
+			originalBytes, _ := exec.Command(os.Args[0], ctlArgs...).Output()
```
echlebek: This is a bit problematic because if there is an error executing the command (say, the connection to the server is interrupted), the dump command will continue on and silently discard information.

It would be nice to detect failure, and pass on the stderr output from the command if that happens.

I'd recommend something like 
```
originalBytes, err := exec.Command(os.Args[0], ctlArgs...).CombinedOutput()
if err != nil {
    // print output and exit
}
```

nikkixdev: I believe I originally implemented this with `CombinedOutput()` but ran into an issue with enterprise resources. A licensed user would have no problem using `sensuctl dump all` since they have access to the injected enterprise resources. An unlicensed user however, might get something like below, where the silently discarded information is invalid and irrelevant. Happy to bounce some ideas around about a solution to this use case.
```
---
Error: whatever error is thrown if an unlicensed user tries to get license info
---
type: CheckConfig
api_version: core/v2
metadata:
  name: default
  namespace: default
spec:
  command: echo default
...
```

echlebek: Yeah I see the problem.. how annoying. If commands returned different status codes based on their error, then we could differentiate, but they don't.

Fixing this problem would require changing the status codes we return on error, or reimplementing everything with API calls, so let's go with what you've done for now. :+1:

Issue #3137, opened 2019-07-12
Template values require restart to populate in proxy checks


sjthespian:
## Expected Behavior
Proxy checks should update any templated values when those values change.

## Current Behavior
A proxy check that contains templated values will not see updates to those values without a restart of sensu-backend.

## Possible Solution
It appears that this is an issue with caching between the backend and etcd.

## Steps to Reproduce (for bugs)
1. Create a proxy check with templated values
2. Allow the check to run at least once
3. Change the label in the entity
4. Wait for the check to run again and see that it is still using the old value

## Context
Sample check:
```
type: CheckConfig
api_version: core/v2
metadata:
  annotations:
    sensu.io.json_attributes: '{"occurrences":3,"refresh":3600,"dependencies":["keepalive"]}'
  name: proxy-http
  namespace: default
spec:
  command: check-http.rb -u {{ .labels.url }}
  handlers:
  - mailer
  interval: 180
  proxy_entity_name: ""
  proxy_requests:
    entity_attributes:
    - entity.labels.has_http_server == 'true'
    splay: true
    splay_coverage: 90
  publish: true
  round_robin: true
  runtime_assets:
  - sensu-ruby-runtime
  - sensu-ruby-runtime-armv7l
  - sensu-plugins-http
  subscriptions:
  - standalone
```

Sample proxy entity:
```
type: Entity
api_version: core/v2
metadata:
  labels:
    has_http_server: "true"
    has_ssh: "true"
    has_ssl_cert: "true"
    needs_fping: "true"
    url: https://www.example.org/
  name: myhost.example.org
  namespace: default
spec:
  deregister: false
  deregistration: {}
  entity_class: proxy
  last_seen: 0
  subscriptions:
  - entity:myhost.example.org
  system:
    hostname: myhost.example.org
    network:
      interfaces: null
```

If you change the URL above with `sensuctl entity edit` it will not reflect in the check that is being run util the backend is restarted. This can be verified by looking at the event in the UI.

## Your Environment
* Sensu version used - sensuctl, sensu-backend, sensu-agent: 5.11.0-4888
* Installation method: Deb x86_64 packages
* Operating System and version: Ubuntu 18.04

jspaleta:
I'm very suspicious that this is a an issue with cache/etcd watcher issue for entities.

Issue #3101, opened 2019-06-28
Keep track of etcd revisions in watchers


palourde:
This is a follow up of https://github.com/sensu/sensu-go/pull/3091.

There's currently a possibility of losing events coming from a watch in case there's a split-brain or even a temporary disconnection, which could lead to an invalid cache.

A possible solution would be to keep track of etcd global revision number so when we need to re-create a watcher, we could replay events from this revision and therefore make sure we didn't lose any events.

We will need to carefully test this functionality though. For example, what happen if multiple compaction happen and the revision we try to use is no longer available? We need to better understand how etcd handles these use cases.

palourde:
From https://github.com/etcd-io/etcd/blob/master/Documentation/learning/api.md#watch-streams

> Compact_Revision - set to the minimum historical revision available to etcd if a watcher tries watching at a compacted revision. This happens when creating a watcher at a compacted revision or the watcher cannot catch up with the progress of the key-value store. The watcher will be canceled; creating new watches with the same start_revision will fail.

PR #3181, opened 2019-08-02
Watcher refactoring

palourde:
## What is this change?

This PR introduces multiple changes/improvements to the watcher framework in order to improve its reliability and accuracy.

In summary:
- The watcher now keeps track of the revisions, so whenever it's restarted, it can pick up the events that were possibly missed.
- It introduces watch events of type `store.WatchError`, which are used to indicate to the consumer that watch events might have been lost
	- The cacher uses these events to force a rebuild of its cache. I've basically copied [Eric's commit](https://github.com/sensu/sensu-go/commit/17f1d951ef5895a3e9f1a32da557ecdfb0de94fc) for that.
- A `store.WatchError` can only be sent in one of these 2 situations:
	1. When `CompactRevision` in the WatchResponse is `!= 0`. It basically means the watcher tried to watch a revision that already got compacted. Therefore, what we do instead is send this error so consumers can act accordingly and we restart the watcher at the revision specified by `CompactRevision`, which is the last revision still available.
	2. When the previous key value is not available on a watch event of type `delete`; we now request the previous key value instead of trying to build a resource using only the key, which was a big hack. Therefore, if the previous key value is no longer available, we no longer try to hack our way around.
- Watch events sent to the consumers are now buffered so we don't block immediately if the last event wasn't consumed yet. 
- The logging around the watcher has been improved so it's easier to identify an unstable watcher.
- Additional integration tests around the watcher & cacher. The `github.com/coreos/etcd/integration` package is now used to emulate an etcd cluster with various failures.

Closes https://github.com/sensu/sensu-go/issues/3101
Closes https://github.com/sensu/sensu-go/issues/3135
Closes https://github.com/sensu/sensu-go/issues/3137

Added!

Nope!

The issues reported around the watcher are near impossible to reproduce in a lab environment.

Not required.

In addition to the new integration tests, I've successfully tested it against sensu-staging with the automated tests. It also successfully ran under load within the SPDC environment.

I don't believe we can easily add any e2e tests, because it's almost impossible to reproduce the required conditions that cause the watcher to fail. Instead, I've used the `github.com/coreos/etcd/integration` to reproduce a cluster partition and other weird failures.


backend/store/etcd/helpers_test.go
```diff
@@ -0,0 +1,57 @@
...
+	"testing"
+
+	"github.com/coreos/etcd/clientv3"
+	"github.com/sensu/sensu-go/backend/etcd"
+	"github.com/sensu/sensu-go/backend/store"
+	"github.com/sensu/sensu-go/types"
+	"github.com/stretchr/testify/require"
+)
+
+func testWithEtcd(t *testing.T, f func(store.Store)) {
```
golangcibot: `testWithEtcd` is unused (from `deadcode`)

backend/store/etcd/helpers_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+	s := NewStore(client, e.Name())
+
+	// Mock a default namespace
+	require.NoError(t, s.CreateNamespace(context.Background(), types.FixtureNamespace("default")))
+
+	f(s)
+}
+
+func testWithEtcdStore(t *testing.T, f func(*Store)) {
```
golangcibot: `testWithEtcdStore` is unused (from `deadcode`)

backend/store/etcd/helpers_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+	s := NewStore(client, e.Name())
+
+	// Mock a default namespace
+	require.NoError(t, s.CreateNamespace(context.Background(), types.FixtureNamespace("default")))
+
+	f(s)
+}
+
+func testWithEtcdClient(t *testing.T, f func(store.Store, *clientv3.Client)) {
```
golangcibot: `testWithEtcdClient` is unused (from `deadcode`)

backend/store/etcd/watcher.go
```diff
@@ -62,25 +71,25 @@ func (w *Watcher) Result() <-chan store.WatchEvent {
...
-	// resultChan
-	var resultChanWG sync.WaitGroup
-
 	// Create a cancellable context for our watcher
 	ctx, cancel := context.WithCancel(w.ctx)
 
 	// Start the watcher
-	logger.Debugf("starting a watcher for key %s", w.key)
-	w.watch(ctx, opts, watchChanStopped, &resultChanWG)
+	w.logger.Debug("starting a watcher")
```
nikkixdev: Does the watcher key not provide any helpful context for the debug logs?

backend/store/cache/cache.go
```diff
@@ -47,6 +47,16 @@ func buildCache(resources []corev2.Resource, synthesize bool) cache {
...
 type resourceSlice []Value
 
+func (s resourceSlice) Find(value Value) Value {
+	idx := sort.Search(len(s), func(i int) bool {
+		return !resourceLT(s[i], value)
+	})
+	if idx < len(s) && s[idx].Resource.GetObjectMeta().Name == value.Resource.GetObjectMeta().Name {
+		return s[idx]
+	}
+	return Value{}
```
echlebek: Should this be `return nil`?

backend/store/etcd/watcher.go
echlebek: Seconded, this is not really useful logger output without the watcher key.

palourde: The key is now part of a field within the logrus entry, see [line 60](https://github.com/sensu/sensu-go/pull/3181/files#diff-2b8fe086ea46203124c5b444c0be9787R60) above, so now all log entries indicate the key without having to format it every time!

palourde:
@echlebek re: performance testing, yes I did, on the SPDC cluster, it didn't cause any noticeable change in terms of performance.

backend/store/cache/cache.go
palourde: AFAIK you cannot use `nil` as type Value here!

PR #3182, opened 2019-08-02
Trace GraphQL query executions

jamesdphillips:
## What is this change?

* Updates `graphql-go/graphql` module for it's "Extensions" feature. "Extensions" are in essence middleware that allow an implementer to wrap query executions and provide functionality, see: https://github.com/graphql-go/graphql/pull/448.
* Introduces a GraphQL middleware that captures the duration it took to perform different operations within a query execution. Parsing, validation, resolving field, and total resolver time are observed. Implementation borrows heavily from [Ruby implementation](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/tracing/prometheus_tracing.rb).
* Observations are written to a Prometheus summary.

Will allow us to get detailed metrics on GraphQL query performance.

Sure.

No.

Hubris.

I don't believe any will be required.

Unit tests

### DIFF
--- a/graphql/middleware.go
+++ b/graphql/middleware.go
@@ -9,7 +9,8 @@ import (
 // Middleware is an interface for extending a service with operations that
 // wrap existing functionality.
 type Middleware interface {
-	// Init is used to help you initialize the extension
+	// Init is called at the beginning of an execution. Helpful for initializing
+	// anything used during the execution of the query.
 	Init(context.Context, *graphql.Params) context.Context
 
 	// Name returns the name of the extension (make sure it's custom)
--- a/graphql/middleware.go
+++ b/graphql/middleware.go
@@ -1,0 +1,35 @@
+package graphql
+
+import (
+	"context"
+
+	"github.com/graphql-go/graphql"
+)
+
+// Middleware is an interface for extending a service with operations that
+// wrap existing functionality.
+type Middleware interface {
+	// Init is used to help you initialize the extension
+	Init(context.Context, *graphql.Params) context.Context
+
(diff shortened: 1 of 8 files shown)


graphql/middleware.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"context"
+
+	"github.com/graphql-go/graphql"
+)
+
+// Middleware is an interface for extending a service with operations that
+// wrap existing functionality.
+type Middleware interface {
+	// Init is used to help you initialize the extension
```
nikkixdev: Nitpick about comments being full sentences: https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

graphql/tracing/prometheus.go
```diff
@@ -0,0 +1,137 @@
+package tracing
+
+import (
+	"context"
+
+	time "github.com/echlebek/timeproxy"
+	"github.com/graphql-go/graphql"
+	"github.com/graphql-go/graphql/gqlerrors"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/sensu/sensu-go/util/strings"
```
nikkixdev: Might want to use an alias here, as not to get confused with `strings` from the standard library.

```suggestion
	utilstrings "github.com/sensu/sensu-go/util/strings"
```

graphql/middleware.go
jamesdphillips: addressed 🙏

graphql/tracing/prometheus.go
jamesdphillips: addressed 🙏

Issue #3179, opened 2019-08-01
Feature Request: Allow dropping of events with no entity


roganartu:
Provide an option to the backend that causes it to drop (and log) events received for entities that do not exist.

## Expected Behavior
Submitting an event via the API for an entity that does not exist should not create an entity or an event.

## Current Behavior
Submitting an event via the API for an entity that does not exists creates a new entity and attaches the submitted event to it.

## Possible Solution
Add a flag to `sensu-backend` to enable this feature.

When the flag is provided and an event is received for an entity that doesn't exist, log a warning/error and drop the event.

## Context
We have tooling that syncs all our entities to Sensu from an existing central source of truth, so there is no situation where we would ever want to create an event for an entity that doesn't already exist. I would be useful if we could disable this automatic entity creation entirely.

This idea was motivated by the impact of #3178 on our environment.

annaplotkin:
Moving so the team can discuss this morning.

roganartu:
We came across another use case for this today: entity names are case sensitive, so an external tool we use to parse syslogs and create Sensu events created a new entity even though it already existed with a downcased name.

Sanitised output demonstrating this:
```
env  sensuctl entity info fooBAR                                                                     
=== fooBAR
Name:                   fooBAR
Entity Class:           proxy
Subscriptions:          entity:fooBAR
Last Seen:              N/A
Hostname:               
OS:                     
Platform:               
Platform Family:        
Platform Version:       
Auto-Deregistration:    false
Deregistration Handler: 

env  sensuctl entity info foobar
=== foobar
Name:                   foobar
Entity Class:           proxy
Subscriptions:          colo:foo, category:network, role:switch, entity:foobar
Last Seen:              2020-08-03 15:13:41 +0000 UTC
Hostname:               foobar
OS:                     
Platform:               baz
Platform Family:        
Platform Version:       
Auto-Deregistration:    false
Deregistration Handler:
```

This only became apparent because our tooling returned Not Found for events when using the lowercase variant. We will address this by downcasing all hostnames on egress from this syslog parsing service, but ultimately would prefer to be able to configure Sensu to return an error to that event creation instead.

I think this feature request could be better summarised as: Allow Sensu to be configured to respect an external source-of-truth for entities. The desire to ignore events from non-existent entities is simply a side-effect of the fact that we populate entities in Sensu from an external source-of-truth, really.

PR #3192, opened 2019-08-09
Add --no-implied-entities option

roganartu:
Add `--no-implied-entities` option to sensu-backend which disables the creation of entities when an event provides one that doesn't exist.

## What is this change?

Add a `--no-implied-entities` option to the backend that causes it to drop (and log) events received for entities that do not exist.

Implements #3179. This option allows installations which sync entities into Sensu from an external source of truth to entirely disable the creation of entities via proxy check results submitted to the agent API.

Yes, added one.

Don't think so, it was pretty straightforward.

Nope.

New docs are probably required.

Added a new unit test for the modified method.

Additionally ran within docker with a backend that has the option set as an ad hoc integration test. Received the following log as expected, and the entity was not created:
```
backend1_1  | {"component":"agentd","error":"foo does not exist and implied proxy entities are disabled","level":"error","msg":"error handling message","payload":"{\"timestamp\":1565296950,\"entity\":{\"entity_class\":\"agent\",\"system\":{\"hostname\":\"agent1\",\"os\":\"linux\",\"platform\":\"alpine\",\"platform_family\":\"alpine\",\"platform_version\":\"3.8.4\",\"network\":{\"interfaces\":[{\"name\":\"lo\",\"addresses\":[\"127.0.0.1/8\"]},{\"name\":\"eth0\",\"mac\":\"02:42:ac:12:00:05\",\"addresses\":[\"172.18.0.5/16\"]}]},\"arch\":\"amd64\"},\"subscriptions\":[\"switches\"],\"last_seen\":1565296858,\"deregister\":false,\"deregistration\":{},\"user\":\"agent\",\"redact\":[\"password\",\"passwd\",\"pass\",\"api_key\",\"api_token\",\"access_key\",\"secret_key\",\"private_key\",\"secret\"],\"metadata\":{\"name\":\"agent1\",\"namespace\":\"default\"}},\"check\":{\"handlers\":[],\"high_flap_threshold\":0,\"interval\":1,\"low_flap_threshold\":0,\"publish\":false,\"runtime_assets\":null,\"subscriptions\":[],\"proxy_entity_name\":\"foo\",\"check_hooks\":null,\"stdin\":false,\"subdue\":null,\"ttl\":0,\"timeout\":0,\"round_robin\":false,\"executed\":1565296950,\"history\":null,\"issued\":0,\"output\":\"Some error\",\"status\":2,\"total_state_change\":0,\"last_ok\":0,\"occurrences\":0,\"occurrences_watermark\":0,\"output_metric_format\":\"\",\"output_metric_handlers\":null,\"env_vars\":null,\"metadata\":{\"name\":\"test-check\",\"namespace\":\"default\"}},\"metadata\":{}}","time":"2019-08-08T20:42:30Z","type":"event"}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ for check hook execution.
 Protobuf will be used for serialization/deserialization unless indicated by the
 backend to use JSON.
 - Added `sensuctl dump` to dump resources to a file or STDOUT.
+- Added `--no-implied-entities` flag to backend to allow events containing
+proxy entities that do not exist to be dropped. This is useful for setups
+that sync entities from an external source-of-truth and don't want to
+allow the entity list in Sensu to deviate from it.
 
 ### Changed
 - The project now uses Go modules instead of dep for dependency management.
--- a/backend/agentd/entity.go
+++ b/backend/agentd/entity.go
@@ -7,6 +7,12 @@ import (
 	"github.com/sensu/sensu-go/types"
 )
 
+type ErrImpliedEntity string
+
+func (e ErrImpliedEntity) Error() string {
+	return fmt.Sprintf("%s does not exist and implied proxy entities are disabled", string(e))
+}
+
 // addEntitySubscription appends the entity subscription (using the format
 // "entity:entityName") to the subscriptions of an entity
 func addEntitySubscription(entityName string, subscriptions []string) []string {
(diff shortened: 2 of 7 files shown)


roganartu:
This was pretty quick to throw together (maybe a couple hours?). I'm not tied to any aspect of the implementation, I mostly did this as a PoC.

I couldn't see any other points where entities were being created for events received, but if there are any let me know and I can update the PR.

portertech:
A CLI and Environment variable for this configuration is not ideal, as it creates an opportunity for multiple Backends to behave differently.

Ideally, I would prefer we approach this problem through Agent session RBAC.

@palourde @echlebek WDYT?

palourde:
I agree that a configuration attribute at the backend level could lead to unexpected behaviors in terms of UX.

Using RBAC, we could indeed easily verify if the agent's user (which is stored within the context) is authorized to create an entity here: https://github.com/sensu/sensu-go/blob/038ea9ce14dd8497da9a482e5848b0d50746680c/backend/agentd/session.go#L349

roganartu:
I didn't even consider using RBAC for this, and I agree that it would make this a more robust solution.

I will rework this hopefully this week.

palourde:
I don't believe we will go ahead and merge this PR in this form so I'll go ahead and close it for now.

Issue #3178, opened 2019-08-01
Checks are scheduled for deleted entities when using splayed proxy checks


roganartu:
A race condition exists in the scheduling of proxy checks, causing them to sometimes be scheduled for entities that have been deleted.

## Expected Behavior
Checks should not be scheduled for an entity after it has been deleted.

## Current Behavior
Checks are sometimes scheduled for deleted entities. The frequency this occurs is proportional to the interval length, number of entities, and the splay percentage used. It can theoretically occur even with a splay percentage of 0%, but it is far easier to observe with a long interval and high splay percentage.

## Possible Solution
The root cause is that Sensu operates off a cached view of entities when scheduling proxy checks: https://github.com/sensu/sensu-go/blob/27ad8a3db009abfee339de2ad649a7f74266291f/backend/schedulerd/executor.go#L248-L257

The `time.Sleep(splay)` is why this is easier to observe with a higher splay percentage.

Perhaps an easy broad solution is to drop events (maybe with a log message) if they are being submitted by an agent from a scheduled execution and the entity does not exist. I'm not sure what field(s) would differentiate these events from ones submitted via the agent `POST /events` endpoint, for which one might want a new entity created.

Another approach that doesn't remove the race entirely would be to check whether the entity still exists before pushing the check request into the message queue.
https://github.com/sensu/sensu-go/blob/27ad8a3db009abfee339de2ad649a7f74266291f/backend/schedulerd/executor.go#L58-L84

Even with proper resource-locking on the entity above this can still occur anyway, which is why I suggested the broad event-dropping mitigation above. Consider the following order of events:
1. Sensu pushes check onto queue
2. Agent picks up check and begins executing it
3. Entity is deleted in Sensu
4. Agent finishes executing check and submits result to Sensu
5. Sensu re-creates the deleted entity

This is clearly undesirable, but isn't fixed by just addressing the race in `schedulerd` with a pre-publish resource check or lock.

## Steps to Reproduce (for bugs)
1. Create two entities
2. Create a proxy check that matches both entities with a large splay percentage and an interval long enough to allow you to perform a manual action (something like 1-2 mins should suffice). This bug will occur no matter what splay percentage or interval you choose, but it is easier to demonstrate with large values as it exacerbates the race condition
3. Watch the (debug level) backend logs for `"component": schedulerd` entries
4. When schedulerd schedules a check for one of the entities, delete the other one with `sensuctl entity delete <name>`
[...]

PR #3195, opened 2019-08-12
Fix deleted proxy entities

roganartu:
## What is this change?

Proposed fix for #3178.

There is a race condition between the scheduling of a check and the processing of that result, where if an entity is deleted before the result is received then that entity will be minimally recreated by the backend.

Yes, added.

Yes, see complications section.

This was surprisingly complicated. I initially started drafting a fix for this after submitting #3192 because it's a similar problem and the solution appeared relatively simple.

I quickly discovered that the agent is quite overzealous in forcing its entity object on all events it sends, which made it effectively impossible for the backend to differentiate between the different types of events it is receiving (specifically: agent event, proxy check submitted via agent API, proxy check scheduled by backend).

I opted to utilise existing fields that were otherwise kind of ignored in order to differentiate these event types. I couldn't see a way to achieve this another way without adding a new field to the event definition. If the team prefers to add a new field, that should be quite easy to apply to this PR too, I just didn't want to make that call myself.

The three possible states result in an entity embedded in the submitted event as follows:
- agent event (eg keepalive or other non-proxy event): agent entity, as previously
- proxy check result submitted via agent `/events` API: if entity is nil, a blank one is added to the event with the agent namespace. If event namespace is empty it is set to agent namespace, and `Event.EntityClass` is set to `"proxy"`
- proxy check scheduled by backend: a blank entity is added to the event with the agent namespace, the proxy entity name, and `Event.EntityClass` set to `"proxy"`.

The backed is then able to use these expected field values to allow the following behaviour:
- if `Event.Entity.Name == store(entity)` event is ok as entity wasn't deleted
- if `Event.Entity.Name == "" and store(entity) == ""` then entity should be created (this only happens for events submitted via the agent `/events` API)
- if `Event.Entity.Name != store(entity)` event is malformed (ie: because the `Event.Entity.ProxyEntityName != Event.Entity.Name`, or the store returned the incorrect entity). This is logged and dropped
- if `Event.Entity.Name != "" and store(entity) == nil` then the entity was deleted after the check was scheduled. Event is logged and dropped.

My main concern with this is that it may be confusing to users who are using the agent `/events` API endpoint, however that's pretty advanced usage in general I think, so documenting the above behaviour may be sufficient to address that.

I didn't update docs, but they should be updated to include the above description if this approach ends up being used.

Added some tests.

Also did some adhoc testing with a modified `docker-compose.yml`.
Steps:
1. Remove all but `backend1` and `agent1` from `docker-compose.yml` to simplify testing
2. Run `docker-compose up`
3. Run `sensuctl configure`
4. Run `sensuctl create -f fixture.txt` using attached file: [fixture.txt](https://github.com/sensu/sensu-go/files/3493252/fixture.txt)
5. Wait for the first `foobar` check to be scheduled by watching logs
6. Immediately after the check is scheduled by the backend, run `sensuctl entity delete foo`
7. Observe the following error appear in the `backend` logs:
```
backend1_1  | {"component":"agentd","error":"foo was deleted after this event was scheduled, dropping event","level":"error","msg":"error handling message","payload":"\u0008\ufffd\ufffd\ufffd\ufffd\u0005\u0012\u001d\n\u0005proxy\u001a\u00022\u0000:\u0000r\u000e\n\u0003foo\u0012\u0007default\u001a4(\u0001j\u0003foo\ufffd\u0001\ufffd\ufffd\ufffd\ufffd\u0005\ufffd\u0001\tsomething\ufffd\u0001\u0002\ufffd\u0002\u0014\n\tcheck-foo\u0012\u0007default*\u0000","time":"2019-08-11T19:19:15Z","type":"event"}
```
8. Observe checks continue to be scheduled and events successfully received for entity `bar`

Something like this should probably be added to your existing integration/e2e test suite.

### DIFF
--- a/agent/check_handler.go
+++ b/agent/check_handler.go
@@ -64,7 +64,32 @@ func (a *Agent) handleCheck(ctx context.Context, payload []byte) error {
 
 	logger.Info("scheduling check execution: ", checkConfig.Name)
 
-	entity := a.getAgentEntity()
+	// In order to protect against processing results for since-deleted entities,
+	// the backend needs to be aware of the correct entity.
+	// If this is a proxy check, the entity should not be the agent but instead
+	// the entity the check is actually for.
+	agentEntity := a.getAgentEntity()
+	var entity *corev2.Entity
+	fmt.Printf("proxy_entity_name: %q\n", request.Config.ProxyEntityName)
+	if request.Config.ProxyEntityName != "" {
+		// Since we don't actually have access to the real entity here, we need to
+		// make a placeholder. The backend will replace it with the entity in the
+		// store that it fetches using Event.Check.ProxyEntityName, so the full details
+		// aren't necessary. The backend specifically requires:
+		//   - namespace (to find the correct entity)
+		//   - name (to allow detection of events submitted for deleted entities)
+		// Additionally, Agent.executeCheck looks at the entity class to see whether
+		// it is a proxy entity.
+		entity = &corev2.Entity{
+			ObjectMeta: corev2.ObjectMeta{
+				Namespace: agentEntity.Namespace,
+				Name:      request.Config.ProxyEntityName,
+			},
+			EntityClass: "proxy",
+		}
+	} else {
+		entity = agentEntity
+	}
 	go a.executeCheck(ctx, request, entity)
 
 	return nil
@@ -225,7 +250,18 @@ func (a *Agent) executeCheck(ctx context.Context, request *corev2.CheckRequest, 
 	event.Check.Duration = checkExec.Duration
 	event.Check.Status = uint32(checkExec.Status)
 
-	event.Entity = a.getAgentEntity()
+	// Proxy checks depend on the entity in the event being the entity the
+	// check was issued to, not the agent. Non-proxy checks depend on it
+	// being the agent.
+	// This entity is overwritten by the backend (after some validation is performed)
+	// so it's not absolutely necessary that it be correct. The fields that are
+	// primarily depended on for differentiation are the namespace and the name.
+	switch entity.EntityClass {
+	case "proxy":
+		event.Entity = entity
+	default:
+		event.Entity = a.getAgentEntity()
+	}
 	event.Timestamp = time.Now().Unix()
 
 	if len(checkHooks) != 0 {
--- a/api/core/v2/entity.go
+++ b/api/core/v2/entity.go
@@ -131,6 +131,14 @@ func FixtureEntity(name string) *Entity {
 	}
 }
 
+// FixtureProxyEntity returns a testing fixture for an proxy Entity object.
+func FixtureProxyEntity(name string) *Entity {
+	e := FixtureEntity(name)
+	e.EntityClass = "proxy"
+	e.ObjectMeta.Name = ""
+	return e
+}
+
 //
 // Sorting
 
--- a/testing/testutil/errors.go
+++ b/testing/testutil/errors.go
@@ -1,6 +1,9 @@
 package testutil
 
-import "testing"
+import (
+	"reflect"
+	"testing"
+)
 
 // CompareError ensures that the provided err is expected
 func CompareError(err error, expected bool, t *testing.T) {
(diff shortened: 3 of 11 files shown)


testing/testutil/errors.go
```diff
@@ -12,3 +15,20 @@ func CompareError(err error, expected bool, t *testing.T) {
...
+	t.Helper()
+
+	if expected != nil {
+		if err == nil {
+			t.Fatal("expected error, got none")
+		}
+		if reflect.TypeOf(err) != expected {
+			t.Fatalf("expected error of type %s, got %T", expected.Name(), err)
+		}
+	} else if expected == nil && err != nil {
```
golangcibot: nilness: tautological condition: nil == nil (from `govet`)

roganartu: good bot, fixed.

api/core/v2/entity.go
```diff
@@ -131,6 +131,14 @@ func FixtureEntity(name string) *Entity {
 	}
 }
 
+// FixtureProxyEntity returns a testing fixture for an proxy Entity object.
+func FixtureProxyEntity(name string) *Entity {
+	e := FixtureEntity(name)
+	e.EntityClass = "proxy"
+	e.ObjectMeta.Name = ""
```
echlebek: What's the rationale of zeroing out the proxy entity's name?

types/aliases.go
```diff
@@ -193,6 +193,7 @@ var (
 	FixtureCheckRequest       = v2.FixtureCheckRequest
 	FixtureCheckConfig        = v2.FixtureCheckConfig
 	FixtureCheck              = v2.FixtureCheck
+	FixtureProxyCheck         = v2.FixtureProxyCheck
```
echlebek: FWIW it's not necessary to add these aliases - they were only added for compatibility and gradual code repair, as the things in core/v2 were moved there from the types package.

agent/check_handler.go
```diff
@@ -64,7 +64,32 @@ func (a *Agent) handleCheck(ctx context.Context, payload []byte) error {
...
 	logger.Info("scheduling check execution: ", checkConfig.Name)
 
-	entity := a.getAgentEntity()
+	// In order to protect against processing results for since-deleted entities,
+	// the backend needs to be aware of the correct entity.
+	// If this is a proxy check, the entity should not be the agent but instead
+	// the entity the check is actually for.
+	agentEntity := a.getAgentEntity()
+	var entity *corev2.Entity
+	fmt.Printf("proxy_entity_name: %q\n", request.Config.ProxyEntityName)
```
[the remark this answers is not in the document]
roganartu: whoops

api/core/v2/entity.go
roganartu: The combination of `EntityClass` and `Entity.ObjectMeta.Name` are what allowed this change to differentiate between the three different types of events the agent forwards to the backend without having to add a new field to the event object:

1. Non-proxy checks (`EntityClass: "host"` and `Entity.ObjectMeta.Name: "foo-agent")
2. Proxy check results submitted via API (`EntityClass: "proxy"` and `Entity.ObjectMeta.Name: "")
3. Proxy checks scheduled by backend (`EntityClass: "proxy" and `EntityObjectMeta.Name: "<proxy-entity-name>"`)

The backend can then, after looking up the entity in the store based on `Event.Check.ProxyEntityName` as it previously already did, compare the entity in the store to the entity name in the submitted event. If there's no entity in the store, but `Event.Entity.ObjectMeta.Name` is non-empty, then the backend knows that the entity was deleted and the event should be dropped.

Perhaps this fixture method could be named better. If you look at the tests you'll see that they use either `FixtureEntity` or `FixtureProxyEntity` depending on which of the above cases they are emulating.

portertech:
Hey @roganartu 👋 I'm hoping [cluster configuration to disable automatic proxy entity creation](https://github.com/sensu/sensu-go/issues/3593) would accomplish what you're after (and more). If so, I would be happy to have Sensu Engineering make it happen (and we can close this).

roganartu:
@portertech yeah setting this at the cluster level would work for us, since we have no desire to ever create implicit entities (we sync entities from an external source of truth into Sensu via a cron). Thanks.

PR #3200, opened 2019-08-13
Pass populated event object to ExecuteHooks

### LINKED ISSUE #3173: Check hooks do not have access to check details
## Expected Behavior

Particularly for checks targeting proxy entities, check hooks may need to use information from the entity or check to operate. To use the website example from the documentation, a check hook that collects extra information probably needs to know the website that was checked and triggered the hook.

I expect this information (e.g. the event object) to be available either on STDIN ([docs](https://docs.sensu.io/sensu-go/5.11/reference/hooks/#how-do-hooks-work)) or as template variables in the command, similar to how check commands are templated.

## Current Behavior

Neither STDIN nor the available template values contain information about the check that triggered the hook. Hooks with `stdin: true` do receive an event, but it is empty:

```json
{"check":{"handlers":[],"high_flap_threshold":0,"interval":0,"low_flap_threshold":0,"publish":false,"runtime_assets":null,"subscriptions":[],"proxy_entity_name":"","check_hooks":null,"stdin":false,"subdue":null,"ttl":0,"timeout":0,"round_robin":false,"executed":0,"history":null,"issued":0,"output":"","status":0,"total_state_change":0,"last_ok":0,"occurrences":0,"occurrences_watermark":0,"output_metric_format":"","output_metric_handlers":null,"env_vars":null,"metadata":{}},"metadata":{}}
```

Using a template string like `{{ .labels.job }}` which works for check commands produces this error:

```
unmatched token: template: :1:117: executing "" at <.labels.job>: map has no entry for key "labels"
```

The code supports this behavior as well; in `executeHook` an empty event is created:

https://github.com/sensu/sensu-go/blob/78aec85ac351bbcfc723cc092ce1816ea56c8a36/agent/hook.go#L55-L57

And then never referenced again until it is passed to the hook. When creating the template environment, only the agent information is available:

https://github.com/sensu/sensu-go/blob/78aec85ac351bbcfc723cc092ce1816ea56c8a36/agent/hook.go#L170-L180

## Possible Solution

When `ExecuteHook` is called during check execution, a mostly complete check event appears to be available. This could be passed in to `ExecuteHook` and used to populate the content of STDIN and/or the template environment.

## Steps to Reproduce

1. `sensuctl check create test-hook -c "/bin/sh -c 'echo fail && exit 1'" --interval 60 --publish --subscriptions entity:sensu-agent`
2. `sensuctl hook create cat -c "cat" --stdin`
3. `sensuctl check set-hooks test-hook -k cat --type 1`
4. `sensuctl event info sensu-agent test-hook --format=json | jq .check.hooks[0].output`

## Context

I've written a check using proxy entities to monitor batch jobs in an external system. When a job fails, I want to collect the logs from the failing job in a check hook, but I need to know the ID of the job (a label on the entity) to query the API.

## Your Environment

* Sensu version used: `version 5.11.1#39d2bf2, build 39d2bf22c416f9a78f22c4817dc74ab62d7f4d63, built 2019-07-18T13:31:59Z` for `sensuctl`, `sensu-agent`, and `sensu-backend`.
* Installation method: Docker
* Operating System: Ubuntu 16.04

bluekeyes:
## What is this change?

Previously, hooks that expected data on stdin received an empty event.  Now they are passed the event object constructed as part of check execution.

Fixes #3173, which also has more details on the motivation.

This fixes the behavior to match the documentation, so I don't think any changes are needed.

I haven't verified this in a real environment yet and the existing tests don't fully cover the functionality, so it is only partially tested.

### DIFF
--- a/agent/check_handler.go
+++ b/agent/check_handler.go
@@ -228,10 +228,6 @@ func (a *Agent) executeCheck(ctx context.Context, request *corev2.CheckRequest, 
 	event.Entity = a.getAgentEntity()
 	event.Timestamp = time.Now().Unix()
 
-	if len(checkHooks) != 0 {
-		event.Check.Hooks = a.ExecuteHooks(ctx, request, checkExec.Status, hookAssets)
-	}
-
 	// Instantiate metrics in the event if the check is attempting to extract metrics
 	if check.OutputMetricFormat != "" || len(check.OutputMetricHandlers) != 0 {
 		event.Metrics = &corev2.Metrics{}
@@ -243,6 +239,11 @@ func (a *Agent) executeCheck(ctx context.Context, request *corev2.CheckRequest, 
 
 	if len(check.OutputMetricHandlers) != 0 {
 		event.Metrics.Handlers = check.OutputMetricHandlers
+	}
+
+	// Execute hooks after we have a completely populated event object
+	if len(checkHooks) != 0 {
+		event.Check.Hooks = a.ExecuteHooks(ctx, request, event, hookAssets)
 	}
 
 	// The check requested that we discard its output before writing back
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ queried in the `GET /events/:entity` endpoint (#3141)
 - Prevent a segmentation violation when running `sensuctl config view` without
 configuration.
 - Added entity name to the interactive sensuctl survey.
+- Check hooks with `stdin: true` now receive actual event data on STDIN instead
+  of an empty event.
 
 ### Removed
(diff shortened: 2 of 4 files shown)


agent/check_handler.go
```diff
@@ -245,6 +241,11 @@ func (a *Agent) executeCheck(ctx context.Context, request *corev2.CheckRequest,
 		event.Metrics.Handlers = check.OutputMetricHandlers
 	}
 
+	// Execute hooks after we have a completely populated event object
```
portertech: This is neat, I could apply a hook to extracted metric data 👌

Issue #3067, opened 2019-06-18
[Feature Enhancement]  Multi-build capable asset resources


jspaleta:
Currently  named assets are effectively tied to a single type of agent host, leading to highly duplicative configuration in heterogeneous environments.

Asset filter works to prevent the wrong binary from being installed on an agent, but there's no mechanism by which an agent to choose among multiple binaries.

Examples: 
1) you would like to run a statically compiled golang check on both osX and linux.
  currently you'd have to have a different named asset and checks for each target OS, in order for the asset download to work correctly.
2) you would like to use Sensu ruby-runtime check on both Alpine Linux and Centos Linux.
Both would require different named assets, and checks for each OS.

## Expected Behavior
Have all possible asset builds in a single Asset resource, so named asset resources can be re-used in check definitions for agents to use regardless of agent's host OS/platform/arch
Ex: 
asset named "sensu-plugins/sensu-plugins-cpu-checks" should list all asset binaries+filters, and should be usable across all checks and subscriptions. The agent should be able to determine which binary to use based on the filters.

## Current Behavior
Difficult to impossible to re-use a named asset resource in a heterogeneous environment.  The result is effectively building up check,asset,subscription naming schemes for different platforms.

## Strawman:
introduce builds list  in Asset spec
each build is a hash of key/value pairs:
```
  url: (required) the location of the binary
  sha512: (required) the sha512sum
  priority: (optional) numeric used by asset manager to help choose if multiple build pass filter test. should be unique for each build, defaults to 0
  name:  (optional) human readable name for the build, defaults to null
  filters: (optional) the list of entity filters that must match, defaults to null
```
See Gist: https://gist.github.com/jspaleta/01c9fa0eb2b22ace25ae03c01e47b368

###  Notes:
* To achieve deterministic behavior when multiple builds pass filter matching, priority should be unique for each build. highest priority build with all filter conditions satisfied wins. 

* This can be implemented as an extension to existing v2 asset api, if in addition the high level spec attributes "name" and "priority" are added as well. Asset agent could be taught to fold the build outside the build list into its build list representation as an additional member.

* Could be implemented as v3 api and drop the high level sha512, url, filters keys and require all information to exist in the builds list.

### Key benefits
* Extends Asset resource in a backward compatible way, so existing asset configs without build list continue to work.
* Allow to encode a single named asset configuration in a namespace for all asset binaries
* Simplifies Bonsai UI, no need for platform/arch modular dialog.
[...]

PR #3206, opened 2019-08-15
Record GOARM value at build time

amdprophet:
## What is this change?

Stores the GOARM value (if set) in the entity system info.

The ability to filter on which ARM version an asset is built for is a necessary for installations with varying versions of ARM processors.

A prerequisite for https://github.com/sensu/sensu-go/issues/3067.

Yes.

I built the backend, agent, & sensuctl and inspected an agent entity running on this code:

Screen Shot 2019-08-15 at 2 30 38 PM


portertech:
Does it make any sense for a more generalized field? e.g. arch_version?

amdprophet:
@portertech I think that makes sense. I'll make the change.

echlebek:
@portertech IMO no, because there is no corresponding Go compiler flag for those arch versions. I believe the Go runtime uses feature detection for Intel/AMD, and I'm not aware of any other arch specific compiler flags for other non-x86.

amdprophet:
@echlebek is there a potential for something similar to happen with other architectures down the road? e.g. mips, sparc, powerpc?

echlebek:
I can't predict the future, and that's why I'd prefer to add a distinct field if and when that scenario arises. :)

amdprophet:
`GOMIPS` exists but is used to control whether or not floats are soft/hard. I'll stick with `ARMVersion` for now.

portertech:
@amdprophet @echlebek just feels strange to have arch: i386, arm_version: "".

echlebek:
Oh, yeah. It should have `omitempty` specified so that it doesn't render for its zero value

echlebek:
@portertech the generated code indicates that `omitempty` is set, so the situation you describe will never arise.

Issue #3139, opened 2019-07-12
Silence expire_on_resolve did not clear event on recovery


roganartu:
The `expire_on_resolve` property of a Silenced entry is intended to cause the silence to automatically be deleted when the event resolves, however I have encountered an event in an OK state with an open Silence that has `expire_on_resolve` set to `True`.

## Expected Behavior
A Silence with `expire_on_resolve` set to `True` should be deleted when the event status is `0`.

## Current Behavior
I have an event with a status of `0` that also has a active Silence with `expire_on_resolve` set to `True`.

Here are the details. I changed the hostname, check name, and output to placeholders for simplicity.
```
╰ sensuctl event info --format table host foo
=== host - foo
Entity:      host
Check:       foo
Output:      some output
Status:      0
History:     3,2,2,2,2,3,2,2,2,2,2,3,2,2,2,2,2,0,0,0,0
Silenced:    true
Silenced By: entity:host:foo
Timestamp:   2019-07-12 16:00:04 +0000 UTC
```

```
╰ sensuctl silenced info --check foo  --subscription entity:host    
{
  "metadata": {
    "name": "entity:host:foo",
    "namespace": "default",
  },
  "expire": -1,
  "expire_on_resolve": true,
  "creator": "admin",
  "check": "foo",
  "reason": "some reason",
  "subscription": "entity:host",
  "begin": 1562709459
}
```
Note the `begin` timestamp is several days ago, while the check only recovered earlier today. I would have expected this Silence to have been deleted when the check transitioned from `status == 2` to `status == 0` in the history above.

## Steps to Reproduce (for bugs)
I have not attempted to reproduce with a new silence and a fake check so I am unsure if this is a bug for all Silenced entries or just some kind of race condition.

There are no entries in the logs matching the term `silence` that aren't from the `apid` component, so I'm not sure where to start looking at this

## Possible Solution
I'm not sure why this is occuring from a quick look over the source here:

https://github.com/sensu/sensu-go/blob/27ad8a3db009abfee339de2ad649a7f74266291f/backend/eventd/silenced.go#L54-L80

However, I would personally argue that if it's a transient error then only applying `expire_on_resolve` to DOWN -> OK transitions makes it confusing to users and less robust. An `expire_on_resolve` silence for an OK alert doesn't really make any sense to me. I would imagine that it was added with the intent of allowing users to create an "ack"-like silence, that only persists as long as the event is in a down state.

[...]

PR #3215, opened 2019-08-16
Fix silence bug on event resolution

nikkixdev:

## What is this change?

Fixes a bug where silences would not expire on event resolution. This is because we were attempting to determine if an event - with no history - was a resolution. The event needed to be merged with the stored event first.

Closes https://github.com/sensu/sensu-go/issues/3139

Yas.

Nah.

Nah.

We good.

https://github.com/sensu/sensu-go-qa-crucible/pull/70

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ configuration.
 - Added entity name to the interactive sensuctl survey.
 - Check hooks with `stdin: true` now receive actual event data on STDIN instead
   of an empty event.
+- Fixed a bug where silences would not expire on event resolution.
 
 ### Removed
 - Removed encoded protobuf payloads from log messages (when decoded, they can reveal
--- a/backend/eventd/eventd.go
+++ b/backend/eventd/eventd.go
@@ -219,13 +219,14 @@ func (e *Eventd) handleMessage(msg interface{}) error {
 	// Add any silenced subscriptions to the event
 	getSilenced(ctx, event, e.silencedCache)
 
-	// Handle expire on resolve silenced entries
-	if err := handleExpireOnResolveEntries(ctx, event, e.store); err != nil {
+	// Merge the new event with the stored event if a match is found
+	event, prevEvent, err := e.eventStore.UpdateEvent(ctx, event)
+	if err != nil {
 		return err
 	}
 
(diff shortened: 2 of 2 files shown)


nikkixdev:
@echlebek I'm not sure I understand how [L77](https://github.com/sensu/sensu-go/blob/27ad8a3db009abfee339de2ad649a7f74266291f/backend/eventd/silenced.go#L77) is related. Silenced entries are only retained and saved within the check when `ExpireOnResolve` is false. Debug statements informed me that the function was returning far earlier than this, but open to hearing why you think that might not be the case.

The [event being passed](https://github.com/sensu/sensu-go/blob/88f30484eb2f68eeb151e53baaf68bd2ac095ac0/backend/eventd/eventd.go#L223) into [handleExpireOnResolveEntries](https://github.com/sensu/sensu-go/blob/27ad8a3db009abfee339de2ad649a7f74266291f/backend/eventd/silenced.go#L54) only has a single check execution history, thus always [returning immediately](https://github.com/sensu/sensu-go/blob/27ad8a3db009abfee339de2ad649a7f74266291f/backend/eventd/silenced.go#L57) since it is not a resolution.

echlebek:
Yes - and without line 77 the side effect of the function would never be persisted, since it would never have been passed to the store to be updated.

nikkixdev:
@echlebek ah yes now I see what you mean. An overly simple solution would be to update the final event in the store after L77, however that would constitute an additional write for all resolution events. I'm not sure that would be totally terrible. What are your thoughts?

echlebek:
I think it's probably necessary. We could also check to see if the updated list of silences differs, and only update if it does.

echlebek:
Also, it would be preferable to include a test that would fail without the bug fix! :pray:

nikkixdev:
I'll probably have to write a new function to update the event without adding an occurrence or merging check history: https://github.com/sensu/sensu-go/blob/6767ea134ca748f913833e5640a573419aa07233/backend/store/etcd/event_store.go#L225-L241

nikkixdev:
@echlebek updated the logic within the store, great catch here! I'm assuming additional work is required here for postgres? I can file an issue if you'd like, not sure exactly what we'd need to do.

PR #3222, opened 2019-08-20
Asset builds support

### LINKED ISSUE #3067: [Feature Enhancement]  Multi-build capable asset resources

amdprophet:
## What is this change?

Adds a new concept of 0 or more builds within an asset.

Fixes #3067.

Yes.

See https://github.com/sensu/sensu-docs/issues/1685.

I have verified that existing asset definitions work along with newer asset definitions, and that the correct build asset was pulled in. For example:

```yaml
---
type: Asset
api_version: core/v2
metadata:
  name: sensu-go-cpu-check
  namespace: default
  labels: {}
  annotations: {}
spec:
  builds:
    - url: https://assets.bonsai.sensu.io/981307deb10ebf1f1433a80da5504c3c53d5c44f/sensu-go-cpu-check_0.0.3_linux_amd64.tar.gz
      sha512: 487ab34b37da8ce76d2657b62d37b35fbbb240c3546dd463fa0c37dc58a72b786ef0ca396a0a12c8d006ac7fa21923e0e9ae63419a4d56aec41fccb574c1a5d3
      filters:
        - entity.system.os == 'linux'
        - entity.system.arch == 'amd64'
    - url: https://assets.bonsai.sensu.io/981307deb10ebf1f1433a80da5504c3c53d5c44f/sensu-go-cpu-check_0.0.3_linux_armv7.tar.gz
      sha512: 70df8b7e9aa36cf942b972e1781af04815fa560441fcdea1d1538374066a4603fc5566737bfd6c7ffa18314edb858a9f93330a57d430deeb7fd6f75670a8c68b
      filters:
        - entity.system.os == 'linux'
        - entity.system.arch == 'arm'
        - entity.system.arm_version == 7
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ backend to use JSON.
 - Added `event.check.name` as a supported field selector.
 - [Web] Added timeline chart to event details view.
 - Added `entity.system.arm_version` to record the value of `GOARM` at compile time.
+- Added `builds` field to the `Asset` type to allow assets to specify different
+URLs for each platform/architecture/arch_version.
 
 ### Changed
 - The project now uses Go modules instead of dep for dependency management.
--- a/api/core/v2/asset.go
+++ b/api/core/v2/asset.go
@@ -45,6 +45,44 @@ func (a *Asset) Validate() error {
 		return errors.New("namespace cannot be empty")
 	}
 
+	if len(a.Builds) == 0 {
+		if a.Sha512 == "" {
+			return errors.New("SHA-512 checksum cannot be empty")
+		}
+
+		if len(a.Sha512) < 128 {
+			return errors.New("SHA-512 checksum must be at least 128 characters")
+		}
+
+		if a.URL == "" {
+			return errors.New("URL cannot be empty")
+		}
+
+		if a.URL != "" {
+			u, err := url.Parse(a.URL)
+			if err != nil {
+				return errors.New("invalid URL provided")
+			}
+
+			if u.Scheme != "https" && u.Scheme != "http" {
+				return errors.New("URL must be HTTP or HTTPS")
+			}
+		}
+
+		return js.ParseExpressions(a.Filters)
+	} else {
+		for _, build := range a.Builds {
+			if err := build.Validate(); err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+// Validate returns an error if the asset contains invalid values.
+func (a *AssetBuild) Validate() error {
 	if a.Sha512 == "" {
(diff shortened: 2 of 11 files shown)


amdprophet:
It should be unit testable. I'll look into adding some tests.

PR #3224, opened 2019-08-21
OIDC exception in RefreshToken middleware

ccressent:
## What is this change?

This change makes it so the `RefreshToken` middleware is able to deal with the new OIDC authentication provider.

Necessary for the new OIDC authentication provider to be able to transparently refresh access tokens.

Yes, for the new `ProviderType` field added to `AuthProviderClaims`.

No.

Ideally we wouldn't need to even mention OIDC, an enterprise feature, in `sensu-go`. We just couldn't find any other way to make it work without _massive_ refactors across the whole authentication and token system.

No, this doesn't change anything for existing users and features.

- Existing unit and integration tests
- `sensuctl` is able to automatically refresh tokens, even when using OIDC as the authn provider


backend/apid/middlewares/refresh_token.go
```diff
@@ -44,6 +44,19 @@ func (m RefreshToken) Then(next http.Handler) http.Handler {
...
 
+		// Retrieve the claims for the access token
+		accessClaims, err := jwt.GetClaims(accessToken)
+		if err != nil {
+			logger.WithError(err).Error("could not parse the access token claims")
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+
+		if accessClaims.Provider.ProviderType == "oidc" {
```
echlebek: Should "oidc" be a constant, and should we accept caps spelling?

ccressent: The actual constant is defined in the enterprise OIDC provider, here:
https://github.com/sensu/sensu-enterprise-go/pull/628/files#diff-ef44a838326105c4d512ea9ba2598a2bR12
But we obviously cannot import enterprise into core!

I suppose we can add a constant at the top of this `refresh_token.go` file? Not sure it would make a big difference since it still wouldn't track the constant that's defined in enterprise.

Cap spelling is not an issue; it should always be lowercase as defined by the constant.

PR #3230, opened 2019-08-22
Add node resolver for event filters

jamesdphillips:

## What is this change?

Registers the GraphQL's EventFilterType schema with the node registry and adds a resolver.

Allows reverse lookup of types and the ability to query filter through the node interface.

```graphql
node(id: "event-filter-global-id") {
  id
  ... on EventFilter {
    action
    statements
  }
}
```

---

I suspect that updating `graphql-go/graphql` may have exposed this error; previous version may have been more permissive.

### DIFF
--- a/backend/apid/graphql/node.go
+++ b/backend/apid/graphql/node.go
@@ -3,6 +3,7 @@ package graphql
 import (
 	"context"
 	"errors"
+	"fmt"
 
 	"github.com/sensu/sensu-go/backend/apid/graphql/globalid"
 	"github.com/sensu/sensu-go/backend/apid/graphql/relay"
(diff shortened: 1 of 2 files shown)


backend/apid/graphql/node.go
```diff
@@ -47,6 +49,11 @@ func (r *nodeResolver) FindType(i interface{}) *graphql.Type {
 
 	components := translator.Encode(i)
 	resolver := r.register.Lookup(components)
+	if resolver == nil {
+		logger := logger.WithField("translator", fmt.Sprintf("%#v", translator))
+		logger.Warn("unable to find node resolver for type")
+		return nil
```
jamesdphillips: maybe it would be better to panic. 🤔

echlebek: At least log at the error level :)

PR #3247, opened 2019-08-30
[Feature] sensuctl asset add & sensuctl asset outdated

amdprophet:
## What is this change?

Adds two new commands to `sensuctl`:
`sensuctl asset add [NAME[:version]]` will fetch an asset from Bonsai. If a version isn't specified, the latest version will be fetched.

`sensuctl asset outdated` will check if updated versions of any assets installed via Bonsai exist.

To simplify the workflow of installing & managing assets from Bonsai.

Yes.

A few - mainly that adding setter methods to all of our types is a bit tedious. We should consider adding support for `setters` to `gogoproto.face`. 

New documentation will be required. I'll update this issue once I've created a sensu-docs issue.

A lot of local testing. Further testing from the reviewers is appreciated.


calebhailey:
Awesome! This looks great! 

A few questions / bits of feedback (my fault for not queuing up a proper spec doc to catch this up front): 

- Can we change the "outdated" command to `sensuctl asset upgrade <asset_name>`; this could also be extended with optional command arguments for dry runs (i.e. `--dry-run`: only check if new versions are available; don't actually upgrade the assets), or for dealing with deprecated assets that redirect to a replacement (e.g. `--upgrade-deprecated`)

- If the user specifies an asset version; e.g. `sensuctl asset add sensu/sensu-slack-handler:1.0.3` can we include the version in the asset name? This may satisfy certain customer requests we've had for "pinning" asset versions. Assets with pinned versions should not be upgraded via `sensuctl asset upgrade`.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Added
+- sensuctl asset add (fetches & adds assets from Bonsai).
+- sensuctl asset oudated (checks for newer versions of assets from Bonsai).
```
palourde: s/oudated/outdated/

palourde:
@amdprophet This resource also needs to be updated 😬 
https://github.com/sensu/sensu-go/blob/master/backend/store/etcd/generic_object_test.go

portertech:
I consider `sensuctl asset upgrade` a separate feature and deliverable, let's not add 😂 that on here, but it's something we should consider in the near future.

@palourde I'm concerned `sensuctl asset list` already has too many columns and already fills the row w/ too much. Perhaps we can compress what's there and work in the outdated info? WDYT?

amdprophet:
Screen Shot 2019-09-05 at 10 04 45 AM
Screen Shot 2019-09-05 at 10 05 04 AM
Screen Shot 2019-09-05 at 10 05 45 AM
Screen Shot 2019-09-05 at 10 05 51 AM
Screen Shot 2019-08-30 at 3 56 48 PM

portertech:
Let's keep to the `sensuctl asset outdated` subcommand implementation, however, we need to add json output format support for consistency.

amdprophet:
Updated the `asset outdated` command to use a new `OutdatedBonsaiAsset` type that can be rendered to a table view, json, yaml, etc.

Screen Shot 2019-09-05 at 1 17 39 PM

cli/commands/asset/add.go
```diff
@@ -0,0 +1,175 @@
...
+	"github.com/sensu/sensu-go/cli"
+	"github.com/spf13/cobra"
+
+	goversion "github.com/hashicorp/go-version"
+	corev2 "github.com/sensu/sensu-go/api/core/v2"
+	"github.com/sensu/sensu-go/cli/commands/create"
+	"github.com/sensu/sensu-go/types"
+)
+
+var rename string
```
nikkixdev: Whats this lil guy doing here?

cli/commands/asset/add.go
```diff
@@ -0,0 +1,175 @@
...
+
+		fmt.Printf("added asset: %s/%s:%s\n", bAsset.Namespace, bAsset.Name, version)
+		return nil
+	}
+}
+
+// ParseResources is a rather heroic function that will parse any number of valid
+// JSON resources. Since it attempts to be intelligent, it likely contains bugs.
+//
+// TODO(amdprophet): This is a modified copy of the ParseResources function
```
nikkixdev: Out of curiosity, why did we have to modify this function?

cli/commands/asset/add.go
amdprophet: This is the variable that the `--rename` flag value is stored in: https://github.com/sensu/sensu-go/blob/bf669448d2a82038ed885a9445e21c77824421f5/cli/commands/asset/add.go#L32.

cli/commands/asset/outdated.go
```diff
@@ -0,0 +1,165 @@
...
+// that have been added from Bonsai.
+func OutdatedCommand(cli *cli.SensuCli) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "outdated",
+		Short: "lists any assets installed from Bonsai that have newer versions available",
+		RunE:  outdatedCommandExecute(cli),
+	}
+
+	helpers.AddFormatFlag(cmd.Flags())
+	helpers.AddAllNamespace(cmd.Flags())
```
nikkixdev: Are these flags all necessary here for this command? `outdated` only refers to bonsai, not sensu resources right? So the namespace, field selector, label selector, and chunk size flags might not be needed.

cli/commands/asset/add.go
amdprophet: `ParseResources` in the `create` command package takes in an `io.Reader` to read a file. I had to create a modified copy of the function to accept a string (of JSON). We also only need to support JSON format in this case.

cli/commands/asset/add.go
nikkixdev: Wondering why its a global variable vs a regular one

cli/commands/asset/outdated.go
amdprophet: Outdated still has to look up assets from the Sensu API in order to find out what assets are installed and which of them are Bonsai assets. I'm unsure on what exactly the field, label & chunk size flags do so we may not need them. This was copied over from the `asset list` command.

ccressent: Conceptually, the flags for {field,label} selectors allows `sensuctl` to ask the backend to only return resources that match certain predicates and the chunk size flag is to make `sensuctl` use the API's pagination facilities to build a resource list chunk by chunk (pages) rather than all in one go.

So you are right, we don't need them here!

(technically we could make use of `--chunk-size` if we wanted to, to fetch the list of current assets from the API, but that would only be useful if one had 1000s of assets and a crappy connection. Let's not bother.)

amdprophet: Agreed - this & the `--all-namespaces` flags can be added in a future release.

cli/commands/asset/add.go
ccressent: If indeed that's exactly the same function with a `string` parameter instead of an `io.Reader`, you can avoid duplication by putting the string into something that meets that interface, eg:

```go
// This gives you a bytes.Reader, which meets, among others, the io.Reader interface
bytes.NewReader([]byte("some string"))
```

Would that be enough to solve your issue?

cli/commands/asset/add.go
amdprophet: This is the pattern that the Cobra documentation shows. I believe our other command packages handle flags differently as they're coupled with our Viper config. If you know of a better way of handling this, please share!

PR #3262, opened 2019-09-06
Fix panic when no asset build filters pass

amdprophet:
## What is this change?

If an asset contains builds and none of the filters pass, the app will crash.

Panics in sensu-agent & sensu-backend are bad.

Yes.

I wrote a new test that reproduced the panic. After making the changes below, the panic no longer occurs.

### DIFF
--- a/asset/filtered_manager.go
+++ b/asset/filtered_manager.go
@@ -30,8 +30,14 @@ type filteredManager struct {
 func (f *filteredManager) Get(ctx context.Context, asset *corev2.Asset) (*RuntimeAsset, error) {
 	var filteredAsset *corev2.Asset
 
+	fields := logrus.Fields{
+		"entity":  f.entity.Name,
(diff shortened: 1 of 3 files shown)


asset/filtered_manager.go
```diff
@@ -30,8 +30,14 @@ type filteredManager struct {
...
+		"asset":   asset.Name,
+		"filters": asset.Filters,
+	}
+
 	if len(asset.Builds) > 0 {
-		fields := logrus.Fields{
+		fields = logrus.Fields{
 			"entity": f.entity.Name,
 			"asset":  asset.Name,
 		}
```
palourde: It appears we no longer need this assignation, since these fields are already defined on lines 33-37.

amdprophet: I didn't want to display the top-level asset's filters as they don't do anything in the case where an asset has builds.

PR #3264, opened 2019-09-09
Use the embedded client when etcd is embedded

### LINKED ISSUE #697: Move to using no-network etcd client in backend
We should move to using the in-memory client to the etcdserver instead of creating localhost connections to the embedded etcdserver.

https://godoc.org/github.com/coreos/etcd/etcdserver/api/v3client

### LINKED ISSUE #3263: Better test harness for etcd tests
## Expected Behavior
Tests do not fail sporadically, for the wrong reasons.

## Current Behavior
Tests can fail if there is a collision when starting up an etcd server, for parallel tests.

## Possible Solution
We could use the embedded client so that we don't need to know which port etcd binds to.

Make a sufficient number of CI builds

echlebek:
## What is this change?

This commit makes Sensu use an embedded client for etcd requests,
instead of connecing localhost. This has several useful effects:
  * Avoids the network stack for etcd requests, for embedded etcd
  * Avoids gRPC serialization, for embedded etcd
  * Simplifies integration testing

The biggest win here is that we can have the integration test etcd
servers listen on ':0', which assigns an arbitrary port instead of
one that we have to pick. This should let the OS pick a port
instead of us, which solves a long-standing CI bug where our tests
could sometimes pick the same port to listen on in parallel,
causing failure.

There should also be some performance wins from avoiding the
network stack, but this is yet to be verified.

Closes #697 
Closes #3263 

No, this is not a user-visible change.

No documentation changes necessary.

Integration testing, which I believe is sufficient here.

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -284,7 +284,7 @@ func Initialize(config *Config) (*Backend, error) {
 		EventStore:          eventStoreProxy,
 		QueueGetter:         queueGetter,
 		TLS:                 config.TLS,
-		Cluster:             clientv3.NewCluster(b.Client),
+		Cluster:             b.Client.Cluster,
 		EtcdClientTLSConfig: etcdClientTLSConfig,
 		Authenticator:       authenticator,
 		ClusterVersion:      clusterVersion,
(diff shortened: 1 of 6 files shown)


echlebek:
Reviewers may notice that I've used things like `client.Cluster` instead of `clientv3.NewCluster(client)` - this is a workaround due to a bug, filed here: https://github.com/etcd-io/etcd/issues/9621

PR #3265, opened 2019-09-09
sensuctl dump redux

### LINKED ISSUE #3236: sensuctl dump silently discards errors
## Expected Behavior
sensuctl dump will report errors when they are encountered

## Current Behavior
sensuctl dump silently discards all errors, and returns a 0 status code

## Possible Solution
Use the API client instead of wrapping the command line tools

1. run sensuctl dump against a backend that is not running
2. observe that no output is produced, and no error status occurs

## Context
When trying to dump postgres configuration, an HTTP error was occurring but I was unable to tell what it was.

## Your Environment
Fedora Linux

echlebek:
## What is this change?

Fixes sensuctl dump again, this time with support for wrapped resources as well.

The original edition was backed out.
Closes #3236

Yes

The original edition of this PR was merged, but backed out due to the enterprise resources not working correctly with it.

New documentation is required, for dumping enterprise resources.

I ran `sensuctl dump all`, and:
```
[eric@localhost sensu-enterprise-go]$ sensuctl dump 'store/v1.PostgresConfig'
type: PostgresConfig
api_version: store/v1
metadata:
  name: postgres
  namespace: default
spec:
  dsn: foo=foo
  pool_size: 0
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 - Only validate check interval/cron when publish true
 
 ### Fixed
+- sensuctl dump no longer silently discards errors.
 - Interactive check create and update modes now have 'none' as the first
 highlighted option, instead of nagios-perfdata.
 - Fixed a bug where silences would not expire on event resolution.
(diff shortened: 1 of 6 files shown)


echlebek:
How do you suggest I document the full resource path for all the resources? Separate command?

echlebek:
Maybe `sensuctl dump --list-types`?

nikkixdev:
I suppose with reflection, it should be relatively easy to dump all the possible types, so your `--list-types` suggestion is definitely sufficient! As far as the literal docs, perhaps a matrix of all supported resource types and their corresponding path? Or a section in their resource reference?

echlebek:
I see, you mean in sensu-docs?

PR #3274, opened 2019-09-12
sensuctl commands support

amdprophet:

## What is this change?

Adds support for sensuctl commands which can be installed to extend the functionality of sensuctl.

This is currently a WIP.

A few things still need to be completed:

- [x] Refactor code duplicated between `cli/cmdmanager/manager.go` & `cli/commands/asset/add.go`
- [x] Support for listing & removing command plugins
- [x] A way to filter out non-command Bonsai assets (e.g. handlers, checks, etc.)
- [x] Other code quality changes

Yes.

### DIFF
--- a/asset/asset.go
+++ b/asset/asset.go
@@ -14,7 +14,10 @@ package asset
 
 import (
 	"context"
+	"fmt"
+	"os"
 	"path/filepath"
+	"strings"
 
 	corev2 "github.com/sensu/sensu-go/api/core/v2"
 )
@@ -57,4 +60,29 @@ func (r *RuntimeAsset) LibDir() string {
 // IncludeDir returns the full path to the asset's include directory.
 func (r *RuntimeAsset) IncludeDir() string {
 	return filepath.Join(r.Path, includeDir)
+}
+
+// Env returns a list of environment variables (e.g. 'PATH=...', 'CPATH=...')
+// with asset-specific paths prepended to the parent environment paths for
+// each variable, allowing an asset to be used during execution.
+func (r *RuntimeAsset) Env() []string {
+	assetEnv := []string{
+		fmt.Sprintf("PATH=%s${PATH}", r.joinPaths((*RuntimeAsset).BinDir)),
+		fmt.Sprintf("LD_LIBRARY_PATH=%s${LD_LIBRARY_PATH}", r.joinPaths((*RuntimeAsset).LibDir)),
+		fmt.Sprintf("CPATH=%s${CPATH}", r.joinPaths((*RuntimeAsset).IncludeDir)),
+	}
+	for i, envVar := range assetEnv {
+		// ExpandEnv replaces ${var} with the contents of var from the current
+		// environment, or an empty string if var doesn't exist.
+		assetEnv[i] = os.ExpandEnv(envVar)
+	}
+	return assetEnv
+}
+
+// joinPaths joins all paths of a given type for each asset in RuntimeAssetSet.
+func (r *RuntimeAsset) joinPaths(pathFunc func(*RuntimeAsset) string) string {
+	var sb strings.Builder
+	sb.WriteString(pathFunc(r))
+	sb.WriteRune(os.PathListSeparator)
+	return sb.String()
 }
--- a/cli/cmdmanager/manager.go
+++ b/cli/cmdmanager/manager.go
@@ -1,0 +1,268 @@
+package cmdmanager
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+	"sync"
+
+	"github.com/sensu/sensu-go/asset"
+	"github.com/sensu/sensu-go/bonsai"
+	"github.com/sensu/sensu-go/command"
+	"github.com/sensu/sensu-go/system"
+	"github.com/sensu/sensu-go/util/environment"
+	"github.com/sensu/sensu-go/util/path"
+
+	goversion "github.com/hashicorp/go-version"
+	corev2 "github.com/sensu/sensu-go/api/core/v2"
+	bolt "go.etcd.io/bbolt"
+)
+
+const (
+	dbName = "commands.db"
+)
+
+var (
+	commandBucketName = []byte("commands")
+)
+
+type CommandManager struct {
+	assetManager *asset.Manager
+	assetGetter  asset.Getter
+	db           *bolt.DB
+}
+
+type CommandPlugin struct {
+	Alias string       `json:"alias"`
+	Asset corev2.Asset `json:"asset"`
+}
+
+func NewCommandManager() (*CommandManager, error) {
+	m := CommandManager{}
+
+	// create an entity for using with command asset filtering
+	systemInfo, err := system.Info()
+	if err != nil {
+		return nil, err
+	}
+	meta := corev2.NewObjectMeta("sensuctl", "")
+	entity := &corev2.Entity{
+		EntityClass: "sensuctl",
+		ObjectMeta:  meta,
+		System:      systemInfo,
+	}
+
+	cacheDir := path.UserCacheDir("sensuctl")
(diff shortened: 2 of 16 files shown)


amdprophet:
I've gone with requiring a binary called `entrypoint` right now (`entrypoint.exe` on Windows) as `command` is an internal Powershell function that will conflict.

portertech:
[sensuctl plugins (command assets)](https://trello.com/c/2HcJteDZ/5-sensuctl-plugins-command-assets)

cli/cmdmanager/manager.go
```diff
@@ -0,0 +1,325 @@
...
+func (m *CommandManager) FetchCommandPlugins() ([]*CommandPlugin, error) {
+	var localCommandPlugins []*CommandPlugin
+
+	if err := m.db.View(func(tx *bolt.Tx) error {
+		b := tx.Bucket([]byte(commandBucketName))
+		if b == nil {
+			return nil
+		}
+
+		b.ForEach(func(k, v []byte) error {
```
golangcibot: Error return value of `b.ForEach` is not checked (from `errcheck`)

asset/asset.go
```diff
@@ -58,3 +61,28 @@ func (r *RuntimeAsset) LibDir() string {
...
+	for i, envVar := range assetEnv {
+		// ExpandEnv replaces ${var} with the contents of var from the current
+		// environment, or an empty string if var doesn't exist.
+		assetEnv[i] = os.ExpandEnv(envVar)
+	}
+	return assetEnv
+}
+
+// joinPaths joins all paths of a given type for each asset in RuntimeAssetSet.
+func (r *RuntimeAsset) joinPaths(pathFunc func(*RuntimeAsset) string) string {
```
echlebek: This looks a lot like `filepath.Join` from stdlib

cli/cmdmanager/manager.go
```diff
@@ -0,0 +1,325 @@
...
+	return nil
+}
+
+func (p *CommandPlugin) SetNamespace(namespace string) {
+	// no-op
+}
+
+func (p *CommandPlugin) SetObjectMeta(meta corev2.ObjectMeta) {
+	// no-op
+}
```
echlebek: What's the purpose of implementing `types.Resource` here?

cli/cmdmanager/manager.go
```diff
@@ -0,0 +1,325 @@
...
+	commandPlugin, err := m.fetchCommandPlugin(alias)
+	if err != nil {
+		return err
+	}
+
+	if commandPlugin == nil {
+		return errors.New("the alias specified does not exist")
+	}
+
+	ctx := context.TODO()
```
echlebek: Would be good to pass this in to `ExecCommand`

cli/cmdmanager/manager.go
amdprophet: It's necessary to use the table output functions. We should find a way to fix this.

asset/asset.go
amdprophet: It's a bit different. This is for joining paths used in the `PATH` variable on both linux & windows. `os.PathListSeparator` will be `:` on *nix and `;` on Windows.

cli/commands/command/list.go
```diff
@@ -0,0 +1,129 @@
...
+	return func(cmd *cobra.Command, args []string) error {
+		// If args are provided print out usage
+		if len(args) > 0 {
+			_ = cmd.Help()
+			return errors.New("invalid argument(s) received")
+		}
+
+		manager, err := cmdmanager.NewCommandManager(cli)
+		if err != nil {
+			return err
```
palourde: If no command plugin is installed, we will return an error that's not really UX friendly (e.g. `Error: open /Users/splourde/Library/Caches/sensu/sensuctl/commands.db: no such file or directory`). We should probably try to wrap this error and make it more user friendly (e.g. `no command plugin found`).

cli/cmdmanager/manager.go
```diff
@@ -0,0 +1,410 @@
...
+			return errors.New("requested asset is not a sensuctl/command asset")
+		}
+	} else {
+		return errors.New("requested asset does not have a provider annotation set")
+	}
+
+	return m.installCommand(alias, &asset)
+}
+
+func (m *CommandManager) InstallCommandFromURL(alias, assetURL, checksum string) error {
```
palourde: I must admit I found the UX of this subcommand a bit weird. Basically you need to link to an URL that points to an asset resource definition, and you also need to specify the checksum of that definition.

I took a look at the specs for this feature and it seems that when using an URL, we should actually point to the _tar.gz_ archive: https://docs.google.com/document/d/1veQ30ONMDFfZOayZLpAjmzgr1V96ahUK62uWWnaQiU4/edit#bookmark=id.yn2wm05esopd. That being said, maybe the specs changed but if so, I found that really strange.

Additionally, could we also support a local archive? I feel like it could be a quick win.

amdprophet: This is a misunderstanding on my part. I think we can change this to point to a single archive but we'll lose the ability to install the correct archive based on what platform sensuctl is running on.

cli/commands/command/install.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+var assetURL string
+var assetChecksum string
+
+// InstallCommand adds command that allows a user to install a command plugin
+// via Bonsai or a URL.
+func InstallCommand(cli *cli.SensuCli) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "install [ALIAS] [NAME][:VERSION]",
```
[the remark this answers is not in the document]
amdprophet: I'll apply these changes.

cli/commands/command/delete.go
```diff
@@ -0,0 +1,51 @@
...
+					return nil
+				}
+			}
+
+			manager, err := cmdmanager.NewCommandManager(cli)
+			if err != nil {
+				return err
+			}
+
+			if err := manager.DeleteCommandPlugin(name); err != nil {
```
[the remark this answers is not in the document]
amdprophet: Is there a way to verify this through the `create` package in sensuctl? I'll take another look.

cli/cmdmanager/manager.go
```diff
@@ -0,0 +1,410 @@
...
+		return nil, err
+	}
+	meta := corev2.NewObjectMeta("sensuctl", "")
+	entity := &corev2.Entity{
+		EntityClass: "sensuctl",
+		ObjectMeta:  meta,
+		System:      systemInfo,
+	}
+
+	cacheDir := path.UserCacheDir("sensuctl")
```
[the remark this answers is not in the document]
amdprophet: Good catch.

cli/cmdmanager/manager.go
```diff
@@ -0,0 +1,410 @@
...
+
+func (p *CommandPlugin) SetNamespace(namespace string) {
+	// no-op
+}
+
+func (p *CommandPlugin) SetObjectMeta(meta corev2.ObjectMeta) {
+	// no-op
+}
+
+func NewCommandManager(cli *cli.SensuCli) (*CommandManager, error) {
```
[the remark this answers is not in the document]
amdprophet: Agreed - I'll add some tests.

cli/cmdmanager/manager.go
palourde: Yeah, I think it's fine IMO; that's an added value of Bonsai, but it still let users easily install command plugins locally or from a private repo

testing/mockassetgetter/mockassetgetter.go
```diff
@@ -0,0 +1,20 @@
...
+	"github.com/stretchr/testify/mock"
+)
+
+// MockAssetGetter ...
+type MockAssetGetter struct {
+	mock.Mock
+}
+
+// Get ...
+func (m MockAssetGetter) Get(ctx context.Context, a *corev2.Asset) (*asset.RuntimeAsset, error) {
```
golangcibot: copylocks: Get passes lock by value: github.com/sensu/sensu-go/testing/mockassetgetter.MockAssetGetter contains github.com/stretchr/testify/mock.Mock contains sync.Mutex (from `govet`)

Issue #1353, opened 2018-04-21
Deprecate pipelined and move pipeline functionality into agentd


echlebek:
Like eventd and keepalived, pipelined represents a fixed size pipeline that we expect to handle any number of events. Unlike eventd and keepalived, pipelined has a somewhat lengthier request scope. Pipelined filters, handlers and mutators can result in processes being spawned, or requests being made to network services.

We may find it necessary to queue events going to pipelines, but each entity should not have its pipeline obstructed by other entities' filters, mutators or handlers.

Every session should include a pipeline goroutine that handles this. Investigate if there is a need for queueing, and if we need _persistent_ queueing.

portertech:
Interesting, I find single queue + dispatcher compelling here for a number of reasons. Looking forward to discussing (but put our notes and findings here).

portertech:
#1 feature of pipelined, no "fork bombing".

portertech:
Making the fixed pipelined concurrency limit configurable would help.

echlebek:
Sorry for taking so long to respond here.

This issue, and the ones similar to it, are driven by the goal of achieving as much parallelism in Sensu as possible. The best way to do that is to minimize information sharing; a shared-nothing system can achieve optimal parallelism.

In our current design, we have to process every event in pipelined. That means that every agent's session must synchronize with pipelined, regardless of the content of the event. However, there is no such requirement to do so; we do not need to serialize events globally.

Instead, events only need to be serialized on a per-entity basis. This means we can process events inside of our session goroutines, or in another separate goroutine for each session. As such, we can reduce blocking for events significantly in the common case.

Another requirement is that we do not exec `N` processes in the execution of filters, mutators and handlers. This was the original stated intent of pipelined, which it achieves; pipelined has a fixed number of goroutines and so can only exec as many processes as there are goroutines.

This requirement can be solved by using a worker pool or token bucket. We simply need to constrain the number of processes that are forked in executing filters, mutators and handlers.

To recap, we can limit information sharing to filters, mutators and handlers that need to be executed as external processes, etcd writes notwithstanding.

PR #3277, opened 2019-09-13
Fix authorization of event & silence updates

jamesdphillips:
## What is this change?

Allow users to update events and silences through GraphQL service.

Fixes an issue reported on [Discourse (#1353)](https://discourse.sensu.io/t/are-these-bugs-in-5-12/1353).

### Replication Steps

1. Set up with limited access.
```
sensuctl user create jeb --password meow1234 --groups linux
sensuctl role create local-admin --verb get,list,create,update,delete --resource assets,checks,entities,events,filters,hooks,silenced --namespace linux
sensuctl role-binding create local-admin --role=local-admin --group linux --namespace linux
```
2. Attempt to resolve an event or create a silence from the web app.

Yup.

It's unclear what the original intent was. Chesterton's fence, et al.

No.

None required.

Unit & manual

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,8 @@ steps.
 - Updated Go version from 1.12.3 to 1.13.1.
 
 ### Fixed
+- [Web] Fixed issue where a user with an appropriate role may have been unable
+to resolve events, queue checks, and create silenced entries.
 - Splayed proxy checks are now executed every interval, instead of every
 `interval + interval * splay_coverage`.
 - [GraphQL] Ensures that proxy entity label & annotations are redacted.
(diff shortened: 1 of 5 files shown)


palourde:
@echlebek The `verb` attribute in the authorization attributes should only contain a single verb (e.g. `update`), which represents whatever action the incoming request does. What happened behind the scenes is that RBAC was probably try to find the string literal `create,update` in roles, which obviously failed, and therefore returned an unauthorized error for all requests.

jamesdphillips:
> This PR lacks a changelog entry

Added an entry!

> we should probably have unit tests around those changes!

I looked through the unit tests again and I think they're already pretty exhaustive. In this case, they didn't catch this problem because the mock authorizator mirror'd the same incorrect configuration.

This to me would indicate that we may want more automated (and/or possibly integration) tests.

I will add some manual tests to our QA suite before we do our v5.14 release.

PR #3282, opened 2019-09-17
Refactoring of agentd to support mTLS

palourde:
## What is this change?

This PR essentially consists of a refactoring to make it possible to configure mTLS from the enterprise codebase. In summary:

- It adds supports for `ClientCAs` & `ClientAuth` attributes in the `ToServerTLSConfig()` method.
- It adds a `AgentTLSOptions` to the backend config, which fallbacks to the common TLS attribute if nil.
- It prevents panics in certain cases when `*corev2.TLSOptions` is nil
- It moves agentd's HTTP server middlewares to the agentd package and make it possible to overwrite them at runtime
- It sends agentd's HTTP server logs to logrus directly instead of being reported as coming from the `etcd` package.
- It splits apid's router initialization into distinct subrouters functions so it's easier to understand what's happening; this change isn't required per se but I got confused at some point and thought I would need access to the authentication subrouter. I still decided to include it within this PR because I feel like it's nicer but I can remove it if required.
- It fixes apid's routers tests; The integration tests failed on my macbook because of `too many open files` errors, which simply happened because we didn't closed the request body.

Required for mTLS enterprise functionality. Related to https://github.com/sensu/sensu-enterprise-go/issues/645.

Already covered by the existing entry within the `Added` section I believe.

Nope

mTLS is hard 😅

Will be done via https://github.com/sensu/sensu-enterprise-go/issues/644.

Will run automated QA on the enterprise branch.


backend/agentd/middlewares.go
```diff
@@ -0,0 +1,129 @@
...
+	corev2 "github.com/sensu/sensu-go/api/core/v2"
+	"github.com/sensu/sensu-go/backend/apid/middlewares"
+	"github.com/sensu/sensu-go/backend/authentication/jwt"
+	"github.com/sensu/sensu-go/backend/authorization"
+	"github.com/sensu/sensu-go/backend/authorization/rbac"
+	"github.com/sensu/sensu-go/backend/store"
+	"github.com/sensu/sensu-go/transport"
+)
+
+// AuthenticationMiddleware represents the middleware used for authentication
```
echlebek: Simon, Nikki and I discussed the rationale for these global variables on Slack. They are to support patching from the enterprise product.

I do think we should include a TODO here about what we intend to do to support graceful degradation in the future.

palourde: @echlebek I'll document the enterprise codebase with the TODO and some links to issues.

Issue #3156, opened 2019-07-22
Round robin check stops getting scheduled


adammcdonagh:
We have 2 round robin checks. For both of them, only our Sensu servers are subscribed. 

I'm not sure if it's only round robin checks that this effects, however both times it has been the same 2 checks that stop getting scheduled.

Twice over the past week, these checks have stopped being scheduled... The only way I have found to resolve the issue is to restart the backend instances... 

A message is logged in the backend log showing this.. But never appears to get rescheduled.

## Expected Behavior
Checks should never stop being scheduled unless they are no longer published.

## Current Behavior
Round robin checks stop being triggered  after a random amount of time.

It looks like its happening after a loss of leader:

```
Jul 21 05:01:19 vlxxxx003 sensu-backend: {"component":"schedulerd","error":"etcdserver: no leader","level":"error","msg":"error scheduling check","name":"check-metrics-remote","namespace":"default","scheduler_type":"round-robin interval","time":"2019-07-21T05:01:19+01:00"}
Jul 21 05:01:29 vlxxxx003 sensu-backend: {"component":"schedulerd","error":"error while starting ring watcher: etcdserver: request timed out","level":"error","msg":"error scheduling check","name":"check-metrics-remote","namespace":"default","scheduler_type":"round-robin interval","time":"2019-07-21T05:01:29+01:00"}
Jul 21 05:01:29 vlxxxx003 sensu-backend: {"component":"schedulerd","level":"warning","msg":"shutting down scheduler","name":"check-metrics-remote","namespace":"default","scheduler_type":"round-robin interval","time":"2019-07-21T05:01:29+01:00"}
```

## Steps to Reproduce (for bugs)
Unknown. Maybe can be triggered by forcing a loss of leader?

## Context
Checks that don't get scheduled need to be manually identified... We have a script to check this, however in this case, the check that is no longer getting scheduled is that script...

## Your Environment
RHEL 7, Sensu Go v5.12

palourde:
I just reproduced this issue by spinning up a 3-node cluster, adding the following proxy check and creating a partition:

```
type: CheckConfig
api_version: core/v2
metadata:
  name: proxy_check
  namespace: default
spec:
  command: echo pong
  interval: 60
  proxy_entity_name: ping
  publish: true
  round_robin: true
  subscriptions:
  - proxy
```

[...]

PR #3283, opened 2019-09-18
Implement an infinite retry mechanism in the ring

nikkixdev:

## What is this change?

Implements an infinite retry mechanism (with backoff) if the ring watcher fails to start. This fixes a bug where round robin scheduled events (such as checks and tessen data collection) would fail to recover after quorum loss.

Closes https://github.com/sensu/sensu-go/issues/3156

Yas.

Nah.

The bug was pretty difficult to reproduce, but the notes in the issue simplified the solution a lot.

We good.

Tested this bug in staging (see steps to reproduce from https://github.com/sensu/sensu-go/issues/3156).


backend/ringv2/ringv2.go
```diff
@@ -355,12 +363,22 @@ func (r *Ring) startWatchers(ctx context.Context, ch chan Event, name string, va
...
-		return
+
+	backoff := retry.ExponentialBackoff{
+		InitialDelayInterval: 10 * time.Millisecond,
+		MaxDelayInterval:     10 * time.Second,
+		Multiplier:           10,
+		Ctx:                  ctx,
 	}
+
+	err = backoff.Retry(func(retry int) (bool, error) {
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

backend/ringv2/ringv2.go
```diff
@@ -355,12 +363,22 @@ func (r *Ring) startWatchers(ctx context.Context, ch chan Event, name string, va
...
+	backoff := retry.ExponentialBackoff{
+		InitialDelayInterval: 10 * time.Millisecond,
+		MaxDelayInterval:     10 * time.Second,
+		Multiplier:           10,
+		Ctx:                  ctx,
 	}
+
+	_ = backoff.Retry(func(retry int) (bool, error) {
+		if err := watcher.ensureActiveTrigger(ctx); err != nil {
+			notifyError(ch, fmt.Errorf("error while starting ring watcher: %s", err))
```
nikkixdev: @echlebek while the bug is fixed with this change, I'm curious if I still need to notify the closing channel and cancel:
```suggestion
			notifyError(ch, fmt.Errorf("error while starting ring watcher: %s", err))
			notifyClosing(ch)
			cancel()
```

backend/ringv2/ringv2.go
```diff
@@ -65,6 +72,7 @@ type Event struct {
 	Err error
 }
 
+// Ring is a ring of items that are triggered in a round robin fashion.
```
echlebek: Ring is a circular queue of items that are cooperatively iterated over by one or more subscribers. The subscribers are notified every time an item goes from the front to the back of the queue, with EventTrigger. Iteration proceeds according to an interval, and is triggered by an etcd lease expiring.

backend/ringv2/ringv2.go
```diff
@@ -355,12 +363,22 @@ func (r *Ring) startWatchers(ctx context.Context, ch chan Event, name string, va
 	r.mu.Lock()
 	r.watchers[watcher.watcherKey] = watcher
 	r.mu.Unlock()
-	if err := watcher.ensureActiveTrigger(ctx); err != nil {
-		notifyError(ch, fmt.Errorf("error while starting ring watcher: %s", err))
-		notifyClosing(ch)
-		cancel()
-		return
+
+	backoff := retry.ExponentialBackoff{
```
echlebek: I think this should go in the `ensureActiveTrigger` method itself - the method is called from two different places.

backend/ringv2/ringv2.go
```diff
@@ -13,17 +13,23 @@ import (
...
 	"github.com/sensu/sensu-go/backend/store"
+	"github.com/sensu/sensu-go/util/retry"
 	"golang.org/x/time/rate"
 )
 
 // EventType is an enum that describes the type of event received by watchers.
 type EventType int
 
 const (
+	// EventError ...
```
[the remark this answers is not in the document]
nikkixdev: Agreed, thanks for supplying the words!

backend/ringv2/ringv2.go
echlebek: I think you should capture the error that comes from `backoff.Retry` and cancel if it's non-nil.

backend/ringv2/ringv2.go
```diff
@@ -318,26 +329,37 @@ func (w *watcher) hasTrigger(ctx context.Context) (bool, string, error) {
...
+		}
+		lease, err := w.grant(ctx)
+		if err != nil {
+			return false, err
+		}
+		nextValue := path.Base(next)
+		triggerOp := clientv3.OpPut(w.triggerKey(), nextValue, clientv3.WithLease(lease.ID))
+		triggerCmp := clientv3.Compare(clientv3.Version(w.triggerKey()), "=", 0)
+
+		resp, err := w.ring.client.Txn(ctx).If(triggerCmp).Then(triggerOp).Commit()
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

backend/ringv2/ringv2.go
```diff
@@ -318,26 +329,40 @@ func (w *watcher) hasTrigger(ctx context.Context) (bool, string, error) {
...
+		if !resp.Succeeded {
```
echlebek: This should come after the error checking.

echlebek:
I added logging for the error conditions, and returned nil from the retry function, as @palourde indicated should be done.

backend/ringv2/ringv2.go
```diff
@@ -318,26 +334,43 @@ func (w *watcher) hasTrigger(ctx context.Context) (bool, string, error) {
...
-	triggerCmp := clientv3.Compare(clientv3.Version(w.triggerKey()), "=", 0)
 
-	resp, err := w.ring.client.Txn(ctx).If(triggerCmp).Then(triggerOp).Commit()
-	if !resp.Succeeded {
-		_, _ = w.ring.client.Revoke(ctx, lease.ID)
-	}
+	err := backoff.Retry(func(retry int) (bool, error) {
+		has, next, err := w.hasTrigger(ctx)
+		if err != nil {
+			logger.WithError(err).Error("retrying")
```
palourde: I'd suggest maybe adding a bit more context to the error than `retrying`, so the operator understand that we could not get the trigger key from etcd etc.

backend/ringv2/ringv2.go
```diff
@@ -318,26 +334,43 @@ func (w *watcher) hasTrigger(ctx context.Context) (bool, string, error) {
...
+			return false, nil
+		}
+		if has || next == "" {
+			// if next == "", there are no ring items
+			return true, nil
+		}
+		lease, err := w.grant(ctx)
+		if err != nil {
+			logger.WithError(err).Error("retrying")
```
palourde: Same as ^

Issue #3201, opened 2019-08-14
Create APIKey API


echlebek:
## Feature Suggestion
Create an API that allows managing API keys. API keys are long-lived authentication tokens that are managed centrally, and persisted in etcd.

Link to spec: https://docs.google.com/document/d/1wdfAhajJsXCEmC0vquifkbdP5O33HSPH9PbjFLC46b4/edit#

## Context
We want to enable automated workflows with Sensu's API. Right now, this is difficult because of the way our authentication system works.

portertech:
![](https://github.trello.services/images/mini-trello-icon.png) [Persistent API access tokens](https://trello.com/c/GpMLqYPO/7-persistent-api-access-tokens)

portertech:
Related #3202 #3203

PR #3303, opened 2019-09-30
Add APIKey and support for http api

nikkixdev:

## What is this change?

Adds the APIKey resource and `api/core/v2/apikeys` API.

Closes https://github.com/sensu/sensu-go/issues/3201

Yas.

Nah.

Nah.

https://github.com/sensu/sensu-docs/issues/1797
https://github.com/sensu/sensu-docs/issues/1798

Unit tests added. Filed https://github.com/sensu/sensu-go-qa-crucible/issues/79.


api/core/v2/apikey.go
```diff
@@ -0,0 +1,75 @@
...
+		"api_key.name":     resource.ObjectMeta.Name,
+		"api_key.username": resource.Username,
+		"api_key.groups":   strings.Join(resource.Groups, ","),
+		"api_key.key":      resource.Key,
+	}
+}
+
+// SetNamespace sets the namespace of the resource.
+func (a *APIKey) SetNamespace(namespace string) {
+	return
```
golangcibot: S1023: redundant `return` statement (from `gosimple`)

api/core/v2/apikey.proto
```diff
@@ -0,0 +1,35 @@
...
+  ObjectMeta metadata = 1 [(gogoproto.jsontag) = "metadata,omitempty", (gogoproto.embed) = true, (gogoproto.nullable) = false];
+
+  // Key is the API key id.
+  string key = 2;
+
+  // Username is the username associated with the API key.
+  string username = 3;
+
+  // Groups are the groups associated with the API key.
+  repeated string groups = 4;
```
ccressent: I don't think groups should be there. I think only `username` should be there, serving as a "pointer" to an actual user in the store, which itself stores all the relevant details about the user, including their groups. That way if a user gains or loses group memberships, it's automatically reflected in the API keys; no need to keep User and API key groups in sync.

There might be (slight) performance costs to doing that, but I believe it's 100% worth it to have a system that cannot diverge by design.

api/core/v2/apikey.proto
```diff
@@ -0,0 +1,35 @@
...
+option (gogoproto.unmarshaler_all) = true;
+option (gogoproto.sizer_all) = true;
+option (gogoproto.testgen_all) = true;
+
+// An APIKey is an api key specification.
+message APIKey {
+  option (gogoproto.face) = true;
+  option (gogoproto.goproto_getters) = false;
+
+  // Metadata contains the name of the APIKey.
```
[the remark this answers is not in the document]
nikkixdev: I think we wanted to at least make use of the `Name` field, however not `Namespace` since these are cluster-wide resources. I can update the comment here 😄

api/core/v2/apikey.proto
nikkixdev: I agree that having both `groups` and `username` defined is a little misleading (following the spec, but unsure why we decided we needed both). If an API key should only be used by one user, than I absolutely agree that we should remove `groups`. However, if we desire that the API key be a little more generically usable, perhaps `groups` is the way to go. Open to discussion, and happy to set a call up to discuss further!

backend/apid/routers/apikeys.go
```diff
@@ -0,0 +1,96 @@
...
+
+func (r *APIKeysRouter) create(w http.ResponseWriter, req *http.Request) {
+	apikey := &corev2.APIKey{}
+	if err := UnmarshalBody(req, apikey); err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	// set/overwrite the key id and created_at time
+	apikey.Key = uuid.New().String()
```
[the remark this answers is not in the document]
nikkixdev: Yeah I'll update that!

api/core/v2/apikey.proto
palourde: Do we rely on the API key groups, or the user groups when determining permissions in RBAC?

1. If we rely on the API key groups, it feels a bit counter intuitive to refer to a user while ignoring its groups.
2. If we rely on the user groups, then we don't need this attribute
3. If we rely on both, it might create some ambiguity and make it hard to understand how a user got permission to whatever resource, and it makes it a bit more complex to remove some permissions (as an operator, I have to remember to modify both the user and its API keys).

TL;DR I might be missing some context about the reason we have a groups attribute here but I feel like it shouldn't probably exist!

echlebek:
Sounds like the `groups` field is a spec bug, so I'm perfectly happy for it to be removed.

api/core/v2/apikey.proto
nikkixdev: Removed the groups field (in favor of username) and the key field (in favor of name).

api/core/v2/apikey.proto
```diff
@@ -0,0 +1,35 @@
...
+// An APIKey is an api key specification.
+message APIKey {
+  option (gogoproto.face) = true;
+  option (gogoproto.goproto_getters) = false;
+
+  // Metadata contains the name, namespace (N/A), labels and annotations of the APIKey.
+  ObjectMeta metadata = 1 [(gogoproto.jsontag) = "metadata,omitempty", (gogoproto.embed) = true, (gogoproto.nullable) = false];
+
+  // Key is the API key id.
+  string key = 2;
```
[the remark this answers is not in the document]
nikkixdev: Removed the groups field (in favor of username) and the key field (in favor of name).

api/core/v2/apikey.go
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+	"fmt"
+	"net/url"
+	"path"
+)
+
+const (
+	// APIKeysResource is the name of this resource type.
+	APIKeysResource = "apikeys"
```
echlebek: Do you think this should be `api-keys`? I did `etcd-replicators` recently, I feel like it reads a bit better to have a dash.

backend/apid/routers/apikeys.go
```diff
@@ -0,0 +1,101 @@
...
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	// set the relative location header
+	w.Header().Set("Location", fmt.Sprintf("%s/%s", req.URL.String(), apikey.Name))
+	w.WriteHeader(http.StatusCreated)
+}
+
+func (r *APIKeysRouter) update(req *http.Request) (interface{}, error) {
```
palourde: Stupid question: The specs mentions we should support updating an API key to update its binding. Considering we really only store the username in the key now, should we still support updating an API key?

api/core/v2/apikey.go
```diff
@@ -0,0 +1,70 @@
...
+// Validate returns an error if the CheckName and Subscription fields are not
+// provided.
+func (a *APIKey) Validate() error {
+	if a.Namespace != "" {
+		return fmt.Errorf("api key cannot have a namespace")
+	}
+
+	if a.Username == "" {
+		return fmt.Errorf("api key must have a username")
+	}
```
palourde: So I'm a bit confused since the specs are no longer in sync with what we appear to be implementing, but I'm wondering if we should validate whether the user exist in the Sensu store; I assume we only want to support API keys for "internal" users, and not users authenticated via an authentication provider, since we can't retrieve their groups on-demand.

backend/apid/routers/apikeys.go
```diff
@@ -0,0 +1,101 @@
...
+		return nil, actions.NewError(actions.NotFound, err)
+	}
+
+	apikey := &corev2.APIKey{}
+	if err := UnmarshalBody(req, apikey); err != nil {
+		return nil, actions.NewError(actions.InvalidArgument, err)
+	}
+
+	// only allow groups to be updated
+	stored.Username = apikey.Username
```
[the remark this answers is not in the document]
nikkixdev: The store already does resource validation: https://github.com/sensu/sensu-go/blob/613f31c430c7d2b8d00de82e8815756f68f118c9/backend/store/etcd/resource.go#L14 Is that sufficient?

api/core/v2/apikey.go
nikkixdev: Readability vs consistency, it's a great question. To me `/clusterrolebindings` is clunky, but it's the naming convention we went with for corev2. Happy to discuss further if we want to break the convention.

api/core/v2/apikey.go
nikkixdev: We would probably validate this in the API router. Not sure why this simple validation isn't in sync with the spec. Could you elaborate?

backend/apid/routers/apikeys.go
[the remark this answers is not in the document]
nikkixdev: We've removed the update functionality for this.

backend/apid/routers/apikeys.go
nikkixdev: I've removed the functionality to update an api key. If we decide we want to support groups (and updating them) in the future, we can reimplement.

api/core/v2/apikey.go
nikkixdev: I've added user validation to the router.

PR #3307, opened 2019-10-02
Fix the --all-namespaces flag in sensuctl dump

### LINKED ISSUE #3306: The --all-namespaces flag no longer work with sensuctl dump
## Expected Behavior
On the v5.13.2 tag of sensu-go:

```
$ sensuctl dump check --format wrapped-json --all-namespaces
{
  "type": "CheckConfig",
  "api_version": "core/v2",
  "metadata": {
    "name": "check-echo",
    "namespace": "default"
  },
  "spec": {
    "check_hooks": null,
    "command": "PATH=/ bash -c 'echo 1'",
    "env_vars": null,
    "handlers": [],
    "high_flap_threshold": 0,
    "interval": 10,
    "low_flap_threshold": 0,
    "output_metric_format": "",
    "output_metric_handlers": null,
    "proxy_entity_name": "",
    "publish": true,
    "round_robin": false,
    "runtime_assets": null,
    "stdin": false,
    "subdue": null,
    "subscriptions": [
      "entity:whisky"
    ],
    "timeout": 0,
    "ttl": 0
  }
}
```

## Current Behavior
On the tip (3c3a68b1db44d64469259db3badcd5c8187fa3a1), nothing is returned:

```
$ sensuctl dump check --format wrapped-json --all-namespaces
$
```

## Possible Solution
Make sure to support the `--all-namespaces` flag.

Create at least 1 check

echlebek:
## What is this change?

Fix a regression that was introduced by the sensuctl dump refactoring.

The --all-namespaces flag works with this change. Closes #3306 

No, as this regression was not shipped in a release.

I manually tested dumping every resource in sensu-go and sensu-enterprise-go with multiple namespaces to make sure they would all work properly.

### DIFF
--- a/api/core/v2/event.go
+++ b/api/core/v2/event.go
@@ -34,6 +34,9 @@ func (e *Event) StorePrefix() string {
 
 // URIPath returns the path component of an event URI.
 func (e *Event) URIPath() string {
+	if e.Namespace == "" {
+		return path.Join(URLPrefix, EventsResource)
+	}
 	if !e.HasCheck() {
 		return ""
 	}
--- a/api/core/v2/event.go
+++ b/api/core/v2/event.go
@@ -34,11 +34,11 @@ func (e *Event) StorePrefix() string {
 
 // URIPath returns the path component of an event URI.
 func (e *Event) URIPath() string {
-	if e.Namespace == "" {
+	if !e.HasCheck() && e.Entity == nil {
 		return path.Join(URLPrefix, EventsResource)
 	}
 	if !e.HasCheck() {
-		return ""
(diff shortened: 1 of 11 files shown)


api/core/v2/event.go
```diff
@@ -34,8 +34,11 @@ func (e *Event) StorePrefix() string {
 
 // URIPath returns the path component of an event URI.
 func (e *Event) URIPath() string {
+	if !e.HasCheck() && e.Entity == nil {
+		return path.Join(URLPrefix, EventsResource)
+	}
 	if !e.HasCheck() {
-		return ""
+		return path.Join(URLPrefix, EventsResource, url.PathEscape(e.Entity.Name))
```
palourde: That wouldn't be a valid resource identifier though, no?

PR #3310, opened 2019-10-03
Fix agent session bugs

### LINKED ISSUE #3168: Backend unable to fully reestablish connection with agents after temporary network outage causing stale checks
## Expected Behavior
When the sensu backend temporary loses network connection for some reason, it should be able to reestablish connection with the sensu agents and continue scheduling checks when the network connection comes back.

## Current Behavior
When the backend comes back from a temporary lost network connection, it might not be able to fully reestablish connection with agents ending up in a state where the keepalive check is OK, but other checks for the agent doesn't get scheduled/executed and end up stale. I can't see any patterns in which checks ending up stale, it seems to random. E.g. I have one host performing 400+ proxy checks where only some of them (random) end up stale. The same is true for non-proxy entities.

## Possible Solution
Only a workaround: Set a TTL on every check to get notified on all the stale checks

1. Run a backend with several agents and checks
2. Plug out network (either cable or disable network card in hyper visor)
3. Plug in network
4. Observe stale checks

## Context
We had a temporary network outage in our data center, backend lost network connection with agents. When network came back up, everything was OK in the sensu dashboard. What we didn't knew was that many of our checks were stale and we had a degraded monitoring system. This was before we set a TTL on every check.

## Your Environment
```
sensu-backend version 5.11.1#39d2bf2, build 39d2bf22c416f9a78f22c4817dc74ab62d7f4d63, built 2019-07-18T13:30:28Z

sensuctl version 5.11.1#39d2bf2, build 39d2bf22c416f9a78f22c4817dc74ab62d7f4d63, built 2019-07-18T13:31:59Z

sensu-agent version 5.11.1#39d2bf2, build 39d2bf22c416f9a78f22c4817dc74ab62d7f4d63, built 2019-07-18T13:31:26Z
```
One server running Sensu backend, installed and configured with the Puppet-module `mod 'sensu-sensu', '3.4.0'`, CentOS 7.

Agents also installed with the same Puppet-module, running mostly on CentOS7 and Windows Server 2012, 2016 and 2019

Backend have:
* ~ 500 checks
* ~ 150 proxy entites
* ~ 150 non-proxy entities

echlebek:
## What is this change?

This commit fixes some concurrency and resource lifetime management bugs in the agent session. In particular, it fixes a bug where the session will continue to operate in a broken state when a connection send error occurs. Now, the session will be torn down on the first error, to force the agent to reconnect.

I've eliminated the subPump concept in the agent session, as I found it to be unnecessarily complex.

Edit: additionally, the `mocktransport.MockTransport` type once again satisfies the `transport.Transport` interface.

Several users have reported that checks will stop executing, while keepalives will continue.

Closes #3168
Closes https://github.com/sensu/sensu-go/issues/3078

Yes, but with the upcoming release I've avoided it to reduce merge conflicts. I will add one later.

I felt the need to perform substantial refactoring on the agent session, as the code was quite crufty, and had several pitfalls that could lead to race conditions. By using contexts for cancellation, I have avoided these pitfalls.

No documentation changes are required.

This is still underway, but I have done integration testing and also manual testing. More testing is need to make sure the backend and agent behave properly in various failure scenarios.


backend/agentd/session_test.go
```diff
@@ -129,7 +132,61 @@ func TestBadSessionConfig(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	tm := &transport.Message{
+		Payload: b,
+		Type:    transport.MessageTypeEvent,
+	}
+
+	conn.On("Receive").After(100*time.Millisecond).Return(tm, nil)
+	conn.On("Send", mock.Anything).Return(transport.ConnectionError{"some horrible network outage"})
```
golangcibot: composites: `github.com/sensu/sensu-go/transport.ConnectionError` composite literal uses unkeyed fields (from `govet`)

backend/agentd/session.go
```diff
@@ -207,58 +199,49 @@ func (s *Session) subPump() {
...
-				switch err := err.(type) {
-				case transport.ConnectionError, transport.ClosedError:
-					return
-				default:
-					logger.WithError(err).Error("send error")
-				}
+		logger.WithField("payload_size", len(msg.Payload)).Debug("session - sending message")
+		if wsc, ok := s.conn.(*transport.WebSocketTransport); ok {
+			deadline := time.Now().Add(time.Second * time.Duration(s.cfg.WriteTimeout))
+			if err := wsc.Connection.SetWriteDeadline(deadline); err != nil {
```
ccressent: Are we calling `SetWriteDeadline()` before every message sent? If so, is that necessary?

echlebek: Yes, and I think it is. SetWriteDeadline takes a time.Time, so it needs to be constantly set to work properly AFAIK.

ccressent: I see, you're right!
See the comment above the `SetDeadline()` method of the [net.Conn interface](https://golang.org/pkg/net/#Conn).

sensu-discourse:
This pull request has been mentioned on **Sensu Community**. There might be relevant details there:

https://discourse.sensu.io/t/just-sharing-a-list-of-problems-ive-found/1389/2

PR #3312, opened 2019-10-03
Implement bounded executors

echlebek:
## What is this change?

A TODO from Greg, this change bounds the number of concurrently
executing processes forked by sensu-go backends and agents to a maximum
size of GOMAXPROCS.

Users can control the parallelism of agent and backend worker processes by
adjusting the GOMAXPROCS environment variable.

In a system that is being loaded to capacity, this change allows the application to exert back pressure. This should allow for better peak throughput of worker processes.

Yes, but I'll add it later, closer to the time of merge.

The existing tests are mostly sufficient, but it would be a nice improvement to add one that explicitly tests the bounded parallelism.

### DIFF
--- a/agent/check_handler.go
+++ b/agent/check_handler.go
@@ -24,7 +24,6 @@ const allowListOnDenyStatus = "allow_list_on_deny_status"
 const allowListOnDenyOutput = "check command denied by the agent allow list"
 
 // handleCheck is the check message handler.
-// TODO(greg): At some point, we're going to need max parallelism.
 func (a *Agent) handleCheck(ctx context.Context, payload []byte) error {
 	request := &corev2.CheckRequest{}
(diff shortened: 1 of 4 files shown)


echlebek:
Yeah, we definitely need some docs here. And more tests.

echlebek:
Sure!

portertech:
Should we really tie pipeline executions to GOMAXPROCS? Doesn't it default to CPU cores? So many handlers are IO bound.

echlebek:
It may be preferable to create a setting in backend.yml.

ccressent:
@echlebek would like to try a different approach:
> I want to use a library that was presented at Gophercon to do this work called [kaburaya](https://github.com/monochromegane/kaburaya)

Closing this PR in favour of that future work.

PR #3314, opened 2019-10-03
[GraphQL] Fixup entity object meta regression

jamesdphillips:
## What is this change?

Fixes regression introduced in #3275.

Closes sensu/web#201

This regression was introduced prior to the last release.

Added an additional unit test.

### DIFF
--- a/backend/apid/graphql/entity.go
+++ b/backend/apid/graphql/entity.go
@@ -35,7 +35,8 @@ func (*entityImpl) ID(p graphql.ResolveParams) (string, error) {
 // Metadata implements response to request for 'metadata' field.
 func (*entityImpl) Metadata(p graphql.ResolveParams) (interface{}, error) {
 	entity := p.Source.(*corev2.Entity)
-	return entity.GetRedactedEntity(), nil
+	entity = entity.GetRedactedEntity()
+	return entity.GetObjectMeta(), nil
 }
 
 // LastSeen implements response to request for 'executed' field.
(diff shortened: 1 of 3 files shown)


backend/apid/graphql/entity.go
```diff
@@ -35,7 +35,8 @@ func (*entityImpl) ID(p graphql.ResolveParams) (string, error) {
 // Metadata implements response to request for 'metadata' field.
 func (*entityImpl) Metadata(p graphql.ResolveParams) (interface{}, error) {
```
ccressent: I'm sure there is a reason, but wouldn't it be safer to have `Metadata()` return `(v2.ObjectMeta, error)` instead? Then this would have become a compile time error.

jamesdphillips: Absolutely agree, in theory these are the sort of mistakes using a strongly typed language should help us avoid entirely. In general I'd always prefer using a concrete type.

However, this is current a deficiency in the code generator package (`graphql/generator`). The _field resolver_ interfaces are generated from the GraphQL IDL definitions found in `api/graphql/schema`, (specifically in this case [schema/entity.graphl](https://github.com/sensu/sensu-go/blob/c5a0ffe/backend/apid/graphql/schema/entity.graphql#L12-L13).) From the generator's perspective, the problem is that there no hints given about which (if any) Golang type the GraphQL type represents, as such when it encounters the `ObjectMeta` GraphQL type it has no way of knowing whether it is referring to the `ObjectMeta` type in `core/v2`, or the `ObjectMeta` type in the Kubernetes package, or whatever other package may export a type by the same name.

The obvious solution would be to implement a typemap or some other mechanism so that the generator can lookup what Golang type the GraphQL type represents. Then we can generate interfaces that are more strict and help us avoid what should be easily avoidable bugs.

PR #3321, opened 2019-10-07
Fix sensuctl create stdin regression

### LINKED ISSUE #3318: valid configuration not created on backend when sent to sensuctl via stdin
## Expected Behavior

Valid configuration sent to `sensuctl create` results in corresponding resources being created in the sensu-backend.

## Current Behavior

As of sensuctl 5.13.0, valid configuration sent to stdin does not result in creation or update of corresponding resources in the sensu-backend

0. set up a sensu-backend and sensuctl 5.13.0 or later, ideally one without any existing configuration (for ease of verifying step 2)
1. Use `cat` to output any known good Sensu Go configuration not already present in the backend,  and pipe it through `sensuctl create` (I've tried both license and asset resource types)
2. Note that the expected resource(s) are not present in the backend
3. Repeat the same with sensuctl 5.12.0 and note that the expected resource(s) are now present in the backend
4.

## Context

This regression is both confusing and potentially breaking in cases where the stdin capabilities of sensuctl are being used to apply changes in a programatic way. 

When providing configuration via stdin, sensuctl gives no output indicating that the operation succeeded or failed. Further more the exit status is 0, so there's really no way to know that things failed other than to verify them after attempting to apply configuration.

The first time i ran into this I thought it was user error -- invalid config or missing yaml document separators. Having run into this a few more times, I was inspired to test against older versions of sensuctl.

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): sensuctl 5.13.0, 5.13.2
* Installation method (packages, binaries, docker etc.): zip filez
* Operating System and version (e.g. Ubuntu 14.04): darwin

echlebek:
## What is this change?

This commit fixes a regression in sensuctl create, where stdin mode
was no longer working. A test has been added to ensure that this
regression does not appear again.

Closes #3318 

Yes

## Aside from unit/integration tests, please describe the e2e steps to verify this change.

`cat check.json | sensuctl create`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,7 @@ is used in `sensuctl asset list`.
 - Fix generic API client's `SetTypeMeta` method. The APIGroup is now correctly
 configured and by virtue unintended authorization denied errs are avoided.
 - Fixed a bug where checks would stop executing after a network error.
+- Fixed a bug where sensuctl create with stdin was not working.
 
 ## [5.13.2] - 2019-09-19
 
--- a/cli/commands/create/create.go
+++ b/cli/commands/create/create.go
@@ -132,6 +132,9 @@ func execute(cli *cli.SensuCli) func(*cobra.Command, []string) error {
 		if err != nil {
 			return err
 		}
+		if len(inputs) == 0 {
+			return processStdin(cli, client)
+		}
 		recurse, err := cmd.Flags().GetBool("recursive")
 		if err != nil {
 			return err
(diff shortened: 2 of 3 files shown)


palourde:
@echlebek Could you create and link a testrail case for the e2e test? It would probably go within the [`sensuctl` section](https://sensuinc.testrail.io/index.php?/suites/view/29&group_by=cases:section_id&group_order=asc&group_id=2713)

echlebek:
OK

echlebek:
https://sensuinc.testrail.io/index.php?/cases/view/11136

Issue #3202, opened 2019-08-14
Add API key support to sensuctl


echlebek:
## Feature Suggestion
Add API key support to sensuctl.

Create commands for sensuctl to facilitate the management of API keys. API keys should have CRUD abilities similar to other resources, but should not be creatable with `senusctl create`.

Link to spec: https://docs.google.com/document/d/1wdfAhajJsXCEmC0vquifkbdP5O33HSPH9PbjFLC46b4/edit#

## Context
We want to be able to manage API tokens that are persisted to the Sensu database.

PR #3324, opened 2019-10-08
Feature/apikeys sensuctl

nikkixdev:
## What is this change?

Adds sensuctl support for the `APIKey` resource.

**Note:** This PR appears to have changes included in https://github.com/sensu/sensu-go/pull/3303, but once that is merged and this is rebased, the diff should be much simpler to read. If you'd like to review anyway, please refer only to files in the `cli/..` package.

Closes https://github.com/sensu/sensu-go/issues/3202.
Requires https://github.com/sensu/sensu-go/pull/3303.

Yas.

Nah.

Nah

We good.

Many unit tests added and e2e done manually. TestRail tests should be done for the actual middleware.


cli/commands/apikey/help.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"github.com/sensu/sensu-go/cli"
+	"github.com/spf13/cobra"
+)
+
+// HelpCommand defines new parent
+func HelpCommand(cli *cli.SensuCli) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "apikey",
```
palourde: I suggest we use `api-key` here and in other subcommands since we did the same for the other compound words (e.g. `cluster-role-binding`). WDYT?

cli/commands/apikey/grant.go
```diff
@@ -0,0 +1,38 @@
...
+
+			apikey := &corev2.APIKey{
+				Username: args[0],
+			}
+
+			if err := cli.Client.Post(apikey.URIPath(), apikey); err != nil {
+				return err
+			}
+
+			fmt.Fprintln(cmd.OutOrStdout(), "Created")
```
palourde: Since the response of the API contains the Location header, should we use it to display the newly created API key? Otherwise a user would have to list all keys and figure out which one he just created, which is not optimal I believe.

cli/commands/apikey/help.go
nikkixdev: Ah, we don't use dashes in the api endpoint, but we do in the sensuctl command. Yes, I'll update that!

cli/commands/apikey/grant.go
nikkixdev: We'll have to write a custom client function for this, rather than refactoring the generic. But I would say it's worth it in this case.

Issue #3248, opened 2019-09-01
new tag  format vX.XX break compilation build.sh


realPy:
The build.sh use the last tag to catch the version number for compilation.
The regex use in version/cmd/version/version.go handle only X.XX
The bugs appears with the git tag v5.12 and must be tag as 5.12 like the other tags

palourde:
Hi @realPy, we've deprecated the `build.sh` script since it was no longer used and maintained by the team. Instead, we added some simple instructions on how to run and build from source: https://github.com/sensu/sensu-go#building-from-source

Hopefully this will make it easier for you!

Issue #3245, opened 2019-08-30
Document how to build the project in README.md


amdprophet:
As we can now build the project without the use of the `build.sh` & `build.ps1` scripts, we should add a "Building" section to `README.md` and document how to build the project using `go build`.

Issue #2971, opened 2019-05-17
Add build instructions to build sensu-go from source to Readme


asachs01:
## Expected Behavior

As an end user, I'd like to have instructions for building sensu-go from source, especially if want to use a purely open-source approach.

## Current Behavior

Instructions are currently missing, and only point to the docs site, which provide instructions for packages currently.

## Possible Solution

Write up some instructions

palourde:
Here's the current instructions we have: https://github.com/sensu/sensu-go/blob/master/CONTRIBUTING.md#building.

That being said, we should probably try to walk away from the `build.sh` script and provide simple Go/bash commands!

asachs01:
@palourde IMO those instructions would be better served living under https://github.com/sensu/sensu-go#installation, rather than in contributing. As it stands, it's not clear from an end-user perspective that I need to go to the contributing doc for instructions on compiling. 

FWIW, the build.sh script is great, but yeah, for me, I'd rather see exactly what I'm doing vs. reading through a script. I also found it _much_ easier to just do a `go build` or `go install`.

PR #3328, opened 2019-10-09
No longer use build.sh to build binaries

palourde:

## What is this change?

This PR removes all the build tools within the build.sh tool, since these are no longer working properly and instead document how Sensu Go should be ran and built from source.

As a side effect, it also does the following things:
- Removes the `sensu/sensu-go/version/cmd/version` package, which was only used by build.sh
- Removes all the unused code in the `sensu/sensu-go/version` package
- Adds a `version.Println()` function so the version output across components is consistent
- Prints `(devel)` as the version when none was provided during compilation instead of something like `sensuctl version #, build , built`
- Removes the `lint`, `dashboard`, `dashboard-ci` tasks in build.sh since they are broken

Closes https://github.com/sensu/sensu-go/issues/2971
Closes https://github.com/sensu/sensu-go/issues/3248
Closes https://github.com/sensu/sensu-go/issues/3245

I don't feel like it's necessary since it doesn't affect the packaged and compiled versions of Sensu Go we ship.

Nope

Nope

Nope

Manually tested all the build tools we offer.


README.md
```diff
@@ -39,6 +39,39 @@ to get started is with the official Docker image, sensu/sensu.
...
+The various components of Sensu Go can be manually built from this repository.
+You will first need [Go](https://golang.org/doc/install#install) installed.
+Then, you should clone this repository **outside** of the GOPATH since Sensu Go
+uses [Go Modules](https://github.com/golang/go/wiki/Modules).
+
+To compile and then run Sensu Go within a single step:
+```
+$ go run ./cmd/sensu-agent/main.go
+$ go run ./cmd/sensu-backend/main.go
+$ go run ./cmd/sensuctl/main.go
```
echlebek:
```suggestion
$ go run ./cmd/sensu-agent
$ go run ./cmd/sensu-backend
$ go run ./cmd/sensuctl
```

echlebek: This makes the instructions less fragile, since if we add more files or change the name of main.go, the command will still work.

version/version_test.go
```diff
@@ -1,289 +1,37 @@
...
-			}
-			if got, want := ver, test.exp; got != want {
-				t.Errorf("bad fullversion: got %q, want %q", got, want)
-			}
-		})
+	for _, tt := range tests {
+		Version = tt.version
+		BuildDate = tt.buildDate
+		BuildSHA = tt.buildSHA
+		Println(tt.component)
```
echlebek: Should this be removed?

calebhailey:
Awesome! 🙌

README.md
calebhailey: Can we also add a note that explains the minimum Golang version required, and these steps (for verbosity): 

```
$ go get github.com/sensu/sensu-go
$ cd $GOPATH/src/github.com/sensu/sensu-go
```

palourde: Ha, I don’t know why but I always thought go run didn’t supported directories. TIL

version/version_test.go
palourde: 🙈

README.md
palourde: We actually recommend now to not use the GOPATH but yeah, adding those steps for verbosity is definitely a good idea, I’ll adapt those!

version/version_test.go
palourde: Like discussed, it’s actually required; it uses this pattern to verify the output: https://golang.org/pkg/testing/#hdr-Examples

PR #3334, opened 2019-10-12
Rework sensuctl edit to work with most resources

### LINKED ISSUE #2265: Interactive CLI - Core Workflow
Background: 
Historically, the interactive CLI was built by hand and to serve the purpose of being a learning tool. In the time since its inception, the scope of it has grown introducing a maintenance and usability burden. Currently we have many fields that need to be confirmed with the checks. Any time a new feature is introduced, we have to manually update the interactive CLI to avoid breakage in the tool. 

Proposal going forward: 

1. Display a copy of the yaml template to the user for each resource individually. 
2. Allow user to modify the copy of the yaml template. 
3. The modified document will be sent via sensuctl create to the Sensu Backend. 

Implementation specification: 

1. For a given resource, serialize the resource's fixture to a temporary file. 
2. The template document format displayed will be compatible with the format sensuctl create expects. 
3. Invoke the editor defined by the user's EDITOR environment variable.
4. When the editor is closed and the file is saved, create the edited resource in the same way that sensuctl create does. 
5. TBD details - Instrument with Tessen. 

Potential implications: 
- Maintenance will be required if any new data types are introduced into the system. 
- We will not need to do maintenance if the fields are updated or removed.

### LINKED ISSUE #2772: sensuctl edit does not work in conjunction with --config-dir flag
## Expected Behavior
I should be able to edit a resource and specify a different config dir, e.g.

```
sensuctl edit check check-cpu --config-dir /Users/splourde/.config/sensu/sensuctl-enterprise
```

## Current Behavior
It returns the following error:

```
Error: GET "/api/core/v2/namespaces/default/checks/check-cpu": failed to request new refresh token; client returned 'The server returned the error: 401 Request unauthorized'
Error: couldn't get resource: exit status 1
exit status 1
```
## Possible Solution
`sensuctl edit` [calls itself](https://github.com/sensu/sensu-go/blob/ccd38e08baa8e4f2c95605f98d9e23e38c6a83a8/cli/commands/edit/edit.go#L57) using `os.Args[0]`, however if the --config-dir flag was passed, it's ignored so the executable called won't be properly configured and therefore won't be authenticated.

We could either make sure to include the config-dir (just like the format) or figure out how to use the CLI client.

See example above.

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): [master](https://github.com/sensu/sensu-go/commit/ccd38e08baa8e4f2c95605f98d9e23e38c6a83a8)

### LINKED ISSUE #3323: Re-work sensuctl edit
## Expected Behavior
Any resource that supports PUT can be edited with sensuctl edit.

## Current Behavior
Many resources cannot be edited. The tool is a hack that I created for the purposes of a demo. :grin: 

## Possible Solution
Use the sensuctl dump machinery to re-implement the edit tool.

`sensuctl edit user admin`

## Context
QA

echlebek:
## What is this change?

Improves `sensuctl edit` somewhat.

Closes #3323 
Closes #2772 
Closes #2265 

The tool only worked with resources that had an `info` command.

Yes

I only did marginal amounts of manual testing, but I did test that users can now be edited.


echlebek:
We should test this with the enterprise resources, and special resources that don't follow the namespace/name convention.

It works for events as expected.

echlebek:
Hmm, I'd argue that it's actually a bug for these resources to fail validation when they have a namespace set.

If they didn't have the field at all, the namespace field would simply be ignored by the unmarshaler.

echlebek:
Either way, I'll add a special case to deal with this, but there's probably no good way to suppress the namespace field from being rendered.

echlebek:
Looks like the namespace field isn't rendered if blank, but unfortunately I don't have a good way to handle this for enterprise resources that have no namespace.

echlebek:
It will work with enterprise resources that have namespaces.

echlebek:
@palourde IMO those resources should simply ignore a namespace when one is set, that would make this PR work properly with them.

palourde:
@echlebek @nikkixdev I've successfully tested this change:

```bash
# checks
$ sensuctl check info check-echo --format yaml > check.yml
$ sensuctl edit check check-echo # not modifying it
$ sensuctl check info check-echo --format yaml > check1.yml
$ diff check.yml check1.yml
$

# cluster role bindings
$ sensuctl cluster-role-binding info cluster-admin --format yaml > binding.yml
$ sensuctl edit clusterrolebindings cluster-admin
$ sensuctl cluster-role-binding info cluster-admin --format yaml > binding1.yml
$ diff binding.yml binding1.yml
$

# authentication provider
$ sensuctl auth info okta --format yaml > provider.yml
$ sensuctl edit authentication/v2.OIDC okta
$ sensuctl auth info okta --format yaml > provider1.yml
$ diff provider.yml provider1.yml
$
```

Please take a quick look and 👍 if you're okay with those changes!

PR #3336, opened 2019-10-15
Fix a panic in wizard bus

echlebek:
## What is this change?

This commit guards against a panic when sending to a closed wizard topic.

In some cases, Sensu can crash after an agent session is terminated.

Yes

The change is unverified as it's very difficult to trigger the conditions that lead to this problem. However, we have verified that in production instances of Sensu, a panic occurs that can be traced back to this chunk of code.

I feel that the change is safe enough to deploy without a reproduction.

### DIFF
--- a/backend/messaging/wizard_topic.go
+++ b/backend/messaging/wizard_topic.go
@@ -24,11 +24,23 @@ func (t *wizardTopic) Send(msg interface{}) {
 
 	for _, subscriber := range subscribers {
 		topicCounter.WithLabelValues(t.id).Set(float64(len(subscriber.Receiver())))
-		select {
-		case subscriber.Receiver() <- msg:
-		case <-t.done:
-			return
-		}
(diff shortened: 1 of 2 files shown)


backend/messaging/wizard_topic.go
```diff
@@ -24,11 +24,23 @@ func (t *wizardTopic) Send(msg interface{}) {
...
+
+// safeSend can attempt to send to a closed channel without panicking.
+// This is only necessary because of a design flaw in wizard bus. Do not
+// use this elsewhere.
+//
+// The topic reads the subscribers and then releases its lock, in Send(). In rare cases,
+// cancelling a subscription can lead to a send on a closed channel.
+func safeSend(c chan<- interface{}, message interface{}, done chan struct{}) {
+	defer func() {
+		recover()
```
golangcibot: Error return value of `recover` is not checked (from `errcheck`)

PR #3345, opened 2019-10-21
Asymmetric JWTs

palourde:
## What is this change?

This PR adds support for asymmetric JWTs. Additionally, it does the following things:

- JWTs validity is decreased from 15 to 5 minutes
- The JWT access list is removed
- JWTs now include a `iss` field, which specifies the URL of the server that delivered the JWT. Clients should always try to use this URL first to renew access tokens, which sensuctl now does.

Closes https://github.com/sensu/sensu-enterprise-go/issues/689

Nope!

Nope!

New documentation will be required: https://github.com/sensu/sensu-docs/issues/1860

Added the required unit tests coverage. Also added some [test cases in TestRail](https://sensuinc.testrail.io/index.php?/suites/view/29&group_by=cases:section_id&group_order=asc&group_id=3046), will [should be automated](https://github.com/sensu/sensu-go-qa-crucible/issues/82).


backend/backend.go
```diff
@@ -292,6 +293,9 @@ func Initialize(config *Config) (*Backend, error) {
 		clusterVersion = b.Etcd.GetClusterVersion()
 	}
 
+	// Load the JWT key pair
+	jwt.LoadKeyPair(viper.GetString(FlagJWTPrivateKeyFile), viper.GetString(FlagJWTPublicKeyFile))
```
golangcibot: Error return value of `jwt.LoadKeyPair` is not checked (from `errcheck`)

cli/client/authentication.go
```diff
@@ -74,12 +76,43 @@ func (client *RestClient) Logout(token string) error {
...
+	_, _, _ = parser.ParseUnverified(tokens.Access, claims)
+
+	switch issuer := claims.Issuer; {
+	case issuer != "":
+		// Attempt to renew the access token with the issuer
+		res, err = client.R().
+			SetHeader("Content-Type", "application/json").
+			SetBody(map[string]string{"refresh_token": tokens.Refresh}).
+			SetResult(&corev2.Tokens{}).
+			Post(claims.Issuer + url)
```
ccressent: I understand why `iss` is a URL instead of just a host now. I'd like to suggest an alternative approach.

Default to `https`. If that fails, try on `http` and issue a big fat warning stating that this is completely insecure when using JWTs. That way `iss` is just the host and we always try to use `https` whenever we can. JWTs over http make as much security sense as passwords over http anyway.

cli/client/authentication.go
```diff
@@ -74,12 +76,43 @@ func (client *RestClient) Logout(token string) error {
...
-		Post("/auth/token")
+func (client *RestClient) RefreshAccessToken(tokens *corev2.Tokens) (*corev2.Tokens, error) {
+	var err error
+	var res *resty.Response
+	url := "/auth/token"
+
+	// Parse the access claims so we can determine the issuer URL
+	parser := new(jwt.Parser)
+	claims := &corev2.Claims{}
+	_, _, _ = parser.ParseUnverified(tokens.Access, claims)
```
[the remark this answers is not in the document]
palourde: Because the client (sensuctl) doesn't necessarily have access to the secret key or the public key to validate the JWT; it will still be validated on the server side though

cli/client/authentication.go
palourde: We can already detect whether the token was issued over HTTP or HTTPS. Right now we can't serve both HTTP _and_ HTTPS traffic over the same API port so trying to request the HTTPS version of an HTTP API that we already know is not secure doesn't do anything is most scenarios.

That being said, your comment about authentication over HTTP totally makes sense but I would argue that this PR doesn't change the current reality and to avoid a scope creep, I'd prefer to open a new issue so we can discuss about the right approach to take to encourage users and operators to use HTTPS.

backend/apid/routers/authentication.go
```diff
@@ -122,3 +123,16 @@ func (a *AuthenticationRouter) token(w http.ResponseWriter, r *http.Request) {
...
+// issuerURL determines the URL used by the client to authenticate and renew its
+// access token, so it can be later re-used by the client in case the access
+// token is used against a different cluster
+func issuerURL(r *http.Request) string {
+	issuerURL := r.Host
+	if r.TLS == nil {
+		issuerURL = "http://" + issuerURL
+	} else {
+		issuerURL = "https://" + issuerURL
+	}
```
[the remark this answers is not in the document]
palourde: Closing in favor of https://github.com/sensu/sensu-go/pull/3345#discussion_r337160715

PR #3351, opened 2019-10-21
Use serializable reads for events

### LINKED ISSUE #3349: Sensu events can use serialized reads instead of linearized reads
According to https://github.com/etcd-io/etcd/blob/master/Documentation/op-guide/performance.md, using serialized reads can confer a large performance advantage.

Sensu stresses etcd very heavily when a lot of events are active. Since events are typically localized to a single backend, there is very little chance for a pipeline execution to actually get stale event data.

Using serialized reads would mean that users accessing the event API could potentially get stale reads, but this is likely not a huge concern from a correctness standpoint.

echlebek:
## What is this change?

Uses `clientv3.WithSerializable` for event reads. This trades the possibility of stale event reads for performance.

I believe that for our use case, this will not have a correctness impact.

Closes #3349 

One should be added once the change is verified.

In progress.

### DIFF
--- a/backend/store/etcd/event_store.go
+++ b/backend/store/etcd/event_store.go
@@ -182,7 +182,7 @@ func (s *Store) GetEventByEntityCheck(ctx context.Context, entityName, checkName
 		return nil, err
 	}
 
-	resp, err := s.client.Get(ctx, path, clientv3.WithPrefix())
+	resp, err := s.client.Get(ctx, path, clientv3.WithPrefix(), clientv3.WithSerializable())
 	if err != nil {
 		return nil, err
 	}


backend/store/etcd/event_store.go
```diff
@@ -182,7 +182,7 @@ func (s *Store) GetEventByEntityCheck(ctx context.Context, entityName, checkName
 		return nil, err
 	}
 
-	resp, err := s.client.Get(ctx, path, clientv3.WithPrefix())
+	resp, err := s.client.Get(ctx, path, clientv3.WithPrefix(), clientv3.WithSerializable())
```
palourde: Could we also add this option to `GetEvents` & `GetEventsByEntity`, or there's no real benefit?

echlebek: We could, but those methods aren't used by the event pipeline, so I expect that adding the option there would have less of an effect.

PR #3354, opened 2019-10-23
Re-use leases instead of creating new ones

### LINKED ISSUE #3355: Keepalives and check TTLs create too many leases
## Expected Behavior
Keepalives and check TTLs should create at most one concurrent lease.

## Current Behavior
Leases are created and then left to expire, instead of being kept alive.

## Possible Solution
Store lease IDs and use etcd keepalives on them instead of granting new ones.

echlebek:
## What is this change?

Recent testing has revealed that we are creating leases at a ferocious rate, and they are not getting cleaned up quickly enough. For instance, in a 6000 entity cluster, the number leases from keepalives alone was over 40,000.

This change tracks leases that are in-use, and attempts to keep them alive if they are still valid. If they are not, a new lease is granted instead.

Etcd <3.4 has a performance problem when many leases are present or expiring, and I believe it is causing stability problems.

Closes #3355

Maybe, more testing needed.

No docs changes needed.

The existing test suite should suffice. I'm going to do some manual testing to see if we get a more reasonable number of leases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ successfully reconnect when using --no-embed-etcd.
 - Check TTL switches are now correctly buried when associated events and entities
 are deleted.
 - Keepalive switches are now correctly buried when the keepalive event is deleted.
+- Sensu now uses far fewer leases for keepalives and check TTLs, resulting in a
+stability improvement for most deployments.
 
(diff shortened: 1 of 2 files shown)


echlebek:
Testing with sensu-perf reveals that this change has a profoundly positive impact, improving cluster stability under load by a large amount.

PR #3361, opened 2019-10-25
Refactor entities and events routers for enterprise entity limit middleware

nikkixdev:
## What is this change?

Refactors the entities and events routers so that enterprise middlewares can be injected into specific routes.

Required by https://github.com/sensu/sensu-enterprise-go/issues/710.

Nah, purely internal.

Nah.

Nah.

We good.

Unit/integration tests. Verified this against sensu-enterprise-go branch.


backend/apid/apid.go
```diff
@@ -180,6 +182,28 @@ func CoreSubrouter(router *mux.Router, cfg Config) *mux.Router {
 	return subrouter
 }
 
+// EntityLimitedCoreSubrouter initializes a subrouter that handles all requests
+// coming to /api/core/v2 that must be gated by entity limits.
+func EntityLimitedCoreSubrouter(router *mux.Router, cfg Config) *mux.Router {
```
palourde: Just so I better understand, the idea here is that we would inject a middleware from the enterprise codebase for this subrouter right?

If that assumption is true, maybe it's too naive but I was just thinking that instead of creating new routers for only specific routes (POST & PUT essentially), we could move the `EntitiesRouter` & `EventsRouter` routers in this new subrouter and then, in the enterprise middleware, we use the [`AuthorizationAttributes`](https://github.com/sensu/sensu-go/blob/ca257bf76a58f1c698bfe16c174ebb45e87c1343/backend/authorization/interface.go#L21:6) or even the HTTP method to determine if we need to act on this request. WDYT?

nikkixdev: Ugh honestly yeah I didn't even think about that. That's much less clunky. I'll revise!

nikkixdev: Wow I like this diff so much better!

PR #3362, opened 2019-10-29
Export placeholder entity limiter middleware for agent websocket connection

nikkixdev:
## What is this change?

Adds a placeholder middleware in core that will be overwritten by enterprise for entity limits on the agent websocket connection.

Required by https://github.com/sensu/sensu-enterprise-go/issues/710.

Nah, purely internal.

Nah.

Nah, thanks @palourde for figuring out this workaround!

We good.

Tested it against enterprise branch.


nikkixdev:
Not sure why circleci is failing on a test that was introduced without fail in another PR. I'll use this time to investigate if it's flaky.

Issue #3350, opened 2019-10-21
[Feature Request] Expose "state" field in check history


roganartu:
The `state` field of [Events](https://docs.sensu.io/sensu-go/5.15/reference/events/#top-level-attributes) describes whether the event is currently considered to be flapping. Include this field in the check history to enable more advanced filtering and handler behaviour, mimicking the flapping start/end notifications that Nagios sends.

## Feature Suggestion
Sensu follows the same algorithm for flap detection as Nagios as outlined here: https://assets.nagios.com/downloads/nagioscore/docs/nagioscore/3/en/flapping.html

One of the features that Nagios exposes is notifications for start/end of flapping. This allows users to perform actions when flapping starts/end. Exposing this `state` field in check history would enable Sensu filters and handlers to differentiate between "this check is flapping" and "this specific event is when the check started/ended flapping".

## Possible Implementation
This could be added to the `Check.MergeWith` method that handles updating check history:
https://github.com/sensu/sensu-go/blob/8a21de8d0cd336d81418bc1cb4c55ed855cfe1bf/api/core/v2/check.go#L229-L249

The field could be added as the `state` string as currently stored in the check, or a `flapping` bool, or similar.

## Context
We have a slack channel that some alerts are sent to depending on filters etc. Instead of simply filtering events where `state=="flapping"` we would like to be able to send "X started flapping" and "X finished flapping: current status is Y" notifications when flapping starts and ends.

echlebek:
We'd like to pick this up internally and add the "state" field to the event history, rather than only a flapping boolean, as we feel this would address more use cases, being more general.

We will also take a look at the issue with low_flap_threshold. Thanks!

PR #3363, opened 2019-10-29
Add flapping field to check history

roganartu:
## What is this change?

Implements #3350, but with a `Flapping` bool instead of exposing the actual state string.

See detailed description in #3350. I also added two new properties to the event object exposed to filters (`is_flapping_start` and `is_flapping_end`) to make it easier for users to create the Nagios-like flapping start/end notifications.

Additionally fixed a bug I found while implementing where flapping events were being erroneously marked as no-longer-flapping when `total_state_change < high_flap_threshold` instead of when `total_state_change < low_flap_threshold` as documented.

Yes, added.

Nope.

What is the correct way to generate the `*.pb.go` model files? I used `go generate ./api/core/v2` which resulted in files that were functional enough to run the tests I added and also do some adhoc testing in a docker container, but it generated so many unrelated changes that I decided to completely omit any changes to those files until I get clarification (or alternatively, leave it up to the Sensu team to generate the files and update this PR).

New documentation will be required. I have not written any, save for docstrings.

Added some tests.

Also did adhoc testing with docker, using the attached config files. Confirmed that the `flapping` field was set as expected, that the `low_flap_threshold` bug found no longer occurs, and that a filter I created using the added flapping properties triggered when expected.

Transitioning from not flapping to flapping:
```
      {
        "status": 1,
        "executed": 1572375719,
        "flapping": false
      },
      {
        "status": 2,
        "executed": 1572375724,
        "flapping": true
      }
    ],
    "issued": 1572375724,
    "output": "Status: 2\n",
    "state": "flapping",
    "status": 2,
    "total_state_change": 26,

backend1_1  | {"assets":null,"check_name":"flap_test","check_namespace":"default","component":"pipelined","entity_name":"foo","entity_namespace":"default","filter":"flap_test","level":"debug","msg":"allowing event that matches filter","time":"2019-10-29T18:44:14Z"}
backend1_1  | {"check_name":"flap_test","check_namespace":"default","component":"pipelined","entity_name":"foo","entity_namespace":"default","filter":"flap_test","handler":"flap_test","level":"debug","msg":"allowing event","time":"2019-10-29T18:44:14Z"}
backend1_1  | {"check_name":"flap_test","check_namespace":"default","component":"pipelined","entity_name":"foo","entity_namespace":"default","handler":"flap_test","level":"info","msg":"sending event to handler","time":"2019-10-29T18:44:14Z"}
```

Confirming state is still `flapping` when `low_flap_threshold > total_state_change > high_flap_threshold`:
```
      {
        "status": 0,
        "executed": 1572376019,
        "flapping": true
      }
    ],
    "issued": 1572376019,
    "output": "",
    "state": "flapping",
    "status": 0,
    "total_state_change": 17,
    "last_ok": 1572375714,
```

Transitioning from flapping to not flapping:
```
      {
        "status": 0,
        "executed": 1572376029,
        "flapping": true
      },
      {
        "status": 0,
        "executed": 1572376034,
        "flapping": false
      },
      {
        "status": 0,
        "executed": 1572376039,
        "flapping": false
      }
    ],
    "issued": 1572376039,
    "output": "",
    "state": "passing",
    "status": 0,
    "total_state_change": 0,
    "last_ok": 1572376039,

backend1_1  | {"component":"etcd","level":"info","msg":"histEntry: {Status:0 Executed:1572376034 Flapping:false XXX_NoUnkeyedLiteral:{} XXX_unrecognized:[] XXX_sizecache:0}\n","pkg":"","time":"2019-10-29T19:07:14Z"}
backend1_1  | {"check_name":"flap_test","check_namespace":"default","component":"pipelined","entity_name":"foo","entity_namespace":"default","hooks":null,"level":"debug","msg":"received event","silenced":[],"time":"2019-10-29T19:07:14Z","timestamp":1572376034}
backend1_1  | {"assets":null,"check_name":"flap_test","check_namespace":"default","component":"pipelined","entity_name":"foo","entity_namespace":"default","filter":"flap_test","level":"debug","msg":"allowing event that matches filter","time":"2019-10-29T19:07:14Z"}
backend1_1  | {"check_name":"flap_test","check_namespace":"default","component":"pipelined","entity_name":"foo","entity_namespace":"default","filter":"flap_test","handler":"flap_test","level":"debug","msg":"allowing event","time":"2019-10-29T19:07:14Z"}
```

Using resources:
```
{
  "type": "Entity",
  "api_version": "core/v2",
  "metadata": {
    "name": "foo",
    "namespace": "default"
  },
  "spec": {
    "entity_class": "proxy",
    "system": {
      "hostname": "foo"
    },
    "subscriptions": [
      "entity:foo"
    ],
    "deregister": false
  }
}
{
  "type": "Handler",
  "api_version": "core/v2",
  "metadata": {
    "name": "flap_test",
    "namespace": "default"
  },
  "spec": {
    "command": "echo triggered",
    "filters": [
      "flap_test"
    ],
    "timeout": 5,
    "type": "pipe"
  }
}
{
  "type": "EventFilter",
  "api_version": "core/v2",
  "metadata": {
    "name": "flap_test",
    "namespace": "default"
  },
  "spec": {
    "action": "allow",
    "expressions": [
      "event.is_flapping_start || event.is_flapping_end"
    ]
  }
}
{
  "type": "CheckConfig",
  "api_version": "core/v2",
  "metadata": {
    "name": "flap_test",
    "namespace": "default"
  },
  "spec": {
    "command": "sh -c 'exit 0; x=$(shuf -i 0-2 -n 1); echo Status: $x; exit $x'",
    "interval": 5,
    "publish": true,
    "handlers": ["flap_test"],
    "subscriptions": [
      "entity:foo"
    ],
    "proxy_entity_name": "foo",
    "high_flap_threshold": 25,
    "low_flap_threshold": 10
  }
}
```
where making the check flap simply involved removing the prefixed `exit 0;` after 21 occurrences.


api/core/v2/flapping.go
```diff
@@ -12,7 +12,8 @@ func isFlapping(check *Check) bool {
 	}
 
 	// Is the check already flapping?
-	if check.State == EventFlappingState {
```
roganartu: This is why `low_flap_threshold` was broken before, as this function is called from inside `updateCheckState` but before `check.State` is updated (so it would've always been comparing `EventFlappingState` to an empty string).

echlebek:
Hi @roganartu, running `go generate` on the `api/core/v2` package is indeed the correct way to generate the protobufs. If you're seeing a lot of differences you may have a different version of protoc installed. That's a bit of a tricky one, we've had some stumbles internally with that as well.

If we incorporate this change we'll need to make associated changes in our enterprise repo as well, to make sure things stay consistent with the postgresql event store, so it may take us a while to merge this.

That said, this is a very welcome PR and a great call-out that our flapping functionality could use some love. Thanks very much for this submission.

roganartu:
Thanks for the kind words @echlebek.

I did some more digging today re successfully generating `pb.go` files without a diff that effects every single model.

The key turned out to be that part of the diff at the top of every file was this change:
```
-const _ = proto.ProtoPackageIsVersion2 // please upgrade the proto package
+const _ = proto.ProtoPackageIsVersion3 // please upgrade the proto package
```
I confirmed that `go generate` was correctly using the vendored versions of all the proto libs.

What it looks like happened is at some point someone updated the vendored version of all your proto packages without running `go generate`, so you're currently running `v1.3.x` of the proto libs with models generated using `v1.2.x`. They're pretty good with backwards compatibility, which I suspect is why no one has noticed yet.

The following fixed the issue and resulted in the expected diff, only adding changes related to the new field:
```
$ go get github.com/gogo/protobuf/protoc-gen-gogofast@v1.2.1
$ go install github.com/gogo/protobuf/protoc-gen-gofast
$ go generate ./api/core/v2
```
The generated files from this successfully build and run.

I will update the PR with the generated changes now that they are correct. I will omit any changes to `go.mod` and let you or someone else coordinate fixing that.

portertech:
Why not include "state" within check history items? The field would capture more than the flapping boolean.

roganartu:
> Why not include "state" within check history items? The field would capture more than the flapping boolean.

At first glance it would seem that way, but afaict there's only 3 unique states currently anyway (passing, failing, flapping): https://github.com/sensu/sensu-go/blob/50fbab6bbe9730589bcf2ad9830acf0d0b570619/api/core/v2/event.go#L19-L26

Passing and failing are just status 0/non-0, so there's only actually 1 bit of information missing from each history entry making a boolean sufficient. If there's plans to have more than 3 unique values for the `state` field then I agree a boolean makes less sense than just storing the raw string. I made the assumption in this PR that space is at a premium here, given the size constraints `etcd` presents, and the fact that whatever is chosen is multiplied by 21 in every full-history event.

echlebek:
We will be merging this, but it's getting pushed to 5.19 so that the enterprise edition can keep parity.

roganartu:
> It looks like this commit re-orders the event history, but that is not correct - the event history should always be the order that events were received in, regardless of timestamp or any other possible ordering criteria.
> 
> We'll need to change that before merging.

I can fix that this week.

echlebek:
@roganartu we've changed sensu-go so that it order events FIFO-style now. A nice side effect of this is that it's now much easier to set the history item.

api/core/v2/check.go
```diff
@@ -256,6 +256,12 @@ func (c *Check) MergeWith(prevCheck *Check) {
 	c.Occurrences = prevCheck.Occurrences
 	c.OccurrencesWatermark = prevCheck.OccurrencesWatermark
 	updateCheckState(c)
+
+	// This has to be done after the call to updateCheckState, as that function is what
+	// sets the value for c.State that is used below, but the order can't be switched
+	// around as updateCheckState relies on the latest item (specifically, its status)
+	// being present in c.History.
+	c.History[len(c.History)-1].Flapping = c.State == EventFlappingState
```
roganartu: nice this is much cleaner than it was before the FIFO ordering was fixed

PR #3366, opened 2019-10-29
[GraphQL] Ensure generation of schema is idempotent

jamesdphillips:
## What is this change?

Fixes bug in GraphQL service definition that could lead to a panic when trying to regenerate the schema.

Allows any upstream implementations or extensions to add types without fear of a panic.

Given scope, it is probably unnecessary.

Added an assertion to integration test.


graphql/service.go
```diff
@@ -76,6 +78,7 @@ func (service *Service) RegisterInterface(t InterfaceDesc, impl InterfaceTypeRes
 func (service *Service) RegisterObject(t ObjectDesc, impl interface{}) {
 	cfg := t.Config()
```
echlebek: This call is now redundant, can we remove it from here and other similar places?

PR #3372, opened 2019-10-31
Expose GlobalID funcs for upstream implementations

jamesdphillips:
## What is this change?

Will allow upstream implementations to register globally unique
identifiers for their resources, as well as adding the ability to fetch
resources based on said IDs.

Brass tacks:

* Makes `RegisterTranslator` func public.
* Exposes `NodeRegister` on graphql.Service.
* Changes `globalid` package's interface to include extra contextual details.
* Adds public `Encode` func to globalid package.

sensu/sensu-enterprise-go#702

### DIFF
--- a/backend/apid/graphql/globalid/components.go
+++ b/backend/apid/graphql/globalid/components.go
@@ -44,14 +44,24 @@ type Components interface {
 	// this is the resource's name.
 	UniqueComponent() string
 
-	// Extra return value for given key within extras dict
-	Extra(key string) string
-
-	// SetExtra sets value for given key within extras dict
-	SetExtra(key, val string)
+	// Extras returns interface with access to extra values
+	Extras() Values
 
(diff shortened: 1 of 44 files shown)


backend/apid/graphql/globalid/components.go
```diff
@@ -139,3 +193,13 @@ func omitEmpty(in []string) (out []string) {
...
 }
+
+type dict struct {
+	url.Values
+}
+
+func (d *dict) Clear() {
+	for key := range d.Values {
+		d.Values.Del(key)
+	}
```
echlebek: This can be rewritten as:

```
d.Values = make(url.Values, len(d.Values))
```

jamesdphillips:
> Any concerns about adding to the Components interface? It will likely break any upstream implementations that rely on it.

Only will be a problem if the upstream changes are not merged.

Issue #3370, opened 2019-10-31
[Feature Request] Include check output in history


roganartu:
Add output from check to history.

## Feature Suggestion
Currently, the history only includes two fields: `status` and `executed`. #3363 proposes adding a `flapping` boolean field to indicate whether the check was flapping at a particular point in time, allowing for more powerful filtering (specifically, start and end flap notifications similar to Nagios).

This request is for adding the `event.check.output` field to each entry in `event.check.history`.

## Possible Implementation
Implementation would be basically identical to that in #3363, ie add a new field and populate it on check merge.

One potential complication that may need to be discussed is whether there should be a max size for the added field, where anything above is truncated by the check merge? History size is hardcoded to 21 entries per event, so this is possibly not a major concern.

## Context
In a similar vein to #3363 enabling more intelligent filtering, there are situations where it would be useful for filters to have access to previous check output in order to make better decisions.

### Generic examples

- Perform a certain action for stale checks, but not on the first occurrence. With the current check history it is not possible to differentiate between a regular warning and one generated due to TTL being exceeded.
- Match a filter only if a certain pattern has been seen in a number of outputs in history. Consider for example a disk check where you wish to ignore a particular mountpoint for certain machines. Check output is likely to contain some kind of structured `mountpoint, usage` output, but having a separate check for each mountpoint makes little sense, nor does baking logic into the check to whitelist certain mountpoints for certain machines. `occurrences` doesn't help with this case either as it also doesn't allow you to historically differentiate between `/mnt/a -> warn | /mnt/b -> ok` and `/mnt/a -> ok | /mnt/b -> warn`. A filter could do this with access to output history, and seems like the logical place for it (at least to me).

### Specific, motivating example
We have a large number of checks that are logically grouped but that make little sense to have a separate check name for each. I previously proposed more complex grouping functionality be added in #2750 but it seems unlikely to happen due to entrenched assumptions throughout various Sensu components about the uniqueness of hostname+check name.

We have solved this by storing an encoded JSON object in our check outputs which contains more detailed information about each component of a check. Take the following (decoded) output from a ping check as a concrete example of what this looks like:
```
{
    "foo": {
        "status": 2,
        "data": {
            "rta": {
                "threshold": {
                    "critical": "disabled",
                    "warning": "disabled"
                },
                "status": "ok",
                "value": 37.1,
                "description": "ok"
            },
            "packet_loss": {
                "threshold": {
                    "critical": ">= 60%",
[...]

PR #3373, opened 2019-11-01
Add output field to check history

roganartu:
## What is this change?

Implements #3370 by adding an `output` field to check history, containing the output from that specific event.

See full description in #3370.

Yes, added.

Nope.

Nope.

Doc update probably necessary, I haven't done it.

I didn't add any new tests, but I can if you think they are necessary.

I confirmed by spinning up a the `docker-compose.yml` and inspecting a keepalive event:
```
$ sensuctl event info agent1 keepalive --format json
...
    "history": [
      {
        "status": 0,
        "executed": 1572619087,
        "output": "Keepalive last sent from agent1 at 2019-11-01 14:38:07 +0000 UTC"
      },
      {
        "status": 0,
        "executed": 1572619092,
        "output": "Keepalive last sent from agent1 at 2019-11-01 14:38:12 +0000 UTC"
      }
    ],
    "issued": 1572619092,
    "output": "Keepalive last sent from agent1 at 2019-11-01 14:38:12 +0000 UTC",
    "state": "passing",
    "status": 0,
...
```
You can see that the last entry in the history matches the current output, and the previous one is from 5s earlier which is the keepalive interval in the docker image.


echlebek:
We would like to prevent stored events from becoming too large, especially when they are being stored in etcd, which has a limitation on value size.

I would suggest storing output out-of-band of Sensu if more complex analysis of the event needs to be done, or doing some continuous analysis via metrics handlers.

PR #3375, opened 2019-11-04
Update changelog for 5.14.2 release

nikkixdev:

I went through the changelog entries for each cherry picked commit on the release/5.14 branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 - Added sensuctl commands to manage the `APIKey` resource.
 - Added support for api keys to be used in api authentication.
 
+## [5.14.2] - 2019-11-04
+
 ### Changed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 - Added sensuctl commands to manage the `APIKey` resource.
 - Added support for api keys to be used in api authentication.
```
ccressent: I'm not sure how/why we have API keys under the Unreleased header... I don't think that should be here at all.

nikkixdev: This change will eventually get merged into master, no? Was hoping to avoid more manual changelog resolutions, but happy to remove this if it doesn't make sense.

echlebek: IMO it is harmless to have stuff in the unreleased header, even if it doesn't exist in the branch.

Issue #3374, opened 2019-11-01
[Feature Request] Support multiple silences for a single entity/subscription


roganartu:
Add support for multiple (potentially overlapping) silences for a single entity/subscription.

## Feature Suggestion
Users should be able to have two silences with different start/end times, but that are otherwise identical.

## Possible Implementation
I don't think this is trivial to achieve due to the way silences are currently structured and processed.

There are strong assumptions about both the uniqueness of names, and the structure of those names are currently used to process the silence. See https://github.com/sensu/sensu-go/issues/3356#issuecomment-545471799 for an example bug this approach has caused however, so perhaps the design of silences should be revisited.

## Context
There are many scenarios where this behaviour is desirable, and arguably necessary. Nagios (and I think every other monitoring tool I've ever used) allows you to have multiple silences for a single check, but Sensu does not.

Duplicate silences with different start/end dates are very useful for controlling alerts and responding to incidents. Consider the following simple-but-common scenario:

1) User has planned some work for some point in the future, and proactively schedules downtime for the checks/entities expected to be affected by creating a silence with a start date in the future.
2) Before that silence time arrives, an incident occurs for which the responding user sets an immediate downtime for the affected check/entity in order to silence alerts while they work to remediate the issue.

Doing this in other monitoring tools involves simply creating a new downtime with a start date set to now. Doing this in Sensu as it currently stands requires either deleting the existing silence first, or editing the existing silence to change the date to start immediately, but both of these require noting the previous silence details in order to recreate them after the fact.

### Example Current Output
```
 sensuctl silenced create --interactive
? Namespace: default
? Subscription: 
? Check: keepalive
? Begin time: now
? Expiry in Seconds: 3600
? Expire on Resolve: No
? Reason: Testing
Created

 sensuctl silenced list                
     Name       Subscription     Check            Begin          Expire   ExpireOnResolve   Creator   Reason    Namespace  
 ───────────── ────────────── ─────────── ───────────────────── ──────── ───────────────── ───────── ───────── ─────────── 
  *:keepalive                  keepalive   01 Nov 19 11:28 EDT   59m55s   false             admin     Testing   default    

 sensuctl silenced create --interactive
? Namespace: default
? Subscription: 
? Check: keepalive
? Begin time: now
? Expiry in Seconds: 4000
? Expire on Resolve: No
? Reason: Testing 2
Error: resource already exists
```

roganartu:
[...]

PR #3376, opened 2019-11-04
Correctly apply silences to events from proxy entities

### LINKED ISSUE #3356: Silences on subscriptions don't always work
## Expected Behavior
If I set a silence on `staging:*` all hosts with this subscription should be silenced

## Current Behavior
There are a number of subscriptions between the various entities checking in. I'm not sure it this is affecting anything but it seems worth noting.

If I set a silence in the UI on `staging:*`, the silence gets listed but the entities aren't showing up as silenced. Other subscriptions i've tested seem to work fine. Additionally `staging` doesn't show up in the dropdown of subscriptions. There are possibly others missing too, which might exhibit the same issue.

Unsure how to reproduce on another system. I have 25+ subscriptions between the various entities. 
Not all of these appear in the dropdown on the silences UI page. 
Setting a silence on one that doesn't appear doesn't seem to actually set a silence.

## Your Environment
Ubuntu 18.04
Sensu 5.14.0-6578
79 entities
~ 4-8 subscriptions per entity
89 unique subscriptions total

roganartu:
Fixes #3356 by removing dependence on silence name for applying silences.

## What is this change?

There is a longer description in this comment: https://github.com/sensu/sensu-go/issues/3356#issuecomment-545471799

The tl;dr is that silences currently depend heavily on the name of the silence to determine whether a silence applies. The name contains the subscription, check, and may or may not have wildcards denoted with `*`.

This PR removes that dependency, instead checking the `subscription` and `check` properties of the silence directly.

Additionally, it adds better support for proxy entities, where subscriptions are used differently to non-proxy entities. Subscriptions on proxy entities may be used to target checks (via the `proxy_entity_attributes` field in check config), while the subscriptions actually attached the check are used to target the agent the check should run on. This prevented silences for them applying correctly.

I have documented all these cases extensively in the code via comments and some additional tests for these missing cases.

See above. I chose to remove the dependency on the `name` property specifically because that felt like a prerequisite for being able to address #3374.

Yes, added.

Nope

Nope, existing tests caught a few issues with the first pass of my implementation that I hadn't considered which was :ok_hand: 

New documentation explaining the difference between how silences are applied proxy and non-proxy entities is probably required. The comments in the code should be sufficiently detailed to inform that for whoever writes it.

- Existing tests pass
- Added new tests for proxy cases
- Ran adhoc test to confirm silence for non-proxy entity still applied as expected
```
 sensuctl create -f check.json

 sensuctl silenced create --interactive
? Namespace: default
? Subscription: switches
? Check: *
? Begin time: now
? Expiry in Seconds: -1
? Expire on Resolve: No
? Reason: testing PR
Created

 sensuctl silenced list
     Name      Subscription   Check          Begin          Expire   ExpireOnResolve   Creator     Reason     Namespace
 ──────────── ────────────── ─────── ───────────────────── ──────── ───────────────── ───────── ──────────── ───────────
  switches:*   switches       *       04 Nov 19 10:37 EST   -1s      false             admin     testing PR   default

 sensuctl event list
  Entity     Check                                  Output                                Status   Silenced             Timestamp
 ──────── ─────────── ────────────────────────────────────────────────────────────────── ──────── ────────── ───────────────────────────────
  agent1   foo                                                                                 1   true       2019-11-04 10:37:36 -0500 EST
  agent1   keepalive   Keepalive last sent from agent1 at 2019-11-04 15:37:35 +0000 UTC        0   false      2019-11-04 10:37:35 -0500 EST
```
where `check.json` is:
```
{
  "type": "CheckConfig",
  "api_version": "core/v2",
  "metadata": {
    "name": "foo",
    "namespace": "default"
  },
  "spec": {
    "command": "sh -c 'exit 1'",
    "interval": 5,
    "publish": true,
    "subscriptions": [
      "switches",
      "entity:agent1"
    ]
  }
}
```

### DIFF
--- a/api/core/v2/silenced.go
+++ b/api/core/v2/silenced.go
@@ -10,6 +10,8 @@ import (
 	"strconv"
 	"strings"
 	"time"
+
+	stringsutil "github.com/sensu/sensu-go/util/strings"
 )
 
 const (
@@ -82,6 +84,27 @@ func (s *Silenced) Prepare(ctx context.Context) {
 			s.Creator = claims.Subject
 		}
 	}
+}
+
+// Matches returns true if the given check name and subscription match the silence.
+//
+// The two properties compared, Subscription and Check, are only compared if they are
+// not empty. An empty value for either of those fields is considered a wildcard,
+// ie: s.Check = "foo" && s.Subscription = "" will return true for s.Matches("foo", <anything>).
+func (s *Silenced) Matches(check, subscription string) bool {
+	if s == nil {
+		return false
+	}
+
+	if !stringsutil.InArray(s.Subscription, []string{"", "*", subscription}) {
+		return false
+	}
+
+	if !stringsutil.InArray(s.Check, []string{"", "*", check}) {
+		return false
+	}
+
+	return true
 }
 
 // NewSilenced creates a new Silenced entry.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ are deleted.
 - Sensu now uses far fewer leases for keepalives and check TTLs, resulting in a
 stability improvement for most deployments.
 - Fixed a minor UX issue in interactive filter commands in sensuctl.
+- Silences now successfully apply to proxy entities where the check doesn't contain
+  the same subscriptions as the entity (#3356)
 
 ## [5.14.1] - 2019-10-16
 
--- a/api/core/v2/entity.go
+++ b/api/core/v2/entity.go
@@ -123,7 +123,7 @@ func GetEntitySubscription(entityName string) string {
 func FixtureEntity(name string) *Entity {
 	return &Entity{
 		EntityClass:   "host",
-		Subscriptions: []string{"linux"},
+		Subscriptions: []string{"linux", GetEntitySubscription(name)},
 		ObjectMeta: ObjectMeta{
 			Namespace: "default",
 			Name:      name,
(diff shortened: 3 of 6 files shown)


api/core/v2/silenced.go
```diff
@@ -84,6 +86,27 @@ func (s *Silenced) Prepare(ctx context.Context) {
...
+//
+// The two properties compared, Subscription and Check, are only compared if they are
+// not empty. An empty value for either of those fields is considered a wildcard,
+// ie: s.Check = "foo" && s.Subscription = "" will return true for s.Matches("foo", <anything>).
+func (s *Silenced) Matches(check, subscription string) bool {
+	if s == nil {
+		return false
+	}
+
+	if !stringsutil.InArray(s.Subscription, []string{"", "*", subscription}) {
```
roganartu: This is a bug fyi @palourde @echlebek, same in the condition below with `s.Check`.

I found this while doing some adhoc testing of the PR comments on https://github.com/sensu/sensu-go/pull/3382#issuecomment-550450740

The first condition in the implementation of `stringsutil.InArray` is:
https://github.com/sensu/sensu-go/blob/3860ab4/util/strings/strings.go#L15-L18

Which means empty strings for either subscription or check are not treated as wildcards as intended.

I will push a separate PR with these two conditions changed to address.

roganartu: Fix is here: https://github.com/sensu/sensu-go/pull/3387

PR #3378, opened 2019-11-05
Add implicit access to namespaces

echlebek:
## What is this change?

Users can now list namespaces that they can access other resources in, even if they don't have explicit permission via rbac to list all namespaces, or explicit permission to get the namespace in question.

This simplifies several aspects of access control, and also simplifies display of namespaces in the web ui.

Closes https://github.com/sensu/sensu-enterprise-go/issues/649

Yes

To accomplish this task, I have pulled code from the enterprise repository into the open source repository. We will need to update the enterprise repository accordingly.

Documentation changes may be required.

Unit tests only so far.

### DIFF
--- a/backend/apid/routers/namespaces.go
+++ b/backend/apid/routers/namespaces.go
@@ -1,20 +1,38 @@
 package routers
 
 import (
+	"context"
+	"net/http"
+
 	"github.com/gorilla/mux"
 	corev2 "github.com/sensu/sensu-go/api/core/v2"
+	"github.com/sensu/sensu-go/backend/apid/actions"
 	"github.com/sensu/sensu-go/backend/apid/handlers"
+	"github.com/sensu/sensu-go/backend/authorization"
+	"github.com/sensu/sensu-go/backend/authorization/rbac"
 	"github.com/sensu/sensu-go/backend/store"
 )
+
+type ListStore interface {
+	ListResources(ctx context.Context, kind string, resources interface{}, pred *store.SelectionPredicate) error
+}
+
+type RuleVisitor interface {
+	VisitRulesFor(ctx context.Context, attrs *authorization.Attributes, fn rbac.RuleVisitFunc)
+}
 
 // NamespacesRouter handles requests for /namespaces
 type NamespacesRouter struct {
 	handlers handlers.Handlers
+	store    ListStore
+	auth     RuleVisitor
 }
 
 // NewNamespacesRouter instantiates new router for controlling check resources
-func NewNamespacesRouter(store store.ResourceStore) *NamespacesRouter {
+func NewNamespacesRouter(store store.ResourceStore, auth RuleVisitor) *NamespacesRouter {
 	return &NamespacesRouter{
+		store: store,
+		auth:  auth,
 		handlers: handlers.Handlers{
 			Resource: &corev2.Namespace{},
 			Store:    store,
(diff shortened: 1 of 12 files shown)


backend/apid/routers/namespaces.go
```diff
@@ -30,8 +48,72 @@ func (r *NamespacesRouter) Mount(parent *mux.Router) {
 	}
 
 	routes.Del(r.handlers.DeleteResource)
-	routes.Get(r.handlers.GetResource)
+	routes.Get(r.get)
```
palourde: The `r.get` method actually lists namespaces

backend/apid/routers/namespaces.go
```diff
@@ -30,8 +48,72 @@ func (r *NamespacesRouter) Mount(parent *mux.Router) {
...
+		namespaceMap[namespace.Name] = namespace
+	}
+
+	namespaces := make([]*corev2.Namespace, 0, len(resources))
+
+	var funcErr error
+
+	// Iterate over the rbac rules to discover namespaces that this
+	// user has read access to.
+	r.auth.VisitRulesFor(req.Context(), attrs, func(binding rbac.RoleBinding, rule corev2.Rule, err error) (terminate bool) {
```
palourde: So this was actually a bug in sensu-enterprise-go; the returned boolean value should actually be named `continue` (it's a reserved keyboard so it needs to be something else similar) and not `terminate`, because if the [return value is `false`, it stops the iteration over the rules](https://github.com/sensu/sensu-go/blob/66ab843069daaca27301b9378f8718f43e47d453/backend/authorization/rbac/rbac.go#L64).

Therefore, we probably need to inverse the return values below

backend/apid/routers/namespaces.go
```diff
@@ -30,8 +48,72 @@ func (r *NamespacesRouter) Mount(parent *mux.Router) {
...
 	routes.Del(r.handlers.DeleteResource)
-	routes.Get(r.handlers.GetResource)
+	routes.Get(r.get)
 	routes.List(r.handlers.ListResources, corev2.NamespaceFields)
 	routes.Post(r.handlers.CreateResource)
 	routes.Put(r.handlers.CreateOrUpdateResource)
 }
+
+func (r *NamespacesRouter) get(req *http.Request) (interface{}, error) {
+	attrs := authorization.GetAttributes(req.Context())
```
palourde: We will probably need to set `attrs.Namespace` to `*` or something like this otherwise it will fail [here](https://github.com/sensu/sensu-go/blob/66ab843069daaca27301b9378f8718f43e47d453/backend/authorization/rbac/rbac.go#L75)

echlebek: I've modified VisitRulesFor to add a special case for the namespace resource, so that the namespace can be empty.

backend/apid/routers/namespaces.go
echlebek: This is now fixed and covered by a unit test.

backend/apid/routers/namespaces.go
echlebek: I realized I actually completely broke the router... oops.

I've fixed it now and reinstated the get unit tests.

echlebek:
@palourde I believe I have finally made this work the way it is supposed to.

echlebek:
Sounds good, I should be able to add some automated QA for this today.

PR #3379, opened 2019-11-05
[GraphQL] Allow queries to skip validation

jamesdphillips:
* When dealing with queries that were generated from another cluster
  node, the query while valid, may not please validators that check for
  user error. (eg. extraneous variables & fragments)
* Opens the door for disabling validation in production, in theory
  saving some CPU cycles.


jamesdphillips:
> I'm not sure it would be wise to implement this. If cluster versions are mismatched, then we could fail to notice validation errors, potentially leading to undefined behaviour.

Version mismatches are an interesting problem but I don't think the request validation really provides any additional value. Regardless of whether the error occurs during validation or during execution as far as the federation cluster is concern the request failed.

Issue #3380, opened 2019-11-05
[Feature Request] Make silence details available in event filters


roganartu:
Make details for silences that have been applied to an event accessible in filter functions.

## Feature Suggestion
#3376 removes the dependence on names of silence entries for applying silences to events, and as such makes it so that the silence name available under `event.check.silenced` field is no longer an accurate description of the silence itself, making the list in this field no more useful than the existing `event.is_silenced` boolean filter available.

Additionally, silences support labels which would be useful to be able to access in filters in order to create more intelligent silencing rules.

## Possible Implementation
Options for implementing this I can think of:

#### Extend `event.check.silenced` to include full silence instead of just name

I don't think this is the correct approach, as it changes an existing data structure and so would be a breaking change

#### Add silence information as event labels

This would work and wouldn't be a breaking change, but it is not very clean and wouldn't handle multiple silences very well. I would think the best way to do this would be to structure the label keys something like `<subscription_name>__<subscription_label_key>`.

The benefit of this approach is the silence data will also be available in handlers, not just filters.

#### Expose matched silences directly to filters

I think this is probably the way to implement this, with the downside of not having detailed silence information in handlers (but I think it's likely less necessary there).

There is already precedent for this in the event synthesization process, where actual and derived attributes are added to the `event` object that is available to filter functions (ie: https://docs.sensu.io/sensu-go/5.15/reference/filters/#event-attributes-available-to-filters).

Exposing detailed silence data under `events.silences` for each matched silence should work well, and is not a breaking change as it is a new property.

## Context
More powerful silencing capabilities.

Consider the following scenario:
- You have a check that runs on all monitored hosts
- A change is rolled out to a subset of those hosts that doesn't align with a specific subscription
- The change causes the check to start failing on those hosts with a specific pattern appearing in the check output (eg a stack trace)

This proposal would allow the following approach to silencing noisy alerts without affecting alerts from the hosts that weren't subject to the bad change:
1) Create a silence with a wildcard subscription and `Silence.Check` set to the affected check name, with a label containing the pattern expected to be present in the output
2) Have a filter that looks for that label in any silences that matched the event, and only suppresses alerts if their output matches the given pattern.

raags:
Another use case is to silence group and silence alerts with arbitrary tags. For e.g. silence alerts for a specific team or project. Right now the entity_name and subscription attributes for a silence are too restrictive.

PR #3382, opened 2019-11-05
Expose silenced details to event filters

roganartu:
## What is this change?

Implements #3380 by adding a new synthesized field `event.silences` which contains a list of full silence details (including subscription, check, and labels) for each silence that has been applied to the event.

See detailed writeup in #3380 re why I proposed this and why I think this is the cleanest way to expose this information to filters.

Yes, added.

Nope.

It required me to call `dynamic.Synthesize` a second time, as it wasn't possible to query the store from `api/core/v2/event.go` as that would introduce a cyclic dependency. As a result, I chose to:

1) keep the original `dynamic.Synthesize` call on `event` (which calls `event.SynthesizeExtras`)
2) add and synthesize the silences to the result

The duplication is close together and documented, so I don't think this is really a problem, just an observation.

Documentation of the new field would be required.

Added new tests (which uncovered the synthesis problem).

Additionally performed an adhoc validation by doing the following:
```
 sensuctl create -f filter.json

 sensuctl create -f handler.json

 sensuctl create -f check.json

 curl -XPOST http:/127.0.0.1:8080/api/core/v2/namespaces/default/silenced -H "Authorization: Bearer $(jq -r '.access_token' < ~/.config/sensu/sensuctl/cluster)" -H "Content-Type: application/json" --data @silence_payload.json

 sensuctl silenced list
  Name    Subscription   Check          Begin          Expire   ExpireOnResolve   Creator     Reason     Namespace
 ─────── ────────────── ─────── ───────────────────── ──────── ───────────────── ───────── ──────────── ───────────
  *:bar                  bar     05 Nov 19 15:49 EST   59m53s   false             admin     pr testing   default
```

```
    "occurrences": 3,
    "occurrences_watermark": 3,
    "silenced": [
      "*:bar"
    ],
    "output_metric_format": "",
    "output_metric_handlers": null,
```

Where `filter.json`, `handler.json`, `check.json`, and `silence_payload.json` look like this:
```
{
  "type": "EventFilter",
  "api_version": "core/v2",
  "metadata": {
    "name": "pr_test",
    "namespace": "default"
  },
  "spec": {
    "action": "deny",
    "expressions": [
      "event.silences.length > 0"
    ]
  }
}
{
  "type": "Handler",
  "api_version": "core/v2",
  "metadata": {
    "name": "pr_test",
    "namespace": "default"
  },
  "spec": {
    "command": "echo triggered",
    "filters": [
      "pr_test"
    ],
    "timeout": 5,
    "type": "pipe"
  }
}
{
  "type": "CheckConfig",
  "api_version": "core/v2",
  "metadata": {
    "name": "bar",
    "namespace": "default"
  },
  "spec": {
    "command": "sh -c 'exit 1'",
    "interval": 5,
    "publish": true,
    "handlers": ["pr_test"],
    "subscriptions": [
      "entity:agent1"
    ]
  }
}
{
    "expire": 3600,
    "reason": "pr testing",
    "metadata": {
        "name": "",
        "namespace": "default",
        "labels": {
            "colo": "foo"
        }
    },
    "expire_on_resolve": false,
    "check": "bar",
    "subscription": null
}
```

And the result was observed to match expected behaviour:
```
backend1_1  | {"check_name":"bar","check_namespace":"default","component":"pipelined","entity_name":"agent1","entity_namespace":"default","hooks":null,"level":"debug","msg":"received event","silenced":["*:bar"],"time":"2019-11-05T20:57:01Z","timestamp":1572987421}
backend1_1  | {"assets":null,"check_name":"bar","check_namespace":"default","component":"pipelined","entity_name":"agent1","entity_namespace":"default","filter":"pr_test","level":"debug","msg":"denying event that matches filter","time":"2019-11-05T20:57:01Z"}
backend1_1  | {"check_name":"bar","check_namespace":"default","component":"pipelined","entity_name":"agent1","entity_namespace":"default","filter":"pr_test","handler":"pr_test","level":"debug","msg":"denying event with custom filter","time":"2019-11-05T20:57:01Z"}
backend1_1  | {"check_name":"bar","check_namespace":"default","component":"pipelined","entity_name":"agent1","entity_namespace":"default","handler":"pr_test","level":"info","msg":"event filtered","time":"2019-11-05T20:57:01Z"}
```

Removing the label from `silence_payload.json` resulted in the handler executing, also as expected.


echlebek:
This PR adds potentially many database reads to every event, so it would likely cause a performance regression, especially if there were a lot of silences mapped to the set of checks.

We already need the silenced entries in another place for processing events. We found that silenced reads were a major contributer to database utilization, so we cached them. I'd suggest working the cache in to this PR, which would avoid the reads entirely.

https://github.com/sensu/sensu-go/blob/master/backend/eventd/silenced.go#L23 for an example where we use the cache for looking up silenced entries.

I haven't looked at how hard it would be to share the cache between eventd and pipelined, so feel free to reach out for help with that if need be!

roganartu:
Updated to have a shared Silenced cache between eventd and pipelined. I did this by pulling the cache creation up into the backend and passing it down into both `eventd` and `pipelined`. I also chose to make the cache a map of caches instead, to allow easier addition of other caches to be shared across the two components (or others) later.

I ran the same adhoc test on this as without the cache and it appears to continue to work as expected. Let me know what you think @echlebek.

roganartu:
Is there anything blocking this getting merged?

palourde:
Hi @roganartu,

We just revisited this PR and decided that it was preferable to take a step back. While we definitely agree with the use case you shared in #3380, we believe it's preferable to take a different approach that would not add more metadata to the event.

For this reason, we will favor https://github.com/sensu/sensu-go/issues/3480 as the universal solution for providing additional metadata, such as silenced entries, to filters. This issue is relatively high priority in our backlog so hopefully it will be tackled soon.

Therefore, I'm gonna close this PR but I want to thank you once again for submitting this patch, we really appreciate it!

Issue #3344, opened 2019-10-18
Allow colons and periods to be used in resource names


portertech:
As a user, I would like to use a colon in my asset resource name to identify a version reference, e.g. `my-asset:1.0.2`.

## Feature Suggestion
Update resource name validation to allow for the `:` and `.` characters.

PR #3384, opened 2019-11-06
Allow colons to be used in asset names

nikkixdev:

## What is this change?

Updates the asset name regex so that colons are allowed.

Closes https://github.com/sensu/sensu-go/issues/3344

Yas.

Yes, I'm actually not sure if https://github.com/sensu/sensu-go/issues/3344 is intended for _all_ resources or just assets.

Nah.

We good.

Unit test and manual testing.
```$ sensuctl asset add sensu/sensu-influxdb-handler --rename sensu-influxdb-handler:3.1.2```


nikkixdev:
Updated the regex for all resource names excluding silences and users.

Issue #3346, opened 2019-10-21
Flap detection weighting is inverted


roganartu:
The calculation for the weighted flap percentage is incorrect, applying a higher weight to older events instead of to newer events.

## Expected Behavior
According to https://assets.nagios.com/downloads/nagioscore/docs/nagioscore/3/en/flapping.html (which is referenced by the docs for high/low `flap_threshold` properties of checks), newer events should have a higher weighting than older events.

## Current Behavior
Older events are weighted higher than newer events.

## Possible Solution
Weighted flap percentage calculation is done here: https://github.com/sensu/sensu-go/blob/73a0f0da35a6f7c394fdf66d8e226fa6de51cddd/api/core/v2/flapping.go#L40-L61

It iterates over events in the history backwards, increasing the weight on each iteration. This operates under the assumption that the last event in the history is the oldest.

History is, however, actually stored with the oldest event first: https://github.com/sensu/sensu-go/blob/87d076b603bae4aeda88025beb716ae16857d885/backend/store/etcd/event_store.go#L300-L324

## Steps to Reproduce (for bugs)
The logic is simple enough to step through the example from the Nagios page by hand to confirm. It claims that the unweighted average is 35% and the weighted average is lower (somewhere around 31%). Running through that example by hand yields a weighted percentage of ~35.9% due to the erroneously increased weight applied to old events.

## Your Environment
Exists on master.

PR #3385, opened 2019-11-06
Fix the order of flap detection weighting for checks

nikkixdev:

## What is this change?

Inverts the order of flap detection weighting for checks so that it is in line with [Nagios](https://assets.nagios.com/downloads/nagioscore/docs/nagioscore/3/en/flapping.html).

Closes https://github.com/sensu/sensu-go/issues/3346

Yas.

Nah.

The Nagios docs were not great, nor the online textbooks and scholarly articles. The example from Nagios claims the flap detection should be ~31%, but the value could be totally arbitrary. The logic that is implemented is consistent with the algorithm.

We good.

Updated the unit test to reflect the example in the Nagios flap detection docs.

statetransitions

statetransitions2

Screen Shot 2019-11-06 at 12 11 07 PM


nikkixdev:
I agree, it's likely an arbitrary number. While making the weight delta 0.01 arrives at 31%, it's not in line with the notion that the newest value should carry 50% more weight than the oldest. (Specifically is spelled wrong in their docs 🙄).
> Specfically, the flap detection routines are currently designed to make the newest possible state change carry 50% more weight than the oldest possible state change.

I'm just going to merge this so we can all stop getting nerd sniped.

PR #3387, opened 2019-11-07
Match empty string correctly

roganartu:
## What is this change?

Correctly handle an empty check/subscription in a silence as a wildcard.

I found a bug in my (now merged but not released) changes to how silences are applied to events.

Details here: https://github.com/sensu/sensu-go/pull/3376/files#r343748169

Not if it's merged before the above PR is released.

Nope.

Not a complication, just context. I decided to do it this way instead of changing `stringsutil.InArray` because it's used in a bunch of places and I don't have any context on why empty strings were chosen to be omitted or what changing that assumption may break.

No.

Tests continue passing, the adhoc test I was doing for https://github.com/sensu/sensu-go/pull/3382 which uncovered this bug started working after I ran a binary with this fix.

### DIFF
--- a/api/core/v2/silenced.go
+++ b/api/core/v2/silenced.go
@@ -96,11 +96,11 @@ func (s *Silenced) Matches(check, subscription string) bool {
 		return false
 	}
 
-	if !stringsutil.InArray(s.Subscription, []string{"", "*", subscription}) {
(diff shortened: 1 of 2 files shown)


echlebek:
@roganartu, I've made a PR against your branch that adds a unit test, which should prove that your patch works. Once you merge it I'll merge this branch to master.

roganartu:
Merged and confirmed tests are green, at least on my machine. Thanks @echlebek

PR #3388, opened 2019-11-07
Only bind pprof to localhost

echlebek:
I tested this manually to ensure that only localhost can serve pprof.

### DIFF
--- a/backend/cmd/start.go
+++ b/backend/cmd/start.go
@@ -237,7 +237,7 @@ func StartCommand(initialize initializeFunc) *cobra.Command {
 
 			if viper.GetBool(flagDebug) {
 				go func() {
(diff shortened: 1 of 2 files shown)


backend/cmd/start.go
```diff
@@ -237,7 +237,7 @@ func StartCommand(initialize initializeFunc) *cobra.Command {
 
 			if viper.GetBool(flagDebug) {
 				go func() {
-					log.Println(http.ListenAndServe(":6060", nil))
+					log.Println(http.ListenAndServe("127.0.0.1:6060", nil))
```
palourde: I know it's an edge case but should we use `[::]` instead of `127.0.0.1`?

PR #3414, opened 2019-11-20
Add the 'sensu-backend init' subcommand

### LINKED ISSUE #3359: Backend Cluster Init Utility
From WIP internal proposal (needs implementation discussion): https://docs.google.com/document/d/14exgupRiOp1DBieDs51Nshl9E8mfKETbvvOWm4tG2Mk/edit?usp=sharing

## Abstract

Sensu Backend should not ship with a built in Admin username and password. A Backend cluster (one or more Backends) should require an initialization, which bootstraps the first Admin user account. The created Admin will be granted the Cluster Admin role. An operator only need to provide a username and password string (they do not need to understand RBAC etc). The Sensu Backend Init Utility provides a command to initialize a fresh Cluster with these details.

## Background

Sensu Backend currently ships with a built-in Admin username and password (admin:P@ssw0rd!). This is problematic but makes things “easy”.

## User Stories

As an operator, I need to deploy a new Sensu Backend cluster and initialize it with a unique and secure admin username and password.

## Proposal

Hashi Vault does something similar: https://www.vaultproject.io/docs/commands/operator/init.html
`sensu-backend init`

env vars:
`SENSU_ADMIN_USERNAME`
`SENSU_ADMIN_PASSWORD`

command line flags:
`--admin-username`
`--admin-password`

interactive mode:
`sensu-backend init --interactive`

Removed the current default admin username and password seed values.

## Rationale

Using a utility makes the installation process consistent across platforms.

## Compatibility

Existing users can upgrade without issue, no need to initialize, since their cluster has already seeded the default admin username and password (hopefully changed since then). New users will require guidance and documentation which includes the addition of a cluster initialization step.

## Implementation

- [ ] Remove the default admin user from fixed seed data (https://github.com/sensu/sensu-go/blob/master/backend/seeds/seeds.go)
- [ ] Create a sensu-backend init subcommand
      - Two required arguments, see above.
      - Two environment variables, see above.
      - Interactive mode, see above.
- [ ] The init subcommand will use the same etcd configuration as defined in `backend.yml`, and the same defaults if undefined

Required by https://github.com/sensu/sensu-staging/issues/164
Required by https://github.com/sensu/sensu-docs/issues/1917
Required by https://github.com/sensu/website/issues/447

echlebek:
## What is this change?

This CL adds the `sensu-backend init` subcommand. The command is used for initializing a backend for the first time, with an admin user.

The backend will no longer initialize itself.

In the process of doing this work, I've removed some flags that were deprecated around a year ago.

Closes #3359 

Yes

I noticed some deprecated code that had been hanging around a while, and have removed it in this PR.

New documentation is required.

I ran the tool in interactive, flag, environment variable, and config file mode.


backend/seeds/seeds.go
```diff
@@ -10,9 +10,16 @@ import (
...
-func SeedInitialData(store store.Store) (err error) {
+type Config struct {
+	// AdminUsername is the username of the cluster admin.
+	AdminUsername string
+
+	// AdminPassword is the password of the cluster admin.
+	AdminPassword string
+}
+
+// SeedClusters seeds the cluster according to the provided config.
```
nikkixdev:
```suggestion
// SeedCluster seeds the cluster according to the provided config.
```

backend/seeds/seeds.go
```diff
@@ -72,6 +85,16 @@ func SeedInitialData(store store.Store) (err error) {
 	return initializer.FlagAsInitialized()
 }
 
+// SeedInitialData will seed a store with initial data. This method is
+// idempotent and can be safely run every time the backend starts.
+func SeedInitialData(store store.Store) (err error) {
+	config := Config{
+		AdminUsername: "admin",
+		AdminPassword: "P@ssw0rd!",
```
nikkixdev: Under what circumstances would we still need to seed `admin`/`P@ssw0rd!` since it's being removed?

palourde:
> Do we still need to seed the agent user? I can't remember if we talked about that. As of now, you would be unable to start an agent with the default agent user.

@nikkixdev Pretty sure we said in our last call that the agent user was out of scope for now, but let's re-confirm with @portertech when he's around!

backend/seeds/seeds.go
echlebek: Two reasons.

1) The integration tests depend on it to some extent.
2) By keeping this function as-is, we avoid breaking the package functionality, and therefore any upstream consumers that depend on it.

palourde:
It might be related to @nikkixdev previous comment, but we should probably add a timeout because the command hang indefinitely if the backend is not reachable, e.g.

```
$ /usr/bin/time go run ./cmd/sensu-backend init --admin-username foo --admin-password bar
^Csignal: interrupt
      942.69 real         4.46 user         2.06 sys
```

echlebek:
> Do we still need to seed the agent user? I can't remember  if we talked about that. As of now, you would be unable to connect an  agent with the default agent user.

The way it works right now is that nothing can connect until `sensu-backend init` is run. Is that not the desired behaviour?

echlebek:
> I guess I'm also a little confused why all of the backend flags are available for sensu-backend init.. what happens if sensu-backend start is already running and the additional flags set by sensu-backend init conflict with those?

Can you explain in detail the scenario you're picturing here? What do you mean conflict? For `sensu-backend init` to work, `sensu-backend start` must already be running.

echlebek:
> Sorry for all the review spam, but in testing this I have hit a hang on the interactive prompt.

It works on my machine :sweat_smile: 

Is `sensu-backend start` up and running?

palourde:
@echlebek I think it might related to https://github.com/sensu/sensu-go/pull/3414#issuecomment-557202945; in my case I intentionally shutdown sensu-backend to figure out what would happen.

nikkixdev:
So the hang happened when the backend was off, however it continues to hang even when the backend is started back up while init is waiting. I think a timeout and log message should be reasonable for this.

For the agent user, I would refer to the proposal/spec. I don’t believe we decided to remove the agent user from the seed data, only the admin user.

echlebek:
Nothing has been removed from the seed data. Rather, the seeding process has been deferred until init is run.

calebhailey:
What resources can be initialized with `sensuctl init`? Just admin username and password? Can it also initialize a cluster with namespaces and other rbac resources?

echlebek:
@calebhailey, this is not `sensuctl init`, rather it is `sensu-backend init`.

`sensu-backend init` connects directly to a running etcd service, creates an admin user with the supplied credentials, creates the default system RBAC policy, and creates an agent user.

No other types of resources can be created, and there is no control over anything other than the admin username and password.

backend/seeds/seeds.go
majormoses: > The integration tests depend on it to some extent.
Shouldn't the integration tests use the same methods that real users should use?

> By keeping this function as-is, we avoid breaking the package functionality, and therefore any upstream consumers that depend on it.

This is already a breaking change in the first place and would go into the next major version unless I am misunderstanding something so what are we really buying here?

majormoses:
A couple of questions/thoughts:
- what options should be supported for supplying the required bits by the user as part of the init? Env vars, a seed file (possibly passed via stdin), interactive, etc.
- should we calling these `admin-...` or perhaps a more descriptive name of `sensu-cluster-admin-(user|password)`
- this would be a breaking change and therefore go into the next major version (6.x at the time of writing this) right?
Apologies if the answers are obvious to someone who is more familiar with go but I find it quite a bit more difficult to grok than a `ruby`, `python`, etc which is what I am most familiar with.

portertech:
1. We’re keeping the first pass simple and focused on solving admin user creation, only two values are required and they can be provided via cli (set and interactive) and environment variables.
2. If we were to change the name, I think cluster-admin would be appropriate (no need for “sensu”)
3. This introduces another step during cluster setup and does not break an API etc, this will not result in a major version bump.

portertech:
I can see this resulting in a major bump in the Chef cookbook version etc. Hmm.

backend/seeds/seeds.go
echlebek: From a Go module compatibility point of view, it costs us little-to-nothing to maintain module compatibility here, which is why I've kept this function as-is, while introducing a new one.

majormoses:
> This introduces another step during cluster setup and does not break an API etc, this will not result in a major version bump.

If someone tears down an existing cluster touching nothing other than bumping the version version it wont work. This fits my definition of what a breaking change is. We agreed to follow semver so lets please do that, it does not break the HTTP api but it breaks the API contract by putting in a new pre-requisite.

portertech:
- A user can update an existing and running cluster without interruption/issue
- A user can install and initialize a new cluster by following the updated documented steps

The area of issue is if a user destroys their existing cluster while upgrading and is unaware of the new cluster initialization step.

majormoses:
> The area of issue is if a user destroys their existing cluster while upgrading and is unaware of the new cluster initialization step.

Exactly if someone has a major outage and needs to rebuild it should be safe to assume that this should be possible with the existing process. Automation is part of disaster recovery and if you change how someone needs to automate it...its breaking and should be a major version bump. 3 AM is not the best time to read through a changelog to discover a breaking change that circumvents the reason to use semver. If we are not gonna use semver properly I think we should just make up our own thing and not call out semver use.

portertech:
We do not believe these changes warrant a major version bump (using Sensu Engineering’s interpretation of SemVer). We also consider a deferred version bump (even patch release) as part of disaster recovery to be a bad and uncommon practice.

backend/cmd/init.go
```diff
@@ -0,0 +1,173 @@
...
+	AdminUsername string `survey:"cluster-admin-username"`
+	AdminPassword string `survey:"cluster-admin-password"`
+}
+
+func (i *initOpts) administerQuestionnaire() error {
+	qs := []*survey.Question{
+		{
+			Name: "cluster-admin-username",
+			Prompt: &survey.Input{
+				Message: "Admin Username:",
```
portertech: "Cluster Admin Username:"?

portertech:
yesjacknicholson

backend/seeds/seeds.go
echlebek: Also, no, the integration tests should not run `sensu-backend init`. Integration tests are different from functional tests, which we have outside of the sensu-go repository. These functional tests cover running the `sensu-backend init` tool the way a user would.

The integration tests are more concerned with receiving a bootstrapped, working backend at the start of the test context.

PR #3428, opened 2019-11-26
Remove dockerfiles from core project

portertech:
Removed the outdated and broken dockerfiles from the core project. These dockerfiles have not been used or maintained since changes were made to the build process. The OSS core project build artifacts are compiled binaries.


georgespatton:
I have a repo that has a working Dockerfile which builds the OSS version of sensu backend and agent here:

https://github.com/georgespatton/sensu-go-oss-docker

PR #3445, opened 2019-12-02
Restart backend daemons on store failure

echlebek:
## What is this change?

This change causes Sensu to shut down all of its internal backend daemons, and restart them, when an error from the store is marked as `ErrInternal`. These errors generally indicate that service has been lost, so Sensu should not keep trying to serve any requests. This will allow clients to select a different backend that may be working properly.

It was discovered under load testing that clients would sometimes remain "stuck" on a server that had failed. In some cases the server never re-established itself as a member of the etcd cluster, leading to clients being connected but never working again, until the connection was severed and the client connected to a different backend.

Yes

This is a bugfix, no documentation changes are required.

I've smoke tested it on my laptop by starting up the system and executing checks, but I will need to test this on the performance cluster to verify that the failure mode works as we desire.

## Is this change a patch?

Yes

### DIFF
--- a/backend/pipeline/handler.go
+++ b/backend/pipeline/handler.go
@@ -1,5 +1,4 @@
-// Package pipelined provides the traditional Sensu event pipeline.
-package pipelined
+package pipeline
 
 import (
 	"context"
@@ -9,6 +8,7 @@ import (
 	"os"
 	"time"
 
+	corev2 "github.com/sensu/sensu-go/api/core/v2"
 	"github.com/sensu/sensu-go/asset"
 	"github.com/sensu/sensu-go/backend/store"
 	"github.com/sensu/sensu-go/command"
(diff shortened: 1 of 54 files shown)


echlebek:
I didn't base the PR correctly, closing in favour of another one.

backend/store/etcd/extension_registry.go
```diff
@@ -55,29 +51,34 @@ func (s *Store) RegisterExtension(ctx context.Context, ext *types.Extension) err
 // DeregisterExtension deregisters an extension
 func (s *Store) DeregisterExtension(ctx context.Context, name string) error {
 	if name == "" {
-		return errors.New("no extension name specified")
+		return &store.ErrNotValid{errors.New("no extension name specified")}
```
golangcibot: composites: `github.com/sensu/sensu-go/backend/store.ErrNotValid` composite literal uses unkeyed fields (from `govet`)

backend/pipeline/handler.go
```diff
@@ -67,13 +62,25 @@ func (p *Pipelined) handleEvent(event *types.Event) error {
 		handler := u.Handler
 		fields["handler"] = handler.Name
 
-		if filtered := p.filterEvent(handler, event); filtered {
+		filtered, err := p.filterEvent(handler, event)
+		if err != nil {
+			if _, ok := err.(*store.ErrInternal); ok {
```
golangcibot: S1020: when ok is true, err can't be nil (from `gosimple`)

Issue #3444, opened 2019-12-01
Unable to create EtcdReplicator using API


treydock:
## Expected Behavior

I'd expect API to be able to create EtcdReplicator.

## Current Behavior

`{"message":"resource name does not match path (rolebinding_replicator != )","code":0}`

## Steps to Reproduce (for bugs)

```
[root@sensu_backend /]# curl -k -X POST -H "Accept: application/json" -H "Content-Type: application/json" -H "Authorization: Bearer $TOKEN" -d '{"spec":
{"ca_cert":"/path/to/ssl/trusted-certificate-authorities2.pem",
"cert":"/path/to/ssl/cert2.pem",
"key":"/path/to/ssl/key2.pem",
"insecure":false,
"url":"http://127.0.0.1:3379",
"api_version":"core/v2",
"resource":"RoleBinding",
"replication_interval_seconds":60},
"metadata":{"name":"rolebinding_replicator"},
"api_version":"federation/v1","type":"EtcdReplicator"
}' https://sensu_backend:8080/api/enterprise/federation/v1/etcd-replicators
```

## Context

I am trying to use the API to manage Puppet resources and unable to add EtcdReplicators.

CC @ghoneycutt

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): `sensu-backend version 5.15.0, build 13884593ee08a7f25d7f66a8b71da61c529de014, built 2019-11-19T20:18:33Z` Also CI builds from 11/30/19
* Installation method (packages, binaries, docker etc.): RPM
* Operating System and version (e.g. Ubuntu 14.04): CentOS 7

echlebek:
Please use
```
https://sensu_backend:8080/api/enterprise/federation/v1/etcd-replicators/rolebinding_replicator
```
as the URL. As with most of our APIs, it is more natural to use PUT, but the server will accept POST as a verb as well.

echlebek:
Apologies, this appears to be a bug.

palourde:
This is actually caused by the exact same problem as https://github.com/sensu/sensu-go/issues/3442. We successfully confirmed the patch we have also fixes this problem.

Issue #3442, opened 2019-11-30
authproviders API nesting metadata under spec


treydock:
## Expected Behavior

I'd expect the `/authproviders` API `GET` to match documentation and have top level key of `metadata`

## Current Behavior

Currently the `metadata` key is empty but there is `metadata` key under `spec` that has `name`.

## Steps to Reproduce (for bugs)

```
[root@sensu-backend ~]# cat ad.json 
{
  "Type": "ad",
  "api_version": "authentication/v2",
  "spec": {
    "servers": [
      {
        "host": "127.0.0.1",
        "binding": {
          "user_dn": "cn=binder,cn=users,dc=acme,dc=org",
          "password": "P@ssw0rd!"
        },
        "group_search": {
          "base_dn": "dc=acme,dc=org"
        },
        "user_search": {
          "base_dn": "dc=acme,dc=org"
        }
      }
    ]
  },
  "metadata": {
    "name": "activedirectory2"
  }
}

[root@sensu-backend ~]# sensuctl create -f ad.json 

[root@sensu-backend ~]# curl -q -k -H "Authorization: Bearer $TOKEN" https://sensu-backend.example.com:8080/api/enterprise/authentication/v2/authproviders 2>/dev/null | python -m json.tool
[
    {
        "api_version": "authentication/v2",
        "metadata": {},
        "spec": {
            "groups_prefix": "",
            "metadata": {
                "name": "activedirectory2"
            },
            "servers": [
                {
                    "binding": {
                        "password": "P@ssw0rd!",
                        "user_dn": "cn=binder,cn=users,dc=acme,dc=org"
                    },
                    "client_cert_file": "",
                    "client_key_file": "",
                    "default_upn_domain": "",
                    "group_search": {
                        "attribute": "member",
                        "base_dn": "dc=acme,dc=org",
                        "name_attribute": "cn",
                        "object_class": "group"
                    },
                    "host": "127.0.0.1",
                    "insecure": false,
                    "port": 636,
                    "security": "tls",
                    "trusted_ca_file": "",
                    "user_search": {
                        "attribute": "sAMAccountName",
                        "base_dn": "dc=acme,dc=org",
                        "name_attribute": "displayName",
                        "object_class": "person"
                    }
                }
            ],
            "username_prefix": ""
        },
        "type": "ad"
    }
]

```

## Your Environment

[...]

PR #3449, opened 2019-12-03
Set the outer ObjectMeta when decoding a wrapper struct

palourde:

## What is this change?

It fixes a bug in the Wrapper's UnmarshalJSON method, which would cause wrapped resources to not properly display their metadata or prevent them from being created.

It also adds more comments and unit tests to the same UnmarshalJSON method.

Fixes https://github.com/sensu/sensu-enterprise-go/issues/704
Fixes https://github.com/sensu/sensu-go/issues/3442
Fixes https://github.com/sensu/sensu-go/issues/3444

Added!

Nope.

Nope.

Nope.

Manually tested that it fixed the reported bugs and added unit tests to cover these cases.

## Is this change a patch?

Yep, but we are not proceeding with a patch release for 5.15.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,14 +14,14 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 when it is not operating as an etcd member. The flag is also used by the new
 sensu-backend init tool.
 - Added the cluster's distribution to Tessen data.
+- Added the `--etcd-discovery` and `--etcd-discovery-srv` flags to
+`sensu-backend`. These are used to take advantage of the embedded etcd's
+auto-discovery features.
 
 ### Fixed
 - Add a timeout to etcd requests when retrieving the nodes health.
 - Show the correct default value for the format flag in `sensuctl dump` help
 usage.
-- Added the `--etcd-discovery` and `--etcd-discovery-srv` flags to
-`sensu-backend`. These are used to take advantage of the embedded etcd's
-auto-discovery features.
 - Installing sensuctl commands via Bonsai will now check for correct labels
 before checking if the asset has 1 or more builds.
 - Listing assets with no results returns an empty array.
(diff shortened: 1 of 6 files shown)


palourde:
@ccressent Feel free to push commits to this branch if you add more/better tests!

palourde:
@echlebek It works for both core and enterprise resources but wrapped resources won't populate the outer metadata. I suppose I could easily modify the encoding methods so we always have both the inner and outer metadata filled internally but only display the external one when decoding; right now [sensuctl does a hack](https://github.com/sensu/sensu-go/blob/0d76ff14412bb2d1b7e80662051943a605fe8dd0/cli/commands/helpers/yaml.go#L20:6) to accomplish that.

palourde:
@echlebek @ccressent Alright, I think I finally figured out how to have a consistent behavior across our codebase with the wrapper.

Essentially, when we decode a `Wrapper`, we first fill the innerMeta with the help of the outerMeta, and then set the outerMeta with the value of innerMeta. By doing so, users can provide the outer and/or inner meta but internally we fill both:
https://github.com/sensu/sensu-go/blob/ba9c8287cc4ad0a214d12589fd08ef7ce3e6865b/types/wrapper.go#L147-L171

Then, I moved the encoding logic from sensuctl to the `MarshalJSON` & `MarshalYAML` methods; This logic is required so the innerMeta is discarded and the YAML format also omits empty fields. From this point, sensuctl and the API will display wrapped resources the same way:
https://github.com/sensu/sensu-go/blob/ba9c8287cc4ad0a214d12589fd08ef7ce3e6865b/types/wrapper.go#L59-L109

echlebek:
I hope this works :joy:

ccressent:
The funny thing is that your outer --> inner --> outer thing is what I arrived at while trying to fix this issue and what I showed Eric before we decided it might be dealt with as a pure display issue! So I suppose it's comforting to know we both independently arrived at the same conclusion :)

Issue #2747, opened 2019-02-22
Feature request: Implement keepalive critical state


hezor:
The old Sensu 1.x had the feature where you could define warning ja critical thresholds for keepalive checks. In Sensu Go, there is only one configuration flag (keepalive-timeout) that sets the keepalive event to a warning state, never critical.

## Expected Behavior

Allow configuring the keepalive-interval (as it can be done now), keepalive-warning-threshold (for example 120 seconds) and keepalive-critical-threshold (for example 240 seconds).

## Current Behavior

Configuration flag keepalive-timeout changes the event to a warning state, never critical.

## Possible Solution

Remove keepalive-timeout flag, add keepalive-warning-threshold and keepalive-critical-threshold flags (or whatever naming convention suits best).

## Steps to Reproduce (for bugs)

Not relevant.

## Context

Every critical event in production pages our on-call officer 24/7, this includes critical keepalive events with Sensu 1.x. This way we can tell if the corresponding host is not working properly. With Sensu Go, the keepalive warning would not trigger this without getting hands dirty by creating some filters and mutators for this use case.

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): 5.2.1
* Installation method (packages, binaries, docker etc.): yum packages from the official repo
* Operating System and version (e.g. Ubuntu 14.04): CentOS 7.6.1810

annaplotkin:
Hi @hezor ! Thanks so much for raising this and the team is engaged in discussion on this and will keep this issue updated. 

In the meantime with Sensu Go, an occurrences filter can be used to achieve this in the meantime: https://docs.sensu.io/sensu-go/5.11/reference/events/#occurrences-and-occurrences-watermark

echlebek:
We won't be removing `keepalive-timeout`, as that would introduce a breaking change. But it seems like it should be fine to have `keepalive-timeout`, and `keepalive-critical-timeout`. What do you think, @hezor?

agm650:
In my use case this would be very handy. I’d like to have that :)

hezor:
> We won't be removing `keepalive-timeout`, as that would introduce a breaking change. But it seems like it should be fine to have `keepalive-timeout`, and `keepalive-critical-timeout`. What do you think, @hezor?

That would be great!

nikkixdev:
[...]

PR #3451, opened 2019-12-04
Add --keepalive-critical-timeout

nikkixdev:

## What is this change?

Added a new flag, `--keepalive-critical-timeout` to define the time after which a critical keepalive event should be created for a dead agent.

Closes https://github.com/sensu/sensu-go/issues/2747

Yas.

Just some product input on the complications below.

Yes, these complications were discussed with @echlebek (the keeper of the keepalives) which I will outline, but we will need some product direction. 

While critical keepalive events are a 1.x parity feature, the implementation of keepalives in Sensu Go is vastly different than that of Classic. The [liveness package](https://github.com/sensu/sensu-go/blob/master/backend/liveness/liveness.go) implements switches with etcd leases that get flipped on life and death events for agent entities. This approach is durable, allowing the keepalive feature to function properly with the loss of a backend or the loss of cluster quorum.

### Refactor switches to accept multiple ttls
Switches currently only accept a single ttl value, where the critical keepalive feature requires a second value. Refactoring of this complex code would be required in order to rework it for critical keepalives, which is possible, but would make the liveness package more error prone. Additionally, iterative calculations will need to be performed on both values to ensure that each keepalive event fires at the right time. This approach offers the highest granularity at the highest cost of complexity.

### Instantiate multiple switches
Rather than refactoring the switches, we could instantiate a second switch for the critical keepalive (in addition to the warning switch that already exists). We believe multiple switches could cause concurrency problems on life and death events, and could cause keepalive events not to fire in certain circumstances. This approach would be simpler to implement but is not recommended.

### Use time since last seen
Time since last seen is most in line with the [classic implementation](https://github.com/sensu/sensu/blob/012c2c0fd35e22817ad030626f75172b4b67cd1e/lib/sensu/server/process.rb#L1112-L1144). I've implemented some simple logic to determine when a warning or critical event should be created/overwritten based on the keepalive check timeout (warning) and the keepalive check ttl (critical). This event will only fire on the original warning timeout, rather than the critical timeout so some granularity is lost in cases where the critical timeout is less than the warning timeout. This is the simplest approach to implement (is implemented in this PR) but offers the lowest granularity.

https://github.com/sensu/sensu-docs/issues/1990

e2e testing was performed for all use cases described in https://github.com/sensu/sensu-go/issues/2747#issuecomment-558407864 and the following test was added to TestRail: https://sensuinc.testrail.io/index.php?/cases/view/13574.

## Is this change a patch?

This feature will go into master.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 when it is not operating as an etcd member. The flag is also used by the new
 sensu-backend init tool.
 - Added the cluster's distribution to Tessen data.
+- Added `--keepalive-critical-timeout` to define the time after which a
+critical keepalive event should be created for an agent.
 
 ### Fixed
 - Add a timeout to etcd requests when retrieving the nodes health.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ sensu-backend init tool.
 - Added the cluster's distribution to Tessen data.
 - Added `--keepalive-critical-timeout` to define the time after which a
 critical keepalive event should be created for an agent.
+- Added `--keepalive-warning-timeout` which is an alias of `--keepalive-timeout`
+for backwards compatibility.
 
 ### Fixed
 - Add a timeout to etcd requests when retrieving the nodes health.
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -31,9 +31,6 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-// KeepaliveCriticalTimeoutLabel is the label key used to store the keepalive critical timeout.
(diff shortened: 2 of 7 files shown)


calebhailey:
@nikkixdev thanks for the detailed breakdown here! I think the approach you've taken is good (until proven otherwise). The only change I would recommend here is to keep the `--keepalive-timeout` flag/attribute (for backwards compat), and add `--keepalive-warning` (same behavior as `--keepalive-timeout`, but overrides the setting of the old flag if both are set), to complement the new `--keepalive-critical` you've already added. This should help with "discoverability" for users (making it more obvious that there are warning and critical thresholds, as opposed to trying to understand how "keepalive-timeout" and "keepalive-critical" are related). 

We should also validate the new critical flag at service start (if set) and fail hard if keepalive-critical is less than keepalive-warning, to protect users from invalid configurations. 

Good job on this!

Issue #3423, opened 2019-11-24
The 'sensuctl cluster health' and API '/health' endpoint hang when etcd is configured for SSL


treydock:
## Expected Behavior

I'd expect `sensuctl cluster health` to continue working when SSL is enabled for Etcd.

## Steps to Reproduce (for bugs)

Generated certs:
```
yum install golang-bin
cd /root
git clone https://github.com/cloudflare/cfssl.git
cfssl/
make
export PATH=/root/cfssl/bin:$PATH

# Generate CA

mkdir -p /etc/sensu/etcd-ssl
cd /etc/sensu/etcd-ssl
echo '{"CN":"CA","key":{"algo":"rsa","size":2048}}' | cfssl gencert -initca - | cfssljson -bare ca -
echo '{"signing":{"default":{"expiry":"43800h","usages":["signing","key encipherment","server auth","client auth"]}}}' > ca-config.json

# Generate certs

export ADDRESS=192.168.52.30,sensu-backend1
export NAME=sensu-backend1
echo '{"CN":"'$NAME'","hosts":[""],"key":{"algo":"rsa","size":2048}}' | cfssl gencert -config=ca-config.json -ca=ca.pem -ca-key=ca-key.pem -hostname="$ADDRESS" -profile=peer - | cfssljson -bare $NAME

export ADDRESS=192.168.52.31,sensu-backend2
export NAME=sensu-backend2
echo '{"CN":"'$NAME'","hosts":[""],"key":{"algo":"rsa","size":2048}}' | cfssl gencert -config=ca-config.json -ca=ca.pem -ca-key=ca-key.pem -hostname="$ADDRESS" -profile=peer - | cfssljson -bare $NAME

export NAME=client
echo '{"CN":"'$NAME'","hosts":[""],"key":{"algo":"rsa","size":2048}}' | cfssl gencert -config=ca-config.json -ca=ca.pem -ca-key=ca-key.pem -hostname="" -profile=client - | cfssljson -bare $NAME
```
Backend config

```
[root@sensu-backend1 ~]# cat /etc/sensu/backend.yml 
---
state-dir: "/var/lib/sensu/sensu-backend"
api-url: https://sensu-backend-peer1.example.com:8080
cert-file: "/etc/sensu/ssl/cert.pem"
key-file: "/etc/sensu/ssl/key.pem"
trusted-ca-file: "/etc/sensu/ssl/ca.crt"
etcd-listen-client-urls: https://0.0.0.0:2379
[...]

PR #3454, opened 2019-12-07
Improve sensuctl cluster health

### LINKED ISSUE #3036: Unexpected error running sensuctl cluster health when cluster is in degraded state.
Running 3 member sensu-backend cluster, fail a member and then sensuctl cluster health times out with  
```
Error: GET "/health": Get http://127.0.0.1:8080/health: net/http: request canceled (Client.Timeout exceeded while awaiting headers)
```

## Expected Behavior
sensuctl cluster health should return a list showing cluster member state

## Current Behavior
sensuctl cluster health  returns  GET "/health" error

1. create cluster
2. fail a member
3. run sensuctl cluster health
4.

## Your Environment

* Sensu version: 5.9
* Installation method binaries
* CentOS 7

echlebek:
## What is this change?

* Fix a bug where the cluster health command would hang
indefinitely if one of the cluster members was not responding.

* Gather cluster health from members concurrently.

* Implement a static timeout of 3s for gathering cluster member
information from members.

* Add MemberIDHex field to the cluster health datatype, so that the
rendered JSON and YAML contain the hex representation as well as
the decimal representation.

Refs #3423 
Closes #3036 

Yes

We'll need to update the API reference for ClusterHealth.

Unverified beyond smoke tests currently.

## Is this change a patch?

No

### DIFF
--- a/api/core/v2/health.go
+++ b/api/core/v2/health.go
@@ -30,8 +30,7 @@ func (h ClusterHealth) MarshalJSON() ([]byte, error) {
 		h.MemberIDHex = fmt.Sprintf("%x", h.MemberID)
 	}
 	type Clone ClusterHealth
-	var clone *Clone
-	clone = (*Clone)(&h)
+	var clone *Clone = (*Clone)(&h)
 	return json.Marshal(clone)
 }
 
--- a/api/core/v2/health.go
+++ b/api/core/v2/health.go
@@ -1,6 +1,9 @@
 package v2
 
 import (
+	"encoding/json"
+	"fmt"
+
 	"github.com/coreos/etcd/etcdserver/etcdserverpb"
 )
 
@@ -8,12 +11,28 @@ import (
 type ClusterHealth struct {
 	// MemberID is the etcd cluster member's ID.
 	MemberID uint64
+
+	// MemberIDHex is the hexadecimal representation of the member's ID.
+	MemberIDHex string
+
 	// Name is the cluster member's name.
 	Name string
-	// Err holds the string representation of any errors encountered while checking the member's health.
+
+	// Err contains any error encountered while checking the member's health.
 	Err string
+
 	// Healthy describes the health of the cluster member.
 	Healthy bool
+}
+
+func (h ClusterHealth) MarshalJSON() ([]byte, error) {
+	if h.MemberIDHex == "" {
+		h.MemberIDHex = fmt.Sprintf("%x", h.MemberID)
+	}
+	type Clone ClusterHealth
+	var clone *Clone
+	clone = (*Clone)(&h)
+	return json.Marshal(clone)
 }
 
 // HealthResponse contains cluster health and cluster alarms.
--- a/backend/apid/routers/cluster.go
+++ b/backend/apid/routers/cluster.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/coreos/etcd/clientv3"
 	"github.com/gorilla/mux"
@@ -69,8 +70,27 @@ func parsePeerAddrs(req *http.Request) ([]string, error) {
 	return strings.Split(peerAddrsValue, ","), nil
 }
 
+func parseTimeout(req *http.Request) (int, error) {
+	val := req.FormValue("timeout")
+	if len(val) == 0 {
+		return 0, nil
+	}
+	return strconv.Atoi(val)
+}
+
 func (r *ClusterRouter) list(w http.ResponseWriter, req *http.Request) {
-	resp, err := r.controller.MemberList(req.Context())
+	timeout, err := parseTimeout(req)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+	ctx := req.Context()
+	if timeout > 0 {
+		tctx, cancel := context.WithTimeout(ctx, time.Duration(timeout)*time.Second)
+		defer cancel()
+		ctx = tctx
+	}
+	resp, err := r.controller.MemberList(ctx)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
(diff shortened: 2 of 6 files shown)


api/core/v2/health.go
```diff
@@ -1,21 +1,40 @@
...
 }
 
+func (h ClusterHealth) MarshalJSON() ([]byte, error) {
+	if h.MemberIDHex == "" {
+		h.MemberIDHex = fmt.Sprintf("%x", h.MemberID)
+	}
+	type Clone ClusterHealth
+	var clone *Clone
+	clone = (*Clone)(&h)
+	return json.Marshal(clone)
```
ccressent: Why is that needed? Wouldn't `return json.Marshal(h)` result in the same `[]byte`? Is that because it creates an infinite recursion of `h.ClusterHealth()`?

backend/apid/routers/cluster.go
```diff
@@ -126,7 +179,18 @@ func (r *ClusterRouter) memberUpdate(w http.ResponseWriter, req *http.Request) {
...
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+	ctx := req.Context()
+	if timeout > 0 {
+		tctx, cancel := context.WithTimeout(ctx, time.Duration(timeout)*time.Second)
+		defer cancel()
+		ctx = tctx
+	}
```
ccressent: This block is repeated so many times: `list()`, `memberUpdate()`, `memberRemove()`, `memberAdd()`, ...
Could it be turned into a function, say `withTimeout(*http.Request) (context, err)`?

backend/store/etcd/health_store.go
```diff
@@ -22,72 +25,100 @@ func isEmbeddedClient(clientURLs []string) bool {
...
+		health.Err = cliErr.Error()
+		return health
+	}
+	defer func() {
+		_ = cli.Close()
+	}()
+
+	_, getErr := cli.Get(ctx, "health")
+
+	if getErr == nil || getErr == rpctypes.ErrPermissionDenied {
```
ccressent: Should a permission denied "swallow" errors and pretend that everything is healthy all the time?

I think it should return and explicit permission denied error so that an operator realizes they need to use the right credentials rather than think, potentially wrongly, that etcd is healthy.

api/core/v2/health.go
echlebek: Yes, the clone is required to avoid the infinite recursion.

backend/store/etcd/health_store.go
echlebek: Ah, this caught my eye as well, since this logic existed before.

What I concluded was that it is correct. If etcd is returning ErrPermissionDenied, then it must be consulting its own store to do so, and therefore _is healthy_!

backend/apid/routers/cluster.go
echlebek: image

backend/store/etcd/health_store.go
ccressent: I suppose if that's the existing logic, let's leave it alone...

While your logic seems to make sense, I would argue it might turn out to be a bit dangerous: "permission denied" doesn't imply "healthy". Say the definition of "healthy" involves checking that some internal metrics are within acceptable bounds. One might get a "permission denied" message, but that doesn't mean the criteria for "healthy" have been checked.

backend/store/etcd/health_store.go
```diff
@@ -22,72 +25,100 @@ func isEmbeddedClient(clientURLs []string) bool {
...
+	sort.Slice(healthResponse.ClusterHealth, func(i, j int) bool {
+		return healthResponse.ClusterHealth[i].Name < healthResponse.ClusterHealth[j].Name
+	})
 
+	alarmResponse, err := s.client.Maintenance.AlarmList(ctx)
 	if err != nil {
 		logger.WithError(err).Error("failed to fetch etcd alarm list")
 	}
 
+	logger.WithField("alarms", len(alarmResponse.Alarms)).Info("cluster alarms")
```
palourde: This will cause a panic if `alarmResponse` is nil because we had an error on line 115, since we only print a log entry and do not return.

backend/apid/routers/cluster.go
```diff
@@ -93,6 +124,17 @@ func (r *ClusterRouter) memberAdd(w http.ResponseWriter, req *http.Request) {
...
+	timeout, err := parseTimeout(req)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+	ctx := req.Context()
+	if timeout > 0 {
+		tctx, cancel := context.WithTimeout(ctx, time.Duration(timeout)*time.Second)
+		defer cancel()
+		ctx = tctx
```
golangcibot: ineffectual assignment to `ctx` (from `ineffassign`)

api/core/v2/health.go
```diff
@@ -1,21 +1,40 @@
...
 	// Healthy describes the health of the cluster member.
 	Healthy bool
 }
 
+func (h ClusterHealth) MarshalJSON() ([]byte, error) {
+	if h.MemberIDHex == "" {
+		h.MemberIDHex = fmt.Sprintf("%x", h.MemberID)
+	}
+	type Clone ClusterHealth
+	var clone *Clone
```
golangcibot: S1021: should merge variable declaration with assignment on next line (from `gosimple`)

backend/apid/routers/cluster.go
```diff
@@ -93,6 +124,17 @@ func (r *ClusterRouter) memberAdd(w http.ResponseWriter, req *http.Request) {
...
 
 func (r *ClusterRouter) memberRemove(w http.ResponseWriter, req *http.Request) {
+	timeout, err := parseTimeout(req)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+	ctx := req.Context()
+	if timeout > 0 {
+		tctx, cancel := context.WithTimeout(ctx, time.Duration(timeout)*time.Second)
```
golangcibot: SA4006: this value of `tctx` is never used (from `staticcheck`)

PR #3456, opened 2019-12-10
OSS only Dockerfile based on Ubuntu 18.04

georgespatton:
## What is this change?

Create Dockerfile for those not needing enterprise features.

For those that only need the non-enterprise OSS only features and who do not want the default 100 node restriction of Sensu Enterprise, this builds a docker container from Ubuntu 18.04 which uses the OSS features and none of the enterprise features.

Yes.

Note that this is not using the docker-scripts/sensu-entrypoint.sh, just calling sensu-backend directly.  You could also use this for running sensu-agent for a docker container for agents.

It may be useful to add the following to the README.md file:

"""To build a OSS only docker image, run these commands:
cd docker-scripts
sudo docker build ."""

Note I have added these to the README.md file.

cd docker-scripts
sudo docker build .
sudo docker run -d new_image_hash
 sudo docker logs new_container_hash

## Is this change a patch?

No.

### DIFF
--- a/docker-scripts/Dockerfile
+++ b/docker-scripts/Dockerfile
@@ -10,7 +10,7 @@ COPY ./sensu-entrypoint.sh /opt/sensu/bin/
 
 RUN apt update &&\
     apt -y upgrade &&\
-    apt -y install git wget 
+    apt -y install git wget
 
 #Download and install Go 1.13.3 for Go modules.
 RUN wget https://dl.google.com/go/go1.13.3.linux-amd64.tar.gz && tar -xvf go1.13.3.linux-amd64.tar.gz &&\
@@ -29,10 +29,8 @@ RUN git clone --depth=1 https://github.com/sensu/sensu-go.git &&\
 
 #State directory lives outside container for persistent data.
 VOLUME [/var/lib/sensu]
-#Port 3000 for Sensu Go UI and port 8081 for sensu-agent communication.  
-EXPOSE 3000 8081
+#Port 3000 for Sensu Go UI, 8080 for Sensu API, and port 8081 for sensu-agent websocket API communication.  
+EXPOSE 3000 8080 8081
 
-#From https://github.com/sensu/sensu-go/blob/master/docker-scripts/sensu-entrypoint.sh
-ENTRYPOINT ["/opt/sensu/bin/sensu-entrypoint.sh"]
 #Use Docker logs to see stdout of sensu-backend (ie. docker logs --details -f  `docker ps -q` )
-CMD ["sensu-backend"]
+CMD ["/opt/sensu/bin/sensu-backend","start"]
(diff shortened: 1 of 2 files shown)


docker-scripts/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+
+WORKDIR /opt/sensu
+COPY ./sensu-entrypoint.sh /opt/sensu/bin/
+
+RUN apt update &&\
+    apt -y upgrade &&\
+    apt -y install git wget
+
+#Download and install Go 1.13.3 for Go modules.
+RUN wget https://dl.google.com/go/go1.13.3.linux-amd64.tar.gz && tar -xvf go1.13.3.linux-amd64.tar.gz &&\
```
majormoses: to make it easier to update it might be good to use an ENV var for the version number to reduce the number of places that are required when we want to build with a newer golang.

docker-scripts/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+    go build -o ../bin/sensuctl ./cmd/sensuctl &&\
+    rm -rf /opt/sensu/sensu-go
+
+#State directory lives outside container for persistent data.
+VOLUME [/var/lib/sensu]
+#Port 3000 for Sensu Go UI, 8080 for Sensu API, and port 8081 for sensu-agent websocket API communication.  
+EXPOSE 3000 8080 8081
+
+#Use Docker logs to see stdout of sensu-backend (ie. docker logs --details -f  `docker ps -q` )
+CMD ["/opt/sensu/bin/sensu-backend","start"]
```
majormoses: I am thinking we likely want a more standard `/entrypoint.sh` or something along the lines of that so that people can choose depending on their needs if they need to run `/opt/sensu/bin/sensu-backend init` prior to starting it to avoid seeding with default credentials.

georgespatton: There is the docker-scripts/sensu-entrypoint.sh, we could use that, or do you suggest another script?

docker-scripts/Dockerfile
georgespatton: I'll fix.

georgespatton: See latest commit https://github.com/sensu/sensu-go/pull/3456/commits/0713c3dd0693466607e4c1dc2d82a56dc22dfcc5

docker-scripts/Dockerfile
majormoses: that one seems reasonable from a quick look.

georgespatton: Ok, I have updated the README.md and Dockerfile for this change.  Should be good to go.

palourde:
Hi @georgespatton 

Thanks for taking the time to create this PR. We really appreciate it. However this is not something we wish to add to the sensu-go repository at this time, because unfortunately we do not have the time to maintain Docker images for the open-source project.

For context, we did have some Dockerfiles [until recently](https://github.com/sensu/sensu-go/pull/3428) but choose to remove them since they were simply broken and unmaintained, and it just caused frustrations.

That being said, we would be more than happy if someone decided to maintain Docker images in the Docker Hub.

Thanks again!

georgespatton:
I don't mind maintaining it.  I have moved this PR to its own repo:

https://github.com/georgespatton/sensu-go-oss-docker

PR #3465, opened 2019-12-12
Add secrets manager and secrets provider interface

nikkixdev:

## What is this change?

Adds support in sensu-go for secrets to be used as check, mutator, and handler environment variables. Secrets are injected directly into the executable resource environments, therefore are not stored/logged. For checks, secrets are sent in `CheckRequest`'s so the agent does not have to pull secrets from providers or store/log them.

Closes https://github.com/sensu/sensu-enterprise-go/issues/779

Yas.

Nah.

No, but some good discussion came about through the implementation.

No, merely offers support for commercial secret providers.

Unit tests were added. e2e testing cannot be done until secrets providers are added.

## Is this change a patch?

No, will go into master.


secrets/provider_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+// GetObjectMeta ...
+func (m *mockProvider) GetObjectMeta() corev2.ObjectMeta {
+	args := m.Called()
+	return args.Get(0).(corev2.ObjectMeta)
+}
+
+// SetObjectMeta ...
+func (m *mockProvider) SetObjectMeta(meta corev2.ObjectMeta) {
+	return
```
golangcibot: S1023: redundant `return` statement (from `gosimple`)

secrets/provider_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+// RBACName ...
+func (m *mockProvider) RBACName() string {
+	args := m.Called()
+	return args.Get(0).(string)
+}
+
+// SetNamespace ...
+func (m *mockProvider) SetNamespace(namespace string) {
+	return
```
golangcibot: S1023: redundant `return` statement (from `gosimple`)

secrets/provider.go
```diff
@@ -0,0 +1,92 @@
...
+	if m.providers == nil {
+		m.providers = map[string]Provider{}
+	}
+
+	m.providers[provider.GetObjectMeta().Name] = provider
+}
+
+// Providers returns the configured providers.
+func (m *ProviderManager) Providers() map[string]Provider {
+	m.mu.RLock()
```
palourde: I'm playing with this PR and I just hit a panic here:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x4e246b9]

goroutine 485 [running]:
sync.(*RWMutex).RLock(...)
	/usr/local/go/src/sync/rwmutex.go:48
github.com/sensu/sensu-go/secrets.(*ProviderManager).Providers(0x0, 0x0)
	/Users/splourde/src/sensu-go/secrets/provider.go:39 +0x49
github.com/sensu/sensu-go/secrets.(*ProviderManager).SubSecrets(0x0, 0xc000f60090, 0x9, 0x6, 0xc000a0f558, 0x0, 0x0, 0x0)
	/Users/splourde/src/sensu-go/secrets/provider.go:74 +0x139
github.com/sensu/sensu-go/backend/schedulerd.buildRequest(0xc000ad01c0, 0x55cadc0, 0xc0004a5300, 0x0, 0x0, 0x0, 0x0)
	/Users/splourde/src/sensu-go/backend/schedulerd/executor.go:342 +0x2b0
```

I'd assume it's simply because the mutex `m.mu` is never initialized!

nikkixdev: Do you recall how you produced this panic?

palourde: I literally did nothing, I just have to wait ~10 seconds. However I do have an interval check created, which I would assume is required so the scheduler tries to schedule it.

nikkixdev: I ran several types of checks before I submitted this PR, and again just now, and I still can't reproduce the panic. The provider manager mirrors that for [auth providers](https://github.com/sensu/sensu-go/blob/517d98798d5972c8d3848c9e4cee415fb652079c/backend/authentication/authenticator.go#L68-L113), and I don't see any initialization of the mutex anywhere. Perhaps I'm missing something..

nikkixdev: In any case, I will make sure it's initialized and we can see if that solves your issue?

nikkixdev: Okay added an initializing function, let me know if that doesn't resolve the panic! @palourde

backend/schedulerd/executor.go
```diff
@@ -21,41 +22,42 @@ var (
...
+	store                  store.Store
+	namespace              string
+	entityCache            *cache.Resource
+	secretsProviderManager *secrets.ProviderManager
 }
 
 // NewCheckExecutor creates a new check executor
-func NewCheckExecutor(bus messaging.MessageBus, namespace string, store store.Store, cache *cache.Resource) *CheckExecutor {
+func NewCheckExecutor(bus messaging.MessageBus, namespace string, store store.Store, cache *cache.Resource, secretsProviderManager *secrets.ProviderManager) *CheckExecutor {
 	return &CheckExecutor{bus: bus, namespace: namespace, store: store, entityCache: cache}
```
palourde: You added `secretsProviderManager` in the parameters but never set it on the `CheckExecutor` struct, which explains why I still see a nil panic!

PR #3477, opened 2019-12-17
Perf optimization for calculating the total number of resources in the cache

nikkixdev:

Optimizes the function to be used by the commercial entity limiter. Requires a PR in sensu-enterprise-go.

Testing on the perf stack now.

### DIFF
--- a/backend/store/cache/cache.go
+++ b/backend/store/cache/cache.go
@@ -176,15 +176,15 @@ func (r *Resource) Get(namespace string) []Value {
 	return r.cache[namespace]
 }
 
-// GetAll returns all cached resources across all namespaces.
-func (r *Resource) GetAll() []Value {
-	values := []Value{}
+// Size returns the size of all cached resources across all namespaces.
+func (r *Resource) Size() int {
+	var size int
 	r.cacheMu.Lock()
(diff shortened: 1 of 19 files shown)


backend/store/cache/cache.go
```diff
@@ -176,15 +176,15 @@ func (r *Resource) Get(namespace string) []Value {
 	return r.cache[namespace]
 }
 
-// GetAll returns all cached resources across all namespaces.
-func (r *Resource) GetAll() []Value {
-	values := []Value{}
+// Size returns the size of all cached resources across all namespaces.
+func (r *Resource) Size() int {
+	var size int
```
echlebek: Let's create a new method instead of replacing GetAll. It is likely still useful for other purposes, and it would be preferable to not remove an exported method.

nikkixdev: We're not using it anywhere 😞as it was specifically implemented for the purpose of getting the total count. I can add it back though if we think there's value there (🤞the linter doesn't complain).

echlebek: Linter will only complain about unexported unused functions, as it doesn't know what is importing the package.

PR #3490, opened 2020-01-09
Bugfix for implicit namespace access in Web UI

palourde:
## What is this change?

This PR fixes a bug where a user with implicit access to a namespace couldn't access that namespace via the Web UI.

Long story short, the Web UI uses the [FetchNamespace](https://github.com/sensu/sensu-go/blob/9aa87c9060ae41198733373d49a22f32d9db08cc/backend/api/namespace.go#L116) method which was never retrofitted with the implicit namespace feature, probably because `sensuctl` doesn't use that API so we forgot to back-port it.

Fixes https://github.com/sensu/sensu-enterprise-go/issues/821

Added

Nope

Nope

Nope

Added unit tests + manual testing.

## Is this change a patch?

Yep!

### DIFF
--- a/backend/api/namespace.go
+++ b/backend/api/namespace.go
@@ -8,6 +8,7 @@ import (
 	"github.com/sensu/sensu-go/backend/authorization"
 	"github.com/sensu/sensu-go/backend/authorization/rbac"
 	"github.com/sensu/sensu-go/backend/store"
+	"github.com/sirupsen/logrus"
 )
 
 type ruleVisitor interface {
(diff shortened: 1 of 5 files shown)


backend/api/namespace.go
```diff
@@ -20,7 +21,7 @@ type NamespaceClient struct {
 	auth   authorization.Authorizer
 }
 
-// NewnamespaceClient creates a new NamespaceClient, given a store and authorizer.
+// NewNamespaceClient creates a new NamespaceClient, given a store and authorizer.
```
echlebek: That's unfortunate :man_facepalming:

palourde:
@echlebek I did some changes so explicit access to a namespace can only be done via a ClusterRole. Additionally, I simplified the unit tests for the namespace router because they actually tested the RBAC logic, which is already covered in the `api` package.

PR #3504, opened 2020-01-14
sensuctl assets installed via Bonsai should have a namespace set

amdprophet:

## What is this change?

Sets the namespace field on Bonsai-installed sensuctl command assets to `sensuctl`.

It is currently impossible to install sensuctl command assets via Bonsai.

Yes.

Altered a test to ensure it fails when a namespace is not set. Hard-code the namespace in the Bonsai-install logic.

## Is this change a patch?

Yes but we're going to ship it with the next minor release.


amdprophet:
@nikkixdev @echlebek sensuctl uses the same asset manager from the agent & backend packages. sensuctl doesn't have a concept of namespaces but needs to adhere to the interfaces provided by the asset manager. These assets are only stored on the system where sensuctl is being run.

Issue #3501, opened 2020-01-14
Sensuctl entity delete returns false confirmation


nikkixdev:
## Expected Behavior
When deleting an entity that does not exist:
```
$ sensuctl entity delete test-entity --skip-confirm
Error: not found
```

## Current Behavior
When deleting an entity that does not exist:
```
$ sensuctl entity delete test-entity --skip-confirm
Deleted
```

## Possible Solution
Return an error on entity delete if that entity does not exist. Check deletion, for example, work as expected but both resources use the generic client. 

## Steps to Reproduce (for bugs)
1. `sensuctl entity delete test-entity --skip-confirm`, where `test-entity` does not exist.

## Context
Found this through a failed sensu-go-qa-crucible test.. this could be a recent regression.

## Your Environment
* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): https://github.com/sensu/sensu-go/commit/dc839c49c603843298112cba2252681c13bc662b
* Installation method (packages, binaries, docker etc.): binaries
* Operating System and version (e.g. Ubuntu 14.04): MacOS

PR #3505, opened 2020-01-15
Add more specific error checking for entity delete

nikkixdev:

## What is this change?

Adds more specific error checking for entity delete.

Closes https://github.com/sensu/sensu-go/issues/3501

Yas.

Nah.

Nah.

We good.

https://github.com/sensu/sensu-go-qa-crucible/blob/e3fba2059f6a5fb9a5494b21c4816daafafdc23d/spec/backend/entities_spec.rb#L31-L42

## Is this change a patch?

Yes, but in an attempt not to scope creep, I will merge this into master.

### DIFF
--- a/backend/apid/actions/delete_events_by_entity.go
+++ b/backend/apid/actions/delete_events_by_entity.go
@@ -11,7 +11,8 @@ import (
 )
 
 type EntityDeleter struct {
-	Store store.Store
+	EntityStore store.EntityStore
+	EventStore  store.EventStore
 }
 
 func (d EntityDeleter) Delete(req *http.Request) (interface{}, error) {
@@ -21,7 +22,7 @@ func (d EntityDeleter) Delete(req *http.Request) (interface{}, error) {
 		return nil, NewError(InvalidArgument, err)
 	}
 
-	events, err := d.Store.GetEventsByEntity(req.Context(), entityName, &store.SelectionPredicate{})
+	events, err := d.EventStore.GetEventsByEntity(req.Context(), entityName, &store.SelectionPredicate{})
 	if err != nil {
 		return nil, fmt.Errorf("error fetching events for entity: %s", err)
 	}
@@ -31,7 +32,7 @@ func (d EntityDeleter) Delete(req *http.Request) (interface{}, error) {
 			// improbable
 			continue
 		}
-		err := d.Store.DeleteEventByEntityCheck(req.Context(), entityName, event.Check.Name)
+		err := d.EventStore.DeleteEventByEntityCheck(req.Context(), entityName, event.Check.Name)
 		if err != nil {
 			logger := logger.WithFields(logrus.Fields{
 				"entity":    entityName,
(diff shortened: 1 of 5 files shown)


backend/apid/actions/delete_events_by_entity.go
```diff
@@ -11,8 +11,7 @@ import (
 )
 
 type EntityDeleter struct {
-	EntityStore store.EntityStore
-	EventStore  store.EventStore
```
echlebek: I'm sorry to say, this will break the EntityDeleter when the postgres event store is used. The separate EventStore needs to stay until we have a wholly swappable Store.

backend/apid/actions/delete_events_by_entity.go
```diff
@@ -43,5 +42,14 @@ func (d EntityDeleter) Delete(req *http.Request) (interface{}, error) {
...
+	result, err := d.Store.GetEntityByName(req.Context(), entityName)
+	if err != nil {
+		return nil, NewError(InternalErr, err)
+	}
+
+	if result == nil {
+		return nil, NewErrorf(NotFound)
+	}
+
+	return nil, d.Store.DeleteEntityByName(req.Context(), entityName)
```
echlebek: I sympathize that it's often confusing to issue a successful delete command, when no such entity existed. However, the original code actually has correct behaviour.

We should not check for existence before issuing the delete, and we should not return an error for requests that issue a delete for a non-existent resource. Doing so would violate the idempotency property of delete requests, and it would also be a breaking change for our HTTP API.

See https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html section 9.1.2 for more information about idempotency.

If other parts of the server exhibit this property, they should also be changed to not return an error, as they would not be conforming to how HTTP delete is supposed to work.

If we want sensuctl to have this behaviour, we could issue a GET followed by a DELETE, if the GET succeeds, in sensuctl.

nikkixdev: I'm a little confused, as we use similar logic for other delete endpoints:
https://github.com/sensu/sensu-go/blob/dae00ac999bb340132af11319a4695649801a4d8/backend/apid/actions/events.go#L71-L85

Additionally, [this test](https://github.com/sensu/sensu-go-qa-crucible/blob/e3fba2059f6a5fb9a5494b21c4816daafafdc23d/spec/backend/entities_spec.rb#L31-L42) (and for all other resources) indicates that we always have returned an error when attempting to delete a resource that does not exist. The fact that this test has recently started to fail indicates that the original behavior must have changed in order to cause this bug.

If idempotency is the issue here, then I would suggest applying that same idea consistently across all resources. For example, the generic client and API functions return errors for the same reason (and entities are a snowflake).
```
$ sensuctl check delete test-check --skip-confirm
Error: not found
```

echlebek: I believe that is incorrect behaviour for an HTTP server.

I can't explain why the test has recently started to fail however. That should definitely be investigated!

echlebek: Edit: I'm wrong about the response code. Idempotency is about effects on the state of the server, not the response code.

echlebek:
I was wrong about the response code for delete, so after the EventStore is put back we can merge this!

backend/apid/actions/delete_events_by_entity.go
nikkixdev: I didn't realize this had implications on postgres, good catch! I've added it back.

Issue #3267, opened 2019-09-09
Handlers for keepalive events should be configurable


cwjohnston:
## Feature Suggestion

It should be possible to send keepalive events to any configured handler.

## Possible Implementation

Allow agent configuration to specify a list of handlers which should receive the agent’s keepalive events.

## Context

This is a 1.x parity feature.

Current implementation in Sensu Go is limited to sending keepalive events to a `keepalive` handler in the agent namespace.

portertech:
We need to implement Agent `keepalive-handlers`, the keepalive generated events can pull the handler list from this.

PR #3507, opened 2020-01-16
Custom handlers for keepalive events

palourde:
## What is this change?

This PR adds support for a custom list of handlers for entities keepalive events, on an per agent basis.

It also updates the description of some flags in the help usage of the agent, so flags that accept a `[]string` have a consistent help usage. 

Closes https://github.com/sensu/sensu-go/issues/3267

Added

I'm a bit shocked it was as simple as this, I'm still trying to figure out if I didn't missed anything!

Nope

Will be documented via https://github.com/sensu/sensu-docs/issues/2108.

Manually tested. I made sure this change was backward compatible.

Will be automatically tested via https://github.com/sensu/sensu-go-qa-crucible/issues/101 since it's a new feature.

## Is this change a patch?

Nope


agent/cmd/start.go
```diff
@@ -37,6 +37,7 @@ const (
 	flagDeregistrationHandler    = "deregistration-handler"
 	flagEventsRateLimit          = "events-rate-limit"
 	flagEventsBurstLimit         = "events-burst-limit"
+	flagKeepaliveHandler         = "keepalive-handler"
```
nikkixdev: This is going to sound like such a nitpick, but wouldn't `keepalive-handlers` make more sense since a user can define multiple handlers? (ex. `--subscriptions` can define 0 to many subscriptions)

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 - Added the secrets provider interface and secrets provider manager to be used
 by commerical secrets providers. Implemented for checks, mutators, and handlers.
 - Added the `secrets` field to checks, mutators, and handlers.
+- Added the `keepalive-handler` configuration flag on the agent to specify the
```
nikkixdev:
```suggestion
- Added the `keepalive-handlers` configuration flag on the agent to specify the
```

palourde:
@echlebek I was planning on adding a test in our QA crucible to really test it from end to end. I could add a simple unit test in the keepalived package but to prove that it's actually getting executed, I would probably need a more complex test which I think is less preferable than a simple e2e test. WDYT?

PR #3510, opened 2020-01-18
Don't fail on check TTL after keepalive fails

### LINKED ISSUE #3304: Check TTL events can be created even if failing keepalive events exist
## Expected Behavior
An entity with a keepalive failure event should not create a check TTL event.

## Current Behavior
Failing events are created for both keepalives and checks.

## Possible Solution
When a check TTL event fails to fire, the handler should consult the store for whether or not a failing keepalive event exists for the entity in question.

1. Create a check with TTL enabled.
2. Take the associated entity offline.
3. Observe failing events for both the keepalive and check.

## Context
Reported by steveberryman in the community slack.

echlebek:
## What is this change?

This commit modifies how eventd responds to failing check TTL
events. Previously, check TTL events would occur even if failing
keepalive events had been recorded. Now, eventd will not issue
failing check TTL events if a keepalive event for the entity has
already occurred.

Closes #3304

Yes

No docs changes required.

I created a check with TTL greater than agent keepalive timeout, got it executing on an agent, and terminated the agent. I observed a keepalive failure occurred. I then observed in the backend logs that the check TTL switch had fired, but no failing event was created for the check.

On the flip side, if the keepalive timeout is longer than the check TTL, then a check TTL event does get created.

## Is this change a patch?

Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ resources.
 - Add support for the `--format` flag in the `sensuctl command list` subcommand.
 - Namespace can be ommited from event when performing an HTTP POST request to
 the `/events` endpoint.
+- Fixed a bug where failing check TTL events could occur event if keepalive
+failures had already occurred.
 
 ## [5.16.1] - 2019-12-18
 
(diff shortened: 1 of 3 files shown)


echlebek:
image

PR #3528, opened 2020-01-29
Bugfixes to curated namespaces

### LINKED ISSUE #3527: Curated namespaces do not work with a RoleBinding/ClusterRole combination

palourde:
## What is this change?

It fixes two bugs in the curated namespace functionality:

- A RoleBinding that refers to a ClusterRole should grant implicit access to its namespace
- A ClusterRoleBinding that refers to a ClusterRole, which grants access to  at least 1 namespaced resource, should grant access to all namespaces

Closes #3527

Added

Nope

Curated namespaces are hard! 😅 

Nope

Unit tests

## Is this change a patch?

Yep


backend/api/namespace.go
```diff
@@ -127,6 +116,29 @@ func (a *NamespaceClient) ListNamespaces(ctx context.Context) ([]*corev2.Namespa
...
+			return true
+		}
+
+		// Determine if this ClusterRoleBinding provides implicit access to
+		// namespaced resources
+		if binding.GetObjectMeta().Namespace == "" {
+			for _, resource := range rule.Resources {
+				if stringsutil.InArray(resource, corev2.CommonCoreResources) {
+					// All resources of type "namespace" are allowed
+					logger.Debugf("all namespaces implicitely authorized by the binding %s", binding.GetObjectMeta().Name)
```
jamesdphillips: at the risk of being a pedant... `s/implicitely/implicitly/g`

palourde: Good catch! 😅

PR #3537, opened 2020-01-30
Add unique IDs to events

### LINKED ISSUE #3125: Event IDs for tracing events through Sensu
- Add mechanism to create a unique ID [UUID] for a given event
- Update Backend and Agents Events' APIs to create a unique event ID on event submission
- Log event field name: event_uuid
- Go through the whole system and update event logging to reference that event ID for context (this will allow us to 'thread the needle' for a particular event)

echlebek:
## What is this change?

This CL adds event IDs to Sensu. The IDs are UUIDs, and are logged at various points throughout the event lifecycle. Support has also been added to sensuctl event list, and sensuctl event info.

Closes #3125

Yes

Protobuf compiler library is now v3, marshal/unmarshal logic is slightly different. Wire format is unaffected.

Docs changes required, we need to add the event_id field to the API docs and reference. cc @hillaryfraley 

Manual testing by watching loging, viewing output of sensuctl.

## Is this change a patch?

No

### DIFF
--- a/cli/commands/event/list.go
+++ b/cli/commands/event/list.go
@@ -129,6 +129,16 @@ func printToTable(results interface{}, writer io.Writer) {
 				return time.String()
 			},
 		},
+		{
+			Title: "ID",
+			CellTransformer: func(data interface{}) string {
+				event, ok := data.(corev2.Event)
+				if !ok {
+					return cli.TypeError
+				}
+				return event.GetUUID().String()
+			},
+		},
 	})
 
 	table.Render(writer, results)
--- a/cli/commands/event/list.go
+++ b/cli/commands/event/list.go
@@ -7,6 +7,7 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/google/uuid"
 	corev2 "github.com/sensu/sensu-go/api/core/v2"
 	"github.com/sensu/sensu-go/cli"
 	"github.com/sensu/sensu-go/cli/client"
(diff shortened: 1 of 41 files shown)


echlebek:
@palourde, I'll add logging context in the areas you mentioned.

@nikkixdev, I'm personally content with the current method of displaying zeroed UUIDs, but I'm open to other ideas if you have some. It's pretty transitory, so I don't see it as being a big deal either way.

nikkixdev:
@echlebek I don't think the zeroed UUIDs is a bad approach, just a little clunky since it's being omitted from the JSON/YAML configs when empty. `n/a` or `00000000-0000-0000-0000-000000000000` or even an empty string would be sufficient in tabular format. Regardless of the decision, we should make sure to document that behavior in the event reference so users don't think their event is broken (just that it hasn't run since the upgrade).

echlebek:
@nikkixdev you raise a good point about it being omitted from JSON/YAML. Let's omit if zeroed from the tabular display as well, which would make the display more coherent between methods.

cli/commands/event/list.go
```diff
@@ -129,6 +130,20 @@ func printToTable(results interface{}, writer io.Writer) {
...
+			CellTransformer: func(data interface{}) string {
+				event, ok := data.(corev2.Event)
+				if !ok {
+					return cli.TypeError
+				}
+				if id := event.GetUUID(); id == uuid.Nil {
+					return ""
+				} else {
+					return id.String()
+				}
```
nikkixdev: 😄

Issue #3425, opened 2019-11-25
sensuctl add asset from Bonsai fails for 'v' prefaced asset versions


nixwiz:
When attempting add an asset from Bonsai via ```sensuctl asset add```, an asset that has its version prefaced with a 'v' will fail to download, as the 'v' is stripped from the requested URL.
## Expected Behavior

For internal sorting (latest/outdated check), stripping of the 'v' is appropriate, but it should not be removed from the requested URL.
## Current Behavior

The 'v' is being stripped from the URL causing a 404 error when attempting to add.
## Steps to Reproduce (for bugs)

1.  ```sensuctl add asset nixwiz/sensu-go-fatigue-check-filter:v0.2.2``` will fail, attempting a GET on the bonsai API for '/nixwiz/sensu-go-fatigue-check-filter/0.2.2/release_asset_builds'
2.  ```sensuctl add asset nixwiz/sensu-go-fatigue-check-filter:0.2.1``` will work

## Context

I recently changed the versioning of my filter asset to including the 'v' prefix on the version.   Following this, I am no longer able to use ```sensuctl asset add``` to add it from Bonsai.
## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent):  sensuctl version 5.15.0, build 13884593ee08a7f25d7f66a8b71da61c529de014, built 2019-11-19T20:20:17Z 
* Installation method (packages, binaries, docker etc.):  binaries
* Operating System and version (e.g. Ubuntu 14.04):

portertech:
We need to revisit the Bonsai work around tags with "v".

PING @jspaleta @amdprophet

julian7:
I just found this error myself. I'm wondering whether it's a relative easy thing to fix:

- version names are thanking you, but they can be fetched from Bonsai
- sensuctl, however, uses Hashicorp's goversion library, which make versions uniform and comparable
- sensuctl then try to use this canonized version name to fetch assets from Bonsai, instead of using the raw version.

It can be fixed either in Bonsai (I can't see its source, but I bet this is not an isolated attribute, therefore changing its behavior might affect other subsystems), or can be fixed in sensuctl itself.

This is the path I see being way simpler: if `(*Asset).HasVersion(*goversion.Version) bool` is replaced with something like `(*Asset).SelectVersion(*goversion.Version) *goversion.Version`, which returns the original version coming directly from Fetch Asset API endpoint, its `.Original()` could be used to fetch Asset Version.

[...]

PR #3545, opened 2020-02-04
Support Bonsai assets versions prefixed with v

palourde:

## What is this change?

It adds supports for Bonsai assets that use the version format `vX.Y.Z`, in addition to semver (`X.Y.Z`).

Closes https://github.com/sensu/sensu-go/issues/3425

Added

I tried my best to make sure I didn't forget anything; it appears this package is only used in sensuctl but I might be forgetting something.

Nope

Nope

Manually tested it and added a unit test where available.

## Is this change a patch?

Yep

### DIFF
--- a/bonsai/bonsai.go
+++ b/bonsai/bonsai.go
@@ -10,6 +10,20 @@ import (
 	corev2 "github.com/sensu/sensu-go/api/core/v2"
 )
 
+const (
+	// NameAnnotation represents a Bonsai asset name
+	NameAnnotation = "io.sensu.bonsai.name"
+
+	// NamespaceAnnotation represents a Bonsai asset namenamespace
+	NamespaceAnnotation = "io.sensu.bonsai.namespace"
+
+	// URLAnnotation represents the Bonsai API URL
+	URLAnnotation = "io.sensu.bonsai.api_url"
+
+	// VersionAnnotation represents a Bonsai asset version
+	VersionAnnotation = "io.sensu.bonsai.version"
+)
+
 // Asset stores information about an asset (metadata, versions, etc.) from Bonsai
 type Asset struct {
 	// Name is the full name (including namespace) of the asset
@@ -86,6 +100,10 @@ func (b *Asset) BonsaiVersion(version *goversion.Version) (*goversion.Version, e
 	if version == nil {
 		// No version was requested, therefore return the latest version
 		v := b.LatestVersion()
+		if v == nil {
+			return nil, fmt.Errorf("asset %q does not have any available versions", b.Name)
+		}
+
 		fmt.Println("no version specified, using latest:", v.Original())
 		return v, nil
 	} else if ok, v := b.HasVersion(version); ok {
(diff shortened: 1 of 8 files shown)


julian7:
The only problem with using `version.Original` is that it is not the version which is equivalent to the version the bonsai asset describes, but it must be exactly in the same format as it is advertised.

Example: if a bonsai asset has v0.0.1 and v0.0.2, but the version "0.0.2" has been specified, there will be a match (since converted "v0.0.2" and "0.0.2" are equivalent), but then it will fail at pulling the selected version. One has to specify "v0.0.2", the exact version string, to download.

I think we should pull the exact string from the version list, if a match is found, and it should use its original to do downloads.

palourde:
@julian7 I pushed a new commit which address this concern:

```sh
$ sensuctl asset add nixwiz/sensu-go-fatigue-check-filter:v0.2.2
fetching bonsai asset: nixwiz/sensu-go-fatigue-check-filter:v0.2.2
added asset: nixwiz/sensu-go-fatigue-check-filter:v0.2.2
$ sensuctl asset add nixwiz/sensu-go-fatigue-check-filter:0.2.2
fetching bonsai asset: nixwiz/sensu-go-fatigue-check-filter:v0.2.2
added asset: nixwiz/sensu-go-fatigue-check-filter:v0.2.2
$ sensuctl asset outdated
               Asset Name                            Bonsai Asset               Current Version   Latest Version
 ────────────────────────────────────── ────────────────────────────────────── ───────────────── ────────────────
  nixwiz/sensu-go-fatigue-check-filter   nixwiz/sensu-go-fatigue-check-filter   v0.2.2            0.4.0
```

palourde:
Here's the expected behavior now:

When installing an asset

| Command        | Bonsai Versions           | Result  |
| ------------- |:-------------:| -----:|
| `sensuctl asset add sensu/foo`     | `0.1.0`, `v0.2.0` | Adding `v0.2.0` |
| `sensuctl asset add sensu/foo:v0.1.0`     | `0.1.0`, `v0.2.0` | Adding `0.1.0` |
| `sensuctl asset add sensu/foo:0.2.0`     | `0.1.0`, `v0.2.0` | Adding `v0.2.0` |
| `sensuctl asset outdated`     | `0.1.0`, `v0.2.0` | `0.1.0`is outdated |
| `sensuctl asset outdated`     | `0.1.0`, `v0.2.0`, `0.3.0` | `0.1.0` & `v0.2.0` are outdated |

julian7:
I approve this change (whatever it means). It already made me smile yesterday, when I went through the code changes.

bonsai/bonsai.go
```diff
@@ -78,14 +93,46 @@ func NewBaseAsset(name string) (*BaseAsset, error) {
...
+// returned if the requested version does not exist
+func (b *Asset) BonsaiVersion(version *goversion.Version) (*goversion.Version, error) {
+	if version == nil {
+		// No version was requested, therefore return the latest version
+		v := b.LatestVersion()
+		if v == nil {
+			return nil, fmt.Errorf("asset %q does not have any available versions", b.Name)
+		}
+
+		fmt.Println("no version specified, using latest:", v.Original())
```
echlebek: Should probably be removed?

palourde: It was actually already present, which produces the following:

```
$ sensuctl asset add nixwiz/sensu-go-fatigue-check-filter
no version specified, using latest: 0.4.0
fetching bonsai asset: nixwiz/sensu-go-fatigue-check-filter:0.4.0
added asset: nixwiz/sensu-go-fatigue-check-filter:0.4.0
```

echlebek: Or maybe moved to somewhere less library-ish?

Issue #3525, opened 2020-01-28
Make entity.metadata.entity_class optional for events published to the events APIs


calebhailey:
## Feature Suggestion

Related to #3081. 

If we implement #3081, and make that default to creating proxy entities, then `entity.metadata.entity_class` should not be required for valid events published to the Events APIs.  

## Possible Implementation

Minimum valid events published to the events APIs should be as follows: 

- `POST /api/core/v2/namespaces/:namespace/events` 

   ```json
   {
     "entity": {
       "metadata": {
         "name": "I-424242"
       }
     },
     "check": {
       "metadata": {
         "name": "helloworld"
       },
       "status": 0,
       "output": "everything is a-ok!"
     }
   }
   ```

   Default values: 
   - `entity.metadata.namespace` defaults to the `:namespace` URL path element
   - `check.metadata.namespace` defaults to the `:namespace` URL path element  
   
   <br>_NOTE: these defaults are not yet implemented._

- `POST /api/core/v2/namespaces/:namespace/events/:entity/:check` 

   ```json
   {
     "check": {
       "status": 0,
       "output": "everything is a-ok!"
     }
   }
   ```

   Default values: 
   - `entity.metadata.name` defaults to the `:entity` URL path element 
   - `entity.metadata.namespace` defaults to the `:namespace` URL path element 
   - `check.metadata.name` defaults to the `:check` URL path element 
   - `check.metadata.namespace` defaults to the `:namespace` URL path element  
   
   <br>_NOTE: these default values are already implemented._

## Context

Let's make the Sensu APIs easier to use! 🙌

echlebek:
@calebhailey, in the next major revision we hope to make all APIs use wrapped data types, which will help us avoid some major technical pitfalls we have with the current system. Is that something you see as being compatible with ease of use?

portertech:
We are going to address #3081 and ensure `event.entity.entity_class` (not `entity.metadata.entity_class`) defaults to "proxy" and is not a required field on API POST/PUT.

calebhailey:
![tenor](https://user-images.githubusercontent.com/141373/74494855-0cc6d780-4e8b-11ea-855a-44c0617ad058.gif)

PR #3551, opened 2020-02-12
Make entity_class optional in /events API

nikkixdev:

## What is this change?

Default `event.entity.entity_class` to `proxy` in the POST/PUT `/events` API.

Closes https://github.com/sensu/sensu-go/issues/3525

Yas.

Nah.

Nah.

We good.

Unit test added. Tested the payload in the original issue.

## Is this change a patch?

Milestone is 5.18.0 so merging this into master.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Changed
 - Updated Go version from 1.13.5 to 1.13.7.
+- Default `event.entity.entity_class` to `proxy` in the POST/PUT `/events` API.
 
 ## [5.17.1] - 2020-01-31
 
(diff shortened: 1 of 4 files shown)


echlebek:
Perhaps we should disable client-side validation in sensuctl, leaving it up to the backend alone.

nikkixdev:
@echlebek good call.. let me see if that is possible without a large amount of refactoring..

nikkixdev:
@echlebek so I've done some local testing and debugging, and the backend already attempts to validate the resource at the store level. Those errors are properly passed back up to the HTTP API. Therefore, `sensuctl create`'s validation call is redundant.. wdyt of just removing that block completely?

https://github.com/sensu/sensu-go/blob/e762d55833e5ce67edd3161ce436bce0e9823db6/cli/commands/create/create.go#L253-L261

Validation errors are still surfaced:
```
Error: error putting resource 0 (/api/core/v2/namespaces/default/events/I-..%2F/helloworld): entity is invalid: entity name cannot contain spaces or special characters
```

echlebek:
@nikkixdev I think it's fine to remove client-side validation, it should reduce friction between minor versions as well.

calebhailey:
🙌 EXCITED 🙌

PR #3555, opened 2020-02-13
Fix sensuctl proxy from environment

### LINKED ISSUE #3538: sensuctl doesn't appear to honor proxy environment variables.
From some testing that I've been doing, it doesn't appear that sensuctl honors the http_proxy or https_proxy variables set in a user's profile. Here's what I've done so far to lead me to this conclusion:

Test env:
* 1 proxy server (squid)
* 1 proxy client

On the proxy server, I ran a tcpdump after configuring squid. I could see traffic flowing through the main interface and I was getting responses from curls, apt-get updates, and other commands after setting the aforementioned variables on my client. However, when using sensuctl from the client, no packets appeared to be flowing from any "asset add" commands. While I could download assets and retrieve information on checks and other resources, none of the traffic seemed to go through the proxy server I had configured.

I believe this to be an issue with how we're using the go-resty library to send calls to Bonsai, and subsequently any Sensu deployment. The library does support the ability to use a proxy (https://github.com/go-resty/resty#proxy-settings---client-as-well-as-at-request-level), but from what I've seen in the code, it doesn't appear like we make any considerations for this.

echlebek:
## What is this change?

This commit upgrades the HTTP client library used by sensuctl. In
doing so, it adds support for proxying HTTP requests from the
environment variables http_proxy, and https_proxy, and no_proxy.

Closes #3538 

The go-resty library had to be upgraded to a new major version.

I used a free proxy to send requests to a sensu-staging instance, and it worked.

## Is this change a patch?

Yes, but targeting for 5.18.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Fixed
+- sensuctl now supports the http_proxy, https_proxy, and no_proxy environment
+variables.
+
 ### Changed
 - Updated Go version from 1.13.5 to 1.13.7.
 
--- a/go.mod
+++ b/go.mod
@@ -23,7 +23,6 @@ require (
 	github.com/frankban/quicktest v1.7.2 // indirect
 	github.com/ghodss/yaml v1.0.0
 	github.com/go-ole/go-ole v0.0.0-20170209151332-de8695c8edbf // indirect
-	github.com/go-resty/resty v0.0.0-20170925192930-9ac9c42358f7
 	github.com/go-resty/resty/v2 v2.1.0
 	github.com/gogo/protobuf v1.3.1
 	github.com/golang/groupcache v0.0.0-20191002201903-404acd9df4cc // indirect
--- a/go.mod
+++ b/go.mod
@@ -24,6 +24,7 @@ require (
 	github.com/ghodss/yaml v1.0.0
 	github.com/go-ole/go-ole v0.0.0-20170209151332-de8695c8edbf // indirect
 	github.com/go-resty/resty v0.0.0-20170925192930-9ac9c42358f7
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Fixed
+- sensuctl now supports the http_proxy, https_proxy, and no_proxy environment
```
ccressent: `https_proxy` is fine; is it `https_proxy` or `HTTPS_PROXY` or both? Same question for `no_proxy`. Is `all_proxy` available? Most programs follow curl/libcurl's convention; see the "ENVIRONMENT" section of `curl(1)` for details.

ccressent: Answered my own question; resty uses [`http.ProxyFromEnvironment`](https://golang.org/pkg/net/http/#ProxyFromEnvironment).

echlebek: The Go documentation mentions the capitalized versions of these environment variables, but I tested it with the lower-cased version, since that's what curl uses.

The Go documentation does not mention `all_proxy`, so I assume it's not supported.

echlebek: I believe resty simply passes through this configuration to the underlying http client.

go.mod
```diff
@@ -24,6 +24,7 @@ require (
 	github.com/ghodss/yaml v1.0.0
 	github.com/go-ole/go-ole v0.0.0-20170209151332-de8695c8edbf // indirect
 	github.com/go-resty/resty v0.0.0-20170925192930-9ac9c42358f7
```
palourde: Could we update resty in `bonsai/client.go` too so we can remove that version?

echlebek: sure

echlebek: I decided I would simply remove resty as a dependency from the bonsai package, as it's not really adding any value, and its use hasn't expanded so much that it would be difficult to remove.

PR #3557, opened 2020-02-13
Rework TestCreds() errors and tests

### LINKED ISSUE #3556: Misleading error message in `TestCreds()`
## Expected Behavior

The error returned by `TestCreds()`, invoked through `sensuctl test-creds` or the API should be clear and correct.

## Current Behavior

1. Most errors lead to an unclear/wrong "basic provider is disabled" message
2. The precise error coming from the provider is swallowed
3. If the basic provider is disabled, it leads to the wrong error message: "invalid username and/or password"
4. The tests around `TestCreds()` are broken

## Possible Solution

1. Swap error messages
2. Bubble up error coming from provider
3. Fix the tests

One way to observe the misleading/incorrect messages is to watch the backend logs when trying to `sensuctl test-creds` with incorrect credentials.

## Context

Noticed weird error while debugging an issue for a customer with CREs.

## Your Environment

* Sensu version: `5.17.1`

ccressent:

Fix #3556 

I suppose. Added.

I started by simply swapping 2 lines to get the right errors out of `TestCreds()`, but I soon realized that the tests for this function were broken in the first place. So I tried to fix them a little bit.

No.

Ran `sensuctl test-creds` and watched the backend's logs.

## Is this change a patch?

I thought it could have been, but Simon said no.

### DIFF
--- a/backend/api/authentication.go
+++ b/backend/api/authentication.go
@@ -71,11 +71,12 @@ func (a *AuthenticationClient) TestCreds(ctx context.Context, username, password
 	if basic, ok := providers[basic.Type]; ok {
 		if _, err := basic.Authenticate(ctx, username, password); err == nil {
 			return nil
+		} else {
+			return err
 		}
-		return errors.New("basic provider is disabled")
 	}
 
-	return errors.New("invalid username and/or password")
+	return errors.New("basic provider is disabled")
 }
 
 // Logout logs a user out. The context must carry the user's access and refresh
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 ### Changed
 - Updated Go version from 1.13.5 to 1.13.7.
 
+### Fixed
+- The `auth/test` endpoint now returns the correct error messages.
+
 ## [5.17.1] - 2020-01-31
 
 ### Fixed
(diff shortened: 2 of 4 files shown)


backend/api/authentication.go
```diff
@@ -71,11 +71,12 @@ func (a *AuthenticationClient) TestCreds(ctx context.Context, username, password
 	if basic, ok := providers[basic.Type]; ok {
 		if _, err := basic.Authenticate(ctx, username, password); err == nil {
 			return nil
+		} else {
+			return err
```
palourde: I don't think we should return the actual error to the user (maybe just log it), simply because we would actually reveal whether a user actually exists or not and if it's because the password is wrong, e.g. https://github.com/sensu/sensu-go/blob/572b4b5facc4a4f7b310dcf0eb34ef711e9583e5/backend/store/etcd/user_store.go#L40-L43

echlebek: Let's log the store error, keep it as is, and write a different error to the user in the HTTP handler.

ccressent: Thanks for raising that @palourde; just like @echlebek I think that "obfuscation" belongs to the HTTP handlers, the boundary between the outside and the inside. Internally, we want the most precise error we can get.

Does that work for you @palourde?

palourde: Ah yes all good, I didn't realized the HTTP handler does not respond with the error!

ccressent: So based on [the current handler code](https://github.com/sensu/sensu-go/blob/master/backend/apid/routers/authentication.go#L70-L89) we agree it's all good? We log the exact error in the backend's logs but we only tell the client it's a 401 status code.

Issue #3559, opened 2020-02-14
Secret resource validation requires namespace


portertech:
When using `sensuctl create -f` to create a number of Sensu resources from YAML files, I get `resource is invalid: secrets require a namespace`.

## Expected Behavior
Secret resources should use the `sensuctl` namespace, either the set default or explicit `--namespace`.

## Current Behavior
`sensuctl create -f` returns an error, `resource is invalid: secrets require a namespace`.

## Context
https://github.com/portertech/sensu-stackdriver-kube/blob/715d4756dbb329cdda60930681e8f767a49bcc19/sensu/pipelines/stackdriver.yml#L10-L17

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): 5.17.1
* Installation method (packages, binaries, docker etc.): Binsary
* Operating System and version (e.g. Ubuntu 14.04): Mac 10.15.2 (19C57)

nikkixdev:
Secrets were the first namespaced resource implemented, and required some refactoring to some of the generic functions. I wouldn't be surprised if some of the cli tooling hasn't been ported over, such as the utility to load the current namespace into each command.

PR #3561, opened 2020-02-14
Fix wrapped resource namespaces

nikkixdev:

## What is this change?

Assigns the namespace of the wrapper if not set.

Closes https://github.com/sensu/sensu-go/issues/3559

Yas.

Nah.

Nah.

We good.

`sensuctl create -f secret.yml `
```
---
type: Secret
api_version: secrets/v1
metadata:
  name: mando
spec:
  id: TEST_SECRET
  provider: env
```

## Is this change a patch?

Targetting 5.18.0 in master

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 variables.
 - returns 401 instead of 500 when issues occur refreshing the access token.
 - Support Bonsai assets versions prefixed with the letter `v`.
+- Fixed a bug where wrapped resources were not getting their namespaces set by
+the default sensuctl configuration.
 
 ### Changed
 - Updated Go version from 1.13.5 to 1.13.7.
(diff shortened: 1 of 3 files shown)


nikkixdev:
@echlebek I have two answers!

**Why did you change the type of the slice throughout the program?**
The slice was changed to a slice of pointers because [setting the namespace within the top level wrapper object meta](https://github.com/sensu/sensu-go/pull/3561/files#diff-8451b2f0dac131c0abaf7fc53aaa8930R252) would otherwise not work.

**SetNamespace() is already invoked when the namespace is not set. Why doesn't SetNamespace work as expected?**
SetNamespace() can only be called on a resource, not a wrapper. The top level object meta of the wrapper is not exposed to the resource itself.

This was the simplest fix I came across for the bug. However I intentionally added you as a reviewer knowing you had the most experience with the core resource data type vs the wrapped resource. If there is a more efficient or generic way to do this, I would be happy to understand what that is!

echlebek:
I see, it's essentially a consequence of ObjectMeta not being a pointer. Thanks!

Issue #3081, opened 2019-06-24
Create proxy entity if an entity doesn't exist for events published to the Events API(s) 


calebhailey:
## Expected Behavior

For all events published to the `sensu-backend` /events API, check if the referenced `entity` exists in the API, and if not, create an entity with `"class": "proxy"`. 

## Current Behavior

We don't automatically create an entity upon processing an event if the event references an entity that doesn't exist in the registry (we used to do this in Sensu 1.x). 

## Possible Solution

Create missing entities (with `"class": "proxy"`) for events published to the `sensu-backend` /events API. 

## Context

The "monitoring event pipeline" workflow from Sensu Classic is compromised in Sensu Go without this change. 

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): any
* Installation method (packages, binaries, docker etc.): any
* Operating System and version (e.g. Ubuntu 14.04): debian stretch

calebhailey:
It's been a while since I've tested this. I have observed that this functionality is not present in `sensu-backend` (i.e. the backend API) as of version 5.10; however I don't recall if the behavior is different in the `sensu-agent` events API. IIRC, the sensu-agent does behave differently (perhaps already implementing this feature), which is great – but we should ideally see both of these /events API endpoints behave more consistently.

cwjohnston:
Adding my two cents here:

* I can confirm that submitting events to the agent API does create a proxy entity when the check specifies a `proxy_entity_name` which does not yet exist
* I agree that the backend `/events` API should implement this same behavior
* From a previous conversation I have noted the following per @portertech:
  > We should create the proxy entity, we should ensure the user posting the event has the right RBAC permissions to create both the event and the entity

roganartu:
Didn't see this when I opened #3179, but it's definitely related (though instead is a feature request to disable this functionality).

There's also #3178 which is a bug caused by the way the agent -> backend interaction creates new entities. It's unlikely an implementation of this would hit the same race condition that causes #3178, however the fix I proposed in #3195 makes significant changes to the `getProxyEntity` function likely to be re-used by the backend API for this, so I thought it was also worth linking here.

calebhailey:
@roganartu I like the idea of a toggle to disable this functionality. The default should be to accept all events (given valid authentication and RBAC), and create proxy entities for entities that don't exist; but if you want to disable this functionality, we add a `--reject-orphaned-events` flag (or similar), as you suggest in #3179.

[...]

PR #3569, opened 2020-02-19
Create proxy entites for events submitted via the backend API

palourde:
## What is this change?

It refactors the proxy entity creation logic to essentially move it from agentd to eventd, so events submitted directly via the backend API also create proxy entities if the given entity does not exist.

Closes https://github.com/sensu/sensu-go/issues/3081

Added

Nope

Nope

Nope

Added unit tests, manually tested and added e2e test via https://github.com/sensu/sensu-go-qa-crucible/pull/107, in additional to the existing tests.

## Is this change a patch?

Nope


backend/agentd/session_test.go
```diff
@@ -190,3 +187,11 @@ func TestSessionTerminateOnSendError(t *testing.T) {
 		t.Fatal("broken session never stopped")
 	}
 }
+
+type channelSubscriber struct {
```
golangcibot: type `channelSubscriber` is unused (from `unused`)

backend/eventd/entity.go
```diff
@@ -0,0 +1,55 @@
...
+	"context"
+
+	corev2 "github.com/sensu/sensu-go/api/core/v2"
+	"github.com/sensu/sensu-go/backend/store"
+)
+
+// createProxyEntity creates a proxy entity for the given event if the entity
+// does not exist already and returns the entity created
+func createProxyEntity(event *corev2.Event, s store.EntityStore) error {
+	entityName := event.Entity.Name
```
echlebek: I think this needs a nil check. If event.Entity is nil, then this will cause a panic.

The feature is kind of weird in general because the idea is that you don't need an entity... but you actually do, to specify its name.

Do we need to look at automatic proxy entity naming as a feature?

palourde: An entity and its name are actually required, since we validate the event just before calling this function: 

https://github.com/sensu/sensu-go/blob/59af146941d78108d4a96864122e498693ef5caa/backend/eventd/eventd.go#L236-L238

There are two use cases for this function:

1. A `proxy_entity_name` is configured in the check, therefore we need to replace the event's entity, which refers to the agent's entity that executed the check, with a brand new entity named using this `proxy_entity_name` attribute as its name
2. The user submitted an event via the backend API, in which case we take whatever entity was provided to create the proxy entity.

Makes sense?

PR #3573, opened 2020-02-20
requiring subscriptions/subscribers on check

### LINKED ISSUE #2932: Users can create empty subscriptions in checks
Users can add empty subscriptions to checks.

## Expected Behavior
Users cannot create checks with empty subscriptions. (`""`)

## Current Behavior
Users can create checks that have subscriptions that are the empty string.

## Possible Solution
Validate subscription names on the backend and optionally on the frontend.

Create a check with one or more blank subscriptions.

## Context
QA

tarcinil:
There is not point of creating a check if nothing can subscribe to it.  This changes `check.Validate()` to check if the length of the `Subscriptions` slice is not equal to zero.

Signed-off-by: Vern Burton <me@vernburton.com>

## What is this change?

fixes #2932 

It creates an organized experience that ensures that checks can be subscribed by entities.

Yes.

No.

The testing changes was larger than I expected.

No.

Created tests cases to ensure that error message was as expected

## Is this change a patch?

No.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ variables.
 the default sensuctl configuration.
 - read/writes `initializationKey` to/from `EtcdRoot`, while support legacy as fallback (read-only)
 - check for a non-200 response when fetching assets
+- ensuring that check/check_config have subscriptions
 
 ### Changed
 - Updated Go version from 1.13.5 to 1.13.7.
--- a/agent/config.go
+++ b/agent/config.go
@@ -213,6 +213,7 @@ func FixtureConfig() (*Config, func()) {
 		KeepaliveWarningTimeout: corev2.DefaultKeepaliveTimeout,
 		Namespace:               DefaultNamespace,
 		Password:                DefaultPassword,
+		Subscriptions: 		     []string{"linux"},
 		Socket: &SocketConfig{
 			Host: DefaultSocketHost,
 			Port: DefaultSocketPort,
--- a/agent/sockets_test.go
+++ b/agent/sockets_test.go
@@ -42,6 +42,7 @@ func TestHandleTCPMessages(t *testing.T) {
 		Name:   "app_01",
 		Output: "could not connect to something",
 		Source: "proxyEnt",
+		Subscribers: []string{"linux"},
 	}
 	bytes, _ := json.Marshal(payload)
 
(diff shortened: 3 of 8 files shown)


echlebek:
Thanks for the PR. However, the original issue was actually pointing at a different problem: checks subscriptions can be set to empty strings, which causes problems elsewhere in the system.

```
{
  "subscriptions": [""]
}
```

We could still accept this PR but it probably requires a slightly different issue.

tarcinil:
@echlebek I think that can be extend into this work.  

If `len(c.Subscriptions) > 0`, iterate over the range and check for an empty string in the slice, and push an error if found.

Does that sound like an appropriate idea?

echlebek:
I had to think about this one for a bit. Where it seems obvious to me that allowing empty strings as subscriptions is a bug, I wasn't so sure about an empty array.

I was a bit worried that accepting this PR would mean a breaking change, but I think it's probably better characterized as a bug fix.

I think the reasoning put forward in the PR makes sense, and I will note there is already a first-class way of creating a check that is not intended to run, called `publish`, so it seems unlikely that someone would depend on this behaviour.

Existing checks in the database can still be retrieved and modified to have subscriptions, even if they don't, so it shouldn't cause any existing check configuration to cause issues.

I'd like to call on @portertech to validate this PR as well before we merge it.

tarcinil:
I think that it would be a breaking change when I am looking at `agent/agent.go` for `newKeepalive`.  While it would create the event, the check contained within the event would likely not have a subscribe.  

The events would need to subscribe to the automatic entity `entity:agent` to be backwards compatible.

palourde:
@tarcinil I think subscribing to the agent's entity for keepalive is an excellent idea!

That being said, I think we should still add an exception for keepalive checks that do not have a subscription, otherwise any agents that do not get upgraded at the same time as the backend would essentially be unable to send keepalives, therefore introducing a breaking change.

portertech:
A check without subscriptions is still valid, usually in combination with `publish: false`, to be used for on-demand executions, see: https://docs.sensu.io/sensu-go/latest/api/checks/#the-checkscheckexecute-api-endpoint

portertech:
I think the validation should take check publish into account, only validating subscriptions if publish is true.

portertech:
As there are valid use cases for checks w/o subscriptions, I'm going to close this issue. We should still validate the subscriptions and check for empty strings.

PR #3580, opened 2020-02-24
Do not set namespace to global wrapped resources

palourde:

## What is this change?

It fixes a bug where sensuctl would attempt to set a namespace to cluster-wide resources, such as authentication providers.

Fix a bug found in staging.

Nope, it fixes an unreleased commit.

Nope

The ValidateResource function was largely untested, so I had to take the time to write proper unit tests.

Nope

Manually tested with sensu-enterprise-go + unit tests.

## Is this change a patch?

Nope

### DIFF
--- a/cli/commands/create/create_test.go
+++ b/cli/commands/create/create_test.go
@@ -247,7 +247,7 @@ func TestValidateResourcesStderr(t *testing.T) {
 	// copy the output in a separate goroutine so printing can't block indefinitely
 	go func() {
 		var buf bytes.Buffer
-		io.Copy(&buf, r)
+		_, _ = io.Copy(&buf, r)
 		ch <- buf.String()
 	}()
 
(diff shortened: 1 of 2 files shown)


cli/commands/create/create_test.go
```diff
@@ -180,3 +181,89 @@ func TestCreateCommandStdin(t *testing.T) {
...
+func TestValidateResourcesStderr(t *testing.T) {
+	oldStderr := os.Stderr
+	r, w, _ := os.Pipe()
+	os.Stderr = w
+
+	ch := make(chan string)
+	// copy the output in a separate goroutine so printing can't block indefinitely
+	go func() {
+		var buf bytes.Buffer
+		io.Copy(&buf, r)
```
golangcibot: Error return value of `io.Copy` is not checked (from `errcheck`)

PR #3581, opened 2020-02-25
Don't do an entity read when entity is an agent

echlebek:

This change avoids an etcd read in eventd when the event's entity is an agent.

It avoids trying to create a proxy entity, assuming instead that the event's entity is already full specified.

No

Unverified.

## Is this change a patch?

Yes

### DIFF
--- a/backend/eventd/eventd.go
+++ b/backend/eventd/eventd.go
@@ -248,10 +248,8 @@ func (e *Eventd) handleMessage(msg interface{}) error {
 
 	// Create a proxy entity if required and update the event's entity with it,
 	// but only if the event's entity is not an agent.
-	if event.Entity.EntityClass != corev2.EntityAgentClass {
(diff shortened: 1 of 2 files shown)


backend/eventd/eventd.go
```diff
@@ -246,9 +246,12 @@ func (e *Eventd) handleMessage(msg interface{}) error {
 
 	ctx := context.WithValue(context.Background(), corev2.NamespaceKey, event.Entity.Namespace)
 
-	// Create a proxy entity if required and update the event's entity with it
-	if err := createProxyEntity(event, e.store); err != nil {
-		return err
+	// Create a proxy entity if required and update the event's entity with it,
+	// but only if the event's entity is not an agent.
+	if event.Entity.EntityClass != corev2.EntityAgentClass {
```
palourde: That won't work for check with `proxy_entity_name` since the agent entity is returned as the event's entity!

PR #3598, opened 2020-03-04
Revert check command validation

echlebek:
## What is this change?

Unfortunately, it seems that check command validation is simply too
much of a change at this point. It breaks several test cases in our
QA suite, breaks examples in the docs, and cannot be done in the
5.x series of sensu-go.

cc @tarcinil 

An entry has been removed from CHANGELOG.md

Docs updates not required.

Unverified.

## Is this change a patch?

No


echlebek:
@nikkixdev I actually liked how the test examples were provided a command, I don't see a reason not to have it there, so that's why I decided to manually revert the change.

Issue #3591, opened 2020-03-03
Object metadata created by (HTTP API)


portertech:
Update all object metadata to contain a "created by" field. This field would be set by the API and override any user provided value. This field adds context that can be leveraged by features and improves optics on user creation and management.

## Feature Suggestion

Update all object metadata to contain a "created by" field. This field adds context that can be leveraged by features and improves optics on user creation and management.

## Possible Implementation
The object metadata created by field could be set by the API and override any user provided value.

nikkixdev:
Limiting the scope of this ticket to just the HTTP API. All resources that are created or updated via the HTTP API will populate the `created_by` field with the sensu user responsible for the changes. Agent side resources will be handled as part of https://github.com/sensu/sensu-go/issues/3601.

Issue #3601, opened 2020-03-05
Object metadata created by (agent side resources)


nikkictl:
## Feature Suggestion
Populate the `created_by` field for all agent side resources. These include but are not limited to:
- registered agent entities
- check events
- keepalive events
- statsd metrics events

## Possible Implementation
When creating different types of events on the agent, set `created_by` to be the agent user, defined by `agent.Config.User`.

## Context
https://github.com/sensu/sensu-go/issues/3591 implements `created_by` for the HTTP API. It would be nice to have this field populated for the aforementioned use cases. While `created_by` is intended for use by `sensuctl prune` and likely not used for agent entities and events, it would be nice to have some consistency for how the field is used across all resources.

PR #3604, opened 2020-03-05
Add created_by field to object metadata

nikkixdev:

## What is this change?

Adds `created_by` to object metadata and populates that field via the HTTP API.

Work is required for the enterprise API to behave the same. Working on that now.

Closes https://github.com/sensu/sensu-go/issues/3591

Yas.

Nah.

Filed https://github.com/sensu/sensu-go/issues/3601

Filed https://github.com/sensu/sensu-docs/issues/2243

Tested HTTP API endpoints for multiple resources, tested `sensuctl create`. QA crucible tests will be in larger part of `sensuctl prune` work.

## Is this change a patch?

Nope, feature work.


nikkixdev:
Have to resolve some tests!

backend/eventd/entity.go
```diff
@@ -45,10 +45,11 @@ func createProxyEntity(event *corev2.Event, s store.EntityStore) error {
...
 			entity.Subscriptions = append(entity.Subscriptions, corev2.GetEntitySubscription(entityName))
 		}
+	}
 
-		if err := s.UpdateEntity(ctx, entity); err != nil {
-			return err
-		}
+	entity.CreatedBy = event.CreatedBy
+	if err := s.UpdateEntity(ctx, entity); err != nil {
+		return err
```
echlebek: Why is this now outside the if statement? This will cause proxy entities to be written to every time an event fires, probably not what we want.

nikkixdev: Otherwise, existing proxy entities will not get updated with the correct `created_by` field.

nikkixdev: We could check to see if `created_by` has changed, and only update if that's the case. Probably more performant that way.

echlebek: I don't think that's quite right either. It makes sense for the case where proxy entities are being created for the first time, but if Alice creates a proxy entity, and then Bob creates an event referencing one, Alice is still the creator of the proxy entity.

nikkixdev: Bob's event could still modify the proxy entity that Alice originally created though?

echlebek: I think the only time Bob's event should result in a proxy entity that Bob owns is when there wasn't one before, and it triggers the automated proxy entity creation. Normally, the entity wouldn't be touched by this process.

nikkixdev: So if Bob were to update metadata labels of an existing proxy entity via the [events API](https://docs.sensu.io/sensu-go/latest/api/events/#events-post), then the desired behavior would be for those changes _not_ to be persisted?

echlebek: Well, desire is in the eye of the user, but Sensu's behaviour up until now has been not to modify entities when processing events, except for keepalives.

nikkixdev: Seems a little strange that's the case, but will revert those changes as to not change the behavior.

echlebek: I do think created_by should be set if the proxy entity is created however.

PR #3607, opened 2020-03-06
Update store to marshal wrapped resources on create

jamesdphillips:
## What is this change?

Updates the behaviour of the store's `Create` method to marshal Wrapped objects using the JSON. Matches the behaviour of the `CreateOrUpdate` method.

Required for sensu/sensu-enterprise-go#870

Why not.

Chesterton's fence. If there was a good reason the method did not have the same behaviour I couldn't find it.

Integration & manual.


jamesdphillips:
Quickly ran through the happy path on staging today and didn't find any regressions.

PR #3622, opened 2020-03-12
Add agent platform discovery

### LINKED ISSUE #3587: Add agent platform auto-discovery
Add the following 4 fields to Entity.System:
```
string libc_type ("glibc", "musl", "apple", "")
string vm_system (described above)
string vm_role ("guest", "host", "")
string cloud_provider ("EC2", "GCP", "Azure", "")
```

Use the design doc to understand how to populate these fields agent-side.

https://docs.google.com/document/d/1HgQWesubiFvNTq-uSBIn5e0hj43PeIUMjdEbFoT-XBo/edit#

echlebek:
## What is this change?

This commit adds features that populate the LibCType, VMSystem,
VMRole, and CloudProvider fields in Entity.System.

LibCType, VMSystem, and VMRole are always populated on agent
startup.

CloudProvider is populated if the --detect-cloud-provider flag is
set. This feature can result in several HTTP requests or DNS
lookups being performed, so it may not be appropriate for all
environments.

There are no automated tests for cloud provider discovery yet.
Since the function requires an attempt to connect to a link-local
IP address, it may be difficult to achieve this.

Closes #3587 

Yes

Detecting the cloud provider required quite a bit of thorny work, involving DNS resolution, HTTP requests, timeouts, and synchronization.

I believe we have already documented the fields; this PR merely sets them.

Manual verification and testing. I'll be adding an automated test to Sensu QA as well.

Example of system output on a Fedora 31 machine:
```
  system:
    arch: amd64
    cloud_provider: ""
    hostname: localhost.localdomain
    libc_type: glibc
    network:
      interfaces:
      - addresses:
        - 127.0.0.1/8
        - ::1/128
        name: lo
      - addresses: null
        mac: <redacted>
        name: enp0s31f6
      - addresses:
        - 192.168.0.110/24
        - fe80::e1c7:8e3c:4a3d:2b3d/64
        mac: <redacted>
        name: wlp82s0
      - addresses:
        - 192.168.122.1/24
        mac: <redacted>
        name: virbr0
      - addresses: null
        mac: <redacted>
        name: virbr0-nic
    os: linux
    platform: fedora
    platform_family: fedora
    platform_version: "31"
    vm_role: host
    vm_system: kvm
```

## Is this change a patch?

No

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -70,6 +70,12 @@ type Agent struct {
 // NewAgent creates a new Agent. It returns non-nil error if there is any error
 // when creating the Agent.
 func NewAgent(config *Config) (*Agent, error) {
+	return NewAgentContext(context.Background(), config)
+}
+
+// NewAgentContext is like NewAgent, but allows threading a context through
+// the system.
+func NewAgentContext(ctx context.Context, config *Config) (*Agent, error) {
 	agent := &Agent{
 		backendSelector: &RandomBackendSelector{Backends: config.BackendURLs},
 		connected:       false,
@@ -89,7 +95,12 @@ func NewAgent(config *Config) (*Agent, error) {
 
 	// We don't check for errors here and let the agent get created regardless
 	// of system info status.
-	_ = agent.refreshSystemInfo()
+	systemInfoCtx, cancel := context.WithTimeout(ctx, time.Duration(DefaultSystemInfoRefreshInterval)*time.Second)
+	defer cancel()
+	_ = agent.refreshSystemInfo(systemInfoCtx)
+	if err := systemInfoCtx.Err(); err != nil {
+		logger.WithError(err).Error("couldn't refresh all system information within deadline")
+	}
 	var err error
 	agent.apiQueue, err = newQueue(config.CacheDir)
 	if err != nil {
@@ -114,10 +125,14 @@ func (a *Agent) sendMessage(msg *transport.Message) {
 	a.sendq <- msg
 }
 
-func (a *Agent) refreshSystemInfo() error {
+func (a *Agent) refreshSystemInfo(ctx context.Context) error {
 	info, err := system.Info()
 	if err != nil {
 		return err
+	}
+
+	if a.config.DetectCloudProvider {
+		info.CloudProvider = system.GetCloudProvider(ctx)
 	}
 
 	a.systemInfoMu.Lock()
@@ -135,7 +150,9 @@ func (a *Agent) refreshSystemInfoPeriodically(ctx context.Context) {
 	for {
 		select {
 		case <-ticker.C:
-			if err := a.refreshSystemInfo(); err != nil {
+			ctx, cancel := context.WithTimeout(ctx, time.Duration(DefaultSystemInfoRefreshInterval)*time.Second/2)
+			defer cancel()
+			if err := a.refreshSystemInfo(ctx); err != nil {
 				logger.WithError(err).Error("failed to refresh system info")
 			}
 		case <-ctx.Done():
--- a/agent/cmd/start.go
+++ b/agent/cmd/start.go
@@ -35,6 +35,7 @@ const (
 	flagConfigFile               = "config-file"
 	flagDeregister               = "deregister"
 	flagDeregistrationHandler    = "deregistration-handler"
+	flagDetectCloudProvider      = "detect-cloud-provider"
 	flagEventsRateLimit          = "events-rate-limit"
 	flagEventsBurstLimit         = "events-burst-limit"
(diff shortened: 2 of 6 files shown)


system/system.go
```diff
@@ -47,13 +61,187 @@ func Info() (types.System, error) {
...
+		return found
+	case found := <-c2:
+		return found
+	case found := <-c3:
+		return found
+	case <-ctx.Done():
+		return ""
+	}
+
+	return ""
```
golangcibot: unreachable: unreachable code (from `govet`)

agent/agent.go
```diff
@@ -89,7 +95,12 @@ func NewAgent(config *Config) (*Agent, error) {
 
 	// We don't check for errors here and let the agent get created regardless
 	// of system info status.
-	_ = agent.refreshSystemInfo()
+	systemInfoCtx, cancel := context.WithTimeout(ctx, time.Duration(DefaultSystemInfoRefreshInterval)*time.Second)
+	defer cancel()
+	_ = agent.refreshSystemInfo(systemInfoCtx)
+	if err := systemInfoCtx.Err(); err != nil {
```
nikkixdev: `systemInfoCtx.Err()` returns the same error as `agent.refreshSystemInfo(systemInfoCtx)`?

echlebek: They can be different. Here, I'm checking systemInfoCtx.Err() to figure out if the context has been canceled.

agent/cmd/start.go
```diff
@@ -220,6 +222,7 @@ func newStartCommand(ctx context.Context, args []string, logger *logrus.Entry) *
 	viper.SetDefault(flagCacheDir, path.SystemCacheDir("sensu-agent"))
 	viper.SetDefault(flagDeregister, false)
 	viper.SetDefault(flagDeregistrationHandler, "")
+	viper.SetDefault(flagDetectCloudProvider, false)
```
nikkixdev: Have you considered leaving this enabled by default? Or is it too expensive to perform the detection functions?

echlebek: Sean and I kicked it back and forth for a bit. Here's the problem:

* There is no way to conclusively do cloud platform detection without sending at least one HTTP request to a metadata service.
* These requests typically go to a link-local IP address.
* Depending on the network, these requests could never complete once started.
* A timeout works, but makes agent startup take longer on non-cloud systems.
* Some users may not appreciate their agents making these network requests.

For this reason, we decided to put cloud provider detection behind a feature flag.

system/system.go
```diff
@@ -47,13 +61,185 @@ func Info() (types.System, error) {
...
+}
+
+// GetCloudProvider inspects local files, looks up hostnames, and makes HTTP
+// requests in order to determine if the local system is running within a cloud
+// provider.
+func GetCloudProvider(ctx context.Context) string {
+	switch runtime.GOOS {
+	case "linux":
+		// Issue a dmidecode command to see if we are on EC2 or GCP
+		logger.Debug("sudo -n dmidecode -s bios-version")
```
nikkixdev: I'm assuming sudo is probably required to run this, but this could pose an issue for agents that don't have that access. Maybe we should just make a point in the docs about this.. wdyt @echlebek cc @hillaryfraley

echlebek: This is only the first attempt. If sudo is not available, then the execution will fail, and a fallback mechanism will be used.

system/system.go
```diff
@@ -47,13 +61,185 @@ func Info() (types.System, error) {
...
+			logger.WithError(err).Debug("couldn't run dmidecode")
+			return cloudMetadataFallback(ctx)
+		}
+		text = strings.ToLower(string(output))
+		if strings.Contains(text, "microsoft") {
+			logger.Debug("Running on Azure")
+			return "Azure"
+		}
+	case "windows":
+		return cloudMetadataFallback(ctx)
```
nikkixdev: We can't detect cloud providers on windows systems? Is this a todo or are we just not gonna support it?

echlebek: cloudMetadataFallback does detect cloud providers, it just uses a more expensive method than running dmidecode or inspecting a file.

agent/cmd/start.go
nikkixdev: Ah that makes sense, thanks for explaining!

system/system.go
nikkixdev: Yes, I see that now. Doesn't look like those fallback mechanisms require sudo so this should be fine!

system/system.go
nikkixdev: Ah so the respective `dial...` functions should be able to retrieve windows info. Sounds good!

PR #3631, opened 2020-03-16
Added a generic client POST function that can pass query params

nikkixdev:

## What is this change?

Adds a generic client POST function that can pass query params, `PostWithParams`. I decided to add a new function to be used for commercial `sensuctl prune` functionality rather than refactoring all calls that use the original `Post` function.

Required by https://github.com/sensu/sensu-enterprise-go/issues/879.

Yas.

Nah.

Nah.

Internal change, no docs change necessary.

Tested against branch for https://github.com/sensu/sensu-enterprise-go/issues/879.

## Is this change a patch?

Nope, required for feature work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ created via `sensuctl create`.
 - Added agent discovery of libc type, VM system/role, and cloud provider.
 - Added `float_type` field to system type to store which float type (softfloat,
   hardfloat) a system is using.
+- Added a generic client POST function that can pass query params.
 
 ### Changed
 - Updated the store so that it may _create_ wrapped resources.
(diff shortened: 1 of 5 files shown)


nikkixdev:
@ccressent We currently only have one use case that requires query params to be passed from the sensuctl client, which doesn't really seem like justification enough to refactor the existing generic functions, so that's why I chose to create a new one. I suppose I could manually construct the path with query params within the `sensuctl prune` command itself, it just seems slightly less elegant since there's already a `SetQueryParamsFromValues` function.

nikkixdev:
Ah, https://golang.org/pkg/net/url/#Values.Encode should be sufficient to replace this!

PR #3633, opened 2020-03-18
Update namespace flag help usage

nikkixdev:
Updates namespace flag help usage based on conversation about pruning resources and how it relates to `--namespace`.


palourde:
Do you think it's still relevant with our latest discussions?

nikkixdev:
@palourde based on our most recent discussion, `--namespace` will behave differently for `sensuctl prune` and that difference will be documented in its specific help usage. Closing this without merge.

PR #3637, opened 2020-03-20
Allow resty backend timeout to be configurable

### LINKED ISSUE #3635: [Proposal] Allowing Resty HTTP Timeout to be configurable
https://github.com/sensu/sensu-go/blob/master/cli/client/client.go#L43

The question/proposal is centered around the above code ^^.  Would a PR be acceptable to
* allow that timeout to be configurable at all?
* allow that timeout to be configurable from the cli?
* allow that timeout configurable item to be persisted in the configuration file

## Possible Implementation

Draft PR here https://github.com/naemono/sensu-go/pull/1

## Context

In heavily used sensu-go installations, sometimes 15 seconds isn't quite enough time to get a response.

naemono:
## What is this change?

Allow resty backend timeout to be configurable

Closes #3635 

The timeout isn't configurable, as gets triggered on highly used installations.

Yes, the changelog was updated in the "Unreleased" section

Documentation likely needs to be updated about this change.  Could someone point me to the right location?

Build sensuctl binary locally, and verified the setting was adjusted, and persisted in configuration file

## Is this change a patch?

No

### DIFF
--- a/cli/client/config/mock.go
+++ b/cli/client/config/mock.go
@@ -1,6 +1,8 @@
 package config
 
 import (
+	"time"
+
 	corev2 "github.com/sensu/sensu-go/api/core/v2"
 
 	"github.com/stretchr/testify/mock"
@@ -38,6 +40,14 @@ func (m *MockConfig) Namespace() string {
 	return args.String(0)
 }
 
+// Timeout mocks the timeout config
+func (m *MockConfig) Timeout() time.Duration {
+	args := m.Called()
+	timeoutString := args.String(0)
+	d, _ := time.ParseDuration(timeoutString)
+	return d
+}
+
 // TrustedCAFile mocks the trusted CA file config
 func (m *MockConfig) TrustedCAFile() string {
 	args := m.Called()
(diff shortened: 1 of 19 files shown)


cli/client/config/mock.go
```diff
@@ -38,6 +40,14 @@ func (m *MockConfig) Namespace() string {
 	return args.String(0)
 }
 
+// Timeout mocks the timeout config
+func (m *MockConfig) Timeout() time.Duration {
+	args := m.Called()
+	timeoutString := args.String(0)
+	d, _ := time.ParseDuration(timeoutString)
+	return d
```
echlebek: What you can do here is 
```
return args.Get(0).(time.Duration)
```

cli/client/testing/mock_config.go
```diff
@@ -43,6 +45,14 @@ func (m *MockConfig) TrustedCAFile() string {
 	return args.String(0)
 }
 
+// Timeout mocks the timeout config
+func (m *MockConfig) Timeout() time.Duration {
+	args := m.Called()
+	timeoutString := args.String(0)
+	d, _ := time.ParseDuration(timeoutString)
+	return d
```
echlebek: Similarly here, you can use 

```
args.Get(0).(time.Duration)
```

cli/client/config/mock.go
naemono: 👍

cli/client/testing/mock_config.go
naemono: 👍

naemono:
@echlebek these issues should be resolved now with latest commit.

Please be aware of https://github.com/etcd-io/bbolt/issues/187 as I was running go 1.14, and ran directly into this when running unit tests.  Mass failure.

echlebek:
To be merged following the 5.19.0 release.

Issue #3634, opened 2020-03-19
Wrong status code returned for PermissionDenied & Unauthenticated errors


palourde:
When a router attempts to return either `actions.PermissionDenied` or `actions.Unauthenticated`:

https://github.com/sensu/sensu-go/blob/22518b342f4659c25f4c13379471e2a1d695cb43/backend/apid/actions/error.go#L36-L40

The backend will return an internal error because these two errors type are not handled here:

https://github.com/sensu/sensu-go/blob/22518b342f4659c25f4c13379471e2a1d695cb43/backend/apid/routers/routers.go#L86-L102

PR #3638, opened 2020-03-20
Return proper HTTP status code

palourde:
## What is this change?

It fixes a bug where controllers that return either `actions.PermissionDenied` or `actions.Unauthenticated` would actually respond with an internal server error (500).

Closes https://github.com/sensu/sensu-go/issues/3634

Added

Nope

Nope

Nope

Manually verified

## Is this change a patch?

Yep

### DIFF
--- a/backend/apid/routers/routers.go
+++ b/backend/apid/routers/routers.go
@@ -96,7 +96,7 @@ func HTTPStatusFromCode(code actions.ErrCode) int {
 	case actions.PaymentRequired:
 		return http.StatusPaymentRequired
 	case actions.PermissionDenied:
-		return http.StatusForbidden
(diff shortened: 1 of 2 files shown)


backend/apid/routers/routers.go
```diff
@@ -95,6 +95,10 @@ func HTTPStatusFromCode(code actions.ErrCode) int {
 		return http.StatusConflict
 	case actions.PaymentRequired:
 		return http.StatusPaymentRequired
+	case actions.PermissionDenied:
+		return http.StatusForbidden
```
echlebek: Should we return http.StatusNotFound instead here?

palourde: Excellent point!

PR #3642, opened 2020-03-23
Disable flapping history for 5.x releases

echlebek:
This commit disables (but not removes) flapping history support.
The effect is that check history still contains the flapping field,
but it is never populated with true values, and it is never
rendered as JSON.

Flapping history support will be introduced in sensu-go 6.0.


echlebek:
I'll need to comment out a test as well here.

api/core/v2/check.proto
```diff
@@ -32,7 +32,8 @@ message CheckRequest {
     int64 Issued = 4 [(gogoproto.jsontag) = "issued"];
 
     // HookAssets is a map of assets required to execute hooks.
-    map<string, AssetList> hook_assets = 5 [(gogoproto.jsontag) = "hook_assets"];
```
nikkixdev: 🤔did vim reformat hook assets and assets?

echlebek: Yeah, it's just the auto-formatter.

PR #3643, opened 2020-03-24
Added a generic client POST function that can return the response body

nikkixdev:

## What is this change?

Adds a generic client POST function that returns the response body.

Required by https://github.com/sensu/sensu-enterprise-go/pull/919

Yas.

Nah.

Deciding if we wanted to refactor `Post` to return a response body or create a new `PostBody` function.

We good.

Testing against https://github.com/sensu/sensu-enterprise-go/tree/sensuctl-prune

## Is this change a patch?

Nope.

### DIFF
--- a/cli/client/generic.go
+++ b/cli/client/generic.go
@@ -133,6 +133,21 @@ func (client *RestClient) Post(path string, obj interface{}) error {
 	return nil
 }
 
+// PostBody sends a POST request with obj as the payload to the given path
+// and additionally returns the response body
+func (client *RestClient) PostBody(path string, obj interface{}) ([]byte, error) {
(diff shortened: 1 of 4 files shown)


cli/client/generic.go
```diff
@@ -133,6 +133,21 @@ func (client *RestClient) Post(path string, obj interface{}) error {
 	return nil
 }
 
+// PostBody sends a POST request with obj as the payload to the given path
+// and additionally returns the response body
+func (client *RestClient) PostBody(path string, obj interface{}) ([]byte, error) {
```
echlebek: Why not return the entire response, that way headers can be inspected as well :)

PR #3650, opened 2020-03-26
Keepalives now work correctly when using postgres

echlebek:
## What is this change?

This commit fixes an issue where keepalives would not always fire
correctly when using the postgres store. It seems that this
regression was introduced several releases ago, but it was not
always apparent for all users.

The keepalive mechanism was always consulting etcd when retrieving
events, and as such, when postgres was enabled, keepalives could be
incorrectly passing.

Unfortunately, due to the current architecture of sensu-go and
sensu-enterprise-go, it is not possible to write a unit test that
proves this commit fixes the issue. A test will need to be added
to the sensu-enterprise-go repo instead.

An unrelated unit test has been added that shores up testing of
check history generation; this has been done to maintain parity
between sensu-go and sensu-enterprise-go for tests.

Yes

The change necessitated updates to the keepalived tests. I also manually verified the change with sensu-enterprise-go.

## Is this change a patch?

Yes

### DIFF
--- a/backend/keepalived/keepalived.go
+++ b/backend/keepalived/keepalived.go
@@ -96,9 +96,9 @@ func New(c Config, opts ...Option) (*Keepalived, error) {
 	ctx, cancel := context.WithCancel(context.Background())
 
 	k := &Keepalived{
-		store:                 c.Store,
-		eventStore:            c.EventStore,
-		bus:                   c.Bus,
+		store:      c.Store,
+		eventStore: c.EventStore,
+		bus:        c.Bus,
 		deregistrationHandler: c.DeregistrationHandler,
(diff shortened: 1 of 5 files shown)


backend/keepalived/keepalived.go
```diff
@@ -480,6 +480,7 @@ func (k *Keepalived) dead(key string, prev liveness.State, leader bool) bool {
 	if entity == nil {
 		// The entity has been deleted, there is no longer a need to
 		// track keepalives for it.
+		logger.Debug("nil entity")
```
jamesdphillips: Minor. Should this be using `lager` so it includes the additional context?

echlebek: Yes, thanks!

PR #3658, opened 2020-03-31
Add the is_silenced field selector for events

palourde:

## What is this change?

It adds both the `event.is_silenced` & `event.check.is_silenced` field selectors; maybe we should only add one, but I feel like `event.is_silenced` is better in terms of UX.

Closes https://github.com/sensu/sensu-go/issues/3552

Added

Nope

Nope

I will open a new docs issue once we know which fields we are adding.

Manually tested in sensu-enterprise-go.

## Is this change a patch?

Nope

### DIFF
--- a/api/core/v2/event.go
+++ b/api/core/v2/event.go
@@ -426,6 +426,8 @@ func EventFields(r Resource) map[string]string {
 	return map[string]string{
 		"event.name":                 resource.ObjectMeta.Name,
 		"event.namespace":            resource.ObjectMeta.Namespace,
+		"event.is_silenced":          isSilenced(resource),
+		"event.check.is_silenced":    isSilenced(resource),
 		"event.check.name":           resource.Check.Name,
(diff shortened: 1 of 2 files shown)


api/core/v2/event.go
```diff
@@ -440,6 +442,13 @@ func EventFields(r Resource) map[string]string {
 	}
 }
 
+func isSilenced(e *Event) string {
+	if len(e.Check.Silenced) > 0 {
+		return "true"
+	}
+	return "false"
+}
+
```
echlebek: Shouldn't this function also consider whether or not its entity is silenced?

palourde: Long story short, we already do but we store the result in `Check` (I have no idea why though): https://github.com/sensu/sensu-go/blob/69c0ef22f3d7d790a3a341d5b7d43489113c027c/backend/eventd/silenced.go#L23-L39

PR #3660, opened 2020-04-01
Lift backend ID out of tessen, into backend proper

echlebek:
## What is this change?

This commit causes the backend to create an ID on startup, and
share it with tessen, instead of tessen creating its own ID. The
ID will be shared with other facilities, including facilities in
the enterprise product.

We need to share backend IDs between different components of Sensu, not just tessen.

No, this is an internal change.

In a sense, this PR is a complication of another change.

I'd consider unit/integration tests to be sufficient.

## Is this change a patch?

No.

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/coreos/etcd/clientv3"
 	"github.com/coreos/etcd/pkg/transport"
+	"github.com/google/uuid"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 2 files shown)


backend/backend.go
```diff
@@ -72,6 +76,16 @@ type Backend struct {
 	runCtx    context.Context
 	runCancel context.CancelFunc
 	cfg       *Config
+	backendID string
```
golangcibot, line 79: field `backendID` is unused (from `unused`)

echlebek:
On second thought, it might be better to leave this as it is.

Issue #2767, opened 2019-03-04
Remove edition header


echlebek:
The edition header was never put to use and will be removed.

PR #3666, opened 2020-04-03
Add Edition to version.go

amdprophet:

## What is this change?

Adds a new `Edition` field to `version.go`.

We want a way to distinguish between opensource builds and commercial builds.

Needed for #3669.

Yes.

No.

No.

I don't believe any new documentation is required.

Added tests, ensured they passed.

## Is this change a patch?

It can be - I've targeted `release/5.19` as a result.


amdprophet:
Some additional discussion & work will need to happen before we can consider merging this. Going to mark it as WIP.

palourde:
@amdprophet Could you schedule a quick discussion about this work?

amdprophet:
Scheduled. :)

amdprophet:
After discussing this issue, we're going to move forward with this implementation. The edition information was previously removed as it was unused (https://github.com/sensu/sensu-go/issues/2767).

We will also need to add this information to Tessen.

version/version.go
```diff
@@ -37,20 +47,44 @@ func Semver() string {
 	return version
 }
 
-// Println prints all available details about the current version in a
-// human-readable format, with an optional component name as the prefix
-func Println(component string) {
+func SemverWithEditionSuffix() string {
```
palourde, line 50: What's the purpose of the edition suffix? Since we already output `community edition` or `enterprise edition`, I'm not totally sure of its purpose!

version/version.go
```diff
@@ -37,20 +47,44 @@ func Semver() string {
...
+		editionSuffix = InvalidEditionSuffix
+	}
+	return fmt.Sprintf("%s+%s", Semver(), editionSuffix)
+}
+
+func EditionOutput() string {
+	if Edition == "community" || Edition == "enterprise" {
+		return fmt.Sprintf("%s edition", Edition)
+	}
+	return fmt.Sprintf("built with an invalid \"edition\" ldflag")
```
palourde, line 67: Why do we need this `invalid` edition and this logic?

version/version.go
amdprophet: This is part of the requirements from the trademark guidelines document. It must be included in the version - and to be clear what it stands for we also print out the full name.

version/version.go
amdprophet: The value should only ever be set to community or enterprise; any other value will be a violation of the trademark guidelines. We will eventually be including the url to the trademark guidelines in the output if an invalid edition is specified.

nikkixdev:
@palourde https://github.com/sensu/sensu-enterprise-go/pull/998

Issue #3632, opened 2020-03-18
List all resource types in sensuctl command


nikkixdev:
##  Feature Suggestion
Sensuctl should have a separate command to list all resource types. The full api name of a resource is required to operate `sensuctl dump/prune`. While `sensuctl dump --types` currently lists all resource types, we should have a separate command that is not associated with the specific dump or prune features.

## Possible Implementation
[kubectl api-resources](https://kubernetes.io/docs/reference/kubectl/overview/#resource-types) does something very similar, we could reuse the logic from `sensuctl dump --types` to implement a command along the lines of `sensuctl api-resources` or `sensuctl resource list`. 

## Context
`sensuctl prune --types core/v2.CheckConfig` requires `sensuctl dump --types` to get the full resource name of the check, which is rather unintuitive. See [fully qualified names](https://docs.sensu.io/sensu-go/latest/sensuctl/reference/#sensuctl-dump-resource-types).

Related to https://github.com/sensu/sensu-enterprise-go/issues/879

nikkixdev:
We decided to move forward with the follow subcommands:
- `sensuctl describe-type`
  - use the api name `sensuctl describe-type core/v2.CheckConfig`
  - use the rbac name `sensuctl describe-type checks`
  - list all types `sensuctl describe-type all`

nikkixdev:
Let's reuse as much logic as possible right now, if during implementation we find that we want to improve/refactor _more_ we should file enhancement/optimization issues then.

PR #3680, opened 2020-04-09
Add sensuctl describe-type command

palourde:
## What is this change?

It adds the `sensuctl describe-type` command, which takes 1 argument, either `all` or a comma-separated list of short names and/or fully qualified names (e.g. `namespace,checks,authentication/v2.Provider`).

The output can be formatted in either the tabular, JSON or YAML format.

I also deprecated the `sensuctl dump --types` flag so it refers to this command, but it still work for now.

Closes https://github.com/sensu/sensu-go/issues/3632.

Added

Nope

Nope

I'll open a new sensu-docs issue so we document this new command and hopefully refer to this command when we list resources in our documentation (https://docs.sensu.io/sensu-go/latest/reference/rbac/#resources, https://docs.sensu.io/sensu-go/latest/sensuctl/reference/#sensuctl-edit-resource-types etc.)

Added rough unit tests and manually tested it on sensu-go  & sensu-enterprise-go.

## Is this change a patch?

Nope

### DIFF
--- a/cli/commands/describetype/describe_type.go
+++ b/cli/commands/describetype/describe_type.go
@@ -1,6 +1,7 @@
 package describetype
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"strconv"
@@ -54,7 +55,17 @@ func execute(cli *cli.SensuCli) func(*cobra.Command, []string) error {
 	return func(cmd *cobra.Command, args []string) error {
 		var resources []apiResource
 
-		for _, resource := range resource.All {
+		if len(args) != 1 {
+			_ = cmd.Help()
+			return errors.New("invalid argument(s) received, expected a single one or a comma-separated list")
+		}
+
+		requested, err := resource.GetResourceRequests(args[0], resource.All)
+		if err != nil {
+			return err
+		}
+
+		for _, resource := range requested {
 			wrapped := types.WrapResource(resource)
 
 			// Short names are only supported for core/v2 resources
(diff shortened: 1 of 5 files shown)


cli/commands/describetype/describe_type.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+	return format
+}
+
+// isNamespaced is a hack to determine whether a resource is global or
+// namespaced, by relying on the SetNamespace method, which is a no-op for
+// global resources, and inspecting the resulting namespace
+func isNamespaced(r corev2.Resource) bool {
+	r.SetNamespace("~sensu")
+	if r.GetObjectMeta().Namespace != "~sensu" {
```
golangcibot, line 112: S1008: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (from `gosimple`)

cli/commands/describetype/describe_type.go
```diff
@@ -0,0 +1,171 @@
...
+$ sensuctl describe-type checks,handlers
+
+You can also use the 'all' qualifier to describe all available types:
+$ sensuctl describe-type all
+`
+
+type apiResource struct {
+	Name       string `json:"name"`
+	ShortName  string `json:"short_name"`
+	APIVersion string `json:"api_version"`
```
echlebek, line 34: This ends up being rendered as "apiversion" instead of "api_version" when output as YAML. Since the `sensuctl create` parser wants `api_version`, I feel it should be consistent here.

palourde:
@echlebek regarding the short names for enterprise resources, this is actually expected; `sensuctl dump` and `sensuctl edit` do not support these short names at the moment (at least it's my understanding). Technically, we could definitely do it I believe, I'm just afraid about the possible name collisions in the future across all API versions.

PR #3683, opened 2020-04-09
Add Processes to Entity.System, export interface for a process getter in the agent

nikkixdev:

## What is this change?

Adds a process struct and adds processes field to the system struct.

Required by https://github.com/sensu/sensu-enterprise-go/issues/956

Yas.

Nah.

There was some general team misunderstanding about the need to store this information. The auto-discovery proposal and discussion cleared this up.

https://github.com/sensu/sensu-docs/issues/2359

https://github.com/sensu/sensu-staging/pull/202
https://github.com/sensu/sensu-go-qa-crucible/issues/113
https://github.com/sensu/sensu-go-qa-crucible/pull/114

## Is this change a patch?

Nope, feature work goes into master.


nikkixdev:
Bumping this with a comment. @ccressent have you had the chance to look at this for review yet? I can pass it along to someone else if needed!

process/process.go
```diff
@@ -0,0 +1,20 @@
...
+	corev2 "github.com/sensu/sensu-go/api/core/v2"
+)
+
+// A Getter is responsible for getting the process info of an agent.
+type Getter interface {
+	Get(context.Context) ([]*corev2.Process, error)
+}
+
+// A ProcGetter is responsible for refreshing process info of an agent.
+type ProcGetter struct{}
```
ccressent, line 15: Maybe this should be called `DummyProcessGetter`, `NoopProcessGetter` or `PlaceholderProcessGetter`?

process/process.go
```diff
@@ -0,0 +1,20 @@
...
+
+// Get is not yet implemented.
+func (p *ProcGetter) Get(ctx context.Context) ([]*corev2.Process, error) {
+	return nil, nil
```
ccressent, line 19: I would either `return nil, ERR_NOT_IMPLEMENTED` or `return []*corev2.Process{}, nil`.

As it stands the caller is likely to check the error, see it's `nil`, and assume it can dereference the first return value, leading to a segmentation fault.

process/process.go
nikkixdev: +1 for noop

process/process.go
nikkixdev: If `Get` returns `nil, nil` there is no dereferencing or seg fault, as info.Processes would remain nil. This can be verified by running the pure OSS agent, or a commercial agent with the flag disabled.
```
proccessInfo, err := a.ProcessGetter.Get(ctx)
if err != nil {
	return err
}
info.Processes = proccessInfo
```
Does it still make sense to return `[]*corev2.Process{}, nil` in this case?

ccressent: Yes, it works _in this case_ because whatever uses `info.Processes` (the serialization code here) is ignoring nil values. But you're essentially forcing callers to null check the first return value.

What I'm suggesting is that maybe it's better to have the following contract: if error is not nil, you can blindly use the first return value. I'm also suggesting that it's what the vast majority of Go code uses; one typically writes, given `f() (a, error)`:

```go
a, err := f()
if err != nil {
    ...
}

// use a, trusting that it can't be nil else err wouldn't have been nil
```
instead of
```go
a, err := f()
if err != nil {
    ...
}

if a != nil {
    // use a
}
```
or
```go
a, err := f()
if err != nil {
    ...
}

if a == nil {
    ...
}

// use a
```

nikkixdev: I understand the argument and agree that it's more inline with Go styling. I wish they had a more formal approach to this documented, ex. https://github.com/golang/go/wiki/CodeReviewComments.

PR #3685, opened 2020-04-13
Support SIGUSR1 log level incrementing

### LINKED ISSUE #1742: Support changing of log level during runtime
Add the ability to change the log level for the agent & backend during runtime. This can be very useful for debugging the daemons when they're in a bad state or an unexpected condition is occurring.

## Expected Behavior
As a user I should be able to either toggle debug logging or set a specific log level for an already running daemon.

## Current Behavior
Changing the log level requires a restart of the daemon.

## Possible Solution
Sensu Enterprise (1.x) supports toggling debug mode by sending `kill -TRAP` to the pid of the daemon. In my personal opinion, it would be nice to have a more versatile approach that allows us to set a specific log level rather than just toggling debug logging.

N/A

## Context
During preparation for a release, we noticed that keepalives were not functioning as expected. The daemons were not running with the debug log level and we were worried that restarting the daemon would resolve the issue (temporarily).

## Your Environment
N/A

echlebek:
## What is this change?

This commit adds support to sensu-backend and sensu-agent for
incrementing the log level with SIGUSR1. When the process receives
this signal, the log level will be incremented in terms of
verbosity. When the level is at TRACE, the most verbose log level,
and the log level is incremented, it will wrap around to error
level.

Closes #1742

Yes

It's unclear at this time how to support this for Windows.

@hillaryfraley, we will need to document that users can now increase the log level at runtime, with the following commands for backends and agents:

```
kill -s SIGUSR1 $(pidof sensu-backend)
```
```
kill -s SIGUSR1 $(pidof sensu-agent)
```

I ran the above commands for the backend and agent.

## Is this change a patch?

No.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 - [Web] Added the ability for labels and annotations with links to images to be
 displayed inline.
 - [Web] Added additional modes for those with colour blindness.
+- Users can now increment the logging level by sending SIGUSR1 to the
+sensu-backend or sensu-agent process.
 
 ### Changed
 - Warning messages from Resty library are now suppressed in sensuctl.
--- a/agent/cmd/logger.go
+++ b/agent/cmd/logger.go
@@ -1,11 +1,6 @@
 package cmd
 
 import (
-	"os"
-	"os/signal"
-	"syscall"
-
-	"github.com/sensu/sensu-go/util/logging"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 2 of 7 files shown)


palourde:
@echlebek looks like it broke the Windows builds 😱

echlebek:
@palourde I moved the signal handling stuff into some files that won't get built on Windows, which seems to have un-broken the build.

palourde:
I figured this is expected, but I'm wondering if there's a way to by-pass the following behavior:

When the log level is incremented from trace --> error, the following entry is logged, which is expected:
```
{"level":"warning","msg":"set log level to error","time":"2020-04-14T18:38:58-04:00"}
```

However, when we go from error --> warning, the same entry never gets logged. I'm wondering if we could add a condition so when the current log level is `error`, the entry gets logged after the log level is changed. e.g. 

```go
logrus.Warnf("set log level to %s", newLevel)
logrus.SetLevel(newLevel)
// Log back the entry if the level passed from error to warning
if logrus.GetLevel() == logrus.WarningLevel {
	logrus.Warnf("set log level to %s", newLevel)
}
```

WDYT?

echlebek:
It's a bit of an oddity, but I like that it makes the message appear in all cases. I'll add it!

Issue #3663, opened 2020-04-02
The init command does not return SSL errors


palourde:
## Expected Behavior

If the SSL certificate for etcd's client traffic is invalid, an explicit error should be returned.

## Current Behavior
The init command only returns the generic error `error connecting to cluster: context deadline exceeded`

Some context: https://sensu.slack.com/archives/C60EEQFH8/p1585835720437200

TL;DR version: The client traffic was configured as following:

```
etcd-advertise-client-urls: https://192.168.156.200:2379
etcd-listen-client-urls: https://0.0.0.0:2379
etcd-cert-file: /etc/pki/tls/certs/fullchain.pem
etcd-key-file: /etc/pki/tls/private/privkey.pem
```

However, this certificate's subject was something like `*.domain.tld`. Because `sensu-backend init` uses `etcd-advertise-client-urls`, the connection was probably rejected because the CN (`*.domain.tld`) didn't matched the URL (`192.168.156.200`), which is expected. However, it would be useful to return an explicit error so it's easier to debug for operators.

## Possible Solution
Figure out, if possible, where the actual error about TLS is returned and use that instead.

## Steps to Reproduce (for bugs)

Setup a new backend with TLS for etcd client traffic, but make the SSL subject mismatch.

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent):
* Installation method (packages, binaries, docker etc.):
* Operating System and version (e.g. Ubuntu 14.04):

palourde:
After a full day of investigation, I'm starting to better understand the problem.

This is essentially coming from https://github.com/grpc/grpc-go/issues/2031.

It appears a workaround [was merged 5 days ago in grpc-go](https://github.com/grpc/grpc-go/pull/3430) to allow blocking callers to surface the actual error, and not just `context deadline exceeded`. Once this patch is released, we would then need to wait for etcd to update to this new version, which might take a while, and may not necessarily be merged into the 3.3 branch of etcd.

A short-term fix could be to rely on the log entry produced by this GRPC unary client interceptor: https://github.com/etcd-io/etcd/blob/1c16c242db884999b495e07e86b5b6ca548a010c/clientv3/retry_interceptor.go#L62-L67, but it requires us to not use a blocking client, therefore removing this line: https://github.com/sensu/sensu-go/blob/72d4d3de4c4e9abaa51e0e2e9df45a0bea5e0675/backend/cmd/init.go#L123

Which would produce the following log entries:
```sh
[...]

PR #3690, opened 2020-04-14
Log TLS failures when initializing the backend

palourde:

## What is this change?

It removes the use of a blocking client when attempting to initialize the backend, in order to have etcd log any TLS error encountered.

See https://github.com/sensu/sensu-go/issues/3663#issuecomment-613656403 for more context.

It also adds a `--timeout` flag to the `sensu-backend init` command.

This is a workaround for https://github.com/sensu/sensu-go/issues/3663. We can probably implement a better solution with a blocking client once [this PR](https://github.com/grpc/grpc-go/pull/3430) is released and etcd is updated to use it.

Added

Nope

go-grpc is not easy 😅 

Nope

Manually tested.

## Is this change a patch?

I'd say so

### DIFF
--- a/backend/cmd/init.go
+++ b/backend/cmd/init.go
@@ -29,7 +29,7 @@ const (
 type seedConfig struct {
 	backend.Config
 	SeedConfig seeds.Config
-	Timeout    uint
+	Timeout    time.Duration
 }
 
 type initOpts struct {
@@ -118,11 +118,11 @@ func InitCommand() *cobra.Command {
 				clientURLs = viper.GetStringSlice(flagEtcdAdvertiseClientURLs)
 			}
 
-			timeout := viper.GetUint(flagTimeout)
+			timeout := viper.GetDuration(flagTimeout)
 
 			client, err := clientv3.New(clientv3.Config{
 				Endpoints:   clientURLs,
-				DialTimeout: time.Duration(timeout) * time.Second,
+				DialTimeout: timeout * time.Second,
 				TLS:         tlsConfig,
 			})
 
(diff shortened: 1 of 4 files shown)


backend/cmd/init.go
```diff
@@ -150,15 +152,37 @@ func InitCommand() *cobra.Command {
...
 
+		seed:
 			return seedCluster(client, seedConfig)
 		},
 	}
 
 	cmd.Flags().String(flagInitAdminUsername, "", "cluster admin username")
 	cmd.Flags().String(flagInitAdminPassword, "", "cluster admin password")
 	cmd.Flags().Bool(flagInteractive, false, "interactive mode")
+	cmd.Flags().Uint(flagTimeout, defaultTimeout, "timeout, in seconds, for failing to establish a connection to etcd")
```
echlebek, line 185: I'd suggest making this a string, and then using viper's facility for parsing durations to get a `time.Duration`.

Viper just calls out to `time.ParseDuration` from stdlib, so the acceptable values are well-known.

It should avoid potential bugs around conversions. There is already technically a bug with using uint here, as converting a large enough uint, which is a 64-bit type on 64 bit platforms, to an int64 (time.Duration), will cause an overflow. Of course realistically, it's unlikely that anyone would try to specify such a large timeout. :laughing:

backend/cmd/init.go
```diff
@@ -15,18 +15,21 @@ import (
...
 	flagInitAdminUsername = "cluster-admin-username"
 	flagInitAdminPassword = "cluster-admin-password"
 	flagInteractive       = "interactive"
+	flagTimeout           = "timeout"
 )
 
 type seedConfig struct {
 	backend.Config
 	SeedConfig seeds.Config
+	Timeout    uint
```
echlebek, line 32: IMO, this should be `time.Duration`, regardless of what the flag's type is. It should simplify setting up tests.

PR #3707, opened 2020-04-19
Fix another bug that caused sensu-backend restarts

### LINKED ISSUE #3571: After #3446 the backend is unstable in the face of agent restarts
## Expected Behavior

Agent behaviour should not affect backend stability.

## Current Behavior

Agents disconnecting and reconnecting can cause the backend to restart. Additionally, the restart is unreliable and etcd sometimes fails to come back up.

## Possible Solution

This doesn't happen every time an agent restarts, so it feels like a race condition of some sort.

1. Occurs with the default config.

2.
```
#!/bin/bash

logfile=backend-$(date +%s).log

./sensu-backend start &>$logfile &
backpid=$!

i=0

while [[ $(( i++ )) -lt 1000 ]] && ! grep -q 'backend stopped working' $logfile; do
    ./sensu-agent start &>/dev/null &
    apid=$!
    sleep 0.1
    kill $apid
done

kill $backpid
```

## Context

After upgrading to v5.17.1 we started experiencing strange half-alive states where many of the components were up and running but etcd was not. I managed to reliably reproduce it in our dev environment, then whittled down the scenario and created a reproducer.

Quickest occurrence (at the default log level): https://gist.githubusercontent.com/flowerysong/ee585a2a69c06baec7045288d0146240/raw/97db4b6107ca8fe9b58e6567c4d7e73e62722470/backend-1582144510.log
Verbose backend log running 572b4b5facc4a4f7b310dcf0eb34ef711e9583e5: https://gist.github.com/flowerysong/e62e60b12e090f1163918930b50e8924/raw/b9a028b35906968a589380ee24b9d70001173a8a/backend-1582144823.log

Verbose backend log running 20fb131a89e55cf395c925988cbad228aa71fa3f (no components appear to fail): https://gist.githubusercontent.com/flowerysong/e62e60b12e090f1163918930b50e8924/raw/b9a028b35906968a589380ee24b9d70001173a8a/backend-1582145452.log

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): First experienced in v5.17.1, bisected down to 572b4b5facc4a4f7b310dcf0eb34ef711e9583e5
* Installation method (packages, binaries, docker etc.): `go build`
* Operating System and version (e.g. Ubuntu 14.04): Amazon Linux 2

flowerysong:
## What is this change?

#3614 was a partial fix for this issue, but only addressed cancellation during authorization; we shouldn't propagate context cancellation errors during authentication either.

Fixes #3571

Yes.

It was difficult to track down the origin of the logged error since that code branch didn't log anything itself, so the only log entry was incredibly generic:

```
{"component":"backend","error":"internal error: internal error: context canceled","level":"error","msg":"backend stopped working and is restarting","time":"2020-03-27T00:55:03-04:00"}
```

No docs.

Manual testing in my nonprod environment with an Ansible playbook that would previously reliably induce a restart.

## Is this change a patch?

Yes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ instance of the event.
 
 ### Fixed
 - `sensu-backend init` now logs any TLS failures encountered.
+- Fixed another bug where sensu-backend would restart when agents disconnect.
 
 ## [5.19.1] - 2020-04-13
 
--- a/backend/agentd/agentd.go
+++ b/backend/agentd/agentd.go
@@ -284,7 +284,15 @@ func (a *Agentd) AuthenticationMiddleware(next http.Handler) http.Handler {
 		// Authenticate against the provider
 		user, err := a.store.AuthenticateUser(r.Context(), username, password)
 		if err != nil {
+			if r.Context().Err() != nil {
+				http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
+				return
+			}
 			if _, ok := err.(*store.ErrInternal); ok {
+				logger.
+					WithField("user", username).
+					WithError(err).
+					Error("unexpected error while authenticating the user")
 				select {
 				case a.errChan <- err:
 				case <-a.ctx.Done():


echlebek:
Thanks, I'll test your patch tomorrow!

echlebek:
Hi, I've just looked at this again with a fresh set of eyes, and while this patch works for errors from the authentication system, it is actually papering over a more widespread problem. I'll try to summarize what I think is going on here, in a typical reproduction of the problem:

* The context the store received was actually an HTTP request context, and the client connection timed out
* This resulted in the context getting canceled, and apparently the store got that error from the etcd client.
* The store says "hey, I have an error from the etcd client, those get marked as internal errors"
* The store returns the error, and its consumer detects that the error is an internal error.
* Internal errors are unrecoverable and they necessitate a restart, so that's bubbled up to the backend, which restarts everything.

So given the above is accurate, what should actually happen is that the store should never wrap a context error with `ErrInternal`. Right now it wraps any error returned by the etcd client, but the client could return `context.Canceled`, which we don't want to wrap.

I'm going to merge the PR, but I'm also going to file a follow-up issue to take care of the bug(s) in the store as well, with a note that we should revert the change once the underlying problem is dealt with. (As it would not be necessary if the store was operating correctly)

Thanks again for investigating this bug and filing this patch.

Issue #2941, opened 2019-05-10
Retry for asset download can be abusive


nixwiz:
I copy/pasted the wrong sha512sum for an asset (a handler asset, to be specific).  When the backend was attempting to download the asset it was failing with the appropriate error, however it was continually trying to download the asset such that it triggered github's abuse detection mechanism.  It was just under an hour once I realized my mistake, but one backend server had attempted this download over 1300 times in that timeframe.

## Expected Behavior

This mismatch should not generate this amount of traffic.

Also, it would be really, really nice if items like this created events (similar to what is requested in #2894) to surface themselves outside of the logs.
## Current Behavior

Too many retries in a short amount of time if an asset is misconfigured.
## Possible Solution

Exponential backoff for these failures?
## Steps to Reproduce (for bugs)

1.  Create a handler asset with an incorrect sha512sum
2.  Watch the log fill up with "failed to retrieve assets for handler" errors

## Context

Once fixed, It kept me from downloading the proper asset until the github abuse detection cleared.
## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent):  sensu-backend version 5.7.0#87a4066, build 87a4066ab68f5eabf666236c8dc2681563fa01c4, built 2019-05-08T19:11:28Z
* Installation method (packages, binaries, docker etc.):  packages
* Operating System and version (e.g. Ubuntu 14.04):  Ubuntu 18.04

echlebek:
We should definitely be using exponential backoff for asset downloads.

majormoses:
While I think that having exponential backoff is something that is needed I think it's needed for a very different reason. Exponential backoff makes sense for errors such as network, auth, etc but I think checksum mismatch is a bit different.

A checksum mismatch essentially falls into several scenarios:
- accidental (most common)
- attacker has been able to modify a check/asset definition
- attacker has been able to modify the asset itself but not the definition

The only scenario from the above I can think of is where fully automating retries would affect change without a direct action to sensu is if you had a compromised asset which prevents you from downloading and you have pushed a sanitized package to replace it. Personally I would recommend to "yank" the compromised package/asset rather than replace it because package managers generally won't be able to detect that there was a change since the version constraints matched what is already installed. I am not sure if sensu currently verifies the checksum of the installed contents and decide if it needs to unpack it again though my gut tells me that is unlikely.

[...]

PR #3716, opened 2020-04-23
Add configurable global rate limit for fetching assets so that asset retries are not abusive

nikkixdev:

## What is this change?

Adds a global rate limit for fetching assets so that asset retries are not abusive (can be configured using `--assets-rate-limit` and `--assets-burst-limit` on the agent and backend).

Closes https://github.com/sensu/sensu-go/issues/2941.

Yas.

Nah.

https://github.com/sensu/sensu-go/issues/2941#issuecomment-618688594

https://github.com/sensu/sensu-docs/issues/2414

Created checks and handlers running every 1s with misconfigured assets.

## Is this change a patch?

Yas.


asset/fetcher.go
```diff
@@ -65,6 +68,18 @@ func (h *httpFetcher) Fetch(ctx context.Context, url string, headers map[string]
 		h.URLGetter = httpGet
 	}
 
+	limiter, ok := limiterMap[url]
+	if !ok {
+		// 0.03 = once every ~30 seconds per unique url
```
echlebek, line 73: Not a bad idea, but I think this would allow for a higher rate than intended if there were many assets being pulled from the same domain. I would expect the domain to rate limit requests for the entire site, not just a single resource.

Perhaps we should have a global rate limit, or perhaps the rate limit should be per domain.

nikkixdev: Ah, you want to parse each url for the domain rather than differentiate by the unique url. I'll have to think on this overnight.. I'm wondering if there are weird edge cases (for example the different ways you can format an s3 bucket as a url) that we would have to take into account. Its a great suggestion, just need to experiment with it more first!

nikkixdev: After some Slack discussion, we decided to set a global rate limit rather than per url or per domain. The global rate limit mimics that of the Github API, ~5000 user to server requests per hour.

PR #3725, opened 2020-04-29
Close event store at end of backend run

echlebek:
## What is this change?

This change ensures that after the backend restarts, the event store is closed. This fixes a bug where postgres connections can linger after the backend restarts internally.

In concert with another bug, where backend restarts occur far too often, connections can leak and overwhelm the server.

Yes

I used the work done in https://github.com/sensu/sensu-go/pull/3723 to trigger internal restarts. It both helped me reproduce the issue and show that this patch fixes it.

## Is this change a patch?

This change is a  patch.

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -451,6 +451,9 @@ func Initialize(ctx context.Context, config *Config) (*Backend, error) {
 }
 
 func (b *Backend) runOnce(sighup <-chan os.Signal) error {
+	eCloser := b.EventStore.(closer)
+	defer eCloser.Close()
+
 	var derr error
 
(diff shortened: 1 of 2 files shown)


backend/backend.go
```diff
@@ -518,9 +518,16 @@ func (b *Backend) runOnce() error {
 
 	_ = b.Client.Close()
 
+	eCloser := b.EventStore.(closer)
+	_ = eCloser.Close()
```
nikkixdev, line 522: Would defer make sense here (in the event more code is added)?

echlebek: Hum, yes it would!

PR #3728, opened 2020-05-01
Don't restart sensu-backend on etcdserver: too many requests

echlebek:
## What is this change?

This PR causes sensu-backend to retry etcd requests if it receives an error saying "too many requests". It retries using exponential backoff.

sensu-backend is too likely to crash when experiencing high load.

Yes

No docs changes required.

Extensive performance testing.

## Is this change a patch?

Yes


palourde:
@echlebek Do we need to do the same for the enterprise codebase?

echlebek:
@palourde no changes required in enterprise as it makes etcd calls through the OSS store.

Issue #3662, opened 2020-04-01
Asset paths should be available in the context of commands


cwjohnston:
##  Feature Suggestion

Resources which invoke commands (e.g. checks, pipe handlers, mutators, hooks) should have access to the full path(s) for assets attached to the resource.

## Possible Implementation

Expose asset paths as environment variables in the command execution context. For example, given this check:

```
---
type: CheckConfig
api_version: core/v2
metadata:
namespace: default
name: win-cpu-check
spec:
command: powershell.exe -ExecutionPolicy ByPass -f %ASSET_PATH%\bin\check-windows-cpu-load.ps1 90
95
subscriptions:
- windows
handlers:
- slack
- email
runtime_assets:
- sensu-plugins-windows
interval: 10
publish: true
```

The resulting command execution would have access to the file path for the named runtime asset as `%ASSET_PATH%`.

For checks with multiple assets, perhaps we could iteratively create incrementing variables like `ASSET_PATH_1`, `ASSET_PATH_2` or similar?

## Context

Sensu reports unexpected check status when executing PowerShell plugins, unless the `-f` or `-File` flag is used. This flag requires the absolute path to the PowerShell script for execution, making it difficult to use this flag with PowerShell scripts distributed as assets.

This is not a new issue in Sensu Go, it was also a consideration in Sensu Core (see https://github.com/sensu/sensu/issues/1919). 

See https://github.com/sensu-plugins/sensu-plugins-windows/issues/101 for examples of how this is affecting Sensu Go users.

echlebek:
I could see this working via token substitution.

```
command: powershell.exe -ExecutionPolicy ByPass -f {{ .assetPath "sensu-plugin-windows" }}\bin\check-windows-cpu-load.ps1 90 95
```

calebhailey:
What would be the benefit of exposing this value as a token vs. as an environment variable? Seems the latter would be the more natural fit?

calebhailey:
Related: https://github.com/sensu/sensu-go/issues/3492

echlebek:
The benefit is that by exposing `assetPath` as a function, we can easily address named assets without having to export multiple environment variables. Additionally, the parameter passed to the `assetPath` could be identical to the name of the asset, rather than needing to know an implicit mapping to an environment variable based on the name.

palourde:
We need to schedule a discussion to expand on Eric's proposal. @nikkixdev will schedule something.

calebhailey:
[...]

PR #3740, opened 2020-05-07
Expose assets paths to consumers

palourde:
## What is this change?

This PR adds the path of an asset to the consumers of assets via the asset's environment variables. Additionally, it introduces a new function to token substitution so it's easier to consume this environment variable.

With this change, you can now retrieve the path of an asset (named `sensu-go-hello-world-0.0.2`) from a check, hook, mutator or handler via:

1. The environment variable itself:
```
echo -n $SENSU_GO_HELLO_WORLD_0_0_2
```

2. The token substitution function (when available)
```
echo -n ${{assetPath "sensu-go-hello-world-0.0.2"}}
```

Closes https://github.com/sensu/sensu-go/issues/3662

Added!

Nope

Implementing a native solution for token substitution proved to be really complicated, while relying on environment variables is way easier.

I'll open an issue.

Unit tests + automated QA

## Is this change a patch?

Nope


asset/asset_test.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+	if want := string(os.PathSeparator) + filepath.Join("tmp", "foo", "lib"); !strings.Contains(got[1], want) {
+		t.Errorf("RuntimeAsset.Env() LD_LIBRARY_PATH = %v, must contains %v", got[1], want)
+	}
+	if want := string(os.PathSeparator) + filepath.Join("tmp", "foo", "include"); !strings.Contains(got[2], want) {
+		t.Errorf("RuntimeAsset.Env() CPATH = %v, must contains %v", got[2], want)
+	}
+	if want := string(os.PathSeparator) + filepath.Join("tmp", "foo"); !strings.Contains(got[3], want) {
+		t.Errorf("RuntimeAsset.Env() asset path = %v, must contains %v", got[3], want)
+	}
```
ccressent, line 41: I wish the env was a `map[string]string` so that we didn't have to rely on indexing `got` like that...
Thanks for adding tests for `PATH`, `LD_LIBRARY_PATH`, ... !

util/environment/environment_test.go
```diff
@@ -99,3 +99,34 @@ func TestMergeEnvironments(t *testing.T) {
...
+func TestKey(t *testing.T) {
+	tests := []struct {
+		name string
+		s    string
+		want string
+	}{
+		{
+			name: "special characters are replaced",
+			s:    "FOO@BAR",
+			want: "FOO_BAR",
```
ccressent, line 112: I think that if we're strictly following POSIX, `@` would be allowed in a variable name since it's a portable character.
See https://stackoverflow.com/a/2821183/5086091, the second quote in particular.

I don't think it's an issue that needs to be fixed right now, but strict POSIX compatibility would mean we allow way more "special" characters.

palourde: Yeah, I was a bit confused by all the mixed information and decided to take the safe approach and only allow alphanumeric characters and `_`. That being said, we can always decide later to allow `@` if it causes any problem.

PR #3742, opened 2020-05-07
Windows agent: fix flag issues, add log rotation

### LINKED ISSUE #2937: Sensu-go Windows 5.7.0 - Log rotation / number of log files to keep
Is there a way to configure the logrotation / size based and number of log files to keep in the log directory, if the file grows bigger eventually a smaller filesystem can easily run out of space and bring down the actual application being monitored on the server.

echlebek:
## What is this change?

This commit makes signifcant fixes and improvements to the Windows
sensu-agent, including log rotation features and improvements to
the way the service installer handlers its arguments.

The service installer now supports all of the arguments that the
regular sensu-agent start command does, and will remember those
arguments when launching the application.

Log rotation is now supported, with support for configurable
log file sizes, and reaping old log files. On rotate, logs are
zipped.

Closes #2937 

Yes

Documentation updates will be required. cc @hillaryfraley

Extensive manual testing, including:

* Testing that `service run` works the same as `service install`.
* Testing that logs are output as expected, in both default and specified locations.
* Testing that logs rotate after the maximum log file size is reached.
* Testing that old logs are reaped according to configuration.
* Testing that arguments to `service install` are persisted and used by the service on restart.
* Testing around restarts and errors.

## Is this change a patch?

No

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -10,8 +10,8 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
-	"log"
 	"net/http"
+	"net/url"
 	"os"
 	"strings"
 	"sync"
@@ -211,6 +211,7 @@ func (a *Agent) buildTransportHeaderMap() http.Header {
 // 8. Start sending periodic keepalives.
 // 9. Start the API server, shutdown the agent if doing so fails.
 func (a *Agent) Run(ctx context.Context) error {
+	ctx, cancel := context.WithCancel(ctx)
 	defer func() {
 		if err := a.apiQueue.Close(); err != nil {
 			logger.WithError(err).Error("error closing API queue")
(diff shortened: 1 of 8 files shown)


echlebek:
@hillaryfraley, here is some information about the new flags, which are available to `sensu-agent service run`, and `sensu-agent service install`:

```      --log-level string                    logging level [panic, fatal, error, warn, info, debug] (default "debug")
      --log-max-size string                 maximum size of log file (default "128 MB")
      --log-retention-duration string       log file retention duration (s, m, h) (default "168h")
      --log-retention-files int             maximum number of archived files to retain (default 10)
```

agent/agent.go
```diff
@@ -219,16 +220,38 @@ func (a *Agent) Run(ctx context.Context) error {
 	a.header = a.buildTransportHeaderMap()
 
 	// Fail the agent after startup if the id is invalid
+	logger.Debug("validating agent name")
```
amdprophet, line 223: Do we still want to keep these debug lines in here?

echlebek: Yeah, I think they're actually useful

echlebek:
Not yet. Stay tuned, you can expect to see that later today.

As for crucible... do you know if we have any tests against Windows agents yet?

nikkixdev:
@echlebek very minimal: https://github.com/sensu/sensu-go-qa-crucible/blob/6b784e51d27df4e612550cd55827b2ed8d5609ee/spec/backend/windows_spec.rb

echlebek:
@nikkixdev that's more of a test of a Linux backend, regarding events coming from a Windows agent.

nikkixdev:
Right, we'd need to create a new spec that only runs on the windows agent. I'm not sure what the rspec sessions would look like for windows..

echlebek:
QA is going to lag behind for this, until we figure out how to do serverspec testing on Windows.

PR #3744, opened 2020-05-08
Add config file for stale bot

nikkixdev:

## What is this change?

Adds config file for a github bot https://probot.github.io/apps/stale/

https://trello.com/c/rTfspzBD/41-start-utilizing-github-bots-https-probotgithubio-apps

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+# Label to use when marking an issue as stale
+staleLabel: wontfix
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
```
echlebek, line 8: can you add "bug" to this?

.github/stale.yml
```diff
@@ -0,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
```
echlebek, line 2: I think this needs to be larger, perhaps 180 or so?

nikkixdev: We could increase it to 180. The bot comments at `daysUntilStale` and then waits `daysUntilClose` until actually closing if no response is received.

nikkixdev:
See also https://github.com/sensu/sensu-enterprise-go/pull/1023

Issue #2434, opened 2018-11-28
Cluster admins need the ability to reset user passwords


palourde:
Any user with the appropriate rights on the `users` resources should be able to reset the password of a user, without knowledge of its current password.

Right now, the `sensuctl user change-password` requires the user to enter the current password, which is not desired for security and practical reasons.

majormoses:
This makes sense I am wondering if we want to create a specific role for user management I don't see one in the [defaults](https://docs.sensu.io/sensu-go/latest/reference/rbac/#default-roles). `admin` and `cluster-admin` would implicitly have access to these. Some other functionality to consider around user management roles would be setting up SAML and such since this is often managed by a different team who do not need general admin access in sensu.

Issue #1768, opened 2018-07-01
Add `sensuctl hashpass` command


echlebek:
Add a command sensuctl that hashes passwords the same way the backend does, so that users can hash passwords instead of sending plaintext passwords over the wire.

Make sure that the method is exactly the same as what the backend does (currently bcrypt).

See #1437 for background.

Requires #1697.

The command can work entirely locally if we accept the premise that encryption algorithms and parameters will not be configurable. Otherwise, it will need to coordinate with the backend to discover the configuration.

Acceptance criteria:

- [ ] `sensuctl hashpass $PASSWD` will output a password hash when called.
- [ ] After a user is created with the hashed password generated by `sensuctl hashpass`, authentication for that user succeeds when the password is provided.

ghoneycutt:
CC @treydock @ghoneycutt

annaplotkin:
Related to #1697

Issue #1697, opened 2018-06-20
Support specifying password hash directly in user json


cyphus:
## Expected Behavior

In order to support creation of users through `sensuctl create -f` (see #1663), we need the ability to specify a user's password as a hash to avoid storing passwords in plaintext.

## Current Behavior
Sensuctl's interactive user creation passes the plaintext password in a TLS encrypted tunnel to the sensu backend where the password is hashed. We should keep this functionality for interactive user creation.

## Possible Solution
_Only_ allow the password to be specified via hash in user json. `sensuctl` should provide functionality to produce a hash from the given password so that users can manually create user json files.

## Context
Required by #1663, which deals with creating users via config management.

ghoneycutt:
We would need sensuctl to be able to create a hash so we do not have to implement bcrypt. The bcrypt library for ruby needs gcc and forcing a compiler to be on all agents will harm adoption. We also want it to be simple for other config management software and automation tools so they do not also have bundle in bcrypt.

Perhaps something like

`sensuctl bcrypt foo`

which would return the following to STDOUT.

`$2a$04$hcu1uRCWZVF4n6OZSTTP1OqE5a/idwP49eR91IBR6/D8jpXkGeYBG`


This would allow automation tools to do a simple string comparison to see if the hash they have is equivalent to the hash returned by sensuctl. It would also allow them to take passwords as input and transform them to the bcrypt hash and use that with sensuctl to manipulate user settings.

ghoneycutt:
CC @treydock

annaplotkin:
- Need to update API to allow you to submit a hash for a user
- Need to update API to determine whether a submission resulted in a modification or not [modifying a user resource]
- Provide a bcrypt util in sensuctl

PR #3753, opened 2020-05-13
Support password hashes

palourde:
## What is this change?

This PR attempts to do the following:

- Support a new `password_hash` field in the core/v2 User resource
- Add a new `sensuctl user hash-password` subcommand to generate hashes
- Add the ability for cluster admins to reset passwords with `sensuctl user reset-password`
	- Previously they could only use the `change-password` subcommand, which required the current password
	- It requires explicit permissions on the `users` resource (even for your own password), so only cluster-admins would have that permission by default.
- Refactored the change-password logic so the verification of the current password is done server-side; previously it would be done on the client-side only, which effectively gave everyone the ability to reset their own password via the API.

Closes https://github.com/sensu/sensu-go/issues/2434
Closes https://github.com/sensu/sensu-go/issues/1768
Closes https://github.com/sensu/sensu-go/issues/1697

Added!

Nope

Touching to authentication is always tricky!

New docs are required, I'll open a couple of issues and link them here.

I had to manually test the backward compatibility for users that still use the `password` field to store the password hash.

The rest has been automated and I'll link the PRs to sensu-staging & sensu-go-qa-crucible here!

## Is this change a patch?

Nope nope nope


backend/apid/actions/users.go
```diff
@@ -74,17 +78,32 @@ func (a UserController) CreateOrReplace(ctx context.Context, user *corev2.User)
 		return NewError(InvalidArgument, err)
 	}
 
-	// Validate password
-	if err := user.ValidatePassword(); err != nil {
-		return NewError(InvalidArgument, err)
-	}
+	// Determine if a hashed and/or cleartext password was provided
+	if user.Password != "" && user.PasswordHash != "" {
```
echlebek, line 82: I think there should be a check to see if these things are equal.

backend/apid/actions/users.go
```diff
@@ -74,17 +78,32 @@ func (a UserController) CreateOrReplace(ctx context.Context, user *corev2.User)
...
-		return NewError(InternalErr, err)
+		// Create a hash for this password
+		hash, err := bcrypt.HashPassword(user.Password)
+		if err != nil {
+			return NewError(InternalErr, err)
+		}
+		user.PasswordHash = hash
+
+		// Override the cleartext password so we don't store it
+		user.Password = ""
```
echlebek, line 104: I think we should not do this. Instead, we should store the hashed password in `user.Password`.

backend/seeds/seeds.go
```diff
@@ -305,9 +305,9 @@ func setupAdminUser(store store.Store, username, password string) error {
 	}
 
 	admin := &types.User{
-		Username: username,
-		Password: hash,
-		Groups:   []string{"cluster-admins"},
+		Username:     username,
+		PasswordHash: hash,
```
echlebek, line 309: I think we should continue to specify the hashed password in the `Password` field here too.

backend/apid/actions/users.go
palourde: We actually do it just below with `bcrypt.CheckPassword(user.PasswordHash, user.Password)`!

backend/apid/actions/users.go
palourde: So you're saying we should the hashed password in both `user.Password` & `user.PasswordHash`?

backend/apid/actions/users.go
echlebek: So correct me if I'm wrong, but `bcrypt.CheckPassword` will compare a hash with a plaintext password. But I think we should check to see if `Password` contains a hash that is equal to `PasswordHash`.

backend/apid/actions/users.go
echlebek: Yes, otherwise users who get created after this change, or change their password, can never be authenticated if the version is rolled back.

backend/apid/actions/users.go
palourde: Right, but `Password` contains here a cleartext password; my understanding is that by design, you can't compare two hashes one of these need to be sent as cleartext.

backend/apid/actions/users.go
palourde: I see, it makes sense.

backend/apid/actions/users.go
echlebek: Oh, of course, since this is CreateOrUpdate, it's not pulling anything from the db.

backend/apid/actions/users.go
```diff
@@ -74,17 +78,33 @@ func (a UserController) CreateOrReplace(ctx context.Context, user *corev2.User)
...
-	// Validate password
-	if err := user.ValidatePassword(); err != nil {
-		return NewError(InvalidArgument, err)
-	}
+	// Determine if a hashed and/or cleartext password was provided
+	if user.Password != "" && user.PasswordHash != "" {
+		// Both the cleartext & hashed passwords were provided, so we need to make
+		// sure they match
+		if ok := bcrypt.CheckPassword(user.PasswordHash, user.Password); !ok {
+			return NewError(InvalidArgument, errors.New("hashed password does not the match the cleartext password"))
```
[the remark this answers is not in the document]
palourde: IMO it's just a safeguard, but I do understand your point and I'm thinking about just modifying the error message to mention that only `password_hash` should be provided.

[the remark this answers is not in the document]
palourde: Updated the message, WDYT?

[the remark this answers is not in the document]
ccressent: Looks good!

backend/seeds/seeds.go
```diff
@@ -305,9 +305,9 @@ func setupAdminUser(store store.Store, username, password string) error {
 	}
 
 	admin := &types.User{
-		Username:     username,
-		PasswordHash: hash,
-		Groups:       []string{"cluster-admins"},
+		Username: username,
+		Password: hash,
+		Groups:   []string{"cluster-admins"},
```
echlebek: This will work, but I wonder if we should also set `PasswordHash` here.

palourde: Added!

tadeboro:
Changes in this PR effectively made it impossible to reset passwords using the web API. Why? Because the client contacting the API needs to know how to generate a password hash. The `sensuctl` command can do this because it shares its guts with the backend, but just about any other client is screwed at this point.

I noticed this when I started testing the [Sensu Go Ansible Collection][collection] against a 5.21.0 and some of the user-related operations started to fail.

   [collection]: https://github.com/sensu/sensu-go-ansible

Now, changing the endpoint from `/api/core/v2/users/<user>/password` to `/api/core/v2/users/<user>/reset_password` is annoying, but doable. Generating password hash that should be sent to the backend is next to impossible.

Are there any instructions on how 3rd party clients should generate a hash? Or how they should reset passwords without having to generate a hash at all?

palourde:
Hi @tadeboro,

The hashes simply need to be generated via the bcrypt function. I believe hashing bcrypt passwords was introduced in Ansible 2.7 (https://github.com/ansible/ansible/pull/21215). They can also be hashed locally via the `sensuctl user password-hash` command.

Unfortunately, in order to solve a potential security vulnerability, we had to modify the `/users/:user/password` endpoint payload, so the current user password is provided in addition to the new hashed password. In your case, the new `reset_password` endpoint might make more sense as long as the user submitting the API requests has explicit permissions on `users`.

I'm now realizing the API documentation wasn't properly updated following this release, therefore I also submitted https://github.com/sensu/sensu-docs/issues/2563.

Please let me know if you have any other question!

tadeboro:
Documenting the hashing algorithm will solve the issue of generating passwords. It is annoying having to implement hashing on each Sensu Go web API client since the crypto libraries are usually not the simplest ones to install, but doable.

What is more annoying is that because the API changed between 5.20 and 5.21, clients will need to conditionally call the API in one of two ways, depending on the backend version. This means that we need to interrogate the backend about the version before we can start using it. Again, doable, but annoying.

Maybe repurposing the `/password` API endpoint to mean "reset password" (as it already did effectively up until this PR) and protecting it with the RBAC would make this a bit less invasive? This way existing integrations can continue working with an added twist that the users performing the operation need to have permission to do so. Adding an `/update_password` API endpoint for the actual password update would be then completely harmless.

The last thing that is puzzling me is *why* the hashing has to be done on the client. When we log in, we still use plain text passwords. This means that if we are not communicating with the backend over secure (https) connection, all of the work in this PR is nullified when the user logs in.

If the scenario we are trying to secure is "prevent storing plain text passwords on a disk", then this is a pretty awkward solution. Most configuration systems have some kind of vault built-in (Ansible has `ansible-vault`, Chef uses encrypted data bags, etc.) or offer integrations for external ones (Hashi's Vault, `pass` password manager, etc.) that allow us to sidestep the issue of plain text secrets.

So, to sum up, while this PR does make things a bit more secure, the cost of that additional security is quite high in this case.

palourde:
> It is annoying having to implement hashing on each Sensu Go web API client

I totally understand that, and this is why we implemented the `sensuctl user password-hash` command, to try to alleviate that challenge.

> What is more annoying is that because the API changed between 5.20 and 5.21, clients will need to conditionally call the API in one of two ways, depending on the backend version. 

Unfortunately in this instance it was required because of a possible security vulnerability.

> Maybe repurposing the /password API endpoint to mean "reset password" (as it already did effectively up until this PR) and protecting it with the RBAC would make this a bit less invasive

This is true from the point of view of an API client, however from the rest of the system, including RBAC, that would also be a breaking change and since 5.21 already shipped, this is unfortunately not something that could be reverted/changed without introducing a new breaking change.

>  If the scenario we are trying to secure is "prevent storing plain text passwords on a disk", then this is a pretty awkward solution. Most configuration systems have some kind of vault built-in [...]

While this is true for users of configuration management tools, you have to realize that not all Sensu users are using one. We do see more and more users driving the Sensu Go configuration via some a git repository and a simple CI/CD pipeline, for example.

Issue #3763, opened 2020-05-15
info level log messages for API requests do not include authorized user


cwjohnston:
## Expected Behavior

As an operator running sensu-backend, I can see which users are making API requests so that I can troubleshoot and report on user activity using service logs at `info` level.

## Current Behavior

All API requests are [logged at `info`](https://github.com/sensu/sensu-go/blob/c031fe6540d225ffb2f76f978d8509fda064e967/backend/apid/middlewares/logging.go#L35), but they do not include the name of the authorized user.

## Possible Solution

Add the user name to the request log message at info level.

## Context

Running at `debug` log level produces RBAC log entries useful for this purpose, but debug is too noisy for normal operation.

nikkictl:
We should be able to retrieve the user from the jwt claims of the request!

PR #3767, opened 2020-05-18
Add username as field in API request logger

nikkictl:

## What is this change?

Adds the username as a field in the API request logger by extracting the jwt claims from the request context. PR required in sensu-enterprise-go to ensure the `SimpleLogger` is called _after_ authentication.

Closes https://github.com/sensu/sensu-go/issues/3763

Yas.

Nah.

Nah.

We good.

Checked the logs on requests using tokens and api keys.
```
{"component":"apid","duration":"1.272ms","level":"info","method":"GET","msg":"request completed","path":"/api/core/v2/namespaces/default/events","size":2832,"status":200,"time":"2020-05-18T12:10:14-07:00","user":"admin"}
```

## Is this change a patch?

Yas, targeting release/5.20 branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Added
+- Added the username as a field to the API request logger.
+
 ## [5.20.1] - 2020-05-15
 *No changelog for this release.*
 
--- a/backend/apid/apid.go
+++ b/backend/apid/apid.go
@@ -135,9 +135,9 @@ func NewRouter() *mux.Router {
 func AuthenticationSubrouter(router *mux.Router, cfg Config) *mux.Router {
 	subrouter := NewSubrouter(
 		router.NewRoute(),
+		middlewares.SimpleLogger{},
 		middlewares.RefreshToken{},
 		middlewares.LimitRequest{},
-		middlewares.SimpleLogger{},
 	)
 
 	mountRouters(subrouter,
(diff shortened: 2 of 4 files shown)


palourde:
I do see a possible problem here: since we are moving the position of the SimpleLogger middleware, any requests that return an error before we have time execute this one will not be logged.

For example, on master, if I try to poke the API with an invalid API token, I will have the following entry:

```
{"component":"apid","duration":"0.082ms","level":"info","method":"GET","msg":"request completed","path":"/api/core/v2/events","size":38,"status":401,"time":"2020-05-19T15:04:16-04:00"}
```

On this branch, I no longer see it =/

nikkictl:
Ah, I see how the order of operations can get tricky here. Well we need the auth middleware to run before we can log the username, otherwise we don't know what the username is from the context. Perhaps a new username logger middleware could be run instead of loading this logic in the existing logger. WDYT @palourde?

nikkictl:
I wonder if there is a way to do this with the hijack logger.. I'll have to look into it more!

nikkictl:
I moved the logger middleware right after the auth layer, and invoked the logger manually on auth errors.

Now, the user is logged when auth is successful:
```
{"component":"apid","duration":"2.593ms","level":"info","method":"GET","msg":"request completed","path":"/api/core/v2/events","size":5649,"status":200,"time":"2020-05-20T10:01:54-07:00","user":"admin"}
```

And a 401 is logged when auth is not successful:
```
{"component":"apid","error":"key /sensu.io/apikeys/68c2027c-f286-4dca-bf9d-c32a1294b2cdd not found","level":"warning","msg":"invalid api key","time":"2020-05-20T10:01:59-07:00"}
{"component":"apid","duration":"0.021ms","level":"info","method":"GET","msg":"request completed","path":"/api/core/v2/events","size":42,"status":401,"time":"2020-05-20T10:01:59-07:00","user":""}
```

PR #3769, opened 2020-05-19
Fix a bug where users could not change password

palourde:

## What is this change?

This PR fixes a bug introduced in https://github.com/sensu/sensu-go/pull/3753.

TL;DR We need to empty the `Password` field once we validated it was the proper one, otherwise the `CreateOrReplace` will refuse it because it differs from the hash.

Fixes a bug found in https://github.com/sensu/sensu-go-qa-crucible/pull/122#pullrequestreview-412794184.

Nope, unreleased stuff.

Nope

Nope

Nope

Deployed on staging & tested with the crucible tests.

## Is this change a patch?

Yes but it's a patch for an unreleased feature.

### DIFF
--- a/backend/apid/routers/users.go
+++ b/backend/apid/routers/users.go
@@ -145,6 +145,9 @@ func (r *UsersRouter) updatePassword(req *http.Request) (interface{}, error) {
 		return nil, err
 	}
 
+	// Remove any old password hash and set the new password hash. The controller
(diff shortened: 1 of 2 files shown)


palourde:
@echlebek I added a unit test for that, in addition to the automated QA that's part of https://github.com/sensu/sensu-go-qa-crucible/pull/122!

PR #3818, opened 2020-06-01
Publish a module for core/v2 and core/v3 APIs

echlebek:
## What is this change?

Publishes modules for core/v2 and core/v3, as sensu-go is not importable as a module. To accomplish this I duplicated a few extraneous packages into core/v2/internal.

This should allow users to import core/v2 and core/v3 at a well defined version.

No, this is somewhat experimental, and not a user-facing change.

We will need to follow up with some release tags for core/v2 and core/v3 modules.

I made sure that building and testing works.

## Is this change a patch?

No.


api/core/v2/go.mod
```diff
@@ -0,0 +1,23 @@
+module github.com/sensu/sensu-go/api/core/v2
+
+go 1.14
```
nikkictl, line 3: I thought there were issues preventing us from upgrading to go 1.14?

echlebek: That's true, I should adjust this to be go 1.13.

PR #3835, opened 2020-06-08
Bugfix/agent panic

### LINKED ISSUE #3779: executeCheck() gets SIGSEGV when command in hook is invalid, and the agent reports nothing
## Expected Behavior

even when configuration is invalid, the agent should not killed by SIGSERV, and should reports the problem.

## Current Behavior

when `command` in hook is invalid, the agent gets SIGSEGV. and the problem is not reported back to user. the status of the check remains `success`.

```
May 21 17:36:23  sensu_agent[2773]: panic: runtime error: invalid memory address or nil pointer dereference
May 21 17:36:23  sensu_agent[2773]: [signal SIGSEGV: segmentation violation code=0x1 addr=0x68 pc=0xe5e0dc]
May 21 17:36:23  sensu_agent[2773]: 
May 21 17:36:23  sensu_agent[2773]: goroutine 87 [running]:
May 21 17:36:23  sensu_agent[2773]: github.com/sensu/sensu-go/agent.(*Agent).ExecuteHooks(0xc0001b6dc0, 0x128b0a0, 0xc00015c540, 0xc0005fc080, 0xc0000f8a80, 0x0, 0x6, 0x8, 0x0)
May 21 17:36:23  sensu_agent[2773]:     /wrkdirs/usr/ports/sysutils/sensu-go-agent/work/sensu-go-5.20.0/agent/hook.go:30 +0x26c
May 21 17:36:23  sensu_agent[2773]: github.com/sensu/sensu-go/agent.(*Agent).executeCheck(0xc0001b6dc0, 0x128b0a0, 0xc00015c540, 0xc0005fc080, 0xc000256000)
May 21 17:36:23  sensu_agent[2773]:     /wrkdirs/usr/ports/sysutils/sensu-go-agent/work/sensu-go-5.20.0/agent/check_handler.go:252 +0xe8f
May 21 17:36:23  sensu_agent[2773]: created by github.com/sensu/sensu-go/agent.(*Agent).handleCheck
May 21 17:36:23  sensu_agent[2773]:     /wrkdirs/usr/ports/sysutils/sensu-go-agent/work/sensu-go-5.20.0/agent/check_handler.go:69 +0x2c8
```

the end result is that, after the hook execution error, the check reports nothing. and the web ui says nothing about the problem.

## Possible Solution

whenever child process is killed, the agent should report the problem.

1. create a hook with command including shell pipe, like this:

```yaml
type: HookConfig
api_version: core/v2
metadata:
  created_by: admin
  name: dmesg
  namespace: default
spec:
  command: dmesg | tail
  runtime_assets: null
  stdin: false
  timeout: 10
```

2. attach the hook to a check

## Context

fixing "nil pointer dereference" that caused SIGSEGV is one thing, but the agent should have not been fragile like this.

I would argue that using pipe is one of common operations for which the hook is designed. but that would be reserved for another bug report.

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): 5.20.0
* Installation method (packages, binaries, docker etc.): my own build
* Operating System and version (e.g. Ubuntu 14.04): FreeBSD 12.1

echlebek:
## What is this change?

```
    Fix a nil panic in sensu-agent
    
    This commit fixes a nil panic in sensu-agent when a hook is
    specified that does not exist. Additionally, other hook config
    errors are now surfaced properly instead of via an out-of-band
    event.
```

Fixes #3779 

Yes

No

The agent hook failure protocol required some review.

The unit test in the first commit of this changelog proves that the fix works.

## Is this change a patch?

Yes

### DIFF
--- a/agent/hook.go
+++ b/agent/hook.go
@@ -3,6 +3,7 @@ package agent
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"strconv"
@@ -27,12 +28,12 @@ func (a *Agent) ExecuteHooks(ctx context.Context, request *corev2.CheckRequest, 
 			// run all the hooks of that type
 			for _, hookName := range hookList.Hooks {
 				hookConfig := getHookConfig(hookName, request.Hooks)
+				if hookConfig == nil {
+					hookConfig = errorHookConfig(a.config.Namespace, hookName, errors.New("missing hook config"))
+				}
 				origCommand := hookConfig.Command
-				if ok := a.prepareHook(hookConfig); !ok {
-					// An error occured during the preparation of the hook and the error
-					// has been sent back to the server. At this point we should not
-					// execute the hook and wait for the next check request
-					continue
+				if err := a.prepareHook(hookConfig); err != nil {
+					hookConfig = errorHookConfig(hookConfig.Namespace, hookConfig.Name, err)
 				}
 				// Do not duplicate hook execution for types that fall into both an exit
 				// code and severity (ex. 0, ok)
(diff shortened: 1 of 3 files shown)


agent/hook.go
```diff
@@ -50,6 +51,19 @@ func (a *Agent) ExecuteHooks(ctx context.Context, request *corev2.CheckRequest,
...
 
+// errorHookConfig will return a hookConfig that will echo its error argument
+// and exit status 2.
+func errorHookConfig(namespace, name string, err error) *corev2.HookConfig {
+	return &corev2.HookConfig{
+		ObjectMeta: corev2.ObjectMeta{
+			Namespace: namespace,
+			Name:      name,
+		},
+		Command: fmt.Sprintf("echo 'hook %s: %s'; exit 2", name, err),
```
palourde, line 62: The resulting command was a bit confusing to me in my tests, e.g.
```
hooks:
    - command: 'echo ''hook prout: missing hook config''; exit 2'
```

Is this the intended behaviour?

echlebek: That looks like yaml serialization weirdness to me. I'm not sure if it's a problem or not...

echlebek: According to the yaml specification, two single quotes are how single quotes are escaped in yaml. https://yaml.org/spec/current.html#id2534365

PR #3850, opened 2020-06-18
Enable linting for our Go modules

amdprophet:

## What is this change?

Piggybacking off of #3849 to add support for our Go modules.

See #3689

Nope!

Nope!

I couldn't find a way to set the version of golangci-lint once and reuse it for each GitHub Actions job.

Nope!

Ran the linter locally.

## Is this change a patch?

No.


amdprophet:
I believe the error in the integration test was never exposed before as the test name was incorrect. I'll look into what's going on with it tomorrow.

echlebek:
Oops, looks like some tests are failing.

amdprophet:
@palourde do you have any opinions on what we should do with the failing test?

echlebek:
@amdprophet was that question intended for me?

PR #3859, opened 2020-06-25
Limit captured output of command execution

jamesdphillips:
## What is this change?

* Uses a ring buffer for capturing output of command.
* Sets upper bounds of output to 1.5 MB. While this value is somewhat arbitrary, etcd key size limits mean that events cannot hold more than 2MB anyways.
* Helps avoids situations where a command writes extremely large amounts of data to STDOUT / STDERR.

https://github.com/armon/circbuf/

Without limits, the agents memory usage can spiral out of control, potentially leading to a crash when it is unable to allocate more memory.

An easy way to observe the negative effect of the previous implementation, is to create that runs the `yes` command and to set a timeout of say `15` seconds. Using pprof you will quickly observe the agent's `inuse_space` explode.

Sure.

1. While using pprof to investigate, I observed that when a command timed out it did not appear that the previous buffer was ever deallocated. I'm not sure if this just a case of the GC having not been run or if there is a memory leak.
2. Unclear to me how I might write a nice cross-platform unit test.

None required.

Manual.

## Is this change a patch?

Yes..?

### DIFF
--- a/command/command.go
+++ b/command/command.go
@@ -2,7 +2,6 @@
 package command
 
 import (
-	"bytes"
 	"context"
 	"os/exec"
 	"strings"
@@ -10,9 +9,12 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/armon/circbuf"
 	"github.com/sensu/sensu-go/types"
 	"github.com/sirupsen/logrus"
 )
+
+const maxOutputLen = 1500000
 
 const undocumentedTestCheckCommand = "!sensu_test_check!"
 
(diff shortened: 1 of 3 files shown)


jamesdphillips:
> Maximum size of stored check outputs. In bytes. When set to a non-zero value, the Sensu backend truncates check outputs larger than this value before storing to etcd. max_output_size does not affect data sent to Sensu filters, mutators, and handlers.

Foot gun or otherwise, we've set the expectation that we'll capture all command output.

This requires more thought.

Issue #2784, opened 2019-03-11
Missing mutators do not cause errors to be reported


echlebek:
## Expected Behavior
When a handler specifies a mutator, but the mutator does not exist, an error is reported when the handler executes.

## Current Behavior
No error is reported. The handler receives 0 bytes.

## Possible Solution
Report an error when a nil mutator is retrieved from the store.

## Steps to Reproduce (for bugs)
1. Create a handler that specs a mutator that does not exist.
2. Associate the handler with a check.
3. Observe the handler producing 0 bytes.

## Your Environment
Alpine Linux (docker)

PR #3865, opened 2020-06-29
Log and return an error if a mutator cannot be found

nikkictl:

## What is this change?

Logs and returns an error if a mutator cannot be found.

Closes https://github.com/sensu/sensu-go/issues/2784

Yas.

Nah.

Nah.

We good.

- Added a unit test that fails without the suggested changes.
- Verified e2e with the steps to reproduce from the original issue.

New enriched log with error:
```
{"check_name":"foo","check_namespace":"default","component":"pipelined","entity_name":"nikkictl","entity_namespace":"default","error":"mutator \"nope\" does not exist","handler":"cat","level":"error","msg":"error mutating event","time":"2020-06-29T12:43:11-04:00","uuid":"5192efe0-8645-4b2d-a55a-1e05ec3d1f20"}
```

## Is this change a patch?

Yes, but TBH I'm not sure what our branch strategy is right now with a few potential upcoming releases. Targeting master for now.

### DIFF
--- a/backend/pipeline/handler.go
+++ b/backend/pipeline/handler.go
@@ -76,7 +76,7 @@ func (p *Pipeline) HandleEvent(ctx context.Context, event *corev2.Event) error {
 
 		eventData, err := p.mutateEvent(handler, event)
 		if err != nil {
-			logger.WithError(err).Warn("error mutating event")
+			logger.WithFields(fields).WithError(err).Warn("error mutating event")
 			if _, ok := err.(*store.ErrInternal); ok {
 				// Fatal error
 				return err
--- a/backend/pipeline/handler.go
+++ b/backend/pipeline/handler.go
@@ -76,7 +76,7 @@ func (p *Pipeline) HandleEvent(ctx context.Context, event *corev2.Event) error {
 
 		eventData, err := p.mutateEvent(handler, event)
 		if err != nil {
-			logger.WithFields(fields).WithError(err).Warn("error mutating event")
+			logger.WithFields(fields).WithError(err).Error("error mutating event")
 			if _, ok := err.(*store.ErrInternal); ok {
 				// Fatal error
 				return err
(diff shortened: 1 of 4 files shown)


backend/pipeline/mutator.go
```diff
@@ -59,7 +60,7 @@ func (p *Pipeline) mutateEvent(handler *corev2.Handler, event *corev2.Event) ([]
 		extension, err := p.store.GetExtension(tctx, handler.Mutator)
 		if err != nil {
 			if err == store.ErrNoExtension {
-				return nil, nil
+				return nil, fmt.Errorf("mutator '%s' does not exist", handler.Mutator)
```
echlebek, line 63:
```suggestion
				return nil, fmt.Errorf("mutator %q does not exist", handler.Mutator)
```

backend/pipeline/handler.go
```diff
@@ -76,7 +76,7 @@ func (p *Pipeline) HandleEvent(ctx context.Context, event *corev2.Event) error {
 
 		eventData, err := p.mutateEvent(handler, event)
 		if err != nil {
-			logger.WithError(err).Warn("error mutating event")
+			logger.WithFields(fields).WithError(err).Warn("error mutating event")
```
echlebek, line 79: Should this be at error level?

backend/pipeline/mutator.go
nikkictl: TIL `%q	a single-quoted character literal safely escaped with Go syntax.`

backend/pipeline/handler.go
nikkictl: Most likely. I wasn't sure why it was warning level to begin with. Considering it didn't even have the fields included, it probably wasn't a very useful log message!

backend/pipeline/mutator.go
nikkictl: The message looks funky with `%q` but will be more parseable with log management: `"error":"mutator \"nope\" does not exist"`.

PR #3879, opened 2020-07-02
Subscribe agentd sessions to entity config updates

palourde:
## What is this change?

This PR modifies the agent sessions in agentd so receive entity config updates via the bus, encode them and send them over the relevant agents via the transport.

Additionally, I removed the `sendq` channel, which is unused.

Closes https://github.com/sensu/sensu-go/issues/3880

There's no changes exposed to users yet.

As discussed, I omitted the integration of the cache/watcher, which will be implemented in a subsequent PR.

Nope

Not yet.

Unit test + manually verified the agent received a mocked entity config update.

## Is this change a patch?

Nope

### DIFF
--- a/backend/agentd/session.go
+++ b/backend/agentd/session.go
@@ -12,6 +12,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/prometheus/client_golang/prometheus"
 	corev2 "github.com/sensu/sensu-go/api/core/v2"
+	corev3 "github.com/sensu/sensu-go/api/core/v3"
 	"github.com/sensu/sensu-go/backend/messaging"
 	"github.com/sensu/sensu-go/backend/ringv2"
 	"github.com/sensu/sensu-go/backend/store"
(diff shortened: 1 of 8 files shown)


palourde:
@echlebek Updated the PR with the changes discussed!

backend/agentd/session.go
```diff
@@ -214,7 +229,48 @@ func (s *Session) sender() {
...
+				// so it can register back
+				s.cancel()
+				continue
+			case store.WatchUnknown:
+				logger.Error("session received unknown watch event")
+				continue
+			}
+
+			// Enforce the entity class to agent
+			if watchEvent.Entity.EntityClass != corev2.EntityAgentClass {
```
echlebek, line 263: In this case, we also need to update the entity in the store, so that it no longer has a non-agent class.

palourde: Should we avoid publishing the entity update in that case and simply publish the watch event that will be result from the store update?

echlebek: Hm, good question. It's kind of a corner case. I don't think it particularly matters, but we could save some time if we avoid publishing, and just wait for the watch event to come in that will be triggered by updating the store.

palourde: Just pushed a new commit that update the entity config when the class is misconfigured!

backend/agentd/session_test.go
```diff
@@ -307,3 +320,64 @@ func TestSessionEntityWatchDeleteAndUnknown(t *testing.T) {
...
 	}
 }
+
+func TestSessionInvalidEntityClassUpdate(t *testing.T) {
+	wait := make(chan struct{})
+
+	conn := new(mocktransport.MockTransport)
+	// Mock the Receive method by blocking it for 100ms and returns an empty
+	// message so it doesn't block our test for too long
+	conn.On("Receive").After(100*time.Millisecond).Return(&transport.Message{}, nil)
```
echlebek, line 330: Does the test still work if you return immediately?

palourde: Oh yeah absolutely, it's just that the actual implementation is a blocking channel so it would constantly call this mocked method; the 100ms wait is just so it spams a bit less but doesn't delay the test execution too much.

PR #3884, opened 2020-07-06
Support agent EntityConfig updates from backend

### LINKED ISSUE #3873: Modify sensu-agent to receive entity updates from the backend
In sensu-go 6.0, the backend will be the source of truth for entity configuration.

Update sensu-agent to expect messages of type "EntityConfig". When sensu-agent gets an EntityConfig message, it means that its configuration has changed in the backend. It should update its cached instance of its configuration and use the cached entity configuration for future event messages and keepalives.

The `getAgentEntity` method should be modified to create an entity from the cached entity config, and the agent's entity state. This method is used in various places by the agent.

echlebek:
## What is this change?

This commit adds support for receiving updates about entity
configuration from sensu-backend to sensu-agent. When a user
updates an entity's configuration, the backend will send a message
to the corresponding connected agent informing it that the entity
configuration has changed. The agent will correspondingly update
its cached version of the entity configuration.

A small unrelated change was also made in the interests of
completeness for Sensu 6.0. The sensu-backend init tool will now
also upgrade the cluster to the latest version, so that new
clusters will start out at the correct version.

Fixes #3873 

Yes

While testing I realized that I did not cause new Sensu installations to start at the most recent database version. I have added some code to the backend init tool to resolve this issue.

cc @hillaryfraley @sensu/reliability, I believe this change will be of particular interest to your team. Please contact me for details if the change is unclear from the description.

Mainly unit testing as there is a lot of work in-flight.

## Is this change a patch?

This is a breaking change.


echlebek:
It seems my tests are passing locally and failing on CI. :shrug:

Issue #3843, opened 2020-06-12
Allow accessing the backend event API from the Javascript filter execution environment


echlebek:
Adapt the backend events API to the Javascript filter execution environment. The backend events API is here: https://github.com/sensu/sensu-go/blob/master/backend/api/event.go

The Javascript environment should provide a way to construct a new API client. The store and bus should be implicitly provided by the backend, and the authorizer should be constructed with a provided API key.

```
var apiKey = "foobar";
var events = Sensu.newEventAPI(apiKey);
```

All of the methods in the events API should be available. eg,
```
var event = events.FetchEvent("router", "network");
```

Notice that the context has been dropped as compared with the Go API. The system should automatically adapt the methods to inject the context, rather than exposing the context concept directly to the user.

The errors returned by Go functions should be converted into Javascript exceptions. The system should behave like a Javascript API, not a Go API.

There is a proof-of-concept for automatically converting any Go function that takes a context as its first argument, and returns an error as its last return value, into this kind of Javascript function. It can be found here: https://github.com/sensu/sensu-go/blob/feature/sensu-sdk/sdk/interpreter.go#L235

Finally, any data structures returned should be converted with `dynamic.Synthesize` so that they are not raw Go pointers, but rather semi-serialized `map[string]interface{}` objects that adhere to the Sensu JSON API.

PR #3887, opened 2020-07-07
Add dynamic.Function

echlebek:
## What is this change?

This commit adds a function that converts a Go function into a
Javascript function, suitable for calling from the Otto Javascript
environment.

Part of #3843 

No


echlebek:
ping @ccressent

Issue #3881, opened 2020-07-02
Send EntityConfig CRUD operations to Wizard Bus


echlebek:
Create a watcher over `*corev3.EntityConfig`, and use it to generate objects of the following type:

```
type WatchEventEntityConfig struct {
  Entity *corev3.EntityConfig
  Action store.WatchActionType
}
```

The type should be created in `backend/store`, if it does not exist already.

The watcher should be instantiated in `backend/backend.go` along with the other top level services.

The watch events should be published to wizard bus, to an entity subscription based on the entity's namespace and name.

PR #3888, opened 2020-07-07
Publish entity config updates to agentd sessions

palourde:
## What is this change?

It adds an etcd watcher to entity configs and publish updates to the bus, so sessions can receive updates about their subscribed entities.

Closes https://github.com/sensu/sensu-go/issues/3881

I believe this change can be considered as covered with the lines added via https://github.com/sensu/sensu-go/pull/3884.

Nope!

Nope!

We will need to cover the entity config update in a separate sensu-docs issue I believe.

Unit tests + manual testing. I'll open an issue on sensu-go-qa-crucible so we can automate some of this testing.

## Is this change a patch?

Nope

### DIFF
--- a/backend/agentd/watcher.go
+++ b/backend/agentd/watcher.go
@@ -1,0 +1,75 @@
+package agentd
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/coreos/etcd/clientv3"
+	"github.com/gogo/protobuf/proto"
+	corev3 "github.com/sensu/sensu-go/api/core/v3"
+	"github.com/sensu/sensu-go/backend/messaging"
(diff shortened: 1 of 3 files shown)


backend/agentd/watcher.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+
+		var (
+			configWrapper wrap.Wrapper
+			entityConfig  corev3.EntityConfig
+		)
+
+		// Decode and unwrap the entity config
+		if err := proto.Unmarshal(response.Object, &configWrapper); err != nil {
+			fmt.Println(err)
```
nikkictl, line 49:
```suggestion
```

PR #3892, opened 2020-07-08
Update eventd to use corev3 and storev2

### LINKED ISSUE #3803: Update eventd check TTL to use core/v3.EntityConfig.
Use `core/v3.EntityConfig` instead of calls to `core/v2` `GetEntityByName` in check TTL callbacks in eventd.

These callbacks are called `dead` and `alive` respectively, in the `eventd` package.

ccressent:
## What is this change?

> Now the dead() callback only uses the corev3.EntityConfig and
storev2.Get() to check if an entity exists.
>
>Proxy entity creation has also been updated to use corev3.EntityConfig
and corev3.EntityState instead of the "monolithic" corev2.Entity. The
proxy creation logic is unchanged (although I feel it could be vastly
simplified). All interactions with the entity store in eventd are now
done through the interfaces provided by corev3 and storev2.
>
>The associated unit tests have been updated and made more "strict" in
that they now have additional checking on the arguments mocked functions
receive.

Closes #3803.

Yes, added.

No.

Updating the `dead()` callback wasn't much work overall, but updating proxy entity creation was, especially the unit testing aspects. We didn't think of it when pointing the issue.

No documentation change or new documentation is needed. Everything works as it did before, none of the interfaces are changed. This is a purely internal change.

All the relevant eventd unit tests have been updated and reworked to be more "strict": their implementation has been updated to match the fact that we now use `corev3` and `storev2` everywhere in eventd (for entities) and mocked functions actually check the value of arguments they receive to ensure they are what we expected.

The integration tests have been updated.

I'm still in the process of deploying the change on a staging environment and running the full suite of automated QA.

## Is this change a patch?

No.


backend/eventd/entity.go
```diff
@@ -20,39 +24,86 @@ func createProxyEntity(event *corev2.Event, s store.EntityStore) error {
 	}
 
 	// Determine if the entity exists
-	ctx := context.WithValue(context.Background(), corev2.NamespaceKey, event.Entity.Namespace)
-	entity, err := s.GetEntityByName(ctx, entityName)
-	if err != nil {
-		return err
-	}
+	//NOTE(ccressent): there is no timeout for this operation?
```
echlebek, line 27: There is an implicit timeout on the store governed by the etcd client, but no explicit timeout for `dead` callbacks. Perhaps we should consider adding one? It's unclear to me if it would help or harm.

backend/eventd/entity.go
```diff
@@ -20,39 +24,86 @@ func createProxyEntity(event *corev2.Event, s store.EntityStore) error {
...
+				// Create a brand new entity since we can't rely on the provided
+				// entity, which represents the agent's entity
+				state.SetMetadata(&entityMeta)
+				config.SetMetadata(&entityMeta)
+			} else {
+				// Use on the provided entity
+				config, state = corev3.V2EntityToV3(event.Entity)
+			}
+
+			state.Metadata.CreatedBy = event.CreatedBy
```
echlebek, line 69: Should we set LastSeen here?

backend/eventd/entity_test.go
```diff
@@ -24,116 +29,237 @@ func TestCreateProxyEntity(t *testing.T) {
...
+				// Assert that CreateOrUpdate() was called with the expected
+				// request and wrapper
+				s.On("CreateOrUpdate", stateReq,
+					mock.MatchedBy(func(w *wrap.Wrapper) bool {
+						return w.String() == wState.String()
+					})).Return(nil)
+
+				// Assert that CreateIfNotExists() was called with the expected
+				// request and wrapper type
+				// TODO(ccressent): can we do something more strict with the
```
echlebek, line 106: Have you abandoned the TODO? I think it's fine as-is.

backend/eventd/entity.go
ccressent: I think we could if we wanted to, but as far as I can tell, the original code doesn't do it. For that reason, and to have the best chance of not breaking something subtle/unforeseen, I'd rather stick to that, at least in this pass.

Also, setting/updating `LastSeen` is probably the responsibility of `keepalived` rather than `eventd`?

backend/eventd/entity_test.go
ccressent: I'm sure we can find a solution, but I'm punting on this...
Before, we would accept `mock.Anything` as argument. Now we only accept `*wrap.Wrapper` with the expected type information. The grail would be check the actual wrapped value on top of that. 

I've mostly abandoned it because other than type information, in a `Wrapper`, we only have a protobuf encoded (by default) `[]byte` to inspect and if I recall correctly, I would get 2 such slices that were not 100% exactly the same, even if they would still decode to the same object, and the test would fail on that ground.

backend/eventd/entity.go
ccressent: Not sure either, I originally left that note because similar operations in `dead()` do have a timeout in their context: https://github.com/sensu/sensu-go/blob/524efa49a13840381dc2a12b2ef0d463f0af462c/backend/eventd/eventd.go#L348-L352

backend/eventd/entity.go
echlebek: Ah you're totally right, that shouldn't happen here.

PR #3894, opened 2020-07-09
Loosen validation for subjects names in RBAC

palourde:
## What is this change?

It modifies the validation rules for ClusterRoleBindings & RoleBindings subjects, in order to allow any Unicode characters in the subject names.

Additionally, it enforces the capitalization of the subject types by automatically capitalizing them before storing them; previously it would accept `group` or `user` but RBAC wouldn't match them during the actual authorization, which is not super optimal in terms of UX. 

Finally, I did some light linting.

Closes https://github.com/sensu/sensu-enterprise-go/issues/1133

Added

Nope

Nope

I don't think we need to update the docs, it didn't specified the restriction around subjects and groups.

Unit tests + manually tested on my local AD. I don't believe we need any e2e tests, since it's really just a matter of validation of subjects.

## Is this change a patch?

Nope


api/core/v2/rbac_test.go
```diff
@@ -197,15 +197,16 @@ func Test_split(t *testing.T) {
...
+		ExpErr       bool
+		Subjects     []Subject
+		WantSubjects []Subject
 	}{
 		{
 			Name: "valid",
 			Subjects: []Subject{
 				{
-					Type: "user",
+					Type: "User",
```
ccressent, line 209: Should `"User"` be replaced with `corev2.UserType` to be consistent?

api/core/v2/rbac.go
```diff
@@ -258,26 +260,34 @@ func (b *RoleBinding) Validate() error {
...
 	if len(subjects) == 0 {
-		return errors.New("a RoleBinding must have at least one subject")
+		return subjects, errors.New("a RoleBinding must have at least one subject")
 	}
 
 	for i, subject := range subjects {
-		if err := ValidateName(subject.Type); err != nil {
-			return fmt.Errorf("subject %d: type not valid: %s", i, err)
+		subjects[i].Type = strings.Title(subject.Type)
+		if subjects[i].Type != GroupType && subjects[i].Type != UserType {
```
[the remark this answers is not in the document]
echlebek: `subject` is a non-pointer, so the assignment would be ineffectual then.

You'd have to:

```
for i := range subjects {
  subject := &subjects[i]
  ...
}
```

api/core/v2/rbac_test.go
palourde: Absolutely!

palourde:
> Does that mean that, before this change, if I created a role with all lower case type ("user" or "group"), it would be broken?

@ccressent That's right 😭 we would let you create it without any error but it would essentially be ignored when determining if the user is authorized.

The same thing is happening right now with the `type` field in RoleRef =/ if you provide `role` rather than `Role`, we will accept it will have no effect during authorization. I'll look into it right away

palourde:
Just pushed a couple of commits to take care of the same issue we had in `roleRef` with the type being ignored if not capitalized!

PR #3898, opened 2020-07-09
Add extension point to pipeline filters for Javascript functions

### LINKED ISSUE #3843: Allow accessing the backend event API from the Javascript filter execution environment

### LINKED ISSUE #3897: Important metadata are completely omitted from events when no labels and/or annotations are undefined
##  Expected Behavior

As a Sensu operator I can write concise event filter expressions which examine the values of labels and/or annotations without implementing boilerplate logic which first checks whether or not the `labels` or `annotations` metadata keys are present.
 
## Current Behavior

An agent with no labels nor annotations configured will generate events where `event.entity.labels` is null, and/or `event.entity.annotations` is null. As the Golang JSON serializer is known to omit keys with null values, operators are somewhat forced to write functions like this one which checks for the presence of `event.entity.annotations` before checking the value of an annotation:

```
(function () {
        var a = event.entity.annotations;
        if(!a) { console.log("there are no annotations") ; return false; } 
        var key = "suppress-opsgenie";
        return a[key] == "true";
      })()
```
## Possible Solution

Ensure that metadata keys for labels and annotations default to empty array/slice/wahtever of key/value pairs instead of null so that operators can safely interrogate keys therein without the boilerplate.

## Context

In my case I'm trying to suppress a handler from firing when entity annotation "suppress-opsgenie" is "true". I want to treat the absence of the attribute as "false".

Seems like one should be able to write a filter like this:

`!!event.entity.annotations['suppress-opsgenie'] || event.entity.annotations['suppress-opsgenie'] != true`

But this presupposes that `event.entity.annotations` is a valid key, which it does not seem to be when the agent has no annotations. I haven't tested this for labels but I assume the behavior is the same.

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): 5.21.0

echlebek:
## What is this change?

This PR adds an extension point for to the `pipeline` package for adding Javascript functions to the event filter execution environment.

Fixes #3843 
Fixes #3897

Yes

While working on this change, I realized I needed a better caching mechanism for Otto VMs, so I fixed and improved the VM cache. I believe it was actually broken previously, and could have resulted in panics at runtime.

@hillaryfraley @cwjohnston we should probably have a call about this feature, and how to document it. I believe it will require some precise explanation and reference.

So far verification has been through unit and integration tests. I need to write up a QA test for this feature and get it into the QA repo.

## Is this change a patch?

Nein


echlebek:
Testing has shown that, unsurprisingly, the auth system is not quite configured properly for filters. I'll need to debug that tomorrow.

portertech:
Please do not merge this until we discuss it on Monday.

backend/pipeline/js.go
```diff
@@ -0,0 +1,62 @@
...
+	// these claims set up authorization credentials for the event client.
+	// the actual roles and role bindings are generated when namespaces are
+	// created, for the purposes of the system.
+	claims := &corev2.Claims{
+		Groups: []string{pipelineRoleName},
+		Provider: corev2.AuthProviderClaims{
+			ProviderID:   "system",
+			ProviderType: "basic",
+			UserID:       pipelineRoleName,
+		},
```
palourde, line 40: The provider is not required here, it's really meant to be consumed the authentication providers when refreshing authentication tokens; the authorization logic does not consult this struct.

PR #3908, opened 2020-07-15
Ensure keepalive entities are consistent

echlebek:
This commit fixes a bug where entity information in keepalive events
was not correct. To achieve this, the following things were done:

* Use a.getAgentEntity() when creating keepalives
* Clear the agent's cached entity on session restart
* Update keepalived to only send an entity notification if the
entity config is out of sync.
* Update keepalived to update keepalive events with the latest
entity config.


echlebek:
I think the problem is actually that the backend session doesn't shut down properly for some reason.

PR #3910, opened 2020-07-16
Various UX improvements in logging

palourde:

## What is this change?

It fixes a couple of bugs/UX issues around logging.

So it's easier to debug various errors.

Added

Nope

Nope

Nope

Manually

## Is this change a patch?

Yes, but no

### DIFF
--- a/backend/schedulerd/executor.go
+++ b/backend/schedulerd/executor.go
@@ -370,7 +370,7 @@ func buildRequest(check *corev2.CheckConfig, s store.Store, secretsProviderManag
 	// Prepare log entry
 	fields := logrus.Fields{
 		"namespace": check.Namespace,
-		"mutator":   check.Name,
+		"check":     check.Name,
 		"assets":    check.RuntimeAssets,
 	}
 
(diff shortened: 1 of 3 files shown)


backend/schedulerd/executor.go
```diff
@@ -370,7 +370,7 @@ func buildRequest(check *corev2.CheckConfig, s store.Store, secretsProviderManag
 	// Prepare log entry
 	fields := logrus.Fields{
 		"namespace": check.Namespace,
-		"mutator":   check.Name,
```
echlebek, line 373: lol

calebhailey:
Looks like this is also a further enhancement on https://github.com/sensu/sensu-enterprise-go/issues/1104? 

🎉

palourde:
@calebhailey The use case that bit me was that I had a secret provider configured & secrets enabled in my checks, but I forgot to enable mTLS so it wasn't working and I had no idea why, even after looking at the logs.

I think there's still some stuff we could do to improve the UX (the agent should return an error in the output if some secrets were expected but never transmitted possibly?)

PR #3914, opened 2020-07-16
Send a close message over the websocket when closing a session

palourde:
## What is this change?

It forces the session to send a close control message over the transport (websocket) when stopping, which allows the peer (agent) to close the transport on its side too.

I also modified a bit the `stop()` logic we don't block forever when stopping the session if the peer never send back a message.

It fixes a bug where the agent would always send back a last keepalive (with a possibly stale entity config) when a session would severe its connection.

Nope

Nope

Nope

Nope

Manually verified, and also applied this patch over Eric's PR while testing.

I probably need a second pair of 👀 just to make sure I'm not missing anything.

## Is this change a patch?

Unreleased stuff.

### DIFF
--- a/backend/agentd/session.go
+++ b/backend/agentd/session.go
@@ -272,8 +272,6 @@ func (s *Session) sender() {
 			// Handle the delete and unknown watch events
 			switch watchEvent.Action {
 			case store.WatchDelete:
-				// The entity was deleted, we should sever the connection to the agent
-				// so it can register back
 				s.cancel()
 				continue
 			case store.WatchUnknown:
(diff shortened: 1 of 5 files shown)


backend/agentd/session.go
```diff
@@ -427,7 +430,28 @@ func (s *Session) stop() {
 		}
 	}()
 
+	// Send a close message to ensure the agent closes its connection if the
+	// connection is not already closed
+	if !s.conn.Closed() {
+		if err := s.conn.SendCloseMessage(); err != nil {
+			logger.WithError(err).Warning("unexpected error while sending a close message to the agent")
+		}
+	}
```
echlebek, line 439: This is a race condition, since the conn could close between checking it and the call to SendCloseMessage. Perhaps we should interrogate the error more carefully, or avoid logging it?

backend/agentd/session.go
```diff
@@ -427,7 +430,28 @@ func (s *Session) stop() {
...
+	go func() {
+		s.wg.Wait()
+		close(done)
+	}()
+	select {
+	case <-done:
+	case <-time.After(closeGracePeriod):
+	}
+
 	s.wg.Wait()
```
echlebek, line 455: This `Wait()` is redundant now

palourde: Ah yes good catch, I thought I removed it 🤦

backend/agentd/session.go
palourde: Yeah, you're right, I don't think we really care about the error here

PR #3924, opened 2020-07-17
Fix a db migration bug

echlebek:
This commit fixes a bug in the database migration that creates
a system:pipeline role and role binding for each namespace. The
type of the role binding subject has been correct from "user" to
"Group".


echlebek:
@ccressent I felt like group made more sense as there will be no explicit `system:pipeline` user.

PR #3925, opened 2020-07-18
Make dynamic.Function work on javascript source

echlebek:
## What is this change?

Makes `dynamic.Function` compile Javascript source code if a string is passed to it.

It supports embedding Javascript functions in Sensu.

No

Unit tests, and the change is being used in the enterprise product.


echlebek:
It is https://github.com/sensu/sensu-enterprise-go/pull/1155

PR #3932, opened 2020-07-21
Fix bug in api package where consumer was unable to retrieve entities

jamesdphillips:
## What is this change?

Previously, the `EntityClient` (api package) used the generic store methods to list/get v2 entities and now with state/config being stored separately, this no longer works in 6.X.

Retain functionality.

No. Regression did not exist in previous release.
 
Unit tests.

### DIFF
--- a/backend/api/entity.go
+++ b/backend/api/entity.go
@@ -17,6 +17,7 @@ import (
 // EntityClient is an API client for entities.
 type EntityClient struct {
 	client     *GenericClient
+	storev1    store.Store
 	storev2    storev2.Interface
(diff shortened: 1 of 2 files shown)


echlebek:
Ah rats. I had forgotten about this generic client business. We'll have to write a version of the generic client that works with store/v2 at some point. For now we'll need to use the store/v1 entity-specific stuff.

We'll have to use the entity store methods for Create and Delete as well, or we'll have corresponding breakage there too.

jamesdphillips:
@echlebek updated

PR #3934, opened 2020-07-21
Allow mocking of system information collection

palourde:
## What is this change?

It permits an agent to optionally disable collection of system information. This is particularly relevant in case of the _loadit_ benchmarking tool.

For performance testing.

Added, although we don't need to ship this in 6.0

Nope

Nope

New docs will be required, just depends on whether we include that change within 6.0.

Manually tested.

## Is this change a patch?

Nope

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ other events within the filter namespace.
 - Log a warning when secrets cannot be sent to an agent because mTLS is not
 enabled.
 - Added the is_silenced boolean key to event.Check object.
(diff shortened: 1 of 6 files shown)


palourde:
@echlebek @nikkictl Added a config attribute to use a mocked system information, which I took from a Vagrant CentOS machine.

Let me know if you think it's preferable to keep the initial collection but just disable the periodic refresh

palourde:
@nikkictl Added a condition to skip the periodic refresh if we are mocking the system information

PR #3938, opened 2020-07-22
Optimize keepalived session registration

palourde:


backend/agentd/session.go
```diff
@@ -390,25 +391,86 @@ func (s *Session) Start() (err error) {
...
+		// Indicate to the agent that this entity does not exist
+		meta := corev2.NewObjectMeta(corev3.EntityNotFound, s.cfg.Namespace)
+		watchEvent := &store.WatchEventEntityConfig{
+			Action: store.WatchCreate,
+			Entity: &corev3.EntityConfig{
+				Metadata:    &meta,
+				EntityClass: corev2.EntityAgentClass,
+			},
+		}
+		err = s.bus.Publish(messaging.EntityConfigTopic(s.cfg.Namespace, s.cfg.AgentName), watchEvent)
```
echlebek, line 433: Rather than send this sentinel entity, why not simply send back the entity config that the agent sent in the first place?

agent/agent.go
```diff
@@ -345,6 +363,18 @@ func (a *Agent) connectionManager(ctx context.Context, cancel context.CancelFunc
 		a.connectedMu.Unlock()
 
 		go a.receiveLoop(ctx, cancel, conn)
```
echlebek, line 365: Don't we need to start the `receiveLoop` after the first message from the server has been received?

backend/agentd/session.go
palourde: The entity does not send its full config, we can only rely on these few headers 😬 
https://github.com/sensu/sensu-go/blob/ec65b9ffcd0a2b40727dde13f2023c2950102d93/backend/agentd/agentd.go#L237-L242

agent/agent.go
palourde: As discussed, `receiveLoop` needs to be started in order to receive the entity config update, but we only add the check request handler once we have the entity config

palourde:
Closing in favour of https://github.com/sensu/sensu-go/pull/3940.

PR #3943, opened 2020-07-24
Unsubscribe session from every configured check subscriptions

palourde:


backend/agentd/session.go
```diff
@@ -522,6 +522,9 @@ func (s *Session) stop() {
 			logger.WithError(err).Error("unable to unsubscribe from message bus")
 		}
 	}
+
```
echlebek, line 525: I think you can remove the loop above now

PR #3947, opened 2020-07-28
Output Metric Tags

### LINKED ISSUE #2160: Add support for output_metric_tags
## Expected Behavior

Add a new `output_metric_tags` attribute to the Check resource/specification. Allowed values should provide dot-notation access to any event attribute (e.g. `entity.id`, `entity.organization`, `entity.system.os`, `check.name`, etc). Here's an example of how it might be implemented:

```yaml
type: Check
api_version: core/v2
metadata:
  namespace: default
  name: check_sensu_io
spec: 
  command: "check_http -u https://sensu.io/ -H sensu.io -N",
  runtime_assets: 
  - check_http_v0.1
  publish: true
  interval: 10
  subscriptions: 
  - docker
  output_metric_format: nagios_perfdata
  output_metric_handlers:
  - influxdb
  output_metric_tags: 
    entity: entity.name 
    namespace: entity.namespace
    app: entity.metadata.labels.application_id
```

Bonus points if we could support access to array objects (e.g. `entity.subscriptions[0]` to get the first subscription), or if we supported arbitrary tags (i.e. not matching an event attribute). 

## Current Behavior
No support is provided for collecting tags during metric output extraction. 

## Possible Solution
See description above.

## Context

Currently, when using metric output extraction with the trusty Nagios `check_http` plugin, for a check output that looks like `HTTP OK: HTTP/1.1 301 Moved Permanently - 280 bytes in 0.043 second response time |time=0.043044s;;;0.000000;10.000000 size=280B;;;0`, the corresponding event will have empty tag values as shown in this event excerpt: 

```
{
  "timestamp": 1539194388,
  "entity": {
    "class": "agent",
    "id": "46e492cdd189",
    "...": "..."
  },
  "check": {
    "name": "check_sensu_io",
    "command": "check_http -u https://sensu.io/ -H sensu.io -N",
    "output": "HTTP OK: HTTP/1.1 301 Moved Permanently - 197 bytes in 0.034 second response time |time=0.033662s;;;0.000000;10.000000 size=197B;;;0\n",
    "...": "...",
    "metrics": {
      "handlers": [
        "influxdb"
      ],
      "points": [
        {
          "name": "time",
          "value": 0.033662,
          "timestamp": 1539194388,
          "tags": []
        },
        {
          "name": "size",
          "value": 197,
          "timestamp": 1539194388,
          "tags": []
        }
      ]
    }
  }
}
```

The challenge with this is that when sent to InfluxDB, these appear as very generically named measurements (i.e. "time" and "size"), with no context about which system or service reported the metrics (because there are no tags). Compared with the graphite-style measurement name you would get with Sensu 1.x (e.g. `46e492cdd189.check_sensu_io.time`), this leaves something to be desired, even though simple measurement names with tags would be a more native data representation in InfluxDB.  

## Your Environment

portertech:
## What is this change?

Added support for configurable custom metric tags to be applied to metric points produced by check output metric extraction.

For example, with this change you can now add an instance tag to Nagios perfdata to Prometheus (with an instance tag).

``` yaml
output_metric_format: nagios_perfdata
output_metric_handlers:
  - prometheus_gateway
output_metric_tags:
  - name: instance
    value: '{{ .name }}'
  - name: prometheus_type
    value: gauge
  - name: service
    value: '{{ .labels.service }}'
```

Some metric line formats don't support tags or a check doesn't contextualize metrics with enough or the right context.

Closes #2160

Yes.

Nope.

Nope. I did see optimization opportunity but avoiding doing any refactoring.

https://github.com/sensu/sensu-docs/issues/2707

Deployed a build on my laptop, created checks for each output metric format, and observed configured output metric tags in the produce metric events.

## Is this change a patch?

No.


calebhailey:
Standby while I find the appropriate GIF for this...

calebhailey:
This one hits me right in the feels! 







Closes #2160 

🙌

calebhailey:
Edited @portertech's issue description with the text "Closes #2160" to link this up with the related issue.

echlebek:
@portertech please request a reviewer when the PR is ready :)

raags:
Can something similar be done for statsd? i.e metrics generated from statsd can be enriched with the entity metadata that sensu client has.

agent/transformers/graphite_test.go
```diff
@@ -87,6 +87,47 @@ func TestParseGraphite(t *testing.T) {
...
+				{
+					Name:  "instance",
+					Value: "hostname",
+				},
+			}
+			graphite := ParseGraphite(event)
+			assert.Equal(tc.expectedFormat, graphite)
+		})
+	}
+}
```
nikkictl, line 129: It might be useful to refactor this test so that `outputMetricTags []*types.MetricTag` is a field in the `testCases` struct. This would allow you to dynamically configure `event.Check.OutputMetricTags = tc.outputMetricTags` and add more test cases (ex. multiple tags, empty tags, nil tags, tags with empty name, tags with empty value) rather than testing just the happy path.

nikkictl:
I've enhanced the unit tests for output metric tags and filed/linked an issue in sensu-docs accordingly. I'm going to merge this so we can tag the minor core/v2 bump and persist that to enterprise for testing!

PR #3951, opened 2020-07-29
Downgrade to etcd 3.3.17 and go 1.13.7

amdprophet:

Attemping to fix performance regressions.


.github/workflows/golangci-lint.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       - name: golangci-lint
```
nikkictl, line 18: `go-version: 1.14.6` on L17.. maybe we should update that too? I know its just for the linter but we should stay consistent.

amdprophet: Good catch. Fixed.

portertech:
america-baby-kick

PR #3953, opened 2020-07-30
Refactor/entity config watcher

nikkictl:
## What is this change?

Use channels for entity config watch events.

Perf optimization.

Nah.

Nah.

Nah.

We good.

Updated unit tests and perf run (see slack for snapshot).

## Is this change a patch?

Yup.

### DIFF
--- a/backend/agentd/watcher.go
+++ b/backend/agentd/watcher.go
@@ -7,7 +7,6 @@ import (
 	"github.com/coreos/etcd/clientv3"
 	"github.com/gogo/protobuf/proto"
 	corev3 "github.com/sensu/sensu-go/api/core/v3"
-	"github.com/sensu/sensu-go/backend/messaging"
 	"github.com/sensu/sensu-go/backend/store"
 	etcdstore "github.com/sensu/sensu-go/backend/store/etcd"
 	storev2 "github.com/sensu/sensu-go/backend/store/v2"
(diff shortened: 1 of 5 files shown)


backend/agentd/watcher.go
```diff
@@ -7,67 +7,59 @@ import (
...
+// GetEntityConfigWatcher watches changes to EntityConfig in etcd and publish them
 // over the bus as store.WatchEventEntityConfig
-func EntityConfigWatcher(ctx context.Context, client *clientv3.Client, bus messaging.MessageBus) {
+func GetEntityConfigWatcher(ctx context.Context, client *clientv3.Client) <-chan store.WatchEventEntityConfig {
 	key := etcdstorev2.StoreKey(storev2.ResourceRequest{
 		Context:   ctx,
 		StoreName: new(corev3.EntityConfig).StoreName(),
 	})
 	w := etcdstore.Watch(ctx, client, key, true)
+	ch := make(chan store.WatchEventEntityConfig, 1)
```
ccressent, line 25: My last question: why a buffer size of 1 and not, something bigger (but still reasonable, after defining that).
8? 32?

That would mean decoupling watch event reception and updating the bus even more, in case there is a spike in watch events or a temporary slow down on the bus side?

nikkictl: Argh sorry I merged this too soon! I would be down to test this theory.

PR #3959, opened 2020-08-04
Don't explicitly store liveness leases

echlebek:
## What is this change?

Leases can be obtained from the keys they were stored with. There
is no reason to explicitly store them in etcd.

This change gets us N fewer keys stored in etcd, where N is the
number of agent entities. It also lets us do puts instead of txns,
which should be a minor efficiency win.

Yes

Verification was only done in a performance testing setting. The change will need to be put through the regular QA process to be validated.

## Is this change a patch?

Yes


nikkictl:
Testing round-robin now..

PR #3962, opened 2020-08-05
Release/5.21.1

nikkictl:
Preparation for the 5.21.1 release. Thanks @portertech for getting the changelog ready!


ccressent:
Can't comment on whether or not that's everything we want in this patch release, but what's there looks good and is congruent with the changelog.

PR #3967, opened 2020-08-07
Use a sync.Map for WizardBus

echlebek:
## What is this change?

Use a sync.Map for the map of names to topics. This should reduce
mutex contention in WizardBus significantly.

In large installations, high mutex contention can lead to the application using more CPU than necessary.

No

## Is this change a patch?

Yes

### DIFF
--- a/backend/messaging/wizard_bus.go
+++ b/backend/messaging/wizard_bus.go
@@ -49,10 +49,9 @@ func init() {
 // message types over a single topic, however, as we do not want to introduce
 // a dependency on reflection to determine the type of the received interface{}.
 type WizardBus struct {
-	running  atomic.Value
-	topicsMu sync.RWMutex
-	topics   map[string]*wizardTopic
-	errchan  chan error
+	running atomic.Value
+	topics  sync.Map
+	errchan chan error
 }
 
 // WizardBusConfig configures a WizardBus
@@ -65,7 +64,6 @@ type WizardOption func(*WizardBus) error
 func NewWizardBus(cfg WizardBusConfig, opts ...WizardOption) (*WizardBus, error) {
 	bus := &WizardBus{
 		errchan: make(chan error, 1),
-		topics:  make(map[string]*wizardTopic),
 	}
 	for _, opt := range opts {
 		if err := opt(bus); err != nil {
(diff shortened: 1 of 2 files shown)


backend/messaging/wizard_bus.go
```diff
@@ -87,11 +85,10 @@ func (b *WizardBus) Start() error {
 func (b *WizardBus) Stop() error {
 	b.running.Store(false)
 	close(b.errchan)
-	b.topicsMu.Lock()
-	for _, wTopic := range b.topics {
-		wTopic.Close()
-	}
-	b.topicsMu.Unlock()
+	b.topics.Range(func(_, value interface{}) bool {
+		value.(*wizardTopic).Close()
```
ccressent, line 89: Everywhere else you use a type assertion, shouldn't you check that the assertion is ok? 

For all we know `value` is null or is not even of type `*wizardTopic`. That would probably be a programming error somewhere else in this case, but this check used to be mostly done automatically via the type of `topics` (`map[string]*WizardTopic`). Now `topics` is essentially of type `map[interface{}]interface{}`.

echlebek: Regarding this being a patch:

It doesn't require a minor version bump, but I didn't want to interrupt release goings-on at the time. I could target the 6.0 release branch.

Regarding the type assertion, I believe it's desirable to get a crash here if the data type isn't `*wizardTopic`. There is no other valid datatype that can be stored in the map.

ccressent: I only partially agree: in my opinion, it's desirable to get a crash in those cases _if and only if_ we are sure to always catch these potential crashes before releasing the software. If we miss cases during our various stages of testing, we could end up releasing software that can crash in production rather than gracefully handle the unexpected.

The type system used to enforce the type of the values in the map at compile time, so we were sure to catch a big class of errors. Now it can't do that for us anymore. That's why I'm in favour of checking the result of the assertion.

If you're confident we have enough coverage to always to skip that check, ship it!

echlebek: I see your point, but I think in this case it would be better if the software just didn't function at all, rather than be partially broken. At least then it's obvious to operators! :)

I'll double-check that the code is well-covered by unit tests.

Issue #3948, opened 2020-07-28
Agent does not provide configurable trusted CA for asset retrieval 


cwjohnston:
## Expected Behavior

As a Sensu user or operator employing assets, I can specify a trusted CA certificate for verifying the TLS connection to an asset server.

## Current Behavior

CA certificates must be added to the system trust store.

## Possible Solution

* Use agent `trusted-ca-file` to configure the HTTP client for asset retrieval
* Add an agent flag for configuring trusted CA for asset retrieval specifically
* Add an agent flag for disabling TLS verification with regard to assets
* Add an attribute to assets so as to control TLS verification on a per-asset basis

## Steps to Reproduce (for bugs)
1. Configure a Sensu agent with `trusted-ca-file` valid for asset server and backend server
2. Add asset definition with URL which will be considered valid using `trusted-ca-file`
3. Observe that the asset is not successfully retrieved due to untrusted CA
4. Add CA to system trust store, observe asset is successfully retrieved

## Context

This is a pain-point for some of our existing customers.

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): 5.21.0
* Installation method (packages, binaries, docker etc.):
* Operating System and version (e.g. Ubuntu 14.04):

portertech:
We are going to use the `trusted-ca-file` configuration option for asset retrieval.

nikkictl:
@cwjohnston @portertech is there any reason we shouldn't support this on the backend as well?

PR #3975, opened 2020-08-18
Add trusted ca file for agent assets

nikkictl:

## What is this change?

Allows agent to provide configurable trusted CA for asset retrieval.

Closes https://github.com/sensu/sensu-go/issues/3948

Yas.

I'm unsure why we wouldn't want this for backend-side assets (ex. handlers, mutators) as well: https://github.com/sensu/sensu-go/issues/3948#issuecomment-675577475

Nah.

Nah.

Tested asset retrieval with and without `--trusted-ca-file` configured on the agent.

## Is this change a patch?

Change in functionality, I wouldn't consider this a patch, therefore targeting master.


nikkictl:
TIL about badssl.com 😱

Issue #3956, opened 2020-08-01
asset PATH not injected in Debian 10


davidvilla:
## Expected Behavior

Directory `{PATH_TO_ASSET}/bin` should be injected into the PATH environment variable as [documentation states
](https://docs.sensu.io/sensu-go/latest/reference/assets/#environment-variables-for-asset-paths)

## Current Behavior

Asset path is not in PATH inside the check command context. Variable `{{assetPath "<ASSET_PATH>"}}` is not there either.

## Steps to Reproduce 

Create a check with command "echo $PATH $LD_LIBRARY_PATH" to see the `PATH_TO_ASSET/bin` is not there. Same for LD_LIBRARY_PATH

## Context

Basic installation with asset sensu/monitoring-plugins, by means; `sensuctl asset add sensu/monitoring-plugins`

## Your Environment
* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): 5.21.0-14262 amd64
* Installation method: debian packages at packagecloud.io
* Operating System and version: Debian GNU/Linux 10 (buster)

![image](https://user-images.githubusercontent.com/838459/89108342-f102be00-d437-11ea-8111-62f39078d662.png)

palourde:
Hi @davidvilla,

Could you make sure the asset is actually getting installed on the agent? You might see something like `entity not filtered from any asset builds, not installing asset` in the agent logs, which would mean that none of the filters applied to the agent's entity. Turning the logs to the `debug` level will also provide more information about that.

Alternatively, you could also try to execute one of the binary in `sensu/monitoring-plugins` to make sure it's actually getting installed.

davidvilla:
Hi:

Yes, the assets are installed and I can run them from console.

Cheers

palourde:
Could you provide us some of your configuration to help us investigate this issue?

Here's various information that could be useful:

```
$ sensuctl entity info $ENTITY_NAME --format yaml

$ sensuctl check info $CHECK_INFO --format yaml

$ sensuctl event info $ENTITY_NAME $CHECK_INFO --format yaml

$ sensu-backend version

# On the related agent node
$ sensu-agent version
```

Thanks!

davidvilla:
Hi:

```
$ sensuctl entity info amy --format yaml
type: Entity
api_version: core/v2
metadata:
  name: amy
  namespace: default
spec:
  deregister: false
  deregistration: {}
  entity_class: agent
  last_seen: 1597219614
  redact:
  - password
[...]

PR #3977, opened 2020-08-19
Improve UX around runtime assets

palourde:

## What is this change?

It adds a warning log in the backend for assets that were configured as runtime_assets but do not exist. Previously, it would just silently ignore those, which would make troubleshooting really hard, since even the agent wouldn't know about any assets that were discarded.

I also adjusted a debug log in the agent depending on whether we have assets or not to install.

Help users troubleshoot issues like https://github.com/sensu/sensu-go/issues/3956.

Added

Nope

Nope

Nope

Manually verified, I don't think we need additional tests.

## Is this change a patch?

Nope

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic
 Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Added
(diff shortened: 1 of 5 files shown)


nikkictl:
We could maybe target the release branch since this is just a logging improvement!

Issue #3981, opened 2020-08-25
Inconsistent UX for resource types in sensuctl commands (ex. dump,prune,list)


nikkictl:
The `sensuctl dump` command has had a few iterations in UX since its inception. While more use cases are accounted for and the UX has certainly improved, we still have some inconsistencies with what we consider valid ways to represent a resource type. In this issue, I will provide a brief history of the design, overview of the current behavior, and proposed solution to the problems presented.

## History

Originally, the command was a wrapper around `sensuctl list` where only short name resource types were supported. For example `sensuctl dump check,entity,role-binding,api-key` would basically run `sensuctl check list && sensuctl entity list && sensuctl role-binding list && sensuctl api-key list`. This approach limited users in how expressive they could be to define a resource type, and had other developmental issues such as hard coded resource names.

Some improvements to this design included @echlebek's approach to generically and dynamically generate resource types by their fully qualified API names. For example `sensuctl dump core/v2.Check` was equivalent to the short name `check` and directly correlated to the Go API package. This was especially advantageous to solidifying our resource type system, but while the approach intended to be backwards compatible, it actually introduced unintentional inconsistencies, for example, support for RBAC short names (ex. `sensuctl dump checks,entities,rolebindings,apikeys`) on top of `sensuctl` command nouns (ex. `sensuctl dump check,entity,role-binding,api-key`).

## Current Behavior
After some refactors to the types system and resource interfaces, `sensuctl dump` exhibited the following behaviors, which have multiple ways to express a resource type inconsistently across resources.

#### Dump checks
✅ `sensuctl dump check`
✅ `sensuctl dump checks`
✅ `sensuctl dump core/v2.Check`

#### Dump entities
✅ `sensuctl dump entity`
✅ `sensuctl dump entities`
✅ `sensuctl dump core/v2.Entity`

#### Dump role bindings
❌ `sensuctl dump rolebinding`
✅ `sensuctl dump rolebindings`
✅ `sensuctl dump role-binding`
✅ `sensuctl dump role-bindings`
✅ `sensuctl dump core/v2.RoleBinding`

#### Dump api keys
❌ `sensuctl dump apikey`
✅ `sensuctl dump apikeys`
❌ `sensuctl dump api-key`
❌ `sensuctl dump api-keys`
❌ `sensuctl dump core/v2.ApiKey`
✅ `sensuctl dump core/v2.APIKey`

[...]

PR #3982, opened 2020-08-25
Add validation around resource types args in dump

nikkictl:

## What is this change?

Adds more validation around acceptable resource types in `sensuctl dump`. Previously, the command would try to resolve providers (ex. `secrets/v1.VaultProvider`, `authentication/v2.LDAP`) but due to the nature of these lifted resource types and our generic API design, they are not compatible.

`sensuctl dump` now only accepts resource type arguments that are displayed in `sensuctl describe-type all`. To display all secrets providers, one would run `sensuctl dump secrets/v1.Provider`.

Closes https://github.com/sensu/sensu-enterprise-go/issues/1216

Yas.

Nah.

Yes, there were actually a few ways to solve this problem, and I discovered more bugs when attempting to resolve specific provider types such as `secrets.v1/VaultProvider`. For that reason, I thought it was best to limit the acceptable resource type arguments to those listed in `sensuctl describe-type all`. See https://github.com/sensu/sensu-go/issues/3981 for full description.

Docs will likely need to be modified if this approach is selected.

Added unit tests.

```
$ sensuctl dump secrets/v1.Env
Error: invalid resource type: secrets/v1.Env
$ sensuctl dump secrets/v1.Provider
type: Env
api_version: secrets/v1
metadata:
  name: env
spec: {}
```

## Is this change a patch?

Yup! Targeting release branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,13 @@ and this project adheres to [Semantic
 Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+
 ### Added
 - A warning is now logged when a runtime asset was requested but does not exist.
 
 ### Fixed
 - Failed check events now get written to the event log file.
+- Added more validation around acceptable resource types in `sensuctl dump`.
 
 ## [6.0.0] - 2020-08-04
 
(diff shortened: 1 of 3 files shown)


nikkictl:
It has been confirmed that this approach would be considered a breaking change https://github.com/sensu/sensu-go/issues/3981#issuecomment-681015034. I will revisit another approach to support lifted providers and close out this WIP PR without merge.

PR #3989, opened 2020-08-31
Support generic query params in sensuctl dump

nikkictl:

## What is this change?

Adds backwards compatible support for a `types` query parameter to be added to the URI to list resources.

Required by https://github.com/sensu/sensu-enterprise-go/issues/1216

No, purely internal. Changelog entry will be noted in enterprise.

Nah.

Nah.

We good.

See enterprise RP.

## Is this change a patch?

No, it supports a minor feature.


nikkictl:
@echlebek I was hoping to avoid refactoring the entire `List` rest client, but if you would be amenable to implementing a new `List` function that accepts query params, I think that would avoid that friction.

Issue #3998, opened 2020-09-08
HTTP PATCH


palourde:
See [the proposal](https://docs.google.com/document/d/1IH-SdBDMN77Xx6OQlLW0kdEF6h8fCX6snw-fn3ypFVw/)

We must support PATCH requests, with an optional ETAG.

- [x] Implement a _Patcher_, probably based on https://github.com/evanphx/json-patch.
- [x] Add a Patch method to the core/v2 generic store
- [x] Add a Patch method to the generic handler
- [ ] Add support for PATCH to all configuration resources
- [x] Support the `Content-Type: application/merge-patch+json` header and fallback to this value; return an error if another value is passed for now (we might want to support `application/json-patch+json` or other strategies later)
- [x] Add a free-standing Etag() function to the store package
- [x] Set the Etag header for all GET requests in APId that resolve an individual resource.

portertech:
We need to guard resource name and namespace alteration. A user should not be able to change the resource name and namespace with PATCH.

palourde:
Test cases:

- A label can be added
- An array is replaced
- Boolean/integer values not specified in the patch are not overridden with the default value
- An etag is returned on 200 OK
- A 204 No Content is returned when no modification has been performed
- Etags are returned for all GET requests
- A namespace cannot be altered
- A resource name cannot be altered
- A 412 Precondition Failed is returned if the etag in If-Match does not match the stored resource

palourde:
Resources to test:

- [ ] apikeys
- [ ] checks
- [ ] clusterrolebindings
- [ ] clusterroles
- [ ] filters
- [ ] hooks
- [ ] mutators
- [ ] namespaces
- [ ] rolebindings
- [ ] roles

Resources to implement
- [ ] entities
- [ ] silenced
- [ ] users

PR #4015, opened 2020-09-21
Add support for the PATCH method on most config resources

palourde:
## What is this change?

This PR adds support for the PATCH method on most configuration resources, with two notable exceptions; silenced entries & users.

In summary:

- The actual Patch mechanism lives in `backend/store/patch`. For now only merge patches are supported but we could easily extend the support to other types.
- There's a new generic handler for patching resources (`backend/apid/handlers/patch.go`), which supports both v2 & v3 resources. It's responsible for initializing the _patcher_, the optional _etag conditions_ and validating the patch to make sure the name & namespaces are not altered.
- We have two new methods for patching in both the v1 & v2 stores. These two work pretty much in a similar fashion; we get the current stored resource, determine if we match the etag conditions (either `If-Match` of `If-None-Match`), we apply the patch on it, validate it and then update it while making sure the stored resource wasn't modified in the mean time.
- The function `RespondWith` now automatically adds the etag header if we have a response body. This etag is determined with the help of the hashstructure lib.
- I made a refactoring around our handling of etcd transactions, because in a lot of cases we were not actually able to determine if and why a condition failed. In order to facilitate that task, I've added the new package `backend/store/etcd/kvc` (key-value client), which also groups some logic we had living in both the v1 & v2 stores.

Closes https://github.com/sensu/sensu-go/issues/3998

Yep

Nope

A lot 😆 

~I'll open a new issue in sensu-docs~

Opened https://github.com/sensu/sensu-docs/issues/2755

I deployed my changes on staging, and also manually tested using these notes: https://github.com/sensu/sensu-go/issues/3998#issuecomment-690560748.

~My next task is to automate these tests in our QA crucible.~ 
Opened https://github.com/sensu/sensu-go-qa-crucible/pull/152

## Is this change a patch?

Nope


echlebek (commented):
I hit an interesting snag while testing your PR.

Although there is a default fallback for content type when it is not set, `curl(1)` sets this header implicitly.

```
[eric@da2 sensu-env]$ curl -i -X PATCH -H 'Authorization: Key 4ed7191b-c939-42ba-a445-363331702ae5' http://localhost:8080/api/core/v2/namespaces/default/checks/ping-localhost -d @patch.json
HTTP/1.1 400 Bad Request
X-Content-Type-Options: nosniff
Date: Mon, 21 Sep 2020 22:56:20 GMT
Content-Length: 89
Content-Type: text/plain; charset=utf-8

{"message":"invalid Content-Type header: \"application/x-www-form-urlencoded\"","code":1}
```

I think this is pretty likely for someone to try. While I don't think we should allow `application/x-www-form-urlencoded` as a Content-Type, maybe we should print out the list of content types we _do_ accept, so that users who hit this snag can get over the issue easily.

echlebek (commented):
One thing I find a bit odd about the implementation is that I have to put quotes around the Etag.

I have to do:

```
-H 'Content-Type: "15585225553805949644"'
```

instead of

```
-H 'Content-Type: 15585225553805949644'
```

echlebek:
Finally, the good news! I tested the PR with `curl` and everything seems to work the way it should. Excellent work!

I think we'll definitely need some tests for the qa-crucible project around this feature. Let me know if you want some help writing them.

palourde:
@echlebek re: etag, in your example you used the `Content-Type` header, but I assume you meant the `If-Match`/`If-None-Match` headers?

Turns out the spec indicates that the actual etag value should be placed between double quotes: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match#Directives.

echlebek:
@palourde yep, I meant "If-Match".

palourde:
@echlebek Alright, I updated the etags to use an hex representation. Also, the error message includes accepted content types (one for now) if a "bad" one was provided.

For the test cases, I already noted some tests I'd like to see in https://github.com/sensu/sensu-go/issues/3998#issuecomment-690560748 but feel free to add some!

echlebek (approved):
Great work!

backend/store/etcd/kvc/client.go
```diff
@@ -0,0 +1,54 @@
...
+
+const (
+	// EtcdRoot is the root of all sensu storage.
+	EtcdRoot = "/sensu.io"
+
+	// EtcdInitialDelay is 100 ms.
+	EtcdInitialDelay = time.Millisecond * 100
+
+	// NamespacesPathPrefix is the path to namespaces in etcds
+	NamespacesPathPrefix = "namespaces"
```
ccressent, line 20: For `NamespacePathPrefix` and `EtcdRoot`, shouldn't those default come from the package that define them (`backend/store/etcd`, though the former would have to be exported)? That way, they can never be inconsistent.

ccressent (approved):
Amazing work!

See my (minor) comments; approved assuming the additional QA tests you're working on do not uncover anything badly broken.

backend/store/etcd/kvc/client.go
palourde: The "problem" is that both the v1 & v2 etcd stores need those, and if we moved them into one of these packages, we would probably have an import cycle hell. WDYT?

Issue #3702, opened 2020-04-16
sensuctl prune multi file handling is busted


jspaleta:
## Expected Behavior
prune multi-file handling should work 

## Current Behavior
using -r  or multple calls to -f  result if prune cleaning out all resources intead of leaving the named resources from the files in place

## Steps to Reproduce (for bugs)

1.  using sensu-community/monitoring-plugins repo
2.  sensuctl create -r -f system/linux/cpu system/linux/disk    (creates 20 checks)
3. sensuctl prune checks -f system/linux/cpu/check-cpu.yaml  -f system/linux/disk/check-disk-usage.yaml     (all checks pruned
4.  sensuctl create -r -f system/linux/cpu system/linux/disk
5. sensuctl prune checks -r -f system/linux/cpu/   ( all checks pruned)

jspaleta:
Shell catting files still works as expected
`cat system/linux/cpu/check-cpu.yaml system/linux/disk/check-disk-usage.yaml | sensuctl prune checks`

Leaves 2 checks behind as expected

sensu-discourse:
This issue has been mentioned on **Sensu Community**. There might be relevant details there:

https://discourse.sensu.io/t/live-webinar-on-april-23-getting-started-with-assets/1696/2

sensu-discourse:
This issue has been mentioned on **Sensu Community**. There might be relevant details there:

https://discourse.sensu.io/t/whats-new-in-sensu-april-16-2020/1687/2

nikkictl:
See https://github.com/sensu/sensu-go/issues/3845 for additional context.

raddessi:
Are there any plans to fix this?

calebhailey:
@raddessi yes! We don't have an ETA as of yet, but it's becoming a priority so I would expect to see some progress here soon. 

In the interim, can you share any details about how this is impacting you? Are you implementing `sensuctl prune` as part of your CI/CD pipeline by any chance? 

Thanks in advance for your feedback!

raddessi:
Great news :) I tried to use prune with multiple files but right now am only using it with a single file. I tried to implement it with multiple files as the management method for my setup but found this bug so I stopped moving in that direction until this is resolved.

PR #4017, opened 2020-09-21
Change the OoO for processing resource input

nikkictl:

## What is this change?

Changes the order of operations for processing resource input. When provided multiple files, urls, or directories, all resources should be read first before the processing command (`create` or `prune`) take place.

Required by https://github.com/sensu/sensu-enterprise-go/pull/1284

Closes https://github.com/sensu/sensu-go/issues/3702

Yes, in enterprise.

Nah.

Nah.

We good.

```
$ sensuctl create -f check1.yaml -f check2.yaml -f check3.yaml 
$ sensuctl check list
   Name    Command    Interval   Cron   Timeout   TTL   Subscriptions   Handlers   Assets   Hooks   Publish?   Stdin?   Metric Format   Metric Handlers  
 ──────── ────────── ────────── ────── ───────── ───── ─────────────── ────────── ──────── ─────── ────────── ──────── ─────────────── ───────────────── 
  check1   echo foo         10                0     0   metrics                                     false      false                                     
  check2   echo foo         10                0     0   metrics                                     false      false                                     
  check3   echo foo         10                0     0   metrics                                     false      false                                     
$ sensuctl prune checks -f check1.yaml -f check2.yaml
[
  {
    "type": "CheckConfig",
    "api_version": "core/v2",
    "name": "check3",
    "namespace": "default",
    "labels": {
      "sensu.io/managed_by": "sensuctl"
    },
    "created_by": "admin"
  }
]

$ sensuctl check list
   Name    Command    Interval   Cron   Timeout   TTL   Subscriptions   Handlers   Assets   Hooks   Publish?   Stdin?   Metric Format   Metric Handlers  
 ──────── ────────── ────────── ────── ───────── ───── ─────────────── ────────── ──────── ─────── ────────── ──────── ─────────────── ───────────────── 
  check1   echo foo         10                0     0   metrics                                     false      false                                     
  check2   echo foo         10                0     0   metrics                                     false      false                                     
```

## Is this change a patch?

It's a bugfix but it somewhat reworks things so I'm targeting master!

### DIFF
--- a/cli/resource/process.go
+++ b/cli/resource/process.go
@@ -28,22 +28,49 @@ type httpDirectory struct {
 }
 
 // Process processes the input.
-func Process(cli *cli.SensuCli, client *http.Client, input string, recurse bool, processor Processor) error {
-	urly, err := url.Parse(input)
-	if err != nil {
+func Process(cli *cli.SensuCli, client *http.Client, inputs []string, recurse bool, processor Processor) error {
+	var resources []*types.Wrapper
+	for _, input := range inputs {
+		res, err := process(client, input, recurse)
+		if err != nil {
+			return err
+		}
+		resources = append(resources, res...)
+	}
+	if err := Validate(resources, cli.Config.Namespace()); err != nil {
 		return err
 	}
+	return processor.Process(cli.Client, resources)
+}
+
+func process(client *http.Client, input string, recurse bool) ([]*types.Wrapper, error) {
+	var resources []*types.Wrapper
+	urly, err := url.Parse(input)
+	if err != nil {
+		return resources, err
+	}
+	var res []*types.Wrapper
 	if urly.Scheme == "" || len(urly.Scheme) == 1 {
 		// We are dealing with a file path
-		return ProcessFile(cli, input, recurse, processor)
+		res, err = ProcessFile(input, recurse)
+		if err != nil {
+			return resources, err
+		}
+	} else {
+		res, err = ProcessURL(client, urly, input, recurse)
+		if err != nil {
+			return resources, err
+		}
 	}
-	return ProcessURL(cli, client, urly, input, recurse, processor)
+	resources = append(resources, res...)
+	return resources, nil
 }
 
 // ProcessFile processes a file.
-func ProcessFile(cli *cli.SensuCli, input string, recurse bool, processor Processor) error {
+func ProcessFile(input string, recurse bool) ([]*types.Wrapper, error) {
+	var resources []*types.Wrapper
 	var tld = true
-	return filepath.Walk(input, func(path string, info os.FileInfo, err error) error {
+	err := filepath.Walk(input, func(path string, info os.FileInfo, err error) error {
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


cli/resource/process.go
```diff
@@ -28,22 +28,49 @@ type httpDirectory struct {
...
+	} else {
+		res, err = ProcessURL(client, urly, input, recurse)
+		if err != nil {
+			return resources, err
+		}
 	}
-	return ProcessURL(cli, client, urly, input, recurse, processor)
+	resources = append(resources, res...)
+	return resources, nil
 }
```
ccressent, line 67: Aren't `resources` and `res` redundant here? I don't understand why `res` is introduced. It looks to me like we could have used `resources` everywhere.

cli/resource/process.go
```diff
@@ -57,59 +84,58 @@ func ProcessFile(cli *cli.SensuCli, input string, recurse bool, processor Proces
 		if err != nil {
 			return err
 		}
-		resources, err := Parse(f)
+		resources, err = Parse(f)
```
ccressent, line 87: Is this not overwriting all the previous Parse, ultimately only keeping the last one?!
Unfortunately the test for `ProcessFile()` won't catch that because it uses a single file, in a single directory.

cli/resource/process.go
```diff
@@ -57,59 +84,58 @@ func ProcessFile(cli *cli.SensuCli, input string, recurse bool, processor Proces
...
-			if err := Process(cli, client, filepath.Join(input, file), recurse, processor); err != nil {
-				return err
+			res, err := process(client, filepath.Join(input, file), recurse)
+			if err != nil {
+				return resources, err
 			}
+			resources = append(resources, res...)
 		}
+	} else {
+		res, err := Parse(resp.Body)
```
ccressent, line 132: Why not `resources, err := Parse()` and then not calling `append()` below?

ccressent (commented):
I'm fairly certain that this change breaks directory processing :(
I haven't tested though.

cli/resource/process.go
echlebek: Yes, I think this would break recursive processing.

cli/resource/process.go
nikkictl: `res` is meant to capture the resources returned by `ProcessFile` or `ProcessURL`.. I introduced the variable to limit the # of lines of code required, ex. I only have to call `resources = append(resources, res...)` once.

cli/resource/process.go
nikkictl: Confirmed, good catch.

cli/resource/process.go
nikkictl: You're right, append is only needed in the previous if clause, not in this one!

cli/resource/process.go
```diff
@@ -57,59 +84,58 @@ func ProcessFile(cli *cli.SensuCli, input string, recurse bool, processor Proces
...
-			if err := Process(cli, client, filepath.Join(input, file), recurse, processor); err != nil {
-				return err
+			res, err := process(client, filepath.Join(input, file), recurse)
+			if err != nil {
+				return resources, err
 			}
+			resources = append(resources, res...)
+		}
+	} else {
+		resources, err := Parse(resp.Body)
```
echlebek, line 133: This is an ineffectual assignment, since it will create a new variable that shadows `resources` in the outer scope. 😬

nikkictl: Shall I revert the changes made per https://github.com/sensu/sensu-go/pull/4017#discussion_r493655336?

ccressent: Sorry Nikki, my comment was confusing; either revert or don't shadow `resources`: `resources, err = Parse()`?

nikkictl: I see the oops now hehe

echlebek (approved):
```
--    .-""-.
   ) (     )
  (   )   (
     /     )
    (_    _)                     0_,-.__
      (_  )_                     |_.-._/
       (    )                    |_--..\
        (__)                     |__--_/
     |''   ``\                   |
     |        \                  |      /b.
     |         \  ,,,---===?A`\  |  ,==y'
   ___,,,,,---==""\        |M] \ | ;|\ |>
           _   _   \   ___,|H,,---==""""bno,
    o  O  (_) (_)   \ /          _     AWAW/
                     /         _(+)_  dMM/
      \@_,,,,,,---=="   \      \\|//  MW/
--''''"                         ===  d/
                                    //
                                    ,'__________________________
   \    \    \     \               ,/~~~~~~~~~~~~~~~~~~~~~~~~~~~
                         _____    ,'  ~~~   .-""-.~~~~~~  .-""-.
      .-""-.           ///==---   /`-._ ..-'      -.__..-'
            `-.__..-' =====\\\\\\ V/  .---\.
 PGMG                 ~~~~~~~~~~~~, _',--/_.\  .-""-.
                            .-""-.___` --  \|         -.__..-
```

SHIPPIT

PR #4023, opened 2020-09-25
[GraphQL] Where possible, avoid scanning all events

jamesdphillips:
Previously, determining an entity's status or fetching it's relevant events would
require retrieving all events in the entity's namespace. In cluster's with a great
many events this was lead to significantly negative performance impact.

---


echlebek (approved):
Awesome! I'll re-test with this on SPDC once merged.

PR #4033, opened 2020-10-06
Expose build info via prometheus metrics

palourde:
## What is this change?

It adds the Sensu Go build info to the prometheus metrics, so it can be easily exposed via Grafana.

Closes https://github.com/sensu/sensu-perf/issues/23

Added

Nope

Nope

Nope

Manually deployed a build from this branch to the SPDC.

## Is this change a patch?

Nope


version/version.go
```diff
@@ -33,8 +35,21 @@ var (
...
 	// (e.g. go1.14.2)
 	GoVersion string = runtime.Version()
+
+	promBuildInfo = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "sensu_go_build_info",
+			Help: "Sensu Go build information",
+		},
+		[]string{"version", "builddate", "buildsha", "goversion"},
+	).WithLabelValues(Version, BuildDate, BuildSHA, GoVersion)
```
ccressent, line 45: I'm not sure that the build date is that valuable or useful; in fact there are situations where it's actively harmful.

Nothing tells you that the clock of the building machine was correctly set, or that it didn't massively change between 2 builds, making the comparison of build date/time unreliable in general.

More importantly, this single nugget of information makes the build not easily reproducible: if you build, checksum the resulting binary, build and checksum again, you'll get 2 different checksums because time is made an implicit build input. This causes problems for some systems that try to deal with software in a reproducible fashion, e.g. Nix/NixOS (the sensu-go "package" there rebuilds from source and scrubs the build date for that reason).

TL;DR: I think we should drop `BuildDate`.

version/version.go
```diff
@@ -33,8 +35,21 @@ var (
...
 )
 
+func init() {
+	prometheus.MustRegister(promBuildInfo)
+	promBuildInfo.Set(1)
```
ccressent, line 50: I don't understand why `.Set(1)` in necessary; what is it for?

ccressent (approved):

version/version.go
palourde: It's required so the `sensu_go_build_info` metric has a value (it could have been any integer value really), otherwise it will not show up in the metrics endpoint

PR #4034, opened 2020-10-06
Don't use leases for silenced entries

### LINKED ISSUE #4014: Silences fail to resolve after cluster restart
## Expected Behavior
After a silence is applied, it should resolve after its period, even through cluster restarts.

## Current Behavior
Silences seem to persist past cluster restarts.

## Possible Solution
Unknown.

1. Create silences for pretty much everything, and do a rolling upgrade of Sensu.
2. Observe that silences don't resolve.

## Context
Reported by a customer.

echlebek:
## What is this change?

This commit changes how silences work in Sensu.

Before, silences were governed by etcd leases. The idea was that a
silenced entry would be created and leased to expire, if it had the
Expire field set.

This was fine in theory, and often in practice, but leases have a subtle
characteristic - they are reset when etcd holds an election, if they are
less than 5 minutes long.

This can cause havoc when Sensu is restarted. For a feature like
silences, it is not reliable enough.

Now, silences store their expiry information directly in their own data
structure. When silences are listed, expired results are simply
discarded from the result set, and asynchronously garbage collected.
This should make silences perfectly reliable, as they will no longer
rely on any state in etcd or Sensu, but rather be self-contained in
their own data structure.

The core/v2.Silenced datatype has a new field, called ExpireAt, which is
the time in seconds since the Unix epoch. When this field is set, Sensu
will use it to determine if the entry has expired. The field will be
automatically set by the store when a new silence is created, and will
therefore be API-compatible with the previous implementation.

If the ExpireAt field is not set on a silenced entry read from the
database, then Sensu will assume that the entry pre-dates the new system,
and fall back to requesting its lease.

Fixes #4014 

Yes

This change creates a new module for api/core/v2. An alpha tag, `api/core/v2.4.0-alpha`, has been created.

The existing integration test suite for silences is quite robust. I believe that the tests show the reworked feature is working as intended.

## Is this change a patch?

Yes


palourde (changes requested):
Looks like staging found a bug with the current implementation:

https://github.com/sensu/sensu-go-qa-crucible/blob/e2f4b34a6d940450357c4602c22ef82cd033a3f4/spec/backend/silencing_spec.rb#L39-L57

TL;DR If you create the following silenced entry, it will never automatically expire: `sensuctl silenced create --reason expire --expire 3 --check test-check`

echlebek:
@palourde I did previously test this use case and it succeeded. I wonder if this is an issue with the enterprise build not having the `core/v2` that contains the new `ExpireAt` field?

palourde:
@echlebek Huh, I was able to reproduce this morning on my local sensu-go setup but looks like it's not longer the case, so I probably had the wrong version running! :shipit:

palourde (approved):

echlebek:
Given the complexity of this change, I'm going to merge it into main instead of the release branch.

PR #4035, opened 2020-10-06
Document web UI separate from backend as of 6.0

hillaryfraley:
## What is this change?

Update the readme to explain that for users who will build from source, the web UI is separate from the backend as of 6.0 and add link to web UI-specific installation instructions.

Document separate web UI repo for OSS and point users to web UI installation instructions.

I don't think so. I think this change would have been properly entered on the changelog for 6.0 and this readme update does not need a separate entry.

No

No

This is already documented in the 6.0 release notes as a breaking change.

I confirmed w/James that we should point users to the web UI-specific instructions.

## Is this change a patch?

No

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,8 @@ to get started is with the official Docker image, sensu/sensu.
 
 See the [installation documentation](https://docs.sensu.io/sensu-go/latest/installation/install-sensu/) to get started.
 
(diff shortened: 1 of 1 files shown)


jamesdphillips (approved):
:tada:

PR #4038, opened 2020-10-07
[backend/api] fill created_by field

jamesdphillips:
Updates the `backend/api` package so that the `ObjectMeta`'s `CreatedBy` field is filled on create & update. 

Fixes sensu/sensu-enterprise-go#1318.

---

### DIFF
--- a/backend/api/generic.go
+++ b/backend/api/generic.go
@@ -56,6 +56,7 @@ func (g *GenericClient) Create(ctx context.Context, value corev2.Resource) error
 	if err := authorize(ctx, g.Auth, attrs); err != nil {
 		return err
 	}
+	setCreatedBy(ctx, value)
 	return g.Store.CreateResource(ctx, value)
 }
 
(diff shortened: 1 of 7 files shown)


backend/api/generic.go
```diff
@@ -95,6 +96,7 @@ func (g *GenericClient) Update(ctx context.Context, value corev2.Resource) error
 	if err := authorize(ctx, g.Auth, attrs); err != nil {
 		return err
 	}
+	setCreatedBy(ctx, value)
```
jamesdphillips, line 99: It seems a bit odd to overwrite the `CreatedBy` field on update, but for now, it at least mirrors the `handler` package.

https://github.com/sensu/sensu-go/blob/master/backend/apid/handlers/update.go#L31-L36
```golang
	meta := resource.GetObjectMeta()
	if claims := jwt.GetClaimsFromContext(r.Context()); claims != nil {
		meta.CreatedBy = claims.StandardClaims.Subject
		resource.SetObjectMeta(meta)
	}
```

echlebek (approved):
Makes sense!

PR #4053, opened 2020-10-13
Backport js vm fixes from 6.x to 5.21

### LINKED ISSUE #4049: Backport the Javascript VM cache bug fix (#4001)
Backport [this fix](https://github.com/sensu/sensu-go/pull/4001) to a release/5.21.3 branch.

echlebek:
## What is this change?

Fixes #4049 

I couldn't preserve the original commit as it was based on newer code. So instead, I simply copied the JS code from 6.0, and this is the result. The 6.0 `js` package changes perserved interfaces, so things should just work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 - Close the response body when done reading from it while downloading assets.
+- Fixed a bug where event filter or asset filter execution could cause a crash.
 
 ## [5.21.2] - 2020-08-31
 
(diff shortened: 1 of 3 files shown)


ccressent (approved):
Looks like a good port to me; didn't test anything though. I'm wondering what type of test coverage we have for the original bug and for JS in general. Are you confident that passing tests are a good indicator of no regression?

Don't forget to add a CHANGELOG entry for this.

echlebek:
I think the original bug fix actually didn't include a test that shows it works. My bad.

The fix was to stop conflating the asset set with the key of the VM. I therefore added a key parameter to functions that need it, which can never be nil, and this fixed the nil pointer dereference.

I'm confident this won't cause any regressions as the entire package is now in line with 6.x.

PR #4065, opened 2020-10-21
Improve formatting of dates & durations in sensuctl silenced

palourde:
## What is this change?

It tweaks the date & duration formats used in `sensuctl silenced [info|list]`.

- The `Begin` date now use the RFC3339 format, so there's no confusion around the year (e.g. `2020-10-13T09:40:09-04:00`)
- The Expire value will now vary depending on a number of factors:
	- If there's no expiration, we will display `-1`
	- If the silenced entry has not yet begun, we display an expiration date in RFC3339 format
	- If the silenced entry has begun, we display the duration (e.g. `1m30s`)

Fixes some bad UX found while investigating a problem around silenced entries.

Nope

We might have to adjust the docs, I'll open an issue if the proposed changes are accepted

Manually verified + unit test.

## Is this change a patch?

Nah


echlebek (approved):
Nice work! Sorry I let the review slip on this.

Issue #4062, opened 2020-10-19
Entity can be listed, but not queried directly


echlebek:
The following entity was created with `sensuctl create`. It can be seen in the output of `sensuctl entity list`, but cannot be queried with `sensuctl entity info`.

```
type: Entity
metadata:
  name: sally
  namespace: default
spec:
  entity_class: agent
```

PR #4067, opened 2020-10-21
Fix entity creation via sensuctl create

palourde:

## What is this change?

It fixes entity creation when done via `sensuctl create`; only the entity_config key would be created, which would prevent entities from being queried.

Now we first try to create an entity when we have an agent entity, and if it already exists then we update the entity config.

Fixes https://github.com/sensu/sensu-go/issues/4062

Added

Nope

Nope

Nope, bug fix

Unit test + manual testing, also ran the QA crucible tests to make sure it didn't break anything.

## Is this change a patch?

Yep

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 - An RBAC rule with the `update` permission now properly authorizes PATCH
 requests.
 - eventd errors now include additional context for debugging.
+- Entities are now properly created using `sensuctl create`.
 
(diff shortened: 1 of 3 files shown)


echlebek (commented):
I'm wondering if this functionality should be incorporated into the store's `CreateOrUpdate` method? Seems like it could also occur through the graphql layer.

palourde:
@echlebek Good point, probably a safe bet to move the logic there!

echlebek (approved):
I was wroooooong

palourde:
TIL we don't support entity creation via the Web UI!

PR #4068, opened 2020-10-21
Fix a bug introduced by #4034

echlebek:
## What is this change?

This commit fixes a bug in the silenced HTTP API where expired silenced
entries can appear, even though they should not.

Fixes an issue introduced in #4034.

No, at the time of this PR the bug has not been released.

Not required.

I created a silenced entry and queried it. The regression is no longer present.

## Is this change a patch?

Nein


palourde (approved):
Successfully tested! 👏

Issue #4073, opened 2020-10-27
Possible race condition in the JS VM cache


palourde:
## Expected Behavior
It should be possible to re-use the same VM cache with concurrent processes without any issue.

## Current Behavior
We receive a fatal error:

```
fatal error: sync: unlock of unlocked mutex

goroutine 64 [running]:
runtime.throw(0x4d27ca6, 0x1e)
	/usr/local/go/src/runtime/panic.go:1116 +0x72 fp=0xc0005ba6d8 sp=0xc0005ba6a8 pc=0x4037d72
sync.throw(0x4d27ca6, 0x1e)
	/usr/local/go/src/runtime/panic.go:1102 +0x35 fp=0xc0005ba6f8 sp=0xc0005ba6d8 pc=0x4037cf5
sync.(*Mutex).unlockSlow(0xc00203f588, 0xffffffff)
	/usr/local/go/src/sync/mutex.go:196 +0xd6 fp=0xc0005ba720 sp=0xc0005ba6f8 pc=0x4073fa6
sync.(*Mutex).Unlock(...)
	/usr/local/go/src/sync/mutex.go:190
github.com/sensu/sensu-go/js.(*vmCache).Dispose(0xc0005a7cb0, 0x0, 0x0)
	/Users/splourde/go/src/github.com/sensu/sensu-go/js/vm_cache.go:96 +0xa3 fp=0xc0005ba788 sp=0xc0005ba720 pc=0x48b2d53
github.com/sensu/sensu-go/js.releaseOttoVM(0x0, 0x0)
	/Users/splourde/go/src/github.com/sensu/sensu-go/js/js.go:79 +0x4b fp=0xc0005ba7b0 sp=0xc0005ba788 pc=0x48b16ab
github.com/sensu/sensu-go/js.Evaluate(0xc0008900e0, 0x1d, 0x4baede0, 0xc0005ba938, 0x0, 0x0, 0xc0005ba900, 0x0, 0x0)
	/Users/splourde/go/src/github.com/sensu/sensu-go/js/js.go:154 +0x2a5 fp=0xc0005ba8c8 sp=0xc0005ba7b0 pc=0x48b1e95
github.com/sensu/sensu-go/asset.(*filteredManager).isFiltered(0xc000279400, 0xc0004381c0, 0x4d0e07f, 0x6, 0xc0005bb318)
[...]

PR #4074, opened 2020-10-27
Fix race condition in the JS VM cache

palourde:

## What is this change?

It fixes a race condition in the JS VM cache.

Fixes https://github.com/sensu/sensu-go/issues/4073

Nope

Nope

Nope

Nope

I've added the unit test before applying the patch, which created a fatal error 100% of the time.

## Is this change a patch?

Yep


echlebek (approved):
Nice work!

Issue #2773, opened 2019-03-06
Backend websocket API should offer unauthenticated health check endpoint


cwjohnston:
## Expected Behavior

As an operator I can put one or more backends behind a load balancer and the load balancer can perform health checks on the API, UI and Agent transport.

## Current Behavior

Health checks against `/` on Agent transport (port 8081) return a 401 response code which load balancers interpret as unhealthy.

## Possible Solution

Provide an unauthenticated endpoint, e.g. `/healthz` for determining the health of the Agent transport

## Steps to Reproduce (for bugs)
1. Deploy one or more Sensu backend nodes behind an ELB or ALB in AWS
2. Configure health check against Agent transport (HTTP port 8081) 
3. Observe that backend nodes are marked as unhealthy

## Context

Some load balancers, e.g. AWS ALB, do not provide overrides for acceptable HTTP responses, so we can't work around this via configuration.

## Your Environment
* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): sensu-backend 5.2.1
* Installation method (packages, binaries, docker etc.): packages
* Operating System and version (e.g. Ubuntu 14.04): Centos 7

portertech:
Backend websocket server `/healthz` endpoint.

PR #4075, opened 2020-10-27
Add health endpoint to backend websocket API

amdprophet:
## What is this change?

Adds an unauthenticated `/health` endpoint to the backend websocket API.

See #2773 

Yes.

No.

I initially misunderstood how the actions controller for health worked.

Yes, new documentation is required.

I wrote a test to ensure that the `/health` endpoint on the agentd websocket port returns a 200 status.

## Is this change a patch?

No.

### DIFF
--- a/backend/agentd/agentd.go
+++ b/backend/agentd/agentd.go
@@ -2,6 +2,7 @@ package agentd
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
 	"log"
 	"net"
@@ -17,7 +18,9 @@ import (
 	"github.com/gorilla/websocket"
 	"github.com/prometheus/client_golang/prometheus"
 	corev2 "github.com/sensu/sensu-go/api/core/v2"
+	"github.com/sensu/sensu-go/backend/apid/actions"
 	"github.com/sensu/sensu-go/backend/apid/middlewares"
+	"github.com/sensu/sensu-go/backend/apid/routers"
 	"github.com/sensu/sensu-go/backend/authentication/jwt"
 	"github.com/sensu/sensu-go/backend/authorization"
 	"github.com/sensu/sensu-go/backend/authorization/rbac"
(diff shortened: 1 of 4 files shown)


backend/agentd/agentd.go
```diff
@@ -141,8 +149,15 @@ func New(c Config, opts ...Option) (*Agentd, error) {
...
 	router := mux.NewRouter()
-	router.HandleFunc("/", a.webSocketHandler)
-	router.Use(authenticate, authorize, entityLimit, agentLimit)
+
+	healthRouter := routers.NewHealthRouter(
+		actions.NewHealthController(a.store, a.client.Cluster, a.etcdClientTLSConfig),
+	)
+	healthRouter.Mount(router)
+
+	route := router.NewRoute().Subrouter()
```
ccressent, line 158: Not sure I understand this line.

ccressent (approved):
I did not test this PR but it seems reasonable to me. I have a few questions though:

- #2773 mentions also UI and "backend" ; that `/health` only covers agentd's websocket though?
- Don't we already have existing health endpoints? Can we not instead relax them a bit so an unauthenticated request can still get barebone information (like just a 200 OK)?

Those are just general interest questions I unfortunately do not have the time to go answer myself right now.

backend/agentd/agentd.go
amdprophet: This creates a new subrouter that apply the authentication & authorization middleware to.

amdprophet:
> * #2773 mentions also UI and "backend" ; that `/health` only covers agentd's websocket though?
I think UI and backend (API) are already covered.

> * Don't we already have existing health endpoints? Can we not instead relax them a bit so an unauthenticated request can still get barebone information (like just a 200 OK)?
User(s) requested a way to verify if the agent transport (websocket) is running. As per @portertech's comment (https://github.com/sensu/sensu-go/issues/2773#issuecomment-696269962) I reused the code for the code that the API's `/health` route is using.

PR #4078, opened 2020-10-28
Ring refactoring

### LINKED ISSUE #4071: Make Sensu's ring implementation abstract and pluggable
We'd like to be able to swap out the ring implementation in Sensu for the enterprise product. To that end, we'll need to do some work to abstract the ring into an interface that can be implemented by any provider, and a way to swap it dynamically at runtime.

Create a type called `ring.Interface` that can provide this abstraction, and figure out a way to replace rings that are being used at runtime.

echlebek:
## What is this change?

```
    Refactor the round-robin ring and consumers
    
    The commit refactors the round-robin ring so that it can be abstracted
    away by its consumers.
    
    The existing interfaces for the ring remain, but new facilities have
    been added (ringv2.Interface, ringv2.RingPool).
    
    The new RingPool allows patching the internal ring implementation from
    outside, so that software like sensu-enterprise-go can replace the
    implementation for another one.
    
    This also should allow us to mock the ring in tests, however, at this
    time, this has not been done.
```

Closes #4071 

Only unit and integration tests. The existing QA for round robin type features should suffice here. As this is simply adding indirection to the ring facility, I don't expect this to cause any difference in behaviour. The ring should work as before, bugs and all.

## Is this change a patch?

The change is arguably at patch level, but doesn't need to target a release branch; it can be applied directly to main.


echlebek:
Looks like there was some existing lint in the `schedulerd` package that's been exposed by this PR. It's not simple to fix, so I'm leaning towards leaving it there.

backend/ringv2/interface_test.go
```diff
@@ -0,0 +1,88 @@
+package ringv2
+
+import "testing"
+
+// interface assertion
+var _ Interface = (*Ring)(nil)
```
ccressent, line 6: I'm not sure what this is for.

backend/ringv2/pool.go
```diff
@@ -25,14 +27,59 @@ func NewPool(client *clientv3.Client) *Pool {
...
 	ring, ok := p.rings[path]
-	p.mu.Unlock()
 	if ok {
 		return ring
 	}
 	ring = New(p.client, path)
-	p.mu.Lock()
 	p.rings[path] = ring
-	p.mu.Unlock()
 	return ring
```
ccressent, line 37: This is correct as it stands, and this is personal preference, but I personally do not like deferring `mu.Unlock()`. I much prefer to see blocks surrounded by `Lock()/Unlock()` that clearly show the critical sections.

backend/ringv2/pool.go
```diff
@@ -25,14 +27,59 @@ func NewPool(client *clientv3.Client) *Pool {
...
+
+// Del deletes the ring corresponding to the given path.
+func (r *RingPool) Del(path string) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	delete(r.rings, path)
+}
+
+// SetNewFunc sets the newer function for the ring pool. It results in the
+// ring cache being cleared.
```
ccressent, line 80: I'm not sure I understand the second part of the comment: "results in the ring cache being cleared".

ccressent (commented):
Looks good, easy to follow and well documented as far as I'm concerned. Great job!
Delaying approval until after CI is fixed.

backend/ringv2/interface_test.go
echlebek: It's a compile time check that `*Ring` is an `Interface`.

backend/ringv2/pool.go
echlebek: Conventional wisdom is that not deferring mutex unlock is more often than not an error. Best-case, it leaves the door open for creating a bug. What's the reason for your dislike of defer?

backend/ringv2/pool.go
echlebek: Ah, that should say pool.

backend/ringv2/pool.go
echlebek: BTW, the change is is actually correcting a race condition on `p.rings[path]`.

backend/ringv2/pool.go
echlebek: This actually exposed a bug, the pool wasn't being reset! I've added a fix and a test.

ccressent (approved):

backend/ringv2/pool.go
ccressent: Interesting that this change fixes a race condition; I don't understand how it could do that though, unless `New()` in the snippet above somehow does something ungodly. If you understand the fix, I would love if you could explain it to me!

> Conventional wisdom is that not deferring mutex unlock is more often than not an error [...] What's the reason for your dislike of defer?

I would be more nuanced than that: forgetting to release a mutex is more often than not an error. Deferring the release happens to save you most of the time; that doesn't mean it's what we should do blindly in my opinion.

I think we can agree that, hand-wavily, `defer X` executes `X` after the function returns. So for `defer Unlock()` in particular, it is akin to saying "my critical section spans from the `Lock()` call all the way to the end of the function". That works most of the time. It can well mean what you intend. But that "here to the end of the function" definition changes over time, or can be missed or misunderstood altogether.

For example:

```
func () {
    Lock()
    defer Unlock()
    [ begin critical section ]
    [ s0: operations that actually need the lock ]
    [ s1: potentially an awful lot of code that doesn't even need to hold the lock ]
    return
    [ end of critical section ]
}
```

My issue is that:
- if the function is not trivial, it's conceivable that someone might edit `s1` and introduce issues, if not outright bugs. For example, calling slow functions unrelated to the lock in that critical section, or trying to reacquire the lock, introducing a deadlock
- if the function is used concurrently or in parallel, it could be holding the lock for much longer than it needs to, blocking the progression of other computations (although sometimes it's better to lock once for a bit too long rather than lock/unlock several times; the one big lock/unlock can waste less time overall. This is highly workload dependent).

I contrast that with explicitly minimizing the span of the critical sections:
```
func () {
    Lock() // begin critical section
    [ s0: operations that actually need the lock ]
    Unlock() // end critical section

    [ s1: potentially an awful lot of code that doesn't even need to hold the lock ]

    Lock() // begin another critical section
    [ ... ]
    Unlock() // end critical section

    return
}
```

The semantics is more precise (objective) and it's clearer to me (subjective) where it is paramount to be extra careful and not be wasteful. In particular, flow-control change should be avoided at all cost there.

If you must have flow-control change in there, then I would agree that `defer Unlock()` is very nice, because it guarantees the unlock as long as you eventually return. But it's not a concurrency get out of jail for free card that we can just play willy-nilly!

echlebek: The advice that I've gotten from people on the Go team is that if critical sections don't span the width of the function, refactor so that they do.

In this case, the entire function is actually the critical section, despite there being no danger of data races, because competing processes can issue faulty writes based on stale reads.

While there was a performance hit to defer in past versions of the compiler, it is now so minimal as to be non-existent, and there are very few reasons to avoid it.

PR #4085, opened 2020-11-05
Disable linter on some legacy schedulerd code

echlebek:


backend/schedulerd/roundrobin_cron.go
```diff
@@ -121,6 +121,7 @@ func (s *RoundRobinCronScheduler) handleEvents(executor *CheckExecutor, ch <-cha
 	}
 }
 
+//nolint
```
ccressent, line 124: I think that should be:
```
//nolint:govet
```
It's the `lostcancel` pass of `go vet` that's raising the issue. If there was a way to only disable that specific pass through a comment, I would do that (I don't think there is a way though). Either way, I'd still run all the other linters!

ccressent (changes requested):
Could you also add a few lines about why you think this is a false positive, warranting an exception?

ccressent (approved):

echlebek:
The tests have failed, but it appears to be unrelated to this PR, so I'm going to merge.

PR #4089, opened 2020-11-16
Add Scheduler field to the core/v2 Check and CheckConfig APIs

echlebek:
## What is this change?

This change adds the Scheduler field to the Check and CheckScheduler APIs.

This field allows us to transmit information about how checks were scheduled to users. Checks can be scheduled with a variety of mechanisms in Sensu, including `memory`, `etcd`, and `postgres` (in the enterprise product).

Yes

Documentation will need updating if this change is accepted.

I verified that events now include this field in their check.

## Is this change a patch?

No


echlebek:
@palourde can you please review?

backend/ringv2/ringv2.go
```diff
@@ -463,25 +463,39 @@ func (r *Ring) startWatchers(ctx context.Context, ch chan Event, name string, va
...
-				if response.Canceled || !ok {
+				if !ok || response.Canceled {
 					// The watcher needs to be reinstated
-					r.startWatchers(ctx, ch, name, values, interval, cron)
-					return
+					if ctx.Err() == nil {
+						r.startWatchers(ctx, ch, name, values, interval, cron)
+						return
+					} else {
+						continue
```
palourde, line 476: Should we use `notifyError` here and below on line 493?

backend/store/etcd/check_store.go
```diff
@@ -26,6 +25,17 @@ func GetCheckConfigsPath(ctx context.Context, name string) string {
 	return checkKeyBuilder.WithContext(ctx).Build(name)
 }
 
+func schedulerFor(c *corev2.CheckConfig) string {
+	if c.Scheduler == "" {
+		if c.RoundRobin {
+			return "etcd"
```
palourde, line 31: Should we use the constants here? e.g. `corev2.EtcdScheduler`

palourde (approved):
Changes look good, only two small suggestions

backend/ringv2/ringv2.go
echlebek: No, `ctx.Err() != nil` only means the context has been canceled. The next iteration will terminate the loop.

backend/store/etcd/check_store.go
echlebek: Yup, absolutely.

PR #4090, opened 2020-11-16
Mock out time with timeproxy to fix flaky test

echlebek:
## What is this change?

This commit mocks out the time package in the prometheus transformer in
order to avoid a flaky test failure.

The test failed in CI with no provocation.

We've used timeproxy successfully in several areas of the codebase, so I'm confident this won't cause any issues.

## Is this change a patch?

Yes, but it only affects developers so I've targeted main.

### DIFF
--- a/agent/transformers/prometheus.go
+++ b/agent/transformers/prometheus.go
@@ -3,7 +3,8 @@ package transformers
 import (
 	"math"
 	"strings"
-	"time"
+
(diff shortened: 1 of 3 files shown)


echlebek:
@portertech ping :sweat_smile:

portertech:
scorpion-king-nice

portertech (approved):

PR #4097, opened 2020-11-20
[GraphQL] Include flapping flag in CheckHistory type

jamesdphillips:


jamesdphillips:
Will re-open when flapping field is persisted.

PR #4104, opened 2020-11-24
Remove managed_by label creation with edit command

### LINKED ISSUE #4087: sensuctl --interactive resource missing sensu.io/managed_by label
Sensuctl resource creation with `--interactive` does not produce resources with the `sensu.io/managed_by` label. Sensuctl prune is unable to prune these resources. `sensuctl edit` does add the  `sensu.io/managed_by` label :sweat_smile: 

## Expected Behavior
Resources created via `sensuctl create --interactive` have the `sensu.io/managed_by` label set to `sensuctl`. These resources can be pruned.

## Current Behavior
Sensuctl resource creation with `--interactive` does not produce resources with the `sensu.io/managed_by` label. Sensuctl prune is unable to prune these resources.

## Possible Solution
Updated sensuctl `--interactive` logic to include the `sensu.io/managed_by` label in the produced object.

`sensuctl handler create --interactive`
`sensuctl prune handlers -r -f example/handlers/`

## Context
Used `--interactive` to create a rogue handler to be pruned, but it could not be pruned.

## Your Environment

ccressent:
## What is this change?

During a previous refactor, parts of the logic for the create, edit and
delete sub-commands were merged together. This resulted in the side
effect of having the edit and delete sub-commands automatically add a
"managed_by" label with a "sensuctl" value to resources they touched.
This is not what we want; only sensuctl create should be automatically
adding this label.

This commit removes the label creation from the code shared by the 3
sub-commands, creates a new Processor that labels resources before
submitting them to the API and uses it as part of the definition of the
create sub-command only.

Closes #4087.

Yes, added.

No.

No.

Unit tests for the new labeling Processor and manual testing showing that:
- `edit` doesn't add the label to existing resources anymore
- `create` adds the label

## Is this change a patch?

Not 100% sure. I think it can go in the next minor. I can target a release
branch if that's deemed more appropriate by the reviewers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ entries in sensuctl.
 
 ### Fixed
 - Fixed a crash in the backend and agent related to Javascript execution.
+- `sensuctl edit` no longer automatically adds the `sensu.io/managed_by` label.
 
 ## [6.1.1] - 2020-10-22
 
--- a/cli/commands/create/create.go
+++ b/cli/commands/create/create.go
@@ -36,7 +36,7 @@ func execute(cli *cli.SensuCli) func(*cobra.Command, []string) error {
 		if err != nil {
 			return err
 		}
-		processor := resource.NewPutter()
+		processor := resource.NewManagedByLabelPutter("sensuctl")
 		if len(inputs) == 0 {
 			return resource.ProcessStdin(cli, client, processor)
 		}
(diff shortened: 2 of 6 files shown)


echlebek (approved):
When your putter isn't quit the right putter

putter

palourde (approved):
Successfully tested!

PR #4106, opened 2020-11-26
Preliminary sensuctl support for core/v3 resources

echlebek:
## What is this change?

This commit adds support for the new core/v3 resource type in sensuctl.
The new core/v3 resources are similar to core/v2 resources and are
specified in the same way by users.

Testing was done with core/v3.EntityConfig.

This will allow us to use the new resource type in core/v3.

No. We are not yet advertising this work as supported.

Not yet.

`sensuctl create -f entity_config.yaml` results in HTTP 404.

## Is this change a patch?

No.


echlebek:
@palourde this PR contains changes necessary for sensuctl to support core/v3 resource types.

It also includes support for core/v3 in apid, but that has yet to be integrated anywhere for now.

backend/apid/handlers/list_v3.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+func (h Handlers) ListV3Resources(ctx context.Context, pred *store.SelectionPredicate) ([]corev3.Resource, error) {
+	req := storev2.NewResourceRequest(ctx, corev2.ContextNamespace(ctx), "", h.V3Resource.StoreName())
+	list, err := h.StoreV2.List(req, pred)
+	if err != nil {
+		return nil, err
+	}
+	return list.Unwrap()
+}
```
palourde, line 19: For reference, listing resources will require a bit more work later because we need to support something equivalent to our [current lister](https://github.com/sensu/sensu-go/blob/main/backend/apid/routers/lister.go); not sure if this is something you intended to do in this PR or in a subsequent one, so just to make sure we don't forget about this!

types/go.mod
```diff
@@ -2,9 +2,14 @@ module github.com/sensu/sensu-go/types
 
 go 1.13
 
+replace (
+	github.com/sensu/sensu-go => ../sensu-go
+	github.com/sensu/sensu-go/api/core/v2 => ../api/core/v2
+)
```
palourde, line 8: Is this intended?

echlebek: Yes, the types package should import whatever the "current" version of sensu-go and core/v2 are.

palourde: Ah I see, thanks for the explanation!

palourde (approved):

PR #4111, opened 2020-12-02
Replace uuid with event id in logging

### LINKED ISSUE #4093: event ID keys are not consistent in sensu-backend logs
## Expected Behavior

In reviewing sensu-backend logs, I can refer to a consistent key name for the event UUID.

## Current Behavior

sensu-backend log messages use inconsistent key names when logging event UUID. Some log messages use key `event_uuid`, whereas others use key `uuid`. 

## Possible Solution

Standardize name for key whose value is the event UUID

1. Turn on debug level logging in your sensu-backend
2. Submit one or more events via the Agent, scheduled or using the agent's /events HTTP API
3. Identify UUID for the event, then grep the UUID from backend-logs
4. Note that although each message uses the same UUID value, the keys which provide that value vary

## Context

Some log messages use key `event_uuid`, whereas others use key `uuid`, and this appears to be inconsistent even for log messages coming from the same backend component but differing log levels:

```
{
  "check_name": "check-sensu-filebeat-process",
  "check_namespace": "monitoring",
  "component": "pipelined",
  "entity_name": "sd-25f8-415b",
  "entity_namespace": "monitoring",
  "handler": "monitoring-email",
  "level": "info",
  "msg": "sending event to handler",
  "time": "2020-11-05T09:30:00-05:00",
  "uuid": "47eedb71-9430-4120-b471-9680a1c2f09d"
}
{
  "assets": [
    "sensu-email-handler"
  ],
  "check": "check-sensu-filebeat-process",
  "component": "pipelined",
  "entity": "sd-25f8-415b",
  "event_uuid": "47eedb71-9430-4120-b471-9680a1c2f09d",
  "handler": "monitoring-email",
  "level": "debug",
  "msg": "fetching assets for handler",
  "namespace": "monitoring",
  "time": "2020-11-05T09:30:00-05:00"
}
```

Further more, the event payload sent to handlers and event-log (when configured) uses key name `id`.

## Your Environment

amdprophet:

## What is this change?

Replaces all instances where `event_uuid` was logged with `event_id`.

Closes #4093.

Yes.

Nope!

Nope!

Just unit/integration tests.

## Is this change a patch?

No.

### DIFF
--- a/util/logging/logging.go
+++ b/util/logging/logging.go
@@ -14,7 +14,7 @@ func EventFields(event *types.Event, debug bool) map[string]interface{} {
 	}
 
 	fields := logrus.Fields{
-		"event_id":         event.ID,
+		"event_id":         event.GetUUID().String(),
 		"entity_name":      event.Entity.Name,
 		"entity_namespace": event.Entity.Namespace,
 	}
--- a/util/logging/logging.go
+++ b/util/logging/logging.go
@@ -14,9 +14,9 @@ func EventFields(event *types.Event, debug bool) map[string]interface{} {
 	}
 
 	fields := logrus.Fields{
+		"event_id":         event.ID,
 		"entity_name":      event.Entity.Name,
 		"entity_namespace": event.Entity.Namespace,
-		"uuid":             event.GetUUID().String(),
 	}
 
 	if event.HasCheck() {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ sensu-backend (`SENSU_BACKEND_CONFIG_FILE`) & sensu-agent (`SENSU_CONFIG_FILE`).
 ### Changed
 - Adjust the date and duration formats used when listing and displaying silenced
 entries in sensuctl.
-- Use event ID in logging instead of UUID.
+- Make `event_id` usage in logging consistent.
 
 ### Fixed
 - The config-file flag is no longer order dependant.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ sensu-backend (`SENSU_BACKEND_CONFIG_FILE`) & sensu-agent (`SENSU_CONFIG_FILE`).
 ### Changed
 - Adjust the date and duration formats used when listing and displaying silenced
 entries in sensuctl.
+- Use event ID in logging instead of UUID.
 
 ### Fixed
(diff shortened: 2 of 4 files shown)


util/logging/logging.go
```diff
@@ -14,9 +14,9 @@ func EventFields(event *types.Event, debug bool) map[string]interface{} {
 	}
 
 	fields := logrus.Fields{
+		"event_id":         event.ID,
 		"entity_name":      event.Entity.Name,
 		"entity_namespace": event.Entity.Namespace,
-		"uuid":             event.GetUUID().String(),
 	}
 
 	if event.HasCheck() {
```
ccressent, line 22: I'm not sure I 100% understand this: I thought only the field what supposed to
be renamed from "uuid" to "event_id". But here, the content of the field also
changes from event.GetUUID().String() to event.ID?

ccressent (commented):
Nice work refactoring to use that `util/logging` package, I wasn't aware of it!
I don't believe we have good test coverage, if any, on the logging and the
fields it uses. So I'll ask: how confident are you this covers all the
relevant logs? This touches Events (obviously) and Handlers (less obvious).
Could there be other types (Mutators, Filters, ...) that could need a similar
treatment?

util/logging/logging.go
ccressent, line 22: I'm not sure I 100% understand this: I thought only the field what supposed to
be renamed from "uuid" to "event_id". But here, the content of the field also
changes from `event.GetUUID().String()` to `event.ID`?

ccressent (commented):
Nice work refactoring to use that `util/logging` package, I wasn't aware of it!
I don't believe we have good test coverage, if any, on the logging and the
fields it uses. So I'll ask: how confident are you this covers all the
relevant logs? This touches Events (obviously) and Handlers (less obvious).
Could there be other types (Mutators, Filters, ...) that could need a similar
treatment?

util/logging/logging.go
amdprophet: I did a recursive search for `uuid` in the backend directory and went through the results to ensure the field was updated where it was being used.

@portertech @cwjohnston It's not clear if we only want to update the field name or not. Do we still want to use `event.GetUUID().String()`? or use `event.ID` for all `event_id` fields.

amdprophet:
@ccressent updated & ready for review.

ccressent (approved):

PR #4115, opened 2020-12-03
Move the compat library to a sensible location

echlebek:
## What is this change?

Tackles some technical debt recently introduced by me, a lazy person.


ccressent (approved):
Test failure appears to be from a known flaky test; change itself looks good, thanks!

Issue #3912, opened 2020-07-16
sensuctl should return status 2  when help message is shown due to malformed command error


jspaleta:
## Current Behavior
Badly formed sensuctl commands return help and status 0
Example:
```
$ sensuctl check > /dev/null; echo $?
0
```

## Expected Behavior
Should return status 2 in these situations where help output displayed without being explicitly requested due to malform command.

nikkictl:
Great find @jspaleta, this is definitely a bug and we should return an error exit status if any sensuctl command fails.

amdprophet:
Closed by #4119.

PR #4119, opened 2020-12-07
Exit non-zero when running sensuctl cmds that only have subcmds

amdprophet:
## What is this change?

Updates sensuctl commands that only contain subcommands to exit 46 (EX_USAGE from `sysexits.h`) when executed.

See https://github.com/sensu/sensu-go/issues/3912.

Yes.

Nope!

Cobra root commands behave differently than subcommands and thus the root command cannot return an error without displaying different output than subcommands.

Tested by running:
* `sensuctl check` (exits 46, displays usage)
* `sensuctl check creat` (exits 46, displays suggestions)
* `sensusctl check create` (works as it did before)
* `sensuctl check help` (exits 0, displays usage)

## Is this change a patch?

No.


command/sysexits.go
```diff
@@ -0,0 +1,82 @@
...
+	ExitNoPermission = 77
+
+	// Something was found in an unconfigured or misconfigured state.
+	ExitConfig = 78
+)
+
+type CommandErrorer interface {
+	ExitStatus()
+	Error()
+}
```
echlebek, line 77: Not sure I understand what this is. Can it be removed?

command/sysexits.go
```diff
@@ -0,0 +1,82 @@
+package command
+
+// Based on sysexits.h from FreeBSD (also used by Linux)
+// https://www.freebsd.org/cgi/man.cgi?query=sysexits&manpath=FreeBSD+13.0-current
+
+var (
+	// The command executed successfully.
+	ExitOk = 0
```
echlebek, line 8: Ideally this would be named ExitOK

cli/commands/helpers/subcommand.go
```diff
@@ -0,0 +1,63 @@
+package helpers
```
echlebek, line 1: I feel this code would be better namespaced in the command package, as "helpers" has less meaning. https://github.com/golang/go/wiki/CodeReviewComments#package-names

echlebek (commented):
Love the docs on the exit constants, great work. I do have a few quibbles about names and I think I may have found some dead code.

command/sysexits.go
amdprophet: Good catch - I meant to use it here: https://github.com/sensu/sensu-go/blob/e47c19449cedeb33fbee0f7ad054e971aa3745d9/cmd/sensuctl/main.go#L25. Updating.

echlebek (approved):
Looks good, can you add a line to the changelog mentioning the help display fixes?

echlebek (approved):

Issue #4114, opened 2020-12-03
Flags of type StringSlice are broken


palourde:
## Expected Behavior

Considering the following patch:
```
 $ git diff
diff --git agent/cmd/start.go agent/cmd/start.go
index bd17186f..bbc2d054 100644
--- agent/cmd/start.go
+++ agent/cmd/start.go
@@ -158,7 +158,9 @@ func NewAgentConfig(cmd *cobra.Command) (*agent.Config, error) {
        }

        cfg.Redact = viper.GetStringSlice(flagRedact)
+       fmt.Printf("    cfg.Redact = %v\n", viper.GetStringSlice(flagRedact))
        cfg.Subscriptions = viper.GetStringSlice(flagSubscriptions)
+       fmt.Printf("cfg.Subscriptions = %v\n", viper.GetStringSlice(flagSubscriptions))

        // Workaround for https://github.com/sensu/sensu-go/issues/2357. Detect if
        // the flags for labels and annotations were changed. If so, use their
```

I'd expect a single item for each configuration slices:
```
$ sensu-agent start --subscriptions unix --redact foo
cfg.Redact = [foo]
cfg.Subscriptions = [unix]
```

## Current Behavior
Values are duplicated:

```
$ sensu-agent --subscriptions unix --redact foo
cfg.Redact = [foo foo]
cfg.Subscriptions = [unix unix]
```

## Possible Solution
I believe my refactoring in https://github.com/sensu/sensu-go/pull/4105 might have caused that.

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent):
* Installation method (packages, binaries, docker etc.):
* Operating System and version (e.g. Ubuntu 14.04):

PR #4122, opened 2020-12-07
Do not parse the same flag set twice

palourde:

## What is this change?

It fixes a bug where configuration values for both sensu-agent & sensu-backend would be duplicated when using slices.

All flags need to be present when we first try to parse the config file attribute, otherwise it will silently fail because undefined flags were defined. However, we should use a different set of flags for the "main" command otherwise they will be parsed twice, which ends up duplicating values in configuration attributes holding a slice.

Fixes https://github.com/sensu/sensu-go/issues/4114

Nope, unreleased bug introduced via https://github.com/sensu/sensu-go/pull/4105

Nope

Nope

Nope

Manually verified. This is not something that's easy to test automatically.

## Is this change a patch?

Nope; it just needs to be included with the PR above.

### DIFF
--- a/agent/cmd/start.go
+++ b/agent/cmd/start.go
@@ -216,15 +216,20 @@ func StartCommandWithError(initialize InitializeFunc) (*cobra.Command, error) {
 	}
 
 	cmd.RunE = NewAgentRunE(initialize, cmd)
-	return cmd, handleConfig(cmd)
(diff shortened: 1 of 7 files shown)


echlebek (approved):
OK, I think this makes sense!

echlebek (approved):
I continue to approve

echlebek (approved):
approved

PR #4125, opened 2020-12-09
Add support for core/v3.Resource to backend/api

echlebek:
## What is this change?

This commit adds support for core/v3.Resource types in the backend API
GenericClient. The support is shimmed through the
core/v3.V2ResourceProxy type.

Users are expected to set up the GenericClient with a store/v2.Interface
and a Kind that is a V2ResourceProxy, if they want to work with core/v3
resources.

Supporting work for sensu-enterprise-go.

No, not a user facing change.

### DIFF
--- a/backend/api/generic.go
+++ b/backend/api/generic.go
@@ -7,8 +7,11 @@ import (
 	"path"
 
 	corev2 "github.com/sensu/sensu-go/api/core/v2"
+	corev3 "github.com/sensu/sensu-go/api/core/v3"
 	"github.com/sensu/sensu-go/backend/authorization"
 	"github.com/sensu/sensu-go/backend/store"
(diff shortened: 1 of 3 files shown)


jamesdphillips (approved):

backend/api/generic.go
```diff
@@ -77,6 +96,20 @@ func (g *GenericClient) SetTypeMeta(meta corev2.TypeMeta) error {
 	return nil
 }
```
jamesdphillips, line 97: Ran into an issue while testing. I believe the `SetTypeMeta` method needs to use `types.ResolveRaw` instead.

jamesdphillips: or, ... the typemap needs to return the proxied resource.

echlebek: I made it work with `ResolveRaw`, should be good to go now :+1:

PR #4129, opened 2020-12-10
Continue to ignore flags parsing errors for now

palourde:

## What is this change?

It reverts a previous change that would handle errors while parsing flags in the `handleConfig` function. Unfortunately, as we recently discovered, this is problematic because it gets called by almost all subcommands, but not all flags are usually defined.

Fixes a bug found on `main`

Nope, unreleased bug.

Nope

Viper & Cobra are hard!

Nope

Manually tested

## Is this change a patch?
Nope


echlebek (approved):
Successfully tested on staging!

PR #4131, opened 2020-12-12
[GraphQL] Generator cleanup

jamesdphillips:
Updates the code generator to omit code comments describing implementation and extraneous interfaces. Doing so greatly reduces the amount of symbols and needless documentation produced.

As a nice bonus, shaves about 400KBs off of the package size.

---


ccressent (approved):
I have to admit I haven't reviewed this in details and that I'm not quite sure of my understanding of the areas this change touches, but I'm approving it in the interest of moving forward instead of having this PR rot.

Issue #4120, opened 2020-12-07
Agent-managed entities


palourde:
###  Background

See the proposal: https://docs.google.com/document/d/1rBag5sWK50kYOhYAmIP_e0mpLsdRNMiIcd5VZ1NRy4U/edit.

See the PoC: https://github.com/sensu/sensu-go/commit/ba121c369c5cbdef9ce3c2889675ef5c30240283

### Acceptance Criteria

- [x] The agent can be started with an opt-in flag to indicate that its entity will be managed by the agent
- [x] Entities managed by agents can update their configuration in the backend via a restart of the sensu-agent process
- [x] Registration events are also supported for entities managed by agents
- [x] An entity that's managed by its agent and deleted in the backend will be re-created with the next keepalive, using the latest entity configuration.
- [x] An entity managed by its agent cannot be modified via the backend API
- [x] An agent managed via the backend can become managed by its agent upon restart.
- [x] An agent managed via its agent can become managed by the backend upon restart.

### Implementation
- The first keepalive sent by agent is identified by a sequence number (TODO: link issue)
- When managed by the agent, an entity will have:
	- the `created_by` field will be set to the agent's username
	- the [`sensu.io/managed_by` label](https://github.com/sensu/sensu-go/blob/74fa91c54f6867838915cd88f07e67de13e4ab60/api/core/v2/meta.go#L6) will be set to `sensu-agent`
- When keepalived determines that an entity is managed by its agent and that this is the first keepalive sent for this entity, the entity config is either created or updated
	- A registration event is created if the entity did not exist
- Entities configuration updates are not sent to agents that do actively manage their entities, and the agents will ignore updates that are still received

### Test Cases
1. Start sensu-agent with --agent-managed-entity
	- The entity should be registered
	- The entity should have the same config as specified in the agent
	- It should be impossible to update the entity via the backend API
2. Restart sensu-agent with --agent-managed-entity --subscriptions foo
	- The entity should be updated in the backend with this new subscription
	- The entity should execute checks for this new subscription
	- It should be impossible to update the entity via the backend API
3. Restart sensu-agent with --subscriptions bar (without --agent-managed-entity)
	- The entity should continue to use its older subscriptions, not “bar”
	- It should be possible to update the entity via the backend API
4. Restart sensu-agent with --agent-managed-entity --subscriptions bar
	- The entity should be updated in the backend with this new subscription
	- The entity should execute checks for this new subscription
	- It should be impossible to update the entity via the backend API

[...]

PR #4132, opened 2020-12-14
Add support for entities managed by agents

palourde:
## What is this change?

It introduces support for entities that are actively managed by agents; this re-introduces the behaviour we previously had with pre-6.0 agents.

sensu-agent now has a new `agent-managed-entity` configuration flag. When used, the entity managed by this agent will use the `sensu.io/managed_by` label to indicate that sensu-agent is now the source of truth for this entity.

keepalived (via its `handleEntityRegistration` method) is now responsible for updating the entity config on startup for entities managed by their agents, by using the new `sequence` field in events and the label mentioned above. In case the agent is restarted and no longer has the `agent-managed-entity` config attribute, we will then update the stored entity config to reflect that (by removing the `sensu.io/managed_by` label).

Finally, we are preventing updates made to entities managed by their agents via the REST API in two ways:
- An entity definition uploaded with the label `sensu.io/managed_by` that equals `sensu-agent` will return HTTP 409 Conflict.
- An entity stored with the label `sensu.io/managed_by` that equals `sensu-agent` will return HTTP 409 Conflict when updated.

Closes https://github.com/sensu/sensu-go/issues/4120

Added

Nope

Yep; the `corev3.V3EntityToV2` currently has a bug which prevents the `CreatedBy` field to be carried over to the corev2 Entity.

Yep, I'll open an issue and link it here.

Update: https://github.com/sensu/sensu-docs/issues/2905

I've created a test cases section in https://github.com/sensu/sensu-go/issues/4120, which I've used to test this feature. Once sensu-staging is working as expected, I'll have to deploy this branch and I'll try to write some tests for our QA crucible.

Update: https://github.com/sensu/sensu-go-qa-crucible/issues/164

## Is this change a patch?

Nope

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ created when in interactive mode.
 - Add support for environment variables to define configuration file paths of
 sensu-backend (`SENSU_BACKEND_CONFIG_FILE`) & sensu-agent (`SENSU_CONFIG_FILE`).
 - Added event sequence numbers.
+- Entities may now be managed exclusively by their agents when sensu-agent is
+started with the new `agent-managed-entity` configuration attribute.
 
 ### Changed
 - Adjust the date and duration formats used when listing and displaying silenced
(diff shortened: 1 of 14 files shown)


echlebek (approved):
Successfully tested :tada:

PR #4133, opened 2020-12-15
Fix CreatedBy not being carried over when converting v3 to v2 entities

ccressent:
## What is this change?

With this change, when converting a (v3.EntityConfig, v3.EntityState) into a
v2.Entity, we carry over the CreatedBy metadata from the config, just like we do
for labels and annotations.

Addresses a bug that @palourde came across while working on something else.

Yes, added.

No.

After careful consideration, a test that asserts that, after a v3 to v2
conversion, the state's metadata ought to be the same as the resulting v2
entity's has been removed because it is not correct: metadata information is
copied from the config part. This change exposed this issue.

No change in documentation is required.

Unit tests only.

## Is this change a patch?

No?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ code 46 when no arguments or incorrect arguments are given.
 authentication.
 - Added description to sensuctl dump command.
 - The sensuctl command descriptions now have consistent capitalization.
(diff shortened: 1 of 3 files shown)


palourde (approved):
👌

PR #4140, opened 2020-12-18
Don't remove entities from nonexistent rings

echlebek:
## What is this change?

This commit fixes a bug in keepalived, where entities are attempted to
be removed from rings that don't exist. Specifically, the entity
subscription.

On etcd this generates extra traffic, although no additional keys are
created. On postgres, this can result in many rows being inserted
needlessly.

Yes

Currently verifying on sensu-perf.

## Is this change a patch?

Yes


palourde (approved):
👌

PR #4149, opened 2020-12-30
Limit agent transport upgrades before hitting etcd

echlebek:
## What is this change?

Do less work in the hard agent limiter, by invoking less middleware.

### DIFF
--- a/backend/agentd/middlewares.go
+++ b/backend/agentd/middlewares.go
@@ -40,12 +40,6 @@ func agentLimit(next http.Handler) http.Handler {
 	return AgentLimiterMiddleware(next)
 }
 
-// entityLimit is the abstraction layer required to be able to change at
-// runtime the actual function assigned to EntityLimiterMiddleware above.
-func entityLimit(next http.Handler) http.Handler {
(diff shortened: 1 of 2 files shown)


backend/agentd/middlewares.go
```diff
@@ -40,12 +40,6 @@ func agentLimit(next http.Handler) http.Handler {
 	return AgentLimiterMiddleware(next)
 }
 
-// entityLimit is the abstraction layer required to be able to change at
-// runtime the actual function assigned to EntityLimiterMiddleware above.
-func entityLimit(next http.Handler) http.Handler {
-	return EntityLimiterMiddleware(next)
```
portertech, line 46: And we don't want to also clean up `EntityLimiterMiddleware`?

portertech: nvm, I see the other references.

portertech (approved):

backend/agentd/middlewares.go
echlebek: Better to leave the exported symbol there to avoid build-breakage downstream.

Issue #4143, opened 2020-12-21
Silenced API in 6.2.0 broke backward-compatibility


tadeboro:
The silencing API up until 6.2.0 returned back a static representation of silencing entries and since 6.2.0, API returns back some non-static calculated values. 

## Expected Behavior

`GET /api/core/v2/namespaces/default/silenced/*:chk` should return a superset of data that was used to create the silencing entry.

## Current Behavior

`GET /api/core/v2/namespaces/default/silenced/*:chk` contains updated data in the `expire` field. The value in the `expire` field is also not static and changes with every `GET` call to the API.

## Possible Solution

In an ideal world, the previous behavior would be restored. And if the currently returned information is needed, it should be added into a new field.

## Steps to Reproduce (for bugs)

Download the script from https://gist.github.com/tadeboro/ad904d7365768f5cf9c94e747e9ade8b. Run:

    $ ./silence.sh 6.2.0  # This will show changing numbers
    $ ./silence.sh 6.1.4  # Output is static for all loop iterations (can also use any other Sensu Go version)

## Context

We discovered this change when we started testing Sensu Go Ansible Collection on 6.2.0. The first failure we encountered was caused by the fact that we were creating some expired silences in our tests and Sensu Go 6.2.0 did not return them back. But this was our error because there is no reason to assume expired entries will remain alive on the backend.

Once we sorted the previous error out, our idempotency checks started failing because the Ansible module for managing silences starting detecting changes where should be none. And we managed to track this down to the change in the `expire` value returned by the API.

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): Tested on 6.2.0 and 6.1.4
* Installation method (packages, binaries, docker etc.): docker

/cc @echlebek because he already helped me on Slack. No good deeds go unpunished ;)

palourde:
For reference, here's a thread about it in the community Slack: https://sensucommunity.slack.com/archives/C9BB9AW7K/p1608552870192700

PR #4152, opened 2021-01-06
Do not override the expire attribute of silenced entries

palourde:
## What is this change?

It fixes a breaking change introduced in 6.2.0, which would set the remaining duration of a silenced entry (its TTL) rather than the configured duration.

Instead, we now make sure to calculate the ExpireAt field value, and use that for display purpose in sensuctl.

Fixes https://github.com/sensu/sensu-go/issues/4143

Added

Nope

Nope

Nope

Manually tested & adjusted the unit tests.

## Is this change a patch?
Yep

### DIFF
--- a/backend/store/etcd/silenced_store.go
+++ b/backend/store/etcd/silenced_store.go
@@ -216,6 +216,13 @@ func (s *Store) arraySilencedEntries(ctx context.Context, resp *clientv3.GetResp
 		leaseID := clientv3.LeaseID(kv.Lease)
 		if leaseID > 0 {
 			// legacy expiry mechanism
+			ttl, err := s.client.TimeToLive(ctx, leaseID)
+			if err != nil {
+				logger.WithError(err).Error("error setting TTL on silenced")
+				continue
+			}
+			silenced.ExpireAt = time.Now().Unix() + ttl.TTL
+
 			result = append(result, silenced)
 		} else if silenced.ExpireAt > 0 {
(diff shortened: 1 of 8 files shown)


backend/store/etcd/silenced_store.go
```diff
@@ -211,27 +211,22 @@ func (s *Store) arraySilencedEntries(ctx context.Context, resp *clientv3.GetResp
...
 		leaseID := clientv3.LeaseID(kv.Lease)
 		if leaseID > 0 {
 			// legacy expiry mechanism
-			leaseID := clientv3.LeaseID(kv.Lease)
-			ttl, err := s.client.TimeToLive(ctx, leaseID)
-			if err != nil {
-				logger.WithError(err).Error("error setting TTL on silenced")
-				continue
-			}
-			silenced.Expire = ttl.TTL
```
echlebek, line 223: So, this code was intentionally left in so that when users upgrade to a newer version, their silences continue to work.

But perhaps it's fine if we take the plunge now.

What do you think?

echlebek: if we decide to drop support for the legacy silences (a breaking change, but a small one), then this whole logical branch should be removed.

palourde: As discussed, I've re-used this code so we calculate instead the ExpireAt value!

echlebek (approved):
Great work! :+1: :+1: :+1:

PR #4158, opened 2021-01-08
Store new events in their own history

### LINKED ISSUE #4145: Incorrect history of an event.
History data is getting updated when I execute the check for second time.

At first time executing the check does not add any history data in the event. This information is misleading while we want to create any filter based on the first history data point.

Below is the history data I got after executing the check for second time.

     "history": [
      {
        "status": 0,
        "executed": 1608785570
      }
      ]

amdprophet:
## What is this change?

Adds the first event received to its own history before persisting it to the store & sending it through the pipeline.

Fixes #4145.

I realized when looking into the issue that the first event received is never stored in the check's history (even when a second event has been received). This PR fixes both issues.

Yes.

No.

No documentation updates are required.

I added some new tests, existing tests pass, and I tested locally by ensuring check history exists when a new event is first received.

## Is this change a patch?

Yes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ seconds, not the remaining duration.
 - Allow agents labels & annotations to be configured via configuration flags.
 - Fixed a bug where a small amount of extra etcd or postgres traffic was
 generated on keepalive failure.
+- Event check history now contains an entry for the first received event.
 
 ## [6.2.0] - 2020-12-17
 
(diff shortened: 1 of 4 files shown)


palourde:
@amdprophet Can you try rebasing this branch off main, because it seems like it's bringing some unexpected commits and I'm worried it might cause some issues when we are merging!

palourde (approved):
Successfully tested this change, we just need to be careful with the extra commits!

PR #4179, opened 2021-01-20
Add managed_by label to cfg.Label

amdprophet:
This fixes an issue where the `managed_by` label wasn't persisted to the agent's config.


palourde:
🙈

PR #4191, opened 2021-01-29
sensu-backend init fixes

amdprophet:
## What is this change?

* Refactors `seeds.go`
* Removes `initializer.Lock()` when seeding etcd competing writes should be ok
* Removed seeding code from separate goroutine as it seemed unnecessary
* Improved error messages

Race conditions & strange behaviour with etcd locking were causing occasional issues when running `sensu-backend init` with `--wait`.

I'm not sure if this requires a changelog or not as it's technically a part of the already-merged `--wait` flag work.

Nope!

Many:
* etcd clientv3 doesn't show detailed error messages when used synchronously
* etcd mutex locking and/or watchers seem outright broken under certain conditions
* cannot easily alter the `CallOptions` that clientv3 uses which results in a massive dump of log messages when there's an error

Rerunning backend-init a bunch of times with etcd being in various states.

## Is this change a patch?

Only a patch meant to fix the current state of `main`.

### DIFF
--- a/backend/cmd/init.go
+++ b/backend/cmd/init.go
@@ -29,10 +29,17 @@ const (
 	flagWait              = "wait"
 )
 
-type seedConfig struct {
+type initConfig struct {
 	backend.Config
 	SeedConfig seeds.Config
 	Timeout    time.Duration
+}
+
+func (c *initConfig) Validate() error {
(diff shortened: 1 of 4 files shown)


amdprophet:
Closing for now until we figure out a solution to the lack of proper locking in etcd.

PR #4193, opened 2021-02-02
Validate resources in storev2

amdprophet:

## What is this change?

Validates V2 & V3 resources before they are wrapped in the storev2 wrap package.

Validation is not happening on any resources in storev2.

Yes.

I don't believe labels or annotations are mandatory so I modified the ObjectMeta Validate method to initialize as an empty map when the values are nil. I'm not sure if this is the best or correct approach; please let me know.

I ensured that the new BSM resources get validated with this change in place.

## Is this change a patch?

No.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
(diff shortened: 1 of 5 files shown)


echlebek (approved):
Nice work!

Issue #4176, opened 2021-01-16
Sensu backend Memory Address issue.


rehanch178:
I have 3 node cluster which was running fine. I checked and found cpu/memory utilization is very low but the cluster suddenly went down. Attaching error logs for the reference.

{"component":"etcd","level":"error","msg":"failed to read 9e83e7f64749072d on stream Message (unexpected EOF)","pkg":"rafthttp","time":"2021-01-16T00:00:46Z"}
{"component":"agentd","error":"","level":"error","msg":"unexpected error while watching for entity config updates","time":"2021-01-16T00:00:47Z"}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x104e3b6]

goroutine 2137 [running]:
github.com/sensu/sensu-go/backend/agentd.(*Agentd).handleEvent(0xc000180840, 0x0, 0x4, 0x1, 0x1)
        /sensu/sensu-go/backend/agentd/agentd.go:249 +0x76
github.com/sensu/sensu-go/backend/agentd.(*Agentd).runWatcher(0xc000180840)
        /sensu/sensu-go/backend/agentd/agentd.go:241 +0x165
created by github.com/sensu/sensu-go/backend/agentd.(*Agentd).Start
        /sensu/sensu-go/backend/agentd/agentd.go:217 +0x21b
{"component":"store","error":"etcdclient: no available endpoints","level":"error","member":8489119027689070460,"msg":"unhealthy cluster member","name":"sensu-backend3","time":"2021-01-16T00:02:50Z"}
{"component":"etcd","level":"error","msg":"failed to dial 4c98a5b754af677d on stream MsgApp v2 (dial tcp: lookup sensu-backend3 on 127.0.0.11:53: no such host)","pkg":"rafthttp","time":"2021-01-16T00:02:56Z"}

Requesting you look at the stability issues.

amdprophet:
This can happen when the etcd client returns a `WatchError`. It seems like etcd was not available judging by the log line: `etcdclient:  no available endpoints`. A crash should not happen as a result of this and it should be considered a bug.

I ran into this exception myself this past week and I've implemented a fix that should go out with the 6.3.0 release.
https://github.com/sensu/sensu-go/compare/bump-etcd-go#diff-bf238bc8d1248a0fa763d9b9b3709d1302a0cd5b53005917b9d65abb8473fe76

raihanchdy:
[...]

PR #4210, opened 2021-02-23
Prevent crash when store watchers have an empty watch event

amdprophet:
## What is this change?

Prevents an edge case, where watcher contents can be nil, from crashing a backend.

Fixes https://github.com/sensu/sensu-go/issues/4176.

Yes.

I haven't seen the same 

## Is this change a patch?

No.

### DIFF
--- a/backend/schedulerd/check_watcher.go
+++ b/backend/schedulerd/check_watcher.go
@@ -111,14 +111,9 @@ func (c *CheckWatcher) startWatcher() {
 	for {
 		select {
 		case watchEvent, ok := <-watchChan:
-			if !ok {
-				return
+			if ok {
+				c.handleWatchEvent(watchEvent)
 			}
-			if watchEvent.CheckConfig == nil {
-				logger.Error("nil check config received from check config watcher")
-				return
-			}
-			c.handleWatchEvent(watchEvent)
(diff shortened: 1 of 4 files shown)


backend/schedulerd/check_watcher.go
```diff
@@ -111,6 +111,10 @@ func (c *CheckWatcher) startWatcher() {
 	for {
 		select {
 		case watchEvent, ok := <-watchChan:
+			if watchEvent.CheckConfig == nil {
+				logger.Error("nil check config received from check config watcher")
+				return
+			}
```
echlebek, line 117: This is probably not quite right, as it doesn't check the `ok` value before accessing `watchEvent.CheckConfig`

amdprophet: Good catch. Fixed!

echlebek: That's not quite right either, I think we need to stop the schedulers if this happens, like when `<-ctx.Done()`. :thinking:

echlebek: Maybe put the stopping logic in a defer.

amdprophet:
Closing in favour of https://github.com/sensu/sensu-go/pull/4248.

PR #4212, opened 2021-02-24
Add ResourceV3 watcher

ccressent:
-------

## What is this change?

This adds `GetResourceV3Watcher()` to the `store/etcd` package. Similarly to the
other `GetXWatcher()` functions, it returns a channel with events related to the
given key, representing a store path to any type of `corev3.Resource`.

This function and the associated `WatchEventResourceV3` type were written for
more specific work in the enterprise code, but they are useful in general and
should be moved here.

Yes, added.

Used in enterprise feature. Unit tests for the various `GetXWatcher()` functions
would be nice, although a bit of refactoring might be necessary.

## Is this change a patch?

No.

### DIFF
--- a/backend/store/etcd/watchers.go
+++ b/backend/store/etcd/watchers.go
@@ -5,8 +5,10 @@ import (
 	"reflect"
 
 	"github.com/coreos/etcd/clientv3"
+
 	corev2 "github.com/sensu/sensu-go/api/core/v2"
 	"github.com/sensu/sensu-go/backend/store"
+	"github.com/sensu/sensu-go/backend/store/v2/wrap"
 )
 
 // GetCheckConfigWatcher returns a channel that emits WatchEventCheckConfig structs notifying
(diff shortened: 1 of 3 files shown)


backend/store/etcd/watchers.go
```diff
@@ -123,3 +125,52 @@ func GetResourceWatcher(ctx context.Context, client *clientv3.Client, key string
...
+
+// GetResourceV3Watcher returns a channel that emits WatchEventResourceV3
+// structs notifying that a corev3.Resource has been created, deleted or
+// updated.
+func GetResourceV3Watcher(ctx context.Context, client *clientv3.Client, key string) <-chan store.WatchEventResourceV3 {
+	w := Watch(ctx, client, key, true)
+	c := make(chan store.WatchEventResourceV3, 1)
+
+	go func() {
+		defer close(c)
```
echlebek, line 137: Can you add a deferred logger statement here at info level, so we can see the watcher shutting down?

backend/store/etcd/watchers.go
```diff
@@ -123,3 +125,52 @@ func GetResourceWatcher(ctx context.Context, client *clientv3.Client, key string
...
+						logger.WithError(err).Error("unable to unwrap wrapper")
+						continue
+					}
+
+					c <- store.WatchEventResourceV3{
+						Action:   action,
+						Resource: component,
+					}
+
+				case action == store.WatchError, action == store.WatchUnknown:
```
echlebek, line 166: Maybe a logger statement here too?

echlebek (approved):

backend/store/etcd/watchers.go
ccressent: In the case for `WatchError`, is `event.Err` carrying the actualy error or not? I wasn't 100% sure after a cursory look at the underlying watcher code...

I've updated the PR with logging, but in doubt I did not include `.WithError(event.Err)`. Let me know if I should correct that.

PR #4213, opened 2021-02-24
Add EntityServiceClass

ccressent:
-------

## What is this change?

Add a constant in the `corev2` package representing an entity that is a BSM
service.

The constant has been defined for BSM related work in the enterprise repository,
but it really belongs here with the other constants representing entity types.

Sure, I added one...

Documentation about what a "service" entity class is is covered by the BSM
documentation.

I didn't, because it's just a new constant definition.

## Is this change a patch?

No.


echlebek (approved):
Uncontroversial! :laughing: :shipit:

PR #4214, opened 2021-02-25
Only calculate build info once

echlebek:
This commit fixes a bug where build information is calculated for each
keepalive generated. This generates some additional work for the agent
which should be marginal in all but the most extreme cases.

The bug manifests itself more clearly when using the loadit tool for
load testing, consuming a large fraction of the CPU when 10,000 agents
are loaded.


agent/config.go
```diff
@@ -17,7 +17,7 @@ const (
 
 	// TCPSocketReadDeadline specifies the maximum time the TCP socket will wait
 	// to receive data.
-	TCPSocketReadDeadline = 500 * time.Millisecond
+	TCPSocketReadDeadline = 5000 * time.Millisecond
```
ccressent, line 20: Curious to know what's the rationale being this 10 fold increase?
Not objecting to the change at all, just curious!

cmd/loadit/main.go
```diff
@@ -27,6 +27,7 @@ var (
 	flagKeepaliveInterval = flag.Int("keepalive-interval", agent.DefaultKeepaliveInterval, "Keepalive interval")
 	flagKeepaliveTimeout  = flag.Int("keepalive-timeout", types.DefaultKeepaliveTimeout, "Keepalive timeout")
 	flagProfilingPort     = flag.Int("pprof-port", 6060, "pprof port to bind to")
+	flagPromPort          = flag.String("prom", ":8080", "binding for prometheus server")
```
ccressent, line 30: s/binding/port?

ccressent (approved):
Looks good to merge, provided the linter issue is resolved.

agent/config.go
echlebek: Well, 500 ms seemed a bit tight

cmd/loadit/main.go
echlebek: Nah, binding, it's host+port

amdprophet: Perhaps we should rename the flag to `flagPromBinding` to accurately reflect what the flag is for?

amdprophet (approved):

PR #4215, opened 2021-02-27
Instantiate ResourceTemplates on namespace create

echlebek:
## What is this change?

This commit causes sensu-go to implicitly create resources in newly
created namespaces when they are defined. So far, there are no
ResourceTemplates defined, so no new resources will be created for OSS.

This is supported work for the BSM enterprise feature. We may leverage it in the future and expose it to users, but for now it is internal only.

Yes.

Unit tests only.

## Is this change a patch?

No.


jamesdphillips (approved):
I didn't realize the namespace router was already leveraging the api package. :tada:

PR #4218, opened 2021-03-05
Make SIGHUP a no-op

### LINKED ISSUE #4217: SIGHUP causes the internal daemons to restart
## Expected Behavior
SIGHUP should do nothing, or rotate the event log, in the case of the enterprise product.

## Current Behavior
SIGHUP causes the daemons to restart.

## Possible Solution
Remove the signal handler for SIGHUP in `backend/backend.go`

1. Run sensu-backend
2. Issue SIGHUP

echlebek:
## What is this change?

This commit removes the functionality that causes sensu-go's internal
daemons to restart on SIGHUP. Instead, SIGHUP is now a no-op.

In the enterprise product, SIGHUP is used to rotate the special event
log file, so the restarting behaviour was undesirable.

We don't anticipate anyone should ever want to force an internal daemon
restart, users should instead restart the whole sensu-backend process if
that's what they want to do. So, support for doing this is now removed.

Fixes #4217 

Yes

I ran sensu-backend and sent SIGHUP to it manually several times. It had no effect, as expected.

## Is this change a patch?

No.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ defined.
 round robin checks.
 - Fixed a bug where build information would get calculated for every keepalive
 in OSS builds.
+- Don't trigger internal restart on SIGHUP.
 
 ## [6.2.3] - 2021-01-21
 
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -453,7 +453,7 @@ func Initialize(ctx context.Context, config *Config) (*Backend, error) {
 	return b, nil
 }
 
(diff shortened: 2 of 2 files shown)


ccressent (commented):
This just stops the internal daemon restarts on `SIGHUP` right; log rotation still works with `HUP`?

echlebek:
@ccressent yes, the log rotation signal handler is defined in sensu-enterprise-go, however.

ccressent (approved):

Issue #3734, opened 2020-05-04
Concatenated YAML does not support CRLF


mjiderhamn:
## Expected Behavior
`sensuctl create -f` with multiple YAML files concatenated into one with `---` delimiter should work the same way whether LF or CRLF line endings are used.

## Current Behavior
`sensuctl create -f` can handle YAML files with either LF or CRLF. If it is a concatenated YAML file however, using CRLF results in only the first resource being processed.

Presumably, [cli/resource/parse.go line 37](https://github.com/sensu/sensu-go/blob/master/cli/resource/parse.go#L37) is the culprit.

## Steps to Reproduce (for bugs)
1. Create file with the following contents with CRLF line endings.
```
type: CheckConfig
api_version: core/v2
metadata:
  name: foo
spec:
  command: echo foo
  interval: 100
---
type: CheckConfig
api_version: core/v2
metadata:
  name: bar
spec:
  command: echo bar
  interval: 200
```
2. Import using `sensuctl create -f`

## Your Environment
* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): 5.19.2

echlebek:
Right you are. I knew this hack would probably cause problems for someone, but it flew under the radar until now. :sweat_smile:

echlebek:
We should move to using `bufio.Scanner` for this purpose, as it handles these issues correctly.

PR #4223, opened 2021-03-11
sensuctl: Support CRLF for concatenated YAML

fguimond:
Changed the file parser to use a Scanner to read line by line instead of using a separator substring which had a hard-coded line ending.

## What is this change?

When using sensuctl with a windows encoded (CRLF) concatenated input yaml file only the first resource was applied.

To make sure files are processed the same way, regardless of their line-feed encoding.

Yes, added the following:
- Concatenated YAML files now support CRLF

No

No

Not required.

Using the input file in the zenhub entry, encoded as a Windows file (CRLF) run the following command:

```
sensuctl create -f ./checks-windows.yml 
```

Verify both checks are created by running the following command.

```
sensuctl check list
```

Make sure `foo` and `bar` are present. 

## Is this change a patch?

No


cli/resource/parse.go
```diff
@@ -78,6 +78,31 @@ func Parse(in io.Reader) ([]*types.Wrapper, error) {
 	return resources, err
 }
 
+// splitResources scans the content of the reader and splits the resources.
+// The resources should be separated by a line containing only "---".
```
echlebek, line 82: This reminds me that there is another bug in how yaml is handled here. The yaml specification allows comments on document separators, e.g. https://www.yaml.info/learn/document.html

If you could modify this slightly to split lines that _start_ with `---`, it would take care of that bug too.

cli/resource/parse.go
```diff
@@ -78,6 +78,31 @@ func Parse(in io.Reader) ([]*types.Wrapper, error) {
...
+// An error will be returned if the
```
echlebek, line 83: Looks like you didn't quite finish the thought here :)

echlebek (changes requested):
Looking great overall, but it would be nice to fix up an additional bug (which is untracked): the parser should be able to handle yaml documents like

```
--- # a comment
resource1
--- # another comment
resource2
```

cli/resource/parse.go
fguimond: Added support for comments on the "---" line and updated unit test.

cli/resource/parse.go
fguimond: Oops 😊

echlebek (approved):
:shipit:

fguimond:
Closes #3734 
Fixes #3734

PR #4226, opened 2021-03-17
Enable namespace creation side effects for updates

echlebek:
## What is this change?

This commit changes the namespace CRUD API to implement the same side
effects for create-or-update as currently exist for create. The
namespace POST API has also been updated to make use of the
functionality.

It fixes an unreleased defect that largely applies to an upcoming enterprise feature.

No.

The change is covered by unit tests and also by automated QA.

## Is this change a patch?

No.

### DIFF
--- a/backend/api/namespace.go
+++ b/backend/api/namespace.go
@@ -364,6 +364,9 @@ func (a *NamespaceClient) UpdateNamespace(ctx context.Context, namespace *corev2
 	if err := a.client.Update(ctx, namespace); err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


echlebek:
@fguimond in coming up with some testing instructions for you I found a bug, so I'm putting this PR into draft status until then :)

echlebek:
@fguimond I was wrong about the bug before. I sent you instructions on how to test this with sensu-enterprise-go via Slack.

For onlookers, it's not directly possible to test this change in sensu-go only (aside from unit testing) as it is currently mostly a hook for the enterprise product.

fguimond (approved):
LGTM

PR #4227, opened 2021-03-18
sensuctl: fix shell auto-completion

fguimond:
## What is this change?

Suggestions provided by shell auto-completion included extraneous flags and arguments (e.g. -f, FILE). This fix removes them.

To only show the available auto-completion arguments.

Yes, added.

No.

No.

No new documentation required.

- Setup bash and zsh auto-completion following the instructions here.
- Rebuild sensuctl
- Restart your shell to make sure the latest auto-completion logic is loaded
- Type `sensuctl <tab><tab>`
- Observe the suggested options and make sure there only valid commands and nothing like `-f`, `[blahblah]`

## Is this change a patch?

No.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ round robin checks.
 in OSS builds.
 - Don't trigger internal restart on SIGHUP.
 - Concatenated YAML files now support CRLF.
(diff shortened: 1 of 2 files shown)


echlebek (approved):
Successfully tested :+1:

PR #4228, opened 2021-03-18
Fix parsing of yaml resources with --- prefix

amdprophet:

## What is this change?

Fixes parsing of yaml resources with an initial `---` prefix.

It fixes a regression that I noticed when attempting to spin up a staging stack.

No as it fixes a regression that has not yet been released.

I created test cases that reproduced the error and then re-ran them after fixing the bug.

## Is this change a patch?

No.


fguimond (approved):
LGTM

jamesdphillips (commented):
Just to confirm, this only happens when the first resource in the file started with `---`? What specific error did this lead to? Was it trying to decode an empty string into a resource?

PR #4243, opened 2021-03-29
Make the agentd health router public so it can be extended.

### LINKED ISSUE #4182: Agentd health endpoint response does not include PostgresHealth
## Expected Behavior

Response from agentd /health endpoint is same as output from HTTP API /health endpoint

## Current Behavior

Response from agentd /health endpoint does not include PostgresHealth data

1. Configure a backend with postgres
2. Query backend HTTP API (8080) /health
3. Query agentd (8081) /health
4. Compare results, note the absence of PostgresHealth

## Context

As a Sensu administrator I want to run health checks on both agent and HTTP API. 

Because the currently implemented /health endpoints return a 200 regardless of the cluster health, I need to inspect the response bodies. 

If the response bodies differ between the two /health endpoints, I can't use the same check interchangeably between the two endpoints.

```
$ curl -k --cert /tmp/sensu/agent.pem --key /tmp/sensu/agent-key.pem https://sensu.example.com:8081/health  
{"Alarms":null,"ClusterHealth":[{"MemberID":9861478486968595124,"MemberIDHex":"88db026f7feb72b4","Name":"backend01","Err":"","Healthy":true},{"MemberID":16828500076473183190,"MemberIDHex":"e98ad7a888d16bd6","Name":"backend02","Err":"","Healthy":true},{"MemberID":848052855499371373,"MemberIDHex":"bc4e39432cbb36d","Name":"backend03","Err":"","Healthy":true}],"Header":{"cluster_id":17701109828877155686,"member_id":9861478486968595124,"raft_term":198}}
$ curl -k --cert /tmp/sensu/agent.pem --key /tmp/sensu/agent-key.pem https://sensu.example.com:8080/health 
{"Alarms":null,"ClusterHealth":[{"MemberID":9861478486968595124,"MemberIDHex":"88db026f7feb72b4","Name":"backend01","Err":"","Healthy":true},{"MemberID":16828500076473183190,"MemberIDHex":"e98ad7a888d16bd6","Name":"backend02","Err":"","Healthy":true},{"MemberID":848052855499371373,"MemberIDHex":"bc4e39432cbb36d","Name":"backend03","Err":"","Healthy":true}],"Header":{"cluster_id":17701109828877155686,"member_id":9861478486968595124,"raft_term":198},"PostgresHealth":[{"Name":"postgres","Active":true,"Healthy":true}]}
```

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): 6.2.3
* Operating System and version (e.g. Ubuntu 14.04): Centos 7

fguimond:

## What is this change?

Add a function to replace the agentd's health controller to be used by sensu-enterprise-go to include Postgres health information.

We need to include the Postgres health information from the agentd health endpoint accessible at http://[host]:8081. For this the default health controller needs to be replaced with an extended one that includes the Postgres health.

Fixes #4182 
Closes #4182 

The releated sensu-enterprise-go change will require one but it is a transparent change in this project.

agentd's HealthRouter needs to be accessible from sensu-enterprise-go to be able to swap the HealthControler. I'm not sure if it is the best way to make the HealthRouter accessible however it seems a similar approach was implemented for agentd's middlewares in backend/agentd/middlewares.go. If there is a better way to do it please let me know 😄 

No.

Not required.

With sensu-go run the following commands and make sure the content is identical.

```
curl http://localhost:8080/health
curl http://localhost:8081/health
```

## Is this change a patch?

No.

### DIFF
--- a/backend/agentd/agentd.go
+++ b/backend/agentd/agentd.go
@@ -89,6 +89,7 @@ type Agentd struct {
 	watcher             <-chan store.WatchEventEntityConfig
 	client              *clientv3.Client
 	etcdClientTLSConfig *tls.Config
+	healthRouter        *routers.HealthRouter
 }
 
 // Config configures an Agentd.
@@ -150,10 +151,10 @@ func New(c Config, opts ...Option) (*Agentd, error) {
 	// runtime, so we need this workaround
 	router := mux.NewRouter()
 
-	HealthRouter = routers.NewHealthRouter(
+	a.healthRouter = routers.NewHealthRouter(
 		actions.NewHealthController(a.store, a.client.Cluster, a.etcdClientTLSConfig),
 	)
-	HealthRouter.Mount(router)
+	a.healthRouter.Mount(router)
 
 	route := router.NewRoute().Subrouter()
 	route.HandleFunc("/", a.webSocketHandler)
@@ -504,4 +505,10 @@ func (a *Agentd) EntityLimiterMiddleware(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		next.ServeHTTP(w, r)
 	})
+}
+
+// ReplaceHealthController replaces the default health controller. It
+// can be replaced by an enterprise controller to include more information.
+func (a *Agentd) ReplaceHealthController(controller routers.HealthController) {
+	a.healthRouter.Swap(controller)
 }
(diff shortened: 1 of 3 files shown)


echlebek (changes requested):
Please do not make this change. Creating a global variable will complicate testing. Let's discuss on Slack how we can achieve your goal in another way! :smile:

echlebek (approved):
This looks great!

PR #4245, opened 2021-03-31
Remove validation for partial V3 resources in PATCH API

### LINKED ISSUE #4241: Regression in PATCH API
A bunch of the PATCH API tests are failing when testing in our staging environment with messages similar to:

```
{"message":"invalid EntityConfig: nil metadata","code":1}
```

I believe that a regression was introduced when fixing V3 resource validation in https://github.com/sensu/sensu-go/commit/0ddd51d87fa3ac3b44deb36139b6f607daf9618c.

amdprophet:
## What is this change?

Removes the validation that was happening for V3 resources in the PATCH API.

The initial validation of V3 resources in the PATCH API would always fail as it was trying to validate incomplete resources.

Closes #4241.

No as this is fixing a regression for code that has not yet been released.

Attempting to patch a resource using a non-existent field will succeed & the non-existent field will be ignored.  I'm not sure if there's some way we can validate that the fields are correct. Errors are still returned if the data type of a field does not match the resource field's data type.

I added specs to reproduce the issue and then implemented a fix.

## Is this change a patch?

No.

### DIFF
--- a/api/core/v3/validation.go
+++ b/api/core/v3/validation.go
@@ -2,6 +2,7 @@ package v3
 
 import (
 	"errors"
+	fmt "fmt"
 
 	corev2 "github.com/sensu/sensu-go/api/core/v2"
 )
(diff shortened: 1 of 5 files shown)


echlebek (approved):
Nice work.

PR #4248, opened 2021-04-01
Nil check pointer fields in some watchers before using them

ccressent:


amdprophet (approved):

echlebek (approved):
LGTM. I'm looking forward to us tackling the larger issue as well.

PR #4249, opened 2021-04-01
Add labels to canned events for load testing

echlebek:
## What is this change?

This commit adds a static set of labels to each event produced in load
testing scenarios. The intent is to make the events a little more
realistic in terms of their payload. Labels are indexed by postgres in
the commercial distribution, and may have an impact on performance that
is currently unmeasured.


portertech:
Not sure if we're going to merge this exact PR. We've built a loadit bin w/ this change and tested the negative performance impact. We need to incorporate something like this into our regular testing.

echlebek:
@portertech should we close this out?

PR #4257, opened 2021-04-08
Register Viper aliases after Cobra commands are added

amdprophet:

## What is this change?

Moves the Viper aliases to their own function which is called after all Cobra commands have been added.

Fixes deprecation warnings from showing when the deprecated configuration options aren't being set. On Windows the `handleConfig` function is called multiple times as its used within the start & service commands.

Yes.

No

No.

I reproduced the deprecation warnings on Windows and then ensured they no longer showed after making this change, with and without using a config.

## Is this change a patch?

No.


echlebek (approved):
Nice

Issue #4258, opened 2021-04-08
Asset specific proxy


agoddard:
## Feature Suggestion
For environments which require use of an HTTP proxy for asset retrieval, setting a blanket http_proxy (even with a noproxy exclude list) isn't idea because the install only wants to use the proxy for asset downloading and no other http operations

## Possible Implementation
implement a proxy setting which is only used in asset downloading

jspaleta:
See linked PR #4259   for prototype implementation to extend asset resources with optional attributes to control proxy usage on either an asset or asset build basis.

two new attributes: 
1. proxy : string to set a proxy url to override environment proxy
2. proxy_disabled:  to disable use of proxy if set by environment or by asset attribute.

amdprophet:
Having per-package proxy overrides is not guaranteed to work as the environment variables used can differ based on the language the assets were written in or how they were implemented.

@agoddard and I had a discussion about this last week and ended up coming to the conclusion that this can already be solved with configuration.

If a proxy is needed for most HTTP/HTTPS operations:
1. Set the `HTTP_PROXY` and/or `HTTPS_PROXY` environment variables for the backend process
2. Set the `NO_PROXY` environment variable for any URIs that should not use the proxy (e.g. Etcd URIs)
3. Set `HTTP_PROXY`/`HTTPS_PROXY` to an empty string or set `NO_PROXY` to exclude URIs at the resource configuration level (e.g. checks, handlers, etc.) when those resources should bypass the proxy.

If a proxy is needed for few resource HTTP/HTTPS operations, simply set `HTTP_PROXY`/`HTTPS_PROXY` per resource configuration.

Does this make sense @jspaleta?

PR #4259, opened 2021-04-08
WIP: adding new proxy attribute to asset and using it if enabled

jspaleta:
## What is this change?

Adding a proxy and proxy_disabled attribute to asset, that can be used to override/disable environment http/https proxy settings on a per asset or asset build definition.

Support different http proxy settings for different asset download sources.  Environment provided HTTP_PROXY isn't fined grained enough for all use cases.
Ref: Issue #4258 

Yes.

I'm not sure how to do go about building the appropriate integration tests to make sure proxy support is working.
I guess the best test is to introspect the transport settings used by the http client, but not clear on how to do that in the test framework.

asset protobuf needs to be regenerated.
Asset reference documentation will be required. I'll file a docs PR once this has had some technical review.  

Undated existing unit tests using empty proxy string.

## Is this change a patch?
No


jspaleta:
Okay the asset scoped override is definitely NOT THE RIGHT WAY {tm}.

But this can be refactored into a agent/backend level scoped override after some design think time.

echlebek:
@jspaleta It seems we've determined that we need another way around this problem, so I'll close this PR if that's OK.

PR #4270, opened 2021-04-26
Syntax error in GitHub workflow

amdprophet:

Just fixes a small syntax error in the golangci-lint GitHub workflow.

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -3,12 +3,12 @@ name: golangci-lint
 on:
   push:
     tags:
-      - v*
+      - 'v*'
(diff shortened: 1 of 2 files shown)


echlebek (approved):
Surprised to see the resty upgrade in go.sum :thinking:

ccressent (changes requested):
I don't think there should be any dependency upgrade in this PR?

amdprophet:
For histoircal purposes, go.sum was updated w/go mod tidy because it didn’t match up with go.mod. See  https://github.com/sensu/sensu-go/runs/2443180765 which was run prior to the `go mod tidy` commit.

Issue #4274, opened 2021-04-28
Sensu Agent failing to reconnect correctly when network interruption occurs


guipe:
## Expected Behavior
When network interruption occurs between Sensu agent and Sensu Backend (mTLS), agent can again register to Sensu backend when interruption is finished.

## Current Behavior
When network interruption occurs, Sensu agent cannot register again to Sensu backend.

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Context
When a network disconnection occurs, agent cannot reconnect to Sensu backend correctly. Keepalive checks failed and notify our OnCall teams.
Sensu backend never accept agent reconnection until we restart Sensu backend container.

**Keepalive failed even if:**
- we restart sensu agent
- we are trying to stop agent 1 hour and start again

When a disconnection occurs, agent cannot reconnect to Sensu backend, and we can see multiples CLOSE_WAIT connections on Sensu-backend container:
`tcp      254      0 ::ffff:172.18.0.6:8081  ::ffff:172.18.0.4:32776 CLOSE_WAIT  -`

**Traefik logs**(multiple lines like that, 172.18.0.6 = sensu-backend, 172.18.0.4 = traefik ):
`reverse-proxy              | time="2021-04-28T09:20:15Z" level=error msg="Error during connection: read tcp 172.18.0.4:41328->172.18.0.6:8081: read: connection reset by peer"
`
**Sensu agent logs:**
```
Apr 27 08:00:35 agent1.example.com sensu-agent[12899]: {"component":"agent","level":"info","msg":"connecting to backend URL \"wss://backend.sensu.example.com:443\"","time":"2021-04-27T08:00:35+02:00"}
Apr 27 08:00:35 agent1.example.com sensu-agent[12899]: {"component":"agent","header":"Accept: application/octet-stream","level":"debug","msg":"setting header","time":"2021-04-27T08:00:35+02:00"}
Apr 27 08:00:50 agent1.example.com sensu-agent[12899]: {"component":"agent","error":"read tcp 192.168.228.20:58402-\u003e159.100.240.146:443: i/o timeout","level":"error","msg":"reconnection attempt failed","time":"2021-04-27T08:00:50+02:00"}
```

**Sensu-backend logs:**
No error logs appears

Could you help us on this issue?

## Your Environment
Docker environment:

    Traefik 2.4 (redirect agent mTLS connection to Sensu backend)
    Sensu backend 6.2.7
    Sensu agent 6.2.7

[...]

Issue #4265, opened 2021-04-14
Keepalive status OK with last seen 44 hours ago


guipe:
## Expected Behavior
When agent keepalive timeout is configured with 120 seconds, and command is schedule every 20 seconds, alert must be generated when timeout occurs.

## Current Behavior
keepalive is OK, and last OK check is 44hr. ago.

No alert is generated, no failed state

## Steps to Reproduce (for bugs)

Currently I cannot reproduce.

Last logs retrieved for this agent:
sensu-backend              | {"addr":"172.18.0.4:54774","agent":"GRI","component":"agentd","error":"Connection closed: websocket: close 1001 (going away): bye","level":"warning","msg":"stopping session","time":"2021-04-13T02:37:32Z"}
sensu-backend              | {"addr":"172.18.0.4:51984","agent":"GRI","component":"agentd","error":"Connection error: websocket: close 1006 (abnormal closure): unexpected EOF","level":"warning","msg":"stopping session","time":"2021-04-13T02:37:33Z"}

## Your Environment

    Sensu version used (sensuctl, sensu-backend, and/or sensu-agent):sensuctl: 6.0.0 sensu-agent:6.0.0 sensu-backend: 6.2.5
    Installation method (packages, binaries, docker etc.): docker
    Operating System and version (e.g. Ubuntu 14.04): CentOS 7.8.2003

guipe:
![image](https://user-images.githubusercontent.com/51151700/114686292-ee6c9e00-9d12-11eb-8c4b-039ea2c70c69.png)

guipe:
Thanks for your help!

PR #4277, opened 2021-04-29
Handle signals in sensu-agent main package

amdprophet:
## What is this change?

Moves context & signal handler to the main package for sensu-agent. Passes context down.

We are handling signals in multiple places between the commercial repo and this repo. This should eliminate any race conditions between signal handlers.

Maybe.

Is this something that should be added to the CHANGELOG?

Nope!

I used this to test our staging & QA test refactoring. I saw some strange behaviour where backends had to be restarted before agents registered (after deleting their entity via sensuctl).

May have been related to unreachable code before I fixed it in https://github.com/sensu/sensu-go/pull/4277/commits/9e4614f83d9ee8c1391351276ea932ebebaafdb3.

May be related to these issues:
* https://github.com/sensu/sensu-go/issues/4265
* https://github.com/sensu/sensu-go/issues/4274

## Is this change a patch?

No.


agent/agent.go
```diff
@@ -404,6 +402,7 @@ func (a *Agent) connectionManager(ctx context.Context, cancel context.CancelFunc
 		case <-ctx.Done():
 			// shutting down
 			return
+		default:
```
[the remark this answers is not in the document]
amdprophet: Indeed, rebased.

agent/agent.go
```diff
@@ -408,7 +401,6 @@ func (a *Agent) connectionManager(ctx context.Context, cancel context.CancelFunc
 		// Make sure the process is not shutting down before trying to connect
 		if ctx.Err() != nil {
 			logger.Warning("not retrying to connect")
-			return
```
echlebek, line 411: I think this creates a bug, the return should remain

agent/agent.go
```diff
@@ -482,6 +474,14 @@ func (a *Agent) connectionManager(ctx context.Context, cancel context.CancelFunc
...
 	defer cancel()
 	for {
+		select {
+		case <-ctx.Done():
+			if err := conn.Close(); err != nil {
+				logger.WithError(err).Error("error closing websocket connection")
+			}
+			return
+		default:
+		}
```
echlebek, line 484: This can be written more simply as:
```
if err := ctx.Err(); err != nil {
    if err := conn.Close(); err != nil {
        logger.WithError(err).Error("error closing websocket connection")
    }
    return
}
```

agent/agent.go
amdprophet: Yeah, I'm not sure why that got removed when I rebased.

agent/agent.go
amdprophet: Done!

amdprophet:
@echlebek should be done now.

echlebek (approved):

PR #4278, opened 2021-04-30
Update etcd to v3.5.0-alpha.0

vooon:
## What is this change?

That PR changes import paths for etcd clientv3 and server modules.

That change required to be able to use gRPC newer than 1.26, which is required to add OpenTelemetry-Go instrumentation.

Not sure. Perhaps yes, since that changes etcd version a lot.

Probably nothing.

No complications, simple search and replace job...

No changes in the documentation is required.

## Is this change a patch?

Not a patch.


vooon:
Yes, i will update. I think our team would celebrate when 3.5 outs :)
Other thing - I'm going to release some OpenTelemetry instrumentation patches, which depends on that PR.
For now it's quick and dirty...

calebhailey:
@vooon thanks for this contribution!! Even while we wait to merge this you should go ahead and register your contribution so we can send you some SWAG! https://sensu.io/register-your-contribution

PR #4281, opened 2021-05-07
command: set Pdeathsig

vooon:
## What is this change?

Sets Pdeathsig to subprocesses to be sure of their termination on agent restart.

Usually that isn't something required, but you'll be more sure that nothing left from previous execution.

Possibly no. It's likely that user won't ever notice that.

No.

No.

No doc changes required.

Runs on our systems.

## Is this change a patch?

That change may be backported, but probably not required.

### DIFF
--- a/command/proc_unix.go
+++ b/command/proc_unix.go
@@ -15,7 +15,7 @@ func Command(ctx context.Context, command string) *exec.Cmd {
 
 // SetProcessGroup sets the process group of the command process
 func SetProcessGroup(cmd *exec.Cmd) {
-	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
+	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true, Pdeathsig: syscall.SIGTERM}
 }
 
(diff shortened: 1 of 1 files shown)


echlebek:
Thanks for your contribution @vooon. We would definitely like to merge this. Can you please sign the commit with DCO? (See https://github.com/apps/dco) This is required for contributions to sensu-go.

For Sensu developers: we should test this change on Windows before merging it, to make sure it doesn't create any issues there. It shouldn't apply, but you never know.

echlebek:
Ah thanks, you're right, I didn't notice the file name :)

echlebek (approved):

ccressent:
Unfortunately this PR is breaking the build for every platform except Windows and Linux because [that `.Pdeathsig` field is available on Linux only](https://golang.org/src/syscall/exec_linux.go) but it has been added in a file that's used for all Unices!

The CI for `sensu-go` didn't catch that because it only runs in a Linux environment.

I think a quick fix to be able to include this work would be to create a new file, `proc_linux.go`, that's specific to Linux and change the build tags so that `proc_unix.go` is not used for Linux.

PR #4283, opened 2021-05-11
Fix an RBAC bug where missing roles are fatal

### LINKED ISSUE #4268: Odd behaviour when RBAC roles cannot be resolved
When a `ClusterRoleBinding` refers to a non existent `Role` or `ClusterRole` it leads to odd behaviour in the WebUI and unhelpful error messages in the CLI.

I suspect that `RoleBinding` suffers from the same issue.

## Expected Behavior

The user gets as clear an error as possible when something doesn't go well. At least enough to be able to continue investigating instead of being dumbfounded.

## Current Behavior

- In the CLI, most operations result in an error: `unexpected error occurred during authorization`
- In the WebUI, we're told the session has expired and we are prompted to log in again, _ad infinitum_

## Possible Solution

Generally: warn about "missing references" when creating objects (that can't be an error though, since objects can be created in arbitrary order; think of `sensuctl create` being fed the content of a directory for example)

Specifically: handle this error in the authorizer or its middleware so that it gives a more precise error message instead of the generic "unexpected authorization error" on the CLI and "Session expired" popup in the WebUI.

1. Create a `ClusterRoleBinding` that refers to a non-existent `ClusterRole`
2. Log in as a user that's a subject of the binding and try to perform operations that require any type of permission

## Context

This was observed while testing an OIDC setup. Everything worked once we finally realized that we were giving the Cluster Role Binding we were trying to setup a reference to a `Role` instead of a `ClusterRole`: `role_ref.type = "Role"` instead of `role_ref.type = "ClusterRole"`.

Through a bit of luck, the output of `sensuctl namespace list` hinted at what the problem was and we were able to solve it pretty quickly from there.

## Your Environment

* Sensu version used: `df4abb44d5b73424423bbb7534bc35d47ff14d1e` but all versions affected.

echlebek:
## What is this change?

This CL fixes a bug where the HTTP API returns a 500 status instead of 403 when an RBAC role specified in a RoleBinding RoleRef is missing. The API now returns status 403 and also explains that the referred role is missing.

Fixes #4268

Yes

It's unclear if anything needs to be done in `backend/api` or in the `graphql` layer.

I created tests that fail without the added changes. I also manually tested that a user bound to a nonexistent role would be given a better error message when using sensuctl.

## Is this change a patch?

Yes, but is not targeted for the current release branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ embedded etcd server.
 - Fixed config deprecation warnings from being shown when deprecated config
 options weren't set.
 - Fixed issue with keepalive status remaining in OK status after agent shutdown.
+- Fixed a bug where role bindings that refer to missing roles would cause the
+wrong status to be returned from the HTTP API, and the dashboard to go into a
+crash loop.
 
 ## [6.3.0] - 2021-04-07
 
--- a/backend/apid/middlewares/authorization.go
+++ b/backend/apid/middlewares/authorization.go
@@ -6,6 +6,7 @@ import (
 	corev2 "github.com/sensu/sensu-go/api/core/v2"
 	"github.com/sensu/sensu-go/backend/apid/actions"
 	"github.com/sensu/sensu-go/backend/authorization"
+	"github.com/sensu/sensu-go/backend/authorization/rbac"
 )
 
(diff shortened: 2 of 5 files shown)


jamesdphillips (approved):
image

PR #4284, opened 2021-05-12
Only reference deathsig on linux platform

jamesdphillips:
Addresses https://github.com/sensu/sensu-go/pull/4281#issuecomment-839367779

> Unfortunately this PR is breaking the build for every platform except Windows and Linux because [that `.Pdeathsig` field is available on Linux only](https://golang.org/src/syscall/exec_linux.go) but it has been added in a file that's used for all Unices!
> 
> The CI for sensu-go didn't catch that because it only runs in a Linux environment.
> 
> I think a quick fix to be able to include this work would be to create a new file, proc_linux.go, that's specific to Linux and change the build tags so that proc_unix.go is not used for Linux.

---

### DIFF
--- a/command/proc_group_linux.go
+++ b/command/proc_group_linux.go
@@ -1,5 +1,3 @@
-// +build linux
-
 package command
 
 import (
--- a/command/proc_group_linux.go
+++ b/command/proc_group_linux.go
@@ -1,0 +1,13 @@
+// +build linux
+
+package command
+
+import (
+	"os/exec"
+	"syscall"
+)
+
+// SetProcessGroup sets the process group of the command process
+func SetProcessGroup(cmd *exec.Cmd) {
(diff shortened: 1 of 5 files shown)


command/proc_group_linux.go
```diff
@@ -0,0 +1,13 @@
+// +build linux
```
echlebek, line 1: Because the file is postfixed with _linux.go, it will only build on linux, no need for this build tag.

command/proc_group_windows.go
```diff
@@ -0,0 +1,13 @@
+// +build windows
```
echlebek, line 1: Similarly, no need to include this build tag.

echlebek (approved):

echlebek (approved):

amdprophet:
https://github.com/sensu/sensu-go/pull/4286 should help identify issues like this more quickly in the future.

PR #4286, opened 2021-05-13
Test compilation on all supported platforms

amdprophet:
Signed-off-by: Justin Kolberg <amd.prophet@gmail.com>## What is this change?

It will help us catch platform issues before we vendor sensu-go in our commercial offering.

No.

Nope!

None.

I verified that tests pass & fail for the platforms we expect. The `main` branch is currently broken for some platforms which is why we should see some failures.

## Is this change a patch?

No.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,7 +49,7 @@ jobs:
     steps:
       - checkout
       - go/mod-download-cached
-      - go build ./cmd/<< parameters.binary >>
+      - run: go build ./cmd/<< parameters.binary >>
 
 workflows:
   version: 2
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -17,14 +17,39 @@ jobs:
       GO111MODULE: 'on'
       GOPROXY: 'https://proxy.golang.org'
     resource_class: large
-    parallelism: 2
```
ccressent, line 20: I'm curious why `parallelism: 2` is gone?

ccressent (approved):

.circleci/config.yml
amdprophet: We're not making proper use of parallelism in CircleCI. As a result, that line runs all of our tests twice.

ccressent: That sounds... fishy haha. I'm OK with removing it for now, but I find it suspicious that tests get run twice; maybe there is a problem with our setup or our tests?

amdprophet: If you take a look at https://app.circleci.com/pipelines/github/sensu/sensu-go/1503/workflows/83568d71-7ab1-4ace-8e79-d2fa3a6655e5/jobs/6997/parallel-runs/0?filterBy=ALL you'll notice test 0 & 1. The test runs are identical because we're not splitting the tests using the circleci command.

If we wanted to split tests and run them in parallel, we can follow the documentation at https://circleci.com/docs/2.0/parallelism-faster-jobs/. That being said, I've tested parallel execution previously and it didn't provide much benefit. Running the unit tests & integration tests in separate jobs will make the most difference.

PR #4287, opened 2021-05-14
Implement init --wait flag

amdprophet:
This reverts commit e38df949e902f3ee0b711a4c0285fa2a8072d894.## What is this change?

Implements the `--wait` flag for `sensu-backend init`. Improves context passing & adds a lease to the lock used with etcd seeding.

Yes.

https://github.com/sensu/sensu-docs/issues/2942.

A lot of `sensu-backend start`, `sensu-backend init --wait`, and `Ctrl-C`.

## Is this change a patch?

No.


echlebek (approved):
It seems to work beautifully. Great work!

tadeboro:
I am sorry for commenting on the closed PR, but it seems that the changes in this PR broke the ability to detect the "thing is already initialized, doing nothing" scenario. We use that functionality in the Sensu Go Ansible COllection to correctly detect when something changed (https://github.com/sensu/sensu-go-ansible/blob/9e825c068b7b56b00c0994ca038eb9dad7476624/roles/backend/tasks/start.yml#L16-L25).

If I understand things correctly, the `seeds.ErrAlreadyInitialized` being handled inside the `InitCommand` is the root of the problem. Previously, this error bubbled up to the main function that set the exit code appropriately (https://github.com/sensu/sensu-go/blob/71288f57931765919232db165acb3e71996f1212/cmd/sensu-backend/main.go#L29-L34).

And while I am here, nagging you all, I might as well ask one more question. Are there any nightly builds available that we could use to test against? This way, we could catch such things before they get into a released version of Sensu Go. I know there is a testing repo on packagecloud, but that repo is updated just a few days before release, so it is of limited usefulness to us.

PR #4290, opened 2021-05-18
Introduce execute verb for check #3458

### LINKED ISSUE #3458: Fine grained create verb RBAC for check execute (POST) for individual checks
Goal create Role with a rules that give narrow access to execute a specifc unpublished check.

Currently with existing RBAC I have to grant 'create' verb permissions to the full check resource.

What's needed is to extend the resource-name logic to work with the 'create' verb so that the POST to a specific check execute endpoint can be covered via role rule

mjiderhamn:
## What is this change?

Untested starting point for implementing `execute` verb to fix #3458.

See #3458

Yes

This is only meant as a starting point, to get you motivated to get the implementation moving. Don't know what other bits may be lacking.

No

https://docs.sensu.io/sensu-go/latest/operations/control-access/rbac/ should be updated.

*Untested starting point only*

## Is this change a patch?

No


echlebek:
Thanks for the patch! We'd need to base this on `main` and not release/6.3, which is meant to be stable. As a starting point, it looks good, but appears to need testing fixups.

mjiderhamn:
Rebased on `main`

calebhailey:
This is great. Thanks!

echlebek:
I've converted the PR to a draft to indicate its partial status.

PR #4296, opened 2021-05-20
Refine the storev2.Interface datatype, add enterprise store hooks

echlebek:
## What is this change?

This commit refactors storev2.Interface to be more abstract. Instead of
relying on the concrete wrap.Wrapper datatype, it defines a wrapper
interface and relies on that instead, opening the door for other types
of wrappers to be used.

A configuration option for the storev2.WrapResource method has been
added, to ask the store to prefer using postgresql when available.

This makes storev2 more pluggable. It reduces the need to treat storage as blobs of protobufs.

Automated testing is sufficient here.

## Is this change a patch?

Yes, but is only needed for the main branch.


jamesdphillips (approved):
image

PR #4297, opened 2021-05-20
disambiguate API URL from WebSocket URL

agoddard:

We refer to the WebSocket URL as "backend URL", leading users to incorrectly think the API here is the websocket port. Calling  it "Backend API URL" might help disambiguate it

Unsure, seems too small :sweat_smile: 

There may be a more specific term, like "Backend HTTP API URL" but that might confuse things more.


echlebek (approved):
```
         /|\
       /__| )
     /____| ))
   /______| )))
 /________|  )))
         _|____))
 \======| o o /
```

PR #4300, opened 2021-05-27
considered silenced started at begin timestamp

jamesdphillips:
* The previous behaviour of the silence's `StartSilence` method appears to have been incorrect, if the silence's begin date was equal to the current date, it would return `false`. Fortunately the only consumers of this method were the GraphQL package.
* I've updated the method with what I believe is the correct behaviour but also marked it as deprecated to discourage its usage.
* I've also removed all references to the method from the rest of the project and settled for simply referencing the Begin timestamp on the Silence. (Additionally, improved unit tests where applicable.)

---


echlebek (approved):
:shipit: :dog:

echlebek:
@jamesdphillips please release a new tag for the core/v2 module after merge, I reckon at patch level.

PR #4304, opened 2021-06-01
Use LogLevel as default EtcdLogLevel

amdprophet:

## What is this change?

If `EtcdLogLevel` is not specified, it will use the value of `LogLevel`, or use `debug` if `LogLevel` is set to `trace` as `trace` is not compatible with the etcd logger.

Having to configure log levels with two separate flags is an inconvenience. This change still allows each log level to be configured independently if so desired.

No, this is a change to unreleased code & there's already a changelog for the addition of `EtcdLogLevel`.

https://github.com/sensu/sensu-docs/issues/3066

## Is this change a patch?

No.


echlebek (approved):
A virtuous change.

PR #4305, opened 2021-06-01
Use level instead of cfg.LogLevel, use proper types

amdprophet:

## What is this change?

`cfg.LogLevel` is not set; compare against `level` variable. Use `logrus.Level` types for comparisons. Logrus also outputs its `warn` level as `warning`; use `warn` in that case.

The `main` branch will fail without this change unless `--etcd-log-level` is explicitly set.


echlebek (approved):
I am once again approving

PR #4310, opened 2021-06-08
Windows process timeouts

### LINKED ISSUE #3979: Windows agent doesn't properly handle timeouts
## Expected Behavior

When setting up a timeout value in a CheckConfig, if the check execution takes longer than the timeout value, I expect the "Execution timed out" to appear after the number of seconds set in the timeout value.

## Current Behavior

Timeout happens when the check ends, not taking into account the timeout value

1. Setup a Windows check (Powershell) with a 30s sleep and a 20s timeout
2. Run the check: after a bit more than 30s the "Execution timed out" message appears
3. Setup a Windows check (Powershell) with a 45s sleep and a 20s timeout
4. Run the check: after a bit more than 45s the "Execution timed out" message appears
sensu-bug-img1
sensu-bug-img2

## Context

Trying to monitor a Windows server using checks https://github.com/sensu/sensu-windows-powershell-checks

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): 6.0.0
* Installation method (packages, binaries, docker etc.): packages for agent and sensuctl, docker for Sensu-backend
* Operating System and version (e.g. Ubuntu 14.04): agent runs on Microsoft Windows Server 2012 R2 Datacenter 6.3.9600 Build 9600, sensuctl on macOs 10.15.6, Sensu-backend runs on Docker19.03.12, build 48a66213fe on Centos 7.8.2003

fguimond:
## What is this change?

Windows child processes were not getting killed correctly when processing the command timeout.

Closes #3979 

To prevent having stale processes on the host.

Added the following:
- Fixed issue with Windows agent not handling command timeouts properly

No

No

Create a check running on a Windows agent with the following command:
`PowerShell.exe Write-Host "start sleep"; Start-Sleep -s 20 ; Write-Host "sleep done"`

and a timeout of 2 seconds. When executing the check result should show the execution time to be close to 2 seconds. Also by looking at the Windows task manager make sure the `powershell.exe` process is terminated properly after 2 seconds.

## Is this change a patch?

No

### DIFF
--- a/command/command.go
+++ b/command/command.go
@@ -4,6 +4,8 @@ package command
 import (
 	"bytes"
 	"context"
+	"fmt"
+	"math"
 	"os/exec"
 	"strings"
 	"sync"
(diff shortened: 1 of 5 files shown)


ccressent (approved):

command/command.go
```diff
@@ -160,36 +162,50 @@ func (e *ExecutionRequest) Execute(ctx context.Context, execution ExecutionReque
...
+	go func() {
+		err = cmd.Wait()
+		waitCh <- struct{}{}
+	}()
+
+	// Wait for the process to complete or the timer to trigger, whichever comes first.
+	var killErr error
+	select {
+	case <-waitCh:
+		break
```
echlebek, line 189: No need for break in selects

command/command.go
```diff
@@ -160,36 +162,50 @@ func (e *ExecutionRequest) Execute(ctx context.Context, execution ExecutionReque
...
 	}
 
-	err := cmd.Wait()
-	if timer != nil {
-		timer.Stop()
+	waitCh := make(chan struct{})
+	var err error
+	go func() {
+		err = cmd.Wait()
+		waitCh <- struct{}{}
```
echlebek, line 182: close(waitCh) is a bit more idiomatic here

command/command.go
```diff
@@ -160,36 +162,50 @@ func (e *ExecutionRequest) Execute(ctx context.Context, execution ExecutionReque
...
 		return resp, err
 	}
 
-	err := cmd.Wait()
-	if timer != nil {
-		timer.Stop()
+	waitCh := make(chan struct{})
+	var err error
+	go func() {
+		err = cmd.Wait()
```
echlebek, line 181: It seems this error is unchecked

echlebek (commented):
Looks good but I think the error from cmd.Wait() needs checking

echlebek:
https://www.youtube.com/watch?v=CdqoNKCCt7A

command/command.go
fguimond: Checked when the process is not killed. I don't believe the error is relevant when the process is killed since we already log and display the kill information.

command/command.go
fguimond: Yes, thank you!

command/command.go
fguimond: Addressed, thank you!

echlebek (approved):

PR #4314, opened 2021-06-10
extend time helper functions for filters

jspaleta:
## What is this change?
Adding additional time helper functions for filter expressions.

- add minute() function output int 0-59 
- add second() function output int 0-59
- add seconds_since() function outputs float64, useful when needing to look at age of timestamp relative to current time. 
- added test for each function.

we only expose hour() and weekday() right now, customer interested in minute() granularity.

Yes and a docs PR as well once this is merged.

nope

New documentation is required, but impact is limited just need to provide example filter examples for the additional functions.

wrote some additional golang tests that call the new functions inside an Otto vm with a fixture event with specified timestamp.

## Is this change a patch?
I think is qualifies as a new feature so, not a patch.

### DIFF
--- a/js/js_test.go
+++ b/js/js_test.go
@@ -25,7 +25,47 @@ func TestEvaluateRaceCondition(t *testing.T) {
 
 func TestTimeFunctions(t *testing.T) {
 	timestamp := int64(7323)
-	t.Run("hour() function", func(t *testing.T) {
+	t.Run("failing hour() function", func(t *testing.T) {
+		entity := corev2.FixtureEvent("foo", "bar")
+		entity.Timestamp = timestamp
+		synth := dynamic.Synthesize(entity)
+		params := map[string]interface{}{"event": synth}
+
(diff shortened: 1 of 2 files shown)


js/js_test.go
```diff
@@ -21,3 +22,87 @@ func TestEvaluateRaceCondition(t *testing.T) {
...
+		assert.NoError(t, err)
+		assert.Equal(t, true, result)
+	})
+	t.Run("passing minute() function", func(t *testing.T) {
+		entity := corev2.FixtureEvent("foo", "bar")
+		entity.Timestamp = timestamp
+		synth := dynamic.Synthesize(entity)
+		params := map[string]interface{}{"event": synth}
+
+		result, err := Evaluate("minute(event.timestamp) == 2", params, nil)
```
echlebek, line 84: I don't really understand this use case. The timestamp is going to be a unix time stamp... seconds since the epoch. Is this how the function is intended to be used?

echlebek (approved):

PR #4320, opened 2021-06-16
Fix a memory leak in eventd

echlebek:
## What is this change?

This commit fixes a defer that happens in an unbounded loop, which was
causing a memory leak on main.

No, not a regression.


amdprophet (approved):

echlebek:
Closing, we are going to fix via a different branch.

PR #4321, opened 2021-06-16
Upgrade to Go 1.16.5 & Etcd 3.5, fix memory leak

amdprophet:
## What is this change?

Updates to Go 1.16.5 & Etcd 3.5.0.

Etcd 3.5 has brought significant increases to stability over Etcd 3.3 and does not suffer the more than 50% performance degredation that we saw with Etcd 3.4.

Yes.

@echlebek identified a memory leak in an unreleased Prometheus metric which had been throwing off the performance testing of this work. After the memory leak was fixed the issues went away.

https://github.com/sensu/sensu-docs/issues/3213

Specs & performance testing on the perf cluster.

## Is this change a patch?

Definitely not.


echlebek (approved):
:shipit:

PR #4323, opened 2021-06-18
Separate event/keepalive websockets

### LINKED ISSUE #4273: Use separate websocket connections for keepalive events and check events
To improve stability, we should use separate websocket connections between the backend and agents for keepalive events and check events.

Rationale:
* Keepalive and check event delivery have different stability outcomes.
* If keepalives cannot be delivered because of a backlog of events, then the system can destabilize.
* The backend needs to be able to create backpressure on agents so that it can prevent too many check events from overwhelming it.
* The backend needs to receive timely keepalive events from the agents, so that they won't be marked as failing.

Proposal:
* Add two new routes to the backend's agent API: /ws/events and /ws/keepalives. Preserve the original /ws route for backwards compatibility.
* When an agent connect to /ws/events, set up a session for events. When an agent connects to /ws/keepalives, set up a session for keepalives.
* When an agent connects to /ws, behave as before.
* ~When the agent receives back pressure on /ws/events such that new check events can't be sent, then the agent should report a keepalive warning to the backend that the backend is overloaded. (This is different than the backend's switch timing out for the agent!)~

Additional considerations:
* The agent may connect to two different physical backends, if the connection is mediated using a load balancer. Implementations should not assume that both connections will be to the same backend.
* There will be independent websocket timeouts for each connection. We can't make any assumptions about one connection given the state of another.
* ~Keeaplived will still be subject to backpressure from eventd, by virtue of sending the event there after resetting the keepalive switch. To fully realize the value of this change, we'll need to come up with a way to avoid blocking there, either via dropping events or prioritizing keepalive events over check events.~ (To become its own issue)

ccressent:
## What is this change?

In the backend, we keep the websocket handler on `/` for backwards compatibility, but there are now 2 extra routes, `/events` and `/keepalives`, that are each able to handle agent traffic for events and keepalives respectively.

In the agent, we now create 2 sessions: one dedicated to check requests/events, using the `/events` endpoint and the other dedicated to keepalives using the `/keepalives` endpoint.

The idea is to have separate websockets for check requests/events and keepalives in order to not block keepalives "behind" events when there is backpressure from the backend, potentially leading to unnecessary disconnection/reconnection of the agent. We think it's especially important in cases where the backend is under heavy load: massive disconnection and reconnection of agents distabilize the cluster.

Closes #4273.

Yes.

Opinions from reviewers about other/better approaches, especially on the backend side.

Two major complications:
- we are now doubling the number of connections between a backend and an agent
- the change in the backend's session code is very likely "dumb" but was aiming at minimal change in the old and critical session code

I don't believe any documentation has to be created/altered.

- Local functional testing
- Load testing on our performance cluster

## Is this change a patch?

No.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -93,6 +93,9 @@ var (
 		},
 		[]string{},
 	)
+
+	eventsWebsocketPath, _     = url.Parse("/events")
+	keepalivesWebsocketPath, _ = url.Parse("/keepalives")
 )
 
 func init() {
@@ -370,6 +373,8 @@ func (a *Agent) Run(ctx context.Context) error {
 		a.StartSocketListeners(ctx)
 	}
 
+	// REVIEW(ccressent): I really do not understand the comment below.  Added
+	// by Simon. I suggest deleting altogether?
 	// Increment the waitgroup counter here too in case none of the components
 	// above were started, and rely on the system info collector to decrement it
 	// once it exits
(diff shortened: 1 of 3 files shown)


agent/agent.go
```diff
@@ -457,17 +474,30 @@ func (a *Agent) connectionManager(ctx context.Context, cancel context.CancelFunc
...
+		// probably want to consider the 2 send loops and "continue" if any one
+		// of the 2 terminates? Even then, I don't think that's the correct
+		// behaviour in the face of socket errors.
+		go func() {
+			if err := a.sendKeepalivesLoop(keepalivesConnCtx, keepalivesConnCancel, keepalivesConn); err != nil && err != keepalivesConnCtx.Err() {
+				logger.WithError(err).Error("error sending keepalives")
+			}
+		}()
+
+		if err := a.sendEventsLoop(eventsConnCtx, eventsConnCancel, eventsConn); err != nil && err != eventsConnCtx.Err() {
```
echlebek, line 500: I agree. I don't think they should be independent, but I don't think they should be transitively dependent either.

I think that if the keepalive socket stops working, then we should cancel both connections and retry.

But if the events socket stops working, and the keepalive socket doesn't, to me that means that the system is overloaded and the network link is probably fine. We can retry on the same system, since we know that events are centralized and another backend is unlikely to be better.

I also think that we should eliminate timeouts for event connections.

agent/agent.go
```diff
@@ -370,6 +373,8 @@ func (a *Agent) Run(ctx context.Context) error {
 		a.StartSocketListeners(ctx)
 	}
 
+	// REVIEW(ccressent): I really do not understand the comment below.  Added
+	// by Simon. I suggest deleting altogether?
```
echlebek, line 377: I believe those counters were removed in a previous bugfix or refactoring. I think it's safe to remove the comment.

portertech:
TODO:
- Another performance test run without PSQL batching
- Figure out why PSQL batching negatively impacts top line performance
- **Figure out why we aren't seeing the expected back-pressure behaviour**

ccressent:
The usefulness of doing this was not proven.
We have decided to shelve the idea for now, to be revived later.

PR #4326, opened 2021-06-21
Add sensu_go_keepalives prometheus counter metric

echlebek:
## What is this change?

Adds a prometheus metric for counting keepalive statuses over time.

This will help identify instability due to keepalive failure storms.

Included.


ccressent (approved):

amdprophet (approved):

guipe:
Hi,

that's possible to do the same thing with silence status.

Adds a prometheus metric for entity/checks silence statuses over time.

Thanks

PR #4331, opened 2021-06-23
Don't emit fatal ring errors from keepalived

echlebek:
## What is this change?

This commit changes the behaviour of keepalived such that roundrobin
ring errors are now logged but don't result in an internal restart being
triggered.

Under extreme load, etcd rings can be highly unstable, but it's usually
preferable to continue on when there are lease faults but K/V operations
are still generally working.

The etcd ring facility can emit errors under extreme load, when the lease that the ring is using expires before a key can be written using it. This circumstance has never been observed under regular operation, and is an attempt to bring greater stability to sensu-go when it is under load.

Yes, included.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,10 @@ and this project adheres to [Semantic
 Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
- - There are currently no unreleased changes.
+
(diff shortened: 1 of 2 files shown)


ccressent (approved):

echlebek:
Failed test is unrelated to the PR and is being investigated.

PR #4334, opened 2021-06-25
Bump gopsutil for darwin/arm64 support

amdprophet:

## What is this change?

Bumps gopsutil to the version we use in our internal repository.

Sense Go won't compile on `darwin/arm64` with the current version of gopsutil that this repository is using.

Yes.

The Sensu Go agent, backend and sensuctl all build successfully with `GOOS=darwin GOARCH=arm64`.

## Is this change a patch?

Yes.


echlebek (approved):
A two year newer gopsutil! Hopefully everything still works as expected! :crossed_fingers:

PR #4336, opened 2021-06-25
add restclient apikey support

fgouteroux:
## What is this change?

This commit adds the support of apikey lifecycle for the restclient.

This change is necessary in order to align the cli with the restclient and also as some others modules are using this module like https://github.com/jtopjian/terraform-provider-sensu.

This change has been tested with the forked project [terraform-provider-sensu](https://github.com/fgouteroux/terraform-provider-sensu/commit/0fd372c57dbd3a132839e2d1a16151d48ca7ecc3)

### DIFF
--- a/cli/client/apikey.go
+++ b/cli/client/apikey.go
@@ -1,5 +1,13 @@
 package client
 
+import (
+	"encoding/json"
+	"net/http"
+	"strings"
+
+	corev2 "github.com/sensu/sensu-go/api/core/v2"
+)
+
 // PostAPIKey sends a POST request with obj as the payload to the given path
 // and returns the location header of the key.
 func (client *RestClient) PostAPIKey(path string, obj interface{}) (string, error) {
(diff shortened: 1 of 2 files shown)


echlebek (commented):
We need automated tests for this work. The package overall is poorly tested, but that will only get worse if we don't add tests for new code. Even if the code works as-is, we'll be unprotected from regressions.

Let's add some tests that are similar to those written for authentication, so that we can move forward with confidence.

fgouteroux:
> We need automated tests for this work. The package overall is poorly tested, but that will only get worse if we don't add tests for new code. Even if the code works as-is, we'll be unprotected from regressions.
> 
> Let's add some tests that are similar to those written for authentication, so that we can move forward with confidence.

Yes, I understand. I added some tests, is that enough ?

echlebek (approved):

echlebek:
Thank you for the great work @fgouteroux!

fgouteroux:
> Thank you for the great work @fgouteroux!

Thanks for the review @echlebek

PR #4339, opened 2021-06-25
Guard against nil etcd response header

amdprophet:
Signed-off-by: Francis Guimond <francis@sensu.io>## What is this change?

Prevents `sensuctl cluster health` from panic when running against a etcd cluster that has lost quorum.

No. It's a fix for unreleased code.

No new documentation required.

Copied the new sensuctl to our QA stack and ran against etcd without quorum.

## Is this change a patch?

Yes.

### DIFF
--- a/cli/commands/cluster/health.go
+++ b/cli/commands/cluster/health.go
@@ -1,6 +1,7 @@
 package cluster
 
 import (
+	"errors"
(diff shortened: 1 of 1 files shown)


echlebek (approved):
:shipit:

PR #4342, opened 2021-07-05
Ringv2: update deps

vooon:
## What is this change?

That PR removes dependency on old cron package, the only place not yet updated to cron/v3.
And fixes `UnPath()` (my eyes was being captured by strange indexes, so i wrote test unit to prove me wrong, but it wasn't).

- Removes outdated dependency
- Fixes bug in a function without any calls...

Probbaly no, because there is no real changes.

No.

No.

No.

Compilled, ran tests.

## Is this change a patch?
Well yes, function fix is a patch, but i do not see any calls from whole codebase to it.

### DIFF
--- a/backend/ringv2/interface.go
+++ b/backend/ringv2/interface.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"errors"
 
-	"github.com/robfig/cron"
+	"github.com/robfig/cron/v3"
 )
 
 // Interface is the interface of a round-robin ring.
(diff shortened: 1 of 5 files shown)


vooon:
Appveyor failure looks unrelated.

jamesdphillips (changes requested):
This looks great, thanks for the submission!

If you don't mind, would you revert the additional test? I think it would be unforntunate to add an additional dependency just to test a function that is currently unused. (Or, otherwise update the test omitting the random elements, maybe we use fuzzing in the future if we want to test a wider range of possible inputs.)

vooon:
@jamesdphillips randstr removed.

jamesdphillips (approved):
Beaut

fguimond (approved):
lgtm

PR #4343, opened 2021-07-05
cli: update survey to survey/v2

vooon:
## What is this change?

That PR update cli to use survey/v2 package.

Removes dependency on survey and survey.v1.

Perhaps no, because that's just a dependency update.

No.

No.

No.

Patched version used on testing site, but not much tested because most of things automated trough API.

## Is this change a patch?
No.

### DIFF
--- a/backend/cmd/init.go
+++ b/backend/cmd/init.go
@@ -7,7 +7,7 @@ import (
 	"os"
 	"time"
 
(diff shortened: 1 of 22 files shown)


amdprophet (approved):

amdprophet:
Looks good! Thanks for the contribution.

PR #4346, opened 2021-07-05
Fix backend init exit code when store already initialized

amdprophet:

## What is this change?

Fixes a regression that shipped with 6.4.0 where `sensu-backend init` returns an exit code of 0 when the store is already initialized.

This is a regression causing issues with configuration management.

Yes.

Tested with a new store and an existing store. New store initialization exits with 0 as expected. Existing store initialization exits with 3 as expected.

## Is this change a patch?

Yes.


fguimond (approved):
Lgtm

Issue #4042, opened 2020-10-07
Keepalive event.check metadata should be configurable


cwjohnston:
## Feature Suggestion

As an operator, I can use configuration to add labels and annotations to keepalive events.

## Possible Implementation

1. Expose agent metadata as check metadata in keepalive events.
2. Add keepalive metadata concept to entities?

## Context

Currently, I can configure labels and annotations on checks, and that information is available as metadata in the `check` scope of the event. I can use the API to request events based on labels, e.g.

```
curl -s -X GET http://sensu.example.com:8080/api/core/v2/namespaces/default/events -H "Authorization: Key $SENSU_API_KEY" -G --data-urlencode 'labelSelector=checktype == "servers" && instance == "instance1"'
```

Compared to events produced by scheduled checks, keepalive events are something of a special case. The backend creates these events them on behalf of an agent -- and near as I can tell there's no affordance for backend to include labels and annotations in the check scope.

This means that to get the complete picture of an entity's state, a second API call is necessary to find any keepalive events:

```
curl -s -X GET http://sensu.example.com:8080/api/core/v2/namespaces/default/events -H "Authorization: Key $SENSU_API_KEY" -G --data-urlencode 'labelSelector=instance == "instance1"' --data-urlencode 'fieldSelector=event.check.name == "keepalive"'
```

calebhailey:
Related support ticket: https://secure.helpscout.net/conversation/1176136379/21329?folderId=1931919

calebhailey:
Related support ticket: https://secure.helpscout.net/conversation/1145187127/20910?folderId=1931916

agm650:
would it be "acceptable" to add new fields in the agent configuration ? something like in agent.yml:
```yaml
keepalive-labels:
    checktype: servers

keepalive-annotations:
    test: value
```

If so, it should be easy to add to the agent, and I think I can try to do a PR for that.

portertech:
@ccressent to look into this.

PR #4357, opened 2021-07-15
Metadata configuration for keepalive

agm650:
Signed-off-by: Luc Dandoy <luc.dandoy@gmail.com>

## What is this change?

This Pull Request is adding the possibility to configure metadata (labels & annotations) for keepalive event

This will solve issue #4042

Yes I add an entry for the modification

I'm not sure if the modification is "efficient" but it's working properly on my tests environment

I will have to open a PR to document the addition of 2 new flags for sensu-agent

- Compile the agent
- Launch sensu-agent without any labels and check at backend level, that keepalive is correctly recieved on backend
- Stop sensu-agent and delete the entity on backend side
- Launch again the agent with labels & annotations
- check that annotations & labels are visible in the event received by the backend.

## Is this change a patch?

No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 - Added `ignore-already-initialized` configuration flag to the sensu-backend
 init command for returning exit code 0 when a cluster has already been
 initialized.
+- Added `keepalive-labels`and `keepalive-annotations` configuration flags to the sensu-agent
 
 ### Changed
 - When keepalived encounters round-robin ring errors, the backend no longer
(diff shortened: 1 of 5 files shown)


echlebek:
Can you help me to understand the value of this feature? Right now users can already add metadata to keepalives via the entity, and these are set via the --labels and --annotations flags (or same in the configuration file).

agm650:
Well, your keep alive will have the same labels as your entity that's right. But in a specific case where you want to add a label to a keepalive (mainly because I cannot do "or" in the event filtering), I need this kind of feature.
I had a discussion with Cameron on this topic (request id #1176136379 which lead to the creation of #4042)

echlebek:
@agm650 I've asked @calebhailey to review this from a product standpoint, thanks!

ccressent:
This work is being pulled in as part of #4602.
Merci Luc!

PR #4361, opened 2021-07-21
Add support for javascript mutators

### LINKED ISSUE #4354: Add support for new Javascript Mutators
## Feature Suggestion

Support javascript mutators. Add a new string field to the OSS mutator resource type, `core/v2.Mutator.Spec.Type` with a default value of "pipe" (matching `core/v2.Handler.Spec.Type`). Add an OSS built-in mutator type to the `pipeline` package, "javascript", using the [Otto VM](https://github.com/robertkrimen/otto) for event data manipulation. Users can now use javascript to efficiently mutate Sensu Go event data. 

## Possible Implementation

Otto VM should be seeded with a single variable: `event`, which should contain a deserialized event object (a valid `core/v2.Event`). The `eval` should end with a return statement (the returned value should be a string), otherwise the default return value is `JSON.stringify(event)`. The Otto VM should be seeded with a single variable: `event`, which should contain a deserialized event object (a valid `core/v2.Event`). The `eval` should end with a return statement (the returned value should be a string), otherwise the default return value is `JSON.stringify(event)`. Otto VM errors – including errors raised in the eval statement (e.g. try/catch statements) will result in a failed mutation, terminating the mutator's pipeline execution. Otto VM errors should be logged by the Sensu Backend.

```yaml
---
type: Mutator
api_version: core/v2
metadata:
  name: example
spec:
  type: javascript
  eval: |
    var mutatedEvent = {};
    # do mutation here…
    return mutatedEvent; 
```

## Context

Proposal: https://docs.google.com/document/d/1rWDJoSFhpXRe9PUB8i7SVNA_eggJ6v5L8nJB7UaEkKc/edit#

echlebek:
## What is this change?

```
    This commit adds a Type field to core/v2.Mutator which can be used to
    specify if the mutator is a pipe mutator, or a javascript mutator. If
    the Type field is unset, then the default is pipe.
    
    Javascript mutators can operate on a pre-defined object called "event".
    Any mutations made to this object will be propagated through the
    pipeline. Alternatively, a wholly new value can be returned, as long
    as it of type string.
    
    Runtime assets are fully supported, so long as they are valid javascript
    assets.
    
    Timeouts are supported, but tight loops may block the runtime system. Be
    mindful of the mutator code you write.
    
    Environment variables are supported and are made available as global
    variables in the execution environment.
    
    Secrets are not supported.
    
    Minimal example:
    
    api_version: core/v2
    type: mutator
    metadata:
      namespace: default
      name: doge
    spec:
      command: 'event.check.labels["doge"] = "woof";'
      type: javascript
```

Closes #4354 

Included

I did a bit of refactoring to support testing runtime assets.

@hillaryfraley documentation updates will be required. Perhaps we should schedule a 1:1?

The work is well-covered by unit and integration tests.

## Is this change a patch?

No.


portertech:
@echlebek need to add the `eval` field, matching BSM rule template, for the Javascript. We do not want to overload `command`.

ccressent (approved):

PR #4363, opened 2021-07-22
Customise settings in loadit for username, password, and test hostname.

skorn:
## What is this change?

Customise settings for user, password, and test hostname.
This does change the default hostname from random uuid, to: test-host-1, test-host-2, test-host-n

The user and password change were required to be able to connect if the default credentials are not being used.
The test host allows easily checking on events and such in api.

### Added
- Added -user and -password flags for loadit to set credentials if required (still defaults to default settings).
- Changed test entity name for loadit to "test-entity-1"... "test-entity-n". The "test-entity" can be customized with new -base-entity-name flag

None

So far no complications.

No documentation for loadit

Verified against my test sensu cluster.

## Is this change a patch?

Yes.

### DIFF
--- a/cmd/loadit/main.go
+++ b/cmd/loadit/main.go
@@ -11,7 +11,6 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/google/uuid"
 	"github.com/sensu/sensu-go/agent"
 	"github.com/sensu/sensu-go/types"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


echlebek (changes requested):
Thanks for the pull request, I agree that we should allow setting the username and password. I also like setting a prefix for the name, because it will allow us to avoid having randomized agent names and instead have numbered agent names, while avoiding conflicts with multiple loadit processes by setting a different prefix for each process.

However, can you please change the base of the PR to main, instead of release/6.4? Thanks!

Issue #4368, opened 2021-07-27
Update embedded etcd default timeout values


calebhailey:
## Feature Suggestion

Increase the default `--etcd-election-timeout` from 100ms => 500ms, and `--etcd-heartbeat-interval` from 1000ms to 5000ms. 

NOTE: our customer reliability engineering team has been advising several customers to adjust these settings for several months, with good results. We're considering these to be safe defaults, and recommended settings for all new Sensu Go clusters (where external etcd is not possible). 

## Possible Implementation

Increase the default values as mentioned above. 

```yaml
etcd-election-timeout: 5000
etcd-heartbeat-interval: 500
```

We will highlight this change in the release notes and documentation, and customers can restore the old defaults with a simple configuration change (as needed). 

## Context

If the Etcd cluster is constantly having leader elections, it's always going to be overrun.

portertech:
The "work" is done, however, before we can merge this, we must understand the effects of the change.

c-kruse:
So far the only thing I've ran into that may be worth bringing up is that the minimum lease TTL is loosely tied to the etcd election timeout. This is what it looks like is causing integration tests to fail - the keepalived integration test is using a [keepalive timeout of 5s](https://github.com/sensu/sensu-go/blob/main/backend/keepalived/integration_test.go#L73) which is exactly the election timeout is set to.

etcd currently sets the minimum lease ttl to 1.5x election timeout https://github.com/etcd-io/etcd/blob/main/server/etcdserver/server.go#L343

PR #4369, opened 2021-07-27
Increase embedded etcd heatbeat interval and election timeout defaults

portertech:
Increased the embedded etcd heatbeat interval and election timeout defaults.

Closes https://github.com/sensu/sensu-go/issues/4368

## What is this change?

Increased the embedded etc heatbeat interval and election timeout defaults, making etcd more latency tolerant for our use-case.

Quality network an more importantly storage is harder to come by then expected. The etcd default values for the heatbeat interval and election timeout are too low, too sensitive to network and storage latency. etcd elections are expensive and destructive, increasing the default values helps avoid them while still not being too high that leader failover takes too long.

Yes.

No.

No.

No.

Local build.

## Is this change a patch?

No.


echlebek (approved):
yolo

echlebek (changes requested):
@portertech this seems to be causing integration test failures. I think we'll need to spend a bit more time on this PR.

ccressent:
For some reason I don't quite understand, this change makes it so the entity doesn't have enough time to deregister [in that part of the test](https://github.com/sensu/sensu-go/blob/71cb3f4d70d8b2b7b02aa3e616f43626a7bb3e4a/backend/keepalived/integration_test.go#L133-L149).

If the timer expires before the entity has been deregistered, the test fails.
If the timer lasts for long enough to have the following line logged, the test passes:
```
time="2021-09-30T12:18:24-07:00" level=info msg="entity deregistered" component=keepalived entity=entity1
```

Bumping [this test's keepalive timeout](https://github.com/sensu/sensu-go/blob/71cb3f4d70d8b2b7b02aa3e616f43626a7bb3e4a/backend/keepalived/integration_test.go#L72) from 5 to 10 seems to do the trick on my machine.

As it stands I'm not committing this change to the test because I don't quite understand how changing the etcd heartbeat form 100ms to 500ms would have caused this test to fail in the first place... My main concern is that there _might_ be other unintended consequences we are just not grasping yet?

portertech:
Until we understand the impact of changing this default value, we aren't comfortable merging this pull-request. Many users already use these values (of greater values), it's not clear what side-effects they are experiencing (likely unaware). Changing the milestone.

c-kruse (commented):
@portertech with the keepalive integration test failure explained by etcd's minimum lease ttl == 1.5x election timeout, do we feel good about moving forward with your idea to compromise on lower defaults that still allow for our 5 second min keepalive ttl - 330ms heartbeat with a 3300ms election timeout?

It looks like those values are consistently passing integration tests on my machine. It looks like for any value for heartbeat > 333ms, the keepalive integration test will fail - which makes sense if you look at etcd's minTTL implementation `time.Duration((3*cfg.ElectionTicks)/2) * heartbeat`

echlebek:
This PR has aged considerably. I'm closing it, we can revisit it at a later date if need be.

Issue #4352, opened 2021-07-14
New core/v2.Pipeline resource, Store & HTTP API


calebhailey:
## Feature Suggestion

Add support for a new Pipeline resource (OSS) for configuring event processing "workflows" including: sets of filters + mutator + handler configuration options. All pipeline primitive (e.g. handler) references require an API version and resource type, this allows a pipeline to reference a commercial handler resource type (i.e. “TCPStreamHandler”). The Pipeline resource is intended to eventually [replace "Set Handlers" (core/v2.Handler.Spec.Type = "set")](https://gist.github.com/calebhailey/1b628fbbcfe5d850abb673fa14bd9380), and add support for a more flexible references to new handler types (i.e. instead of trying to wedge forthcoming `TCPStreamHandler` references into `core/v2.CheckConfig.Spec.Handlers` (a string slice field), we will add a new `core/v2.CheckConfig.Spec` field called `Pipelines` (see below for more information). 

## Possible Implementation

```yaml
---
type: Pipeline
api_version: core/v2 
metadata:
  name: ops
spec:
  workflows:
  - name: elastic
    filters:
    - api_version: core/v2
      name: is_incident 
    - api_version: core/v2
      name: state_change_only 
    mutator:
      api_version: core/v2
      name: add_labels
    handler:
      api_version: commercial/v1
      type: TCPStreamHandler
      name: logstash
```

## Context

Proposal: https://docs.google.com/document/d/1rWDJoSFhpXRe9PUB8i7SVNA_eggJ6v5L8nJB7UaEkKc/edit#

portertech:
TODO:

- [ ] Create the new core/v3.Pipeline resource
- [ ] Update the store(s) for Pipelines
- [ ] Create the new Pipeline HTTP API

Acceptance criteria:

- [ ] As a user, I can CRUD Pipelines (but I can't yet use them within the product)

amdprophet:
The store & HTTP API still need to be updated.

amdprophet:
@ccressent this can be closed now, correct?

ccressent:
Yep!

PR #4371, opened 2021-07-27
Add core/v3.Pipeline resource

amdprophet:
## What is this change?

Adds a new Pipeline resource.

Part of what's needed to close https://github.com/sensu/sensu-go/issues/4352.

Yes.

No.

I was initially going to implement support in `backend/apid` for this resource but I discovered that it doesn't currently support `core/v3` resources. @echlebek will be working on adding support separately.

New documentation will be required once this feature is closer to completion.

Added new tests, existing tests pass.

## Is this change a patch?

No.


api/core/v3/pipeline.proto
```diff
@@ -0,0 +1,55 @@
...
+option (gogoproto.sizer_all) = true;
+option (gogoproto.testgen_all) = true;
+
+// Pipeline represents a named collection of pipeline workflows.
+message Pipeline {
+  // Metadata contains the name, namespace, labels and annotations of the
+  // pipeline.
+  sensu.core.v2.ObjectMeta Metadata = 1 [ (gogoproto.jsontag) = "metadata", (gogoproto.nullable) = true ];
+
+  // Pipelines contains one or more pipelines.
```
ccressent, line 22: I think this should read "Workflows contains one or more pipelines"?

amdprophet: "Pipelines contain one or more workflows".

amdprophet: Updated to `Workflows contains one or more pipeline workflows`.

ccressent (approved):

PR #4382, opened 2021-08-11
Event bytes per second

### LINKED ISSUE #4298: Add event size histogram metrics
## Feature Suggestion

Instrument the backend to measure bytes/s received/processed by `eventd`. 

## Possible Implementation

We've done this before! https://github.com/sensu/sensu-go/blob/6555322be77ff04b2a176a13be694868e408fa09/backend/eventd/eventd.go#L89-L96

1. Measure event bytes/s in two places, agentd sessions, and the store. Use metric labels to indicate if the event(s) will be persisted etc (contains a check resource).

   - Agent event bytes/s for events w/ check data (aggregated)
   - Agent event bytes/s for events w/ metrics data (aggregated)
   - Agent event bytes/s for events w/ check+metrics data (aggregated)
   - Store event bytes/s for events w/ check data
   - Store event bytes/s for events w/ metrics data
   - Store event bytes/s for events w/ check+metrics data

2. Add event size histogram metrics to metrics logger (see #4299)

## Context

We have observed that some customers are processing unusually large event payloads, in which case certain deployment/architecture considerations may need to be taken to tune the system for optimal performance. Evaluating events/s throughput with the additional context of bytes/s will help identify these opportunities if a user experiences degraded performance.

amdprophet:
## What is this change?

Adds event size, in bytes, Prometheus metrics for check, metric, and check+metric events in both the agentd & store (etcd/postgres) packages.

Closes #4298.

Yes.

I don't believe that we document all of the metrics exposed by the `/metrics` endpoint, so no new documentation should be required.

Deployed to the perf cluster.

## Is this change a patch?

No.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,9 @@ a decimal multiplier value.
 - Added ProcessedBy field to check results. The ProcessedBy field indicates which
 agent processed a particular event.
 - Added `core/v2.Pipeline` resource for configuring Pipeline resources.
-- Added `pipelines` field to `Check` and `CheckConfig`
+- Added `pipelines` field to `Check` and `CheckConfig`.
+- Added `sensu_go_agentd_event_bytes` & `sensu_go_store_event_bytes` summary
+metrics to the `/metrics` endpoint.
 
 ### Changed
 - When deleting resource with sensuctl, the resource type will now be displayed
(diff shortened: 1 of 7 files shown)


ccressent (approved):
Looks good to me, although I'm not 100% sure on the details for Prometheus summary vectors. I know there are quite a few subtleties with them and histograms.

PR #4389, opened 2021-08-18
Cleanup asset downloads

### LINKED ISSUE #4180: Agent does not clean up failed asset downloads
## Expected Behavior

Sensu agent gracefully handles asset download failures by deleting temporary files when a download fails.

## Current Behavior

Sensu agent does not delete temporary files when a download attempt exceeds a timeout.

## Context

Given a system with limited disk space and sufficiently slow or unreliable internet, it's possible for Sensu agent to fill the disk.

A customer reports that agent filled their disk with incomplete downloads and the agent log reflects messages indicative of a timeout being reached:

```
error getting assets for event: error fetching asset: Get https://example.com/asset.tgz: context canceled
```

2021-01-20 at 4 44 PM

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): 6.0.0
* Operating System and version (e.g. Ubuntu 14.04): Windows

ccressent:
## What is this change?

In some cases, it was possible for the agent to leave temporary files behind
when downloading assets. That shouldn't be the case anymore now.

Fix #4180

Yes, added.

No.

Issue was inaccurate. Otherwise no.

No documentation change needed.

Local testing to the extent possible.

## Is this change a patch?

No.

### DIFF
--- a/asset/fetcher.go
+++ b/asset/fetcher.go
@@ -124,21 +124,28 @@ func (h *httpFetcher) Fetch(ctx context.Context, url string, headers map[string]
 		return nil, fmt.Errorf("can't open tmp file for asset: %s", err)
 	}
 
+	cleanup := func() {
+		tmpFile.Close()
+		_ = os.Remove(tmpFile.Name())
+	}
+
 	buffered := bufio.NewWriter(tmpFile)
 	if _, err = io.Copy(buffered, resp); err != nil {
+		cleanup()
 		return nil, fmt.Errorf("error downloading asset: %s", err)
 	}
 	if err := buffered.Flush(); err != nil {
+		cleanup()
 		return nil, fmt.Errorf("error downloading asset: %s", err)
 	}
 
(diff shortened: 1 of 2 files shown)


asset/fetcher.go
```diff
@@ -124,21 +124,28 @@ func (h *httpFetcher) Fetch(ctx context.Context, url string, headers map[string]
 		return nil, fmt.Errorf("can't open tmp file for asset: %s", err)
 	}
 
+	cleanup := func() {
+		tmpFile.Close()
+		_ = os.Remove(tmpFile.Name())
+	}
+
```
echlebek, line 131: This is good, but a pattern that is more resistant to future regressions might be:

```
defer func() {
    if err != nil {
        tmpFile.Close()
        _ = os.Remove(tmpFile.Name())
    }
}()
```

Where err is a named return variable. Then the function can return from anywhere, without the risk of someone forgetting to call cleanup().

amdprophet (approved):

PR #4391, opened 2021-08-19
Support cluster admin API keys on cluster init

### LINKED ISSUE #4162: Add support to initialize a cluster with an API KEY
## Feature Suggestion

Add support for optionally supplying an API Key via `sensu-backend init`. If paired with #4012 this would simply new-cluster provisioning. New clusters could be initialized and seeded with default configuration with less "orchestration"; i.e. instead of initializing the cluster, configuring `sensuctl` with username and password, and then executing resource management commands w/ sensuctl... you can eliminate the `sensuctl` configure step and just execute `sensuctl` commands with two flags `--api-key=` and `--api-url=`. 

## Possible Implementation

- Accept any valid UUIDv4 via a new `--sensu-cluster-admin-apikey` flag and/or `$SENSU_CLUSTER_ADMIN_APIKEY` environment variable. 
- Like other `sensu-backend init` functions, this cluster admin API Key should only be granted when initializing a fresh cluster

## Context

Monitoring as code for new cluster provisioning.

echlebek:
## What is this change?

Adds the ability to init the cluster with an API key.

Closes #4162

Included

@hillaryfraley we should update the documentation to reflect that API keys are now supported by sensu-backend init.

Manual testing in interactive and CLI mode.

## Is this change a patch?

No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ a decimal multiplier value.
 agent processed a particular event.
 - Added `core/v2.Pipeline` resource for configuring Pipeline resources.
 - Added `pipelines` field to `Check` and `CheckConfig`
+- Added API key support for admin user at cluster init time.
 
 ### Changed
 - When deleting resource with sensuctl, the resource type will now be displayed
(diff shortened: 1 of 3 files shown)


fguimond (approved):
lgtm

PR #4409, opened 2021-09-09
Use temp dirs in sensuctl tests which create configs

amdprophet:

## What is this change?

Creates a temporary directories for any sensuctl tests which result in a config being created.

Running tests prior to this would result in dirtying the git state as it was using the local directory whenever `config-dir` was empty.

No.

Tests pass and no longer dirty the git state.

## Is this change a patch?

No as this only affects the `main` branch at the moment.


fguimond (approved):
Didn't know about this, nice addition 👍

PR #4410, opened 2021-09-13
Update package dgrijalva/jwt-go to golang-jwt/jwt

sspaink:
Signed-off-by: Sebastian Spaink <sspaink@influxdata.com>

## What is this change?

The package https://github.com/dgrijalva/jwt-go has been archived and contains a CVE (https://github.com/advisories/GHSA-w73w-5m7g-f7qc), the project has migrated to https://github.com/golang-jwt/jwt. This pull request just replaces the dependency to use the new project following the migration guide: https://github.com/golang-jwt/jwt/blob/main/MIGRATION_GUIDE.md this should be backwards compatible.

My motivation for this change is that the Telegraf project has a output plugin for [Sensu](https://github.com/influxdata/telegraf/tree/master/plugins/outputs/sensu) and uses sensu-go which is indirectly pulling in `dgrijalva/jwt-go`.

Yes, changelog has been updated

Not sure to be honest, but I did run `./build.sh`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic
 Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Security
+- Migrated the package dgrijalva/jwt-go to golang-jwt/jwt to address [CVE-2020-26160](https://github.com/advisories/GHSA-w73w-5m7g-f7qc)
+
 ## [6.4.1] - 2021-08-24
 
 ### Added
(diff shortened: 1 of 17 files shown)


sspaink:
@echlebek I'm not entirely sure why the linters are failing with `Can't run linter unused: buildir: failed to load package : could not load export data: no export data for \"github.com/golang-jwt/jwt/v4\""`, any idea why?

echlebek:
@sspaink I'm not sure why either. It could be a network issue, or availability problem with the Go module proxy. I'll try re-running the workflow, and if it fails again will dig deeper.

echlebek:
Looks like the core/v2 module needs to be updated, I can do that.

echlebek (approved):

echlebek:
@sspaink many thanks for the PR. We'll be releasing new tags for the `api/core/v2` and `types` modules soon, so that you can make use of this downstream if you are relying only on the domain types.

We'll be releasing version 6.5 of sensu-go in the next few weeks, so if you were depending on the entire project, it will take a bit longer.

PR #4413, opened 2021-09-15
Add deadline to etcd health field resolver

jamesdphillips:
Closes https://github.com/sensu/sensu-enterprise-go/issues/1781

---

* Previously no deadline was provided when querying etcd health from the GraphQL service,
  this meant that in cases where the health was irretrievable the http server's writetimeout would
  trigger before the query completed. In the web client this led to the offline banner being
  displayed briefly and the health dialog being stuck in an infinite loading state.
* I've also updated the GraphQL endpoint with a deadline and now respond with a 504 if the service
  takes too long to finish resolving the query. This change should make it easier to capture, and possibly
  retry the request later.

---

### DIFF
--- a/backend/apid/graphql/health.go
+++ b/backend/apid/graphql/health.go
@@ -1,10 +1,13 @@
 package graphql
 
 import (
+	"context"
 	"strconv"
+	"time"
 
 	corev2 "github.com/sensu/sensu-go/api/core/v2"
 	"github.com/sensu/sensu-go/backend/apid/graphql/schema"
+	"github.com/sensu/sensu-go/backend/store"
 	"github.com/sensu/sensu-go/graphql"
 	"go.etcd.io/etcd/api/v3/etcdserverpb"
 )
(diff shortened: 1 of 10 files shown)


backend/apid/graphql/health.go
```diff
@@ -18,11 +21,16 @@ var _ schema.EtcdClusterMemberHealthFieldResolvers = (*etcdClusterMemberHealthIm
...
+type clusterHealthImpl struct {
+	healthController EtcdHealthController
+}
 
 // Etcd implements response to request for 'etcd' field.
-func (r *clusterHealthImpl) Etcd(p graphql.ResolveParams) (interface{}, error) {
-	return p.Source, nil
+func (r *clusterHealthImpl) Etcd(p schema.ClusterHealthEtcdFieldResolverParams) (interface{}, error) {
+	timeout := time.Duration(p.Args.Timeout) * time.Millisecond
+	ctx := context.WithValue(p.Context, store.ContextKeyTimeout, timeout)
```
amdprophet, line 31: Can we use `context.WithTimeout(p.Context, timeout)` here instead or is the store expecting a context with a value for `store.ContextKeyTimeout`?

jamesdphillips: Admittedly I'm somewhat blindly following what we do in the health router: https://github.com/sensu/sensu-go/blob/0e648b821240a05b2d967b76b6e302da137b2a6f/backend/apid/routers/health.go#L46-L48

Looks like the timeout value is extracted and used for each request to the remote etcd nodes: https://github.com/sensu/sensu-go/blob/0e648b821240a05b2d967b76b6e302da137b2a6f/backend/store/etcd/health_store.go#L78-L86

amdprophet (approved):

backend/apid/graphql/health.go
echlebek: The health router is probably doing the wrong thing here, I'd follow Justin's advice

jamesdphillips: I couldn't quite figure out why we wanted multiple timeouts, so I switched to WithTimeout as suggested. Bit of a _Chesterton's Fence_ scenario but at least this is easier to reason about.

PR #4417, opened 2021-09-21
Add sensuctl pipeline commands

amdprophet:
## What is this change?

Adds pipeline commands to `sensuctl`:
* `sensuctl pipeline list`
* `sensuctl pipeline info [NAME]`
* `sensuctl pipeline delete [NAME]`

It allows for quick viewing of configured pipelines & deleting pipelines without needing to use `sensuctl delete`.

Yes.

Yes.

I created pipeline resources using `sensuctl create -f` and used the new commands against those resources.

## Is this change a patch?

No.


ccressent (approved):
Awesome work!

PR #4419, opened 2021-09-23
Add parallel JSON encoding to event logger

echlebek:
## What is this change?

This change re-implements speedups that were delivered in the enterprise product in the 6.4 series, and resolves a deadlock.

Added


backend/eventd/logger.go
```diff
@@ -138,25 +166,46 @@ func (l *rawLogger) ringBuffer() {
...
-		case l.output <- v:
+		case l.encoderInput <- v:
 		case <-time.After(l.wait):
-			// The new event could not be placed on the outgoing channel, therefore
-			// take the oldest event in the buffer, drop it, and place the new event
-			// at its place
-			<-l.output
-			l.output <- v
+			dropped := false
+			select {
```
fguimond, line 172: We had discussed there could be unwanted dropped events. Did you want to rework that logic in a separate fix?

fguimond (approved):
Aside from comment LGTM. I'll let you be the judge of what you want to do about it.

backend/eventd/logger.go
echlebek: Yes, if we rework this I think it needs a separate effort. :+1:

PR #4426, opened 2021-10-01
Partition silenced entry transactions

### LINKED ISSUE #4384: Silenced entries can fail to be deleted if there are many for an entity
## Expected Behavior
Expired silenced entries will be deleted, no matter how many there are.

## Current Behavior
```
{"component":"store","error":"internal error: etcdserver: too many operations in txn request","level":"error","msg":"error deleting expired silenced entries","time":"2021-08-10T20:57:55Z"}
```
It seems that when the number of silenced entries for an entity pass a certain threshold, trying to delete all of them in a transaction produces an error.

## Possible Solution
Batch the operations in smaller amounts.

echlebek:
## What is this change?

The number of silenced entries in a given transaction could exceed the
default number of etcd transaction operations, which is 128. This commit
changes the silenced entry store so that it makes at most 64 transaction
operations, and handles silenced expiration by creating multiple
transactions that are each no longer than 64 items.

Closes #4384

Yes

On the contrary, this change was easy to make, and that's what I decided to make it ahead of 6.6.

I tested that the partitioning logic I wrote functions as expected. Beyond that, I expect the existing integration tests to hold water.

## Is this change a patch?

Yes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,8 @@ environment variables.
 `SENSU_TIMEOUT` variables to the command's environment.
 - Fixed a crash when running the backend on `darwin/arm64` when compressing a
 wrapped resource.
+- Fixed a bug where large number of silences could cause etcd errors by not
+exceeding etcd'd default maximum number of transaction operations.
 
 ### Changed
 - Upgraded Go version from 1.16.5 to 1.17.1.
--- a/backend/store/etcd/misc_test.go
+++ b/backend/store/etcd/misc_test.go
@@ -1,0 +1,25 @@
+package etcd
+
+import (
+	"reflect"
+	"testing"
+)
+
+func testPartitionStrings(t *testing.T) {
+	data := []string{"a", "a", "a", "a", "a", "a", "a", "a", "a"}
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -34,6 +34,8 @@ environment variables.
 `SENSU_TIMEOUT` variables to the command's environment.
 - Fixed a crash when running the backend on `darwin/arm64` when compressing a
 wrapped resource.
+- Fixed a bug where large number of silences could cause etcd errors by not
+exceeding etcd'd default maximum number of transaction operations.
```
ccressent, line 38: Should this read "by exceeding" instead of "by not exceeding"?

ccressent (commented):
I'm wondering if this could maybe apply to other things than silences? In other words, what is special about silences that this treatment is needed for these only?

CHANGELOG.md
echlebek: No, although the sentence is ambiguous... I mean, by not exceeding the maximum ops per transaction, I've fixed a bug :)

ccressent (approved):

PR #4427, opened 2021-10-04
Don't crash when metrics log not writable

echlebek:
## What is this change?

Logs an error but continues operation when the metrics log file cannot be opened.

Necessary for compatibility with older sensu versions.

No, this is net new functionality.


jamesdphillips (approved):
🎉 looks good to me

PR #4468, opened 2021-10-29
Fix a bug where asset expansion err not handled

echlebek:
A recent PR made a change to assets that changes the way expansion
happens, but the error from the new function call was not handled.


amdprophet (commented):
Can you add a changelog entry?

portertech:
Ah shoot, I'll add the changelog entry as part of the patch release.

PR #4471, opened 2021-10-30
Add a tool for checking nested module freshness

echlebek:
The checkmod tool can be run on any system that has the go compiler and
git. It will check each nested module in sensu-go to see if any require
releasing.


amdprophet (approved):
Looks good. We may want to update to this to use a pure Go implementation of a git library in the future. Depending on the output of commands should be avoided as they can differ between versions.

echlebek:
Yeah go-git would be a nice improvement. We could also likely remove the dependency on the go compiler.

PR #4474, opened 2021-11-01
Add flag to configure API & dashboard's write timeouts

jamesdphillips:
Adds `api-write-timeout` & `dashboard-write-timeout` flags to allow users to configure the respective HTTP server's write timeout. Useful when some requests might reasonably take more than a few seconds to complete.

---

### DIFF
--- a/backend/cmd/start.go
+++ b/backend/cmd/start.go
@@ -222,14 +222,14 @@ func StartCommand(initialize InitializeFunc) *cobra.Command {
 				APIListenAddress:      viper.GetString(flagAPIListenAddress),
 				APIRequestLimit:       viper.GetInt64(flagAPIRequestLimit),
 				APIURL:                viper.GetString(flagAPIURL),
-				APIWriteTimeout:       viper.GetInt64(flagAPIWriteTimeout),
+				APIWriteTimeout:       viper.GetDuration(flagAPIWriteTimeout),
 				AssetsRateLimit:       rate.Limit(viper.GetFloat64(flagAssetsRateLimit)),
 				AssetsBurstLimit:      viper.GetInt(flagAssetsBurstLimit),
 				DashboardHost:         viper.GetString(flagDashboardHost),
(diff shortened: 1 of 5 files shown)


amdprophet (approved):

ccressent (commented):
I thought we were sort of standardizing on using `time.Duration` for these types of flags going forward? Then the flag can take values like `5ms`, `15s`, ...

echlebek:
Yes please @ccressent! I think using the string representation of time.Duration should be our standard, considering it has first class support in our argument parser.

backend/cmd/start.go
```diff
@@ -512,12 +518,14 @@ func flagSet(server bool) *pflag.FlagSet {
 		flagSet.String(flagAPIListenAddress, viper.GetString(flagAPIListenAddress), "address to listen on for api traffic")
 		flagSet.Int64(flagAPIRequestLimit, viper.GetInt64(flagAPIRequestLimit), "maximum API request body size, in bytes")
 		flagSet.String(flagAPIURL, viper.GetString(flagAPIURL), "url of the api to connect to")
+		flagSet.Int64(flagAPIWriteTimeout, viper.GetInt64(flagAPIWriteTimeout), "maximum duration before timing out writes of responses")
```
amdprophet, line 521: As the others have mentioned, let's use `viper.GetDuration()` here instead and update the types surrounding this flag to `time.Duration`.

amdprophet (changes requested):

jamesdphillips:
@ccressent good call! updated to use `GetDuration`.

jamesdphillips:
Bit of a bummer but I can't think of an easy way to migrate `--agent-write-timeout` to use a duration. If we care about compatibility of `backend.Config` that is.

echlebek:
@jamesdphillips yeah it's fine to leave it as is. We can add it to the list of things we want to break with 7.x...

amdprophet (approved):

ccressent (approved):

PR #4476, opened 2021-11-02
Avoid overwriting etcd defaults

### LINKED ISSUE #3952: Avoid setting etcd.Config values with backend.Config values when nil/empty
## Expected Behavior
The defaults that etcd sets should be used unless an override is specifically set by Sensu.

## Current Behavior
Sensu always overrides the default values of `etcd.Config` when `NewEtcd()` is called: https://github.com/sensu/sensu-go/blob/v5.21.0/backend/etcd/etcd.go#L156.

## Possible Solution
Add conditionals to check if Sensu config values are not nil or empty.

ccressent:
## What is this change?

This change makes it so etcd related configuration flags parsed by Sensu are only applied if they are not a zero value (`""`, 0, empty slice, ...). This is to avoid overwriting default values set by the etcd developers.

Fix #3952.

Added.

I think it's a good change to make, but I'm not sure if this had caused any problem so far; the issue (#3952) doesn't give information.

No.

I don't think any documentation needs to be changed.

- All unit tests passing
- All integration tests passing
- Manual tests

## Is this change a patch?

No.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Changed
+
+- Empty and zero value configuration parameters for `etcd` do not overwrite
+  defaults anymore.
+
 ## [6.5.3, 6.5.4] - 2021-10-29
 
 ### Added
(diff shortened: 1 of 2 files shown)


amdprophet (approved):
Looks good! We'll need to check to see if any changes to our Dockerfiles are needed before we ship this in a release.

fguimond (approved):
lgtm

ccressent:
@amdprophet 
I've had a look at all the Dockerfiles I could find, and I don't think there are any changes needed there.
Famous last words...

PR #4482, opened 2021-11-08
Add backend label to logged metrics

### LINKED ISSUE #4450: Add backend label to logged metrics
We need the ability to associate metrics from the metrics log file with the backend they were generated on. This may be tricky as I don't think we'd want this label to present itself in API calls to `/metrics` due to it being populated by Prometheus itself when it scrapes the exporter.

fguimond:

## What is this change?

Add the backend label to the logged metrics.

Closes #4450 

We need the ability to associate metrics from the metrics log file with the backend they were generated on.

Yes, added.

No.

No.

No.

Unit test. Also tested manually by adding the following to my backend.yml:
```
platform-metrics-log-file: "/Users/fguimond/sensu/sensu-backend/logs/backend-stats.log"
```

## Is this change a patch?

No.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,13 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Added
+- Add backend label to logged metrics.
+
 ### Fixed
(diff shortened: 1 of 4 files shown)


echlebek (approved):
Looks good to me!

Issue #4174, opened 2021-01-15
Reflect external etcd version in /version API results


cwjohnston:
## Feature Suggestion

When configured with external etcd, the response body of /version API endpoint should reflect the external etcd version.

## Context

Output from backend /version API doesn't reflect version of etcd running on external cluster, so the output could be misleading.

calebhailey:
We should also include the etcd location (i.e. "embedded" vs "external") in the /status output.

portertech:
Implementation would entail using the etcd client to query external etcd for its version, when the backend is running with no embed.

Francis mentioned he gets a version when using external etcd, we suspect that it's not reporting the correct etcd.

c-kruse:
From what I have gathered, it doesn't look like our current [Version API Model](https://github.com/sensu/sensu-go/blob/79021374d3c0f1c882639df0992a0323976b237a/api/core/v2/version.go#L9) is entirely coherent with sensu backend running with an external etcd cluster.

Sensu's `v2.Version` has an `Etcd` field of type [go.etcd.io/etcd/api/v3/version.Versions](https://pkg.go.dev/go.etcd.io/etcd/api/v3/version#Versions), which is intended to express both etcd Server and Cluster versions of a particular etcd server instance. The Cluster and Server version can be different during a cluster upgrade. It makes sense to communicate both when sensu has an embedded etcd processes, but when using an external cluster I don't think that there is any particular etcd server we care about.

Could use advice on how we aught to handle this. I see a few options for squeaking by using the existing version API.
1. "Lie" and set etcdserver and etcdcluster version to cluster version
2. omit etcdserver or set it to some "N/A"-type status

PR #4484, opened 2021-11-09
Set External etcd version in /version API

c-kruse:
## What is this change?

When a sensu backend is initialized with an external etcd service `--no-embed-etcd`, pull the etcd cluster version from the remote service to reflect in the sensu `/version` endpoint. 

https://github.com/sensu/sensu-go/issues/4174

The version API is currently tailored towards the situation where etcd is running embedded in sensu-backend.

It reflects both the version of that particular embedded instance of etcd, `etcdserver`, and the lowest common version of all cluster members, `etcdcluster`.

**Example with single sensu-backend instance:**
```
{
  "etcd": {
    "etcdserver": "3.5.0",
    "etcdcluster": "3.5.0"
  },
  "sensu_backend": "(devel)"
}
```

Today when using external etcd endpoints the output is misleading, sharing the etcd client version as `etcdserver` and leaving `etcdcluster` blank.

**Example version api output today with external etcd 3.3.8 (single instance):**
```
{
  "etcd": {
    "etcdserver": "3.5.0",
    "etcdcluster": ""
  },
  "sensu_backend": "(devel)"
}
```

This proposed solution pulls an etcd cluster version from the remote endpoint at startup:
**Resulting /version api output with external etcd 3.3.8 (single instance):**
```
{
  "etcd": {
    "etcdserver": "3.5.0", # more akin to etcd client version - admittedly still confusing
    "etcdcluster": "3.3.8"
  },
  "sensu_backend": "(devel)"
}
```

Probably, yes!

I definitely took the shortest path on this - A few areas I'd like y'all to consider to help us zero in on a solution.

1. Should we include etcd version info for remote clusters at all? I am actually leaning _no_ on this, but would like to hear from folks with more context.
2. Assuming we do want to display cluster version, what should we do about etcdserver version in this case?
3. Thoughts on moving this out of the initialization path and into the /version http handler? Presumably the cluster could be upgraded within the lifetime of a sensu backend instance.

The /version api docs were pretty sparse - I think we are probably okay without documenting this behavior.

Manually. Note: may be worth a todo to port the ansible updates for sensu-perf w/ external etcd over to sensu-staging.

## Is this change a patch?

I _think_ it is, but could be convinced that it's a minor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 ### Fixed
 - Duration metrics for assets, pipeline, and eventd have been updated to use
 milliseconds to match other duration metrics.
+- The sensu-backend's /version API has updated to reflect the version of a an
+external etcd cluster.
 
 ## [6.5.3, 6.5.4] - 2021-10-29
 
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -548,8 +548,19 @@ func Initialize(ctx context.Context, config *Config) (*Backend, error) {
 	authenticator.AddProvider(basic)
 
 	var clusterVersion string
-	// only retrieve the cluster version if etcd is embedded
-	if !config.NoEmbedEtcd {
+	if config.NoEmbedEtcd {
+		// get cluster version from first available etcd endpoint
+		endpoints := b.Client.Endpoints()
+		for _, ep := range endpoints {
+			status, err := b.Client.Status(ctx, ep)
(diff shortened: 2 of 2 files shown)


echlebek (approved):
Ship it!

```
                   ,:',:`,:'
                __||_||_||_||__
           ____["""""""""""""""]____
           \ " '''''''''''''''''''' |
    ~~jgs~^~^~^^~^~^~^~^~^~^~^~~^~^~^^~~^~^
```

PR #4493, opened 2021-11-15
Fix several websocket handling bugs

### LINKED ISSUE #4492: sensu-backend does not handle agent websocket connections correctly
There are a number of bugs in how sensu-backend handles websocket connections.

1. sensu-backend can do concurrent writes to websocket connections, as the SendCloseMessage method is not protected by a mutex.
2. The transport incorrectly uses a single sync.RWMutex to serialize reads and writes, when it should actually use two regular mutexes, one for reading and one for writing. This leads to all kinds of incorrect stuff where methods try to lock and unlock mutexes multiple times, leading to racy code.
3. The Close() method of the transport never tries to close the actual connection. It just sends a close message and then returns, leaving the connection open. This could possibly contribute to issues with open file handles on some systems.

This is a longstanding issue and has persisted since before GA. Efforts have been made to resolve it, but I believe we were thrown off by the transport package's idiosyncratic handling of websockets.

echlebek:
## What is this change?

This commit changes the way websocket connections are handled, by both
the backend and the agent.

It replaces one sync.RWMutex with two sync.Mutex values, one for reads,
and one for writes. The gorilla websocket docs state that reads and
writes can be concurrent, so it's not correct to use a single mutex for
both resources.

The commit also avoids setting the internal "closed" flag to true when a
connection error is encountered. Instead, it only sets the flag to true
when a ClosedError is found on receive.

The commit also makes sure the write mutex is locked when sending a
close message.

The commit also removes sync.Pool use for transport messages, which was
a speculative optimization we applied. However, it seems unlikely that
this optimization is really necessary, so I have removed it in service
of making the code clearer.

Closes #4492 

Yes

So far I've only run the existing tests to make sure everything is still working.

## Is this change a patch?

Yes

### DIFF
--- a/transport/transport.go
+++ b/transport/transport.go
@@ -12,15 +12,8 @@ import (
 )
 
 var (
-	sep     = []byte("\n")
-	msgPool sync.Pool
+	sep = []byte("\n")
 )
-
-func init() {
-	msgPool.New = func() interface{} {
-		return &Message{}
-	}
-}
 
 const (
 	// MessageTypeKeepalive is the message type sent for keepalives--which are just an
@@ -132,8 +125,9 @@ type Transport interface {
 // WebSocket.
 type WebSocketTransport struct {
 	Connection *websocket.Conn
-	closed     bool
-	mutex      *sync.RWMutex
+	closed     bool // protected by writeMu
+	readMu     sync.Mutex
+	writeMu    sync.Mutex
 }
 
 // NewTransport creates an initialized Transport and return its pointer.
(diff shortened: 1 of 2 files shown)


transport/transport.go
```diff
@@ -228,22 +223,19 @@ func (t *WebSocketTransport) Heartbeat(ctx context.Context, interval, timeout in
 // a ClosedError or a ConnectionError if unable to receive a message. Receive
 // blocks until the connection has a message ready or a timeout is reached.
 func (t *WebSocketTransport) Receive() (*Message, error) {
-	t.mutex.RLock()
-	if t.closed {
-		t.mutex.RUnlock()
+	t.readMu.Lock()
+	defer t.readMu.Unlock()
+
+	if t.Closed() {
```
fguimond, line 229: There is still the infinitissimal risk that the connection will be closed between the time t.Closed() is called and the time ReadMessage() is called. However since ReadMessage() will return the error anyways I don't believe this is an issue anyways.

echlebek: Yes, I believe this will actually occur quite frequently in practice. So we have an instance of optimistic locking for the common case, with the less optimal case being that we actually try the connection, which fails predictably, like you say. I _think_ it's OK...

fguimond: Agreed. But you had me think... a long Send() - whether it's a large message or slow network - can block the reader(s) since they still require a lock to `t.writeMu`. I'm not sure in practice how impactful this will be.

ccressent (approved):

Issue #4487, opened 2021-11-10
Update events resolver to use store ordering


jamesdphillips:
After https://github.com/sensu/sensu-go/issues/4485 has been implemented, we'll want to update the events GraphQL resolver. I believe the steps to be as follows:

1. Since the etcd store will not be implementing ordering and filtering, I think we should avoid using the dataloader in the events field resolver. This is the because the data loader implicitly caches the results from the store for the duration of the session and differences between the store implementations may lead to subtle bugs. As such I think the resolver should be updated to directly call the loader's [listEvents()](https://github.com/sensu/sensu-go/blob/main/backend/apid/graphql/dataloader.go#L148-L168) function instead.
1. In the field resolver we'll want to lift the order and offset into the context before making a call to the [listEvents()](https://github.com/sensu/sensu-go/blob/main/backend/apid/graphql/dataloader.go#L148-L168) function.
1. In the field resolver we'll want to check if the [events store supports filtering](https://sumologic.slack.com/archives/C024XK35Z3Q/p1637694698190400) and if so make a call to the store's `EventsCount` method, we can use the result we can set the [totalCount](https://github.com/sensu/sensu-go/blob/main/backend/apid/graphql/namespace.go#L291) field.

Voila!

jamesdphillips:
It occurred to me that there is a slight wrinkle in this plan... The GraphQL resolver's signature expects pagination to be offset based while the store API is cursor based; if we actually want to get away from scanning the entire events keyspace / table we'll need to address this. Unfortunately due to federation we'll need to do this is a way that doesn't break the existing API.

A proposed solution:

* soft deprecate the event resolvers `offset` argument; any values provider are ignored.
* add a new `after` argument to the revolvers that will be equivalent to the continue token.
* update the resolvers response such have the `totalCount` field always be zero.
* add a `endCursor` field to the `OffsetPageInfo` type; we will fill this value with the continue token. (We could also duplicate the `OffsetPageInfo` type creating a new type named `EventConnectionPageInfo`, mark the `offset` field as deprecated, and then also add the `endCursor` field. This way we avoid polluting the existing type.)
* update the frontend to select and use the `endCursor`

```graphql
type Namespace {
  # ...
  
  "List events associated with the namespace."
  events(
    # moving forward we'll ignore any value given to the 'offset' arg
    offset: Int = 0 @deprecated(reason: "please use the after argument instead"),
    # proposed new field; equivalent to the "continue" token
    after: String,
[...]

PR #4512, opened 2021-11-22
Add CountEvents store method

echlebek:
## What is this change?

The CountEvents store method will be used by the Web UI to know how many
pages of results are available, when using pagination.

Supporting work for ongoing maintenance related to #4487

No

Integration testing.

## Is this change a patch?

No.

### DIFF
--- a/backend/store/etcd/event_store.go
+++ b/backend/store/etcd/event_store.go
@@ -418,3 +418,23 @@ func handleExpireOnResolveEntries(ctx context.Context, event *corev2.Event, st s
 
 	return nil
 }
+
+// CountEvents counts events in the namespace. The SelectionPredicate is not
+// supported.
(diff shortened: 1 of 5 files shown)


fguimond (approved):
LGTM

ccressent (commented):
Can we leverage [this existing method](https://github.com/sensu/sensu-go/blob/58481edfabb48ca7bd2bf3840bc4a0cbbf7230eb/backend/store/etcd/store.go#L240-L258)?

echlebek:
@ccressent unfortunately it causes an import cycle, but I can copypasta the implementation so that it's the same.

echlebek:
@ccressent yes!

jamesdphillips (approved):

PR #4513, opened 2021-11-23
Event ordering preparation

fguimond:

## What is this change?

In preparation for Postgres ordering/filtering/pagination:

* Add sort fields to event api
* Add offset and ordering to SelectionPredicate

In preparation for Postgres ordering/filtering/pagination.

No.

No.

No.

not required.

Unit tests.

## Is this change a patch?

No.


backend/store/store.go
```diff
@@ -99,14 +99,18 @@ type SelectionPredicate struct {
 	// returned empty from the store, it indicates that there's no additional
 	// resources available
 	Continue string
+	// Offset indicates the index of the first resource to retrieve
+	Offset int64
```
jamesdphillips, line 103: Did we decide if we wanted to do this or not?

api/core/v2/event_sort.go
```diff
@@ -0,0 +1,226 @@
...
+	case EventSortLastOk:
+		sortIf = EventsByLastOk(events, asc)
+	case EventSortSeverity:
+		sortIf = EventsBySeverity(events, asc)
+	default:
+		sortIf = EventsByTimestamp(events, asc)
+	}
+
+	sort.Sort(sortIf)
+}
```
jamesdphillips, line 49: 👍

backend/store/store.go
jamesdphillips: never mind me

backend/store/store.go
```diff
@@ -99,14 +99,18 @@ type SelectionPredicate struct {
...
 	// Limit indicates the number of resources to retrieve
 	Limit int64
 	// Subcollection represents a sub-collection of the primary collection
 	Subcollection string
+	// Ordering indicates the property to sort on, if supported by the store
+	Ordering string
```
ccressent, line 109: Can you also add `OrderingDirection string` (which will hold one of the constants you've defined fore `ASC/DESC`)?

backend/store/store.go
ccressent: My take is that we shouldn't need this field so you can remove it.

jamesdphillips: Unless I'm mistaken, my only concern is that the resolver would need to fabricate a continue token

jamesdphillips (approved):

api/core/v2/event_sort_test.go
```diff
@@ -0,0 +1,355 @@
...
+			input:    []*Event{critical, noCheck, ok},
+			expected: []*Event{critical, ok, noCheck},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			sort.Sort(EventsBySeverity(tc.input, false))
+			fmt.Println("EXP:", eventsToId(tc.expected))
+			fmt.Println("RES:", eventsToId(tc.input))
```
echlebek, line 245: tests should not fmt.Println please :)

backend/store/store.go
```diff
@@ -103,10 +103,14 @@ type SelectionPredicate struct {
...
+	// OrderDirection indicates the sorting direction
+	OrderingDirection string
```
echlebek, line 109: Not to bikeshed too much here but this really only needs to be a boolean.

What about OrderingDirection bool with ASC and DESC constants for true/false, or a field named Descending which defaults to false by way of zero value?

api/core/v2/event_sort_test.go
fguimond: Yes agreed, thanks for catching it!

backend/store/store.go
fguimond: I like the bool idea. It probably would be up to the store itself to convert it to whatever format it needs to.

fguimond: Updated.

echlebek (approved):

PR #4515, opened 2021-11-24
Store Based Event Filtering Aware Graphql Field Resolver

c-kruse:
## What is this change?

Overwrites the Namespace Events graphql field resolver for situations where the backing event store supports filtering and sorting (i.e. postgres in the enterprise product)

Prevents the web ui from using excessive memory by loading large collections of events.

? 😬 

## Is this change a patch?

No

### DIFF
--- a/backend/apid/graphql/namespace.go
+++ b/backend/apid/graphql/namespace.go
@@ -295,7 +295,7 @@ func (r *namespaceImpl) eventsWithInStoreFiltering(p schema.NamespaceEventsField
 
 // Events implements response to request for 'events' field.
 func (r *namespaceImpl) Events(p schema.NamespaceEventsFieldResolverParams) (interface{}, error) {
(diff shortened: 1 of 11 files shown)


backend/apid/graphql/namespace.go
```diff
@@ -7,18 +7,26 @@ import (
...
 	"github.com/sensu/sensu-go/graphql"
 )
 
 var _ schema.NamespaceFieldResolvers = (*namespaceImpl)(nil)
 
+const (
+	storeOffset   = "io.sensu.store.filtering.offset"
+	storeLimit    = "io.sensu.store.filtering.limit"
+	storeOrdering = "io.sensu.store.filtering.ordering"
+)
```
jamesdphillips, line 20: 🔪

jamesdphillips (approved):
beaut

echlebek (approved):
Approved! It will necessitate a similar change in sensu-enterprise-go.

PR #4531, opened 2021-12-03
Disable default excludes in golangci-lint, add default staticcheck rules/exclusions

amdprophet:

## What is this change?

Removes the default excludes from golangci-lint and adds default staticcheck rules/exclusions to match the operation of `staticcheck ./...`.

It enables the desired staticcheck rules from running against the codebase.

No.

All staticcheck/stylecheck rules will be run except the following:

### ST1000
Incorrect or missing package comment
https://staticcheck.io/docs/checks#ST1000

### ST1003
Poorly chosen identifier
https://staticcheck.io/docs/checks#ST1003

### ST1016
Use consistent method receiver names
https://staticcheck.io/docs/checks#ST1016

### ST1020
The documentation of an exported function should start with the function’s name
https://staticcheck.io/docs/checks#ST1020

### ST1021
The documentation of an exported type should start with type’s name
https://staticcheck.io/docs/checks#ST1021

### ST1022
The documentation of an exported variable or constant should start with variable’s name
https://staticcheck.io/docs/checks#ST1022

### ST1023
Redundant type in variable declaration
https://staticcheck.io/docs/checks#ST1023

Do we want to include any of these? Are there any other rules we'd like to exclude?

The same lint rules fail between golangci-lint w/staticcheck & stylecheck vs staticcheck.

## Is this change a patch?

No.


amdprophet:
We've decided to stop using golangci-lint and solely use staticcheck for now.

PR #4532, opened 2021-12-05
Add PrometheusOutputMetricFormat to aliases.go

3l1k:
Need to add PrometheusOutputMetricFormat , as plugins which export to prometheus wont work.

## What is this change?

Need to add PrometheusOutputMetricFormat , as plugins which export to prometheus wont work.

plugins which export to prometheus wont work.

No

No

No required

with push gateway handler [](https://github.com/portertech/sensu-prometheus-pushgateway-handler)
## Is this change a patch?

Looks like

### DIFF
--- a/types/aliases.go
+++ b/types/aliases.go
@@ -171,7 +171,11 @@ const (
 	// InfluxDBOutputMetricFormat is the accepted string to represent the output metric format of
 	// InfluxDB Line
 	InfluxDBOutputMetricFormat = v2.InfluxDBOutputMetricFormat
-
+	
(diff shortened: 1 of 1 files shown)


echlebek:
Hi, thanks for taking the time to submit this PR.

This change shouldn't be necessary. These aliases were only added to support the move of types and functions from `package types` to `core/v2` long ago. `PrometheusOutputMetricFormat` was added after this change and never existed in the `types` package.

These aliases are deprecated and they will be removed in the next major release. Please import `core/v2` instead.

PR #4539, opened 2021-12-07
Hack21 opamp server

fguimond:
## What is this change?

Implements a very basic web socket server to process OpAMP requests from an agent. When a message is received on the WS it is passed to the message handler which should implement the OpAMP protocol. At this time only the StatusReport is handled but the processing doesn't do anything per say, it simply returns a ServerToAgent message.

TODO:
* Start implementing OpAMP protocol
* Leverage Christian configuration work
* Better manage connections

Hackathon 2021!


backend/opampd/agent.go
```diff
@@ -0,0 +1,4 @@
+package opampd
+
+type Agent struct {
```
c-kruse, line 3: Appears to be unused. Did you have a plan for this?

c-kruse (approved):

PR #4542, opened 2021-12-09
Create a new entity when an agent comes online.

fguimond:
## What is this change?

Create a new Sensu entity when a new opamp agent connects to the WS server.

### DIFF
--- a/backend/opampd/protocol.go
+++ b/backend/opampd/protocol.go
@@ -5,17 +5,27 @@ import (
 	"fmt"
 
 	"github.com/open-telemetry/opamp-go/protobufs"
+	corev2 "github.com/sensu/sensu-go/api/core/v2"
 	"github.com/sensu/sensu-go/backend/store"
+)
+
+const (
+	entityNamespace = "default"
 )
 
 const supportedCapabilities = protobufs.ServerCapabilities_AcceptsStatus | protobufs.ServerCapabilities_OffersRemoteConfig
 
 type Protocol struct {
-	Store     store.OpampStore
+	Store     store.Store
 	PartyMode bool
 }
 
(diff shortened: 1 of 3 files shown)


backend/opampd/protocol.go
```diff
@@ -56,3 +67,31 @@ func (p *Protocol) OnAgentDisconnect(instanceUid string, disconnect *protobufs.A
 	//TODO implement me
 	return nil, fmt.Errorf("implement me")
 }
+
+func (p *Protocol) createEntity(instanceUid string, report *protobufs.StatusReport) error {
+	entity := &corev2.Entity{
+		EntityClass:        corev2.EntityAgentClass,
```
ccressent, line 73: I don't think it's very important, but I think we said we would create a new entity class altogether, like `OT Agent` or something.

ccressent (approved):

backend/opampd/protocol.go
portertech: It does matter for filtering etc.

backend/opampd/protocol.go
```diff
@@ -56,3 +67,31 @@ func (p *Protocol) OnAgentDisconnect(instanceUid string, disconnect *protobufs.A
...
+			Name:      instanceUid,
+			Namespace: entityNamespace,
+			Labels: map[string]string{
+				"opamp-instanceid": instanceUid,
+				"opamp-agent-type": report.AgentDescription.AgentType,
+			},
+			Annotations: nil,
+			CreatedBy:   "",
+		},
+		SensuAgentVersion: report.AgentDescription.AgentVersion,
```
portertech, line 92: We shouldn't reuse this class and sensu agent specific fields. Something closure to a proxy entity makes more sense.

fguimond: For now is it ok if I add this information as a `opamp-agent-version` label?

PR #4544, opened 2021-12-09
Add PoC opampd metrics event

ccressent:
This makes `opampd` collect some simple statistics and emit a metrics event with them every 30 seconds.

### DIFF
--- a/backend/opampd/opampd.go
+++ b/backend/opampd/opampd.go
@@ -44,10 +44,11 @@ var (
 )
 
 //
-func emitMetricsEvent(bus messaging.MessageBus) error {
+func emitMetricsEvent(bus messaging.MessageBus, entity *corev2.Entity) error {
 	event := &corev2.Event{}
 	event.Name = "opampd-metrics"
 	event.Namespace = "default"
+	event.Entity = entity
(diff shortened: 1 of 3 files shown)


backend/opampd/opampd.go
```diff
@@ -7,19 +7,78 @@ import (
...
+	// to OpenTelemetry agents.
+	agentConfigsSent uint64
+
+	// errorCount is the total number of errors encountered while exchanging
+	// messages with OpenTelemetry agents.
+	errorCount uint64
+)
+
+//
+func emitMetricsEvent(bus messaging.MessageBus) error {
```
fguimond, line 47: It probably is fine but do we support events without an entity?

fguimond: oh never mind, just saw sean's comment

fguimond (approved):
Aside from the missing event entity the PR looks good to me :)

backend/opampd/opampd.go
ccressent: We don't, good catch!
Sean also mentioned it; I've just pushed a commit that uses the backend as the entity for that metrics event.

PR #4546, opened 2021-12-10
Hack21 opamp broadcast

fguimond:
## What is this change?

Better manage incoming connections and provide a way to broadcast a message to all connected OT agents.

Note that for now connection cleanup is not implemented.


portertech:
Closing but leaving the branch alive.

PR #4550, opened 2021-12-13
Revoke orphaned leases on bury

amdprophet:
## What is this change?

Adds a new `BuryAndRevokeLease` method to SwitchSet which is used by keepalived instead of `Bury`. This allows the lease associated with a keepalive SwitchSet to be revoked when buried, preventing leases from being orphaned.

It eliminates leases from being leaked when agents continuously reconnect.

Yes.

Our agent registration & lease code is tough to grok.

The first keepalive that the backend processes for an agent is generated by the backend when the session starts. This keepalive is generated using information from the HTTP headers on connection, which does not include keepalive thresholds. If we had access to the keepalive thresholds, we could conditionally bury keys depending on if the TTL has changed.

I wrote a tool to track lease usage (https://github.com/amdprophet/etcd-lease-tool). After this change, no leases are orphaned when stopping & starting an agent.

I ran benchmarks for 6.6.2 and 6.6.2 with this change. Notice the difference between active leases in both runs. Backend heap bytes allocated is significantly more stable after this change as well.

Etcd by Prometheus (3 5) - Grafana (2)
Sensu Benchmarks (_= 6 6 0) - Grafana (2)
PostgreSQL Database - Grafana (4)
PostgreSQL Statistics - Grafana (4)

## Is this change a patch?

Yes.

### DIFF
--- a/backend/liveness/liveness.go
+++ b/backend/liveness/liveness.go
@@ -180,11 +180,10 @@ func (t *SwitchSet) BuryAndRevokeLease(ctx context.Context, id string) error {
 
 	// revoke the lease from earlier if it exists
 	if leaseID != 0 {
-		if err := kvc.Backoff(ctx).Retry(func(n int) (done bool, err error) {
-			_, err = t.client.Revoke(ctx, leaseID)
-			return kvc.RetryRequest(n, err)
-		}); err != nil {
-			return fmt.Errorf("error burying switch: %s", err)
+		if _, err := t.client.Revoke(ctx, leaseID); err != nil {
+			t.logger.Debugf(
+				"error revoking lease for buried switch, lease may have already been revoked: %s",
+				err)
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


backend/liveness/liveness.go
```diff
@@ -152,26 +155,64 @@ func (t *SwitchSet) Alive(ctx context.Context, id string, ttl int64) error {
...
+
+	// revoke the lease from earlier if it exists
+	if leaseID != 0 {
+		if err := kvc.Backoff(ctx).Retry(func(n int) (done bool, err error) {
+			_, err = t.client.Revoke(ctx, leaseID)
+			return kvc.RetryRequest(n, err)
+		}); err != nil {
+			return fmt.Errorf("error burying switch: %s", err)
+		}
+	}
```
echlebek, line 189: I have some concerns about this section. If the lease revokes on etcd's end, or on another backend, before this backend's Revoke request, then this would produce a retry of the operation. Perhaps the revocation should be optimistic, and not produce an error if the operation fails.

amdprophet: Makes sense. Update incoming.

echlebek (approved):
:shipit: awesome work!

PR #4553, opened 2021-12-13
[GraphQL] Improve entity list perf

jamesdphillips:
Improvements to entity list response time & memory consumption.

* Deprecate the LASTSEEN sort order as this ordering has been unused by the web ui since June of 2019. This leaves us with sorting of entities by name (desc & asc).
* Update the list resolver to rely on etcd's sorting capabilities, this allow us to avoid having to retrieve the entire list of entities from the store. We can then page through entities from the store until we have the requested amount.
* For the sake of providing a `totalCount` we continue page up to additional 500 matches. This means that in the default case (no filter applied) we will retrieve no more than 500 entities from the store.
* Only the matched entities within the requested range are kept in memory, significantly reducing the memory required for each request.
* The data loaders have also now been updated so that they will only retrieve resources up to a hard limit (for most resources this is 2.5k). This should also have a significant impact memory usage but reduces precision.
* I've also updated the events list resolver so that it no longer uses the data loader, it's unlikely that we want to dedupe this request anyways. This should have a positive impact on memory usage overall.

Closes sensu/sensu-enterprise-go#1943

---


echlebek (approved):
Looks good!

PR #4556, opened 2021-12-14
On daemon stop, crash when not responding

echlebek:
## What is this change?

sensu-backend now crashes with a panic when daemons are asked to
stop but do not respond.

If daemons hang, then the overall ability to operate sensu-backend is compromised. This change ensures that when asked to stop, sensu-backend will do so within 30s.

Yes

Verification on SPDC pending.

## Is this change a patch?

Yes.


amdprophet (approved):
Just for clarity, we wait 30 seconds for each daemon to stop, not 30s for all daemons to stop?

echlebek:
@amdprophet yes, each daemon is given 30s, which may be a bit too generous. WDYT?

PR #4558, opened 2021-12-14
Use gofmt to add new build constraint syntax

c-kruse:
## What is this change?

Ran `go fmt` on most of sensu-go. Wanted to be careful to not format generated code unnecessarily, so only targeted directories with go build constraints. To reproduce - pardon my terrible bash: `rg '\+build' -l . | xargs dirname | uniq | xargs go fmt`

Go has introduced a new build constraint syntax that it is migrating towards. More info out here: https://go.googlesource.com/proposal/+/master/design/draft-gobuild.md#transition

Probably not.

N

N

Y

N/A - build pipeline is what likely matters here.

## Is this change a patch?

Y

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Changed
 - Removed unused rate limiting code in the liveness package.
+- Include new Go 1.17+ build constraint syntax by running gofmt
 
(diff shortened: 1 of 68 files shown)


amdprophet (approved):

amdprophet:
Does this need to target `release/6.6`? If not, let's update this to target `main`.

echlebek (approved):

PR #4559, opened 2021-12-15
Fix an infinite loop that can occur in keepalives

echlebek:
## What is this change?

This commit fixes an infinite loop that can occur when another backend
revokes a lease for an agent, or if an operator manually revokes the
lease.

Yes

## Is this change a patch?

Yes

### DIFF
--- a/backend/liveness/liveness.go
+++ b/backend/liveness/liveness.go
@@ -13,7 +13,6 @@ import (
 	"github.com/sensu/sensu-go/backend/store/etcd/kvc"
 	"github.com/sirupsen/logrus"
 	"go.etcd.io/etcd/api/v3/mvccpb"
-	rpctypes "go.etcd.io/etcd/api/v3/v3rpc/rpctypes"
 	clientv3 "go.etcd.io/etcd/client/v3"
 	"golang.org/x/time/rate"
 )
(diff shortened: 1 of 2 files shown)


amdprophet (approved):
Great work!

backend/liveness/liveness.go
```diff
@@ -258,22 +259,42 @@ func (t *SwitchSet) ping(ctx context.Context, id string, ttl int64, alive bool)
...
+				if err != nil {
+					etcd.LeaseOperationsCounter.WithLabelValues("liveness", etcd.LeaseOperationTypePut, etcd.LeaseStatusFor(err)).Inc()
+				}
+				return kvc.RetryRequest(n, err)
+			}
+			etcd.LeaseOperationsCounter.WithLabelValues("liveness", etcd.LeaseOperationTypePut, etcd.LeaseStatusFor(err)).Inc()
+			return kvc.RetryRequest(n, err)
+		}
+		leaseID := clientv3.LeaseID(resp.PrevKv.Lease)
+		_, err = t.client.KeepAliveOnce(ctx, leaseID)
```
jamesdphillips, line 289: Is this a separate fix?

echlebek: All part of the same whole

amdprophet (approved):

PR #4572, opened 2021-12-17
Create proxy entity state when missing if entity config exists

amdprophet:
## What is this change?

Proxy entity state is now created if it does not exist but matching proxy entity config does exist.

There seems to be a race condition when switching the store from Etcd to Postgres where entity state is not migrated from Etcd to Postgres. This can cause events for proxy entities return an error when attempting to handle them.

Yes.

I've only run the unit & integration tests.

## Is this change a patch?

Yes.

### DIFF
--- a/backend/eventd/entity.go
+++ b/backend/eventd/entity.go
@@ -64,6 +64,31 @@ func createProxyEntity(event *corev2.Event, s storev2.Interface) (fErr error) {
 		// order to create a fully formed corev2.Entity for the event.
 		wState, err = s.Get(stateReq)
 		if err != nil {
+			switch err.(type) {
+			case *store.ErrNotFound:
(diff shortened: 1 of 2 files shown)


backend/eventd/entity.go
```diff
@@ -64,6 +64,31 @@ func createProxyEntity(event *corev2.Event, s storev2.Interface) (fErr error) {
...
+				if err != nil {
+					return err
+				}
+				if err := s.CreateOrUpdate(stateReq, wState); err != nil {
+					return err
+				}
+			default:
+				return err
+			}
 			return err
```
echlebek, line 92: This kills the bug fix :)

echlebek (approved):
Oof, probably needs a test, but let's do it!

Issue #4236, opened 2021-03-24
Nil maps are not correctly being represented


mgibson323:
When using `index` function in token substitution against an empty `labels` object, it does not correctly evaluate to `false`. 

## Expected Behavior

On a check that uses token substitution when executed on an entity with no labels, the following token substitution should return "foobar":

` '{{ index .labels "foo.bar" | default "foobar" }}'`

## Current Behavior

Token substitution returns this error: 

`error calling index: index of untyped nil`

## Context

Labels containing dots "`.`" must use the `index` function in token substitution.
Sensu internal reference: https://secure.helpscout.net/conversation/1460992000/25017?folderId=1211658

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): **6.2.5**
* Installation method (packages, binaries, docker etc.): **packages**
* Operating System and version (e.g. Ubuntu 14.04): **Ubuntu 20.04**

calebhailey:
Thanks for filing!

PR #4579, opened 2022-01-05
Use typecasted nil for empty maps in dynamic pkg

amdprophet:

## What is this change?

Updates the dynamic pkg to set fields (with omitempty) to a map-typed nil value instead of not setting the field at all.

It fixes a bug where token substitution can result in an unexpected error when attempting to use a key that doesn't exist in labels.

Closes https://github.com/sensu/sensu-go/issues/4236.

Yes.

I wrote new specs to confirm the bug in the issue linked above. After adding the changes the new specs (& old specs) pass.

## Is this change a patch?

It can be if we'd like it to be but I've targeted main for now.

### DIFF
--- a/types/dynamic/dynamic.go
+++ b/types/dynamic/dynamic.go
@@ -217,6 +217,10 @@ func synthesizeStruct(value reflect.Value) map[string]interface{} {
 		// Don't add empty/nil fields to the map if omitempty is specified
 		empty := isEmpty(fieldValue)
 		if empty && omitEmpty {
+			// If value is a map, set field to nil with map type
+			if fieldValue.Kind() == reflect.Map {
+				out[fieldName] = (map[string]interface{})(nil)
+			}
 			continue
 		}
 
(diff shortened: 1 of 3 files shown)


types/dynamic/dynamic.go
```diff
@@ -217,6 +217,10 @@ func synthesizeStruct(value reflect.Value) map[string]interface{} {
 		// Don't add empty/nil fields to the map if omitempty is specified
 		empty := isEmpty(fieldValue)
 		if empty && omitEmpty {
+			// If value is a map, set field to nil with map type
+			if fieldValue.Kind() == reflect.Map {
+				out[fieldName] = (map[string]interface{})(nil)
+			}
```
echlebek, line 223: I was thinking something more like

```
if fieldValue.Kind() != reflect.Ptr {
  out[fieldName] = reflect.Zero(fieldValue.Type()).Interface()
}
```

echlebek (approved):
:shipit:

amdprophet:
I spun up a staging stack and ran QA against this. Everything looks good.

PR #4584, opened 2022-01-12
Check for nil handler in legacy pipeline

amdprophet:

## What is this change?

Adds a conditional to the legacy pipeline to ensure that a handler isn't nil after attempting to fetch the handler from the store.

A backend will currently crash if the handler does not exist.

Yes.

Only unit/integration tests.

## Is this change a patch?

Yes.


jamesdphillips (approved):
team effort

PR #4592, opened 2022-01-24
Add max-session-length flag and logic

### LINKED ISSUE #3329: Periodically disconnect agents from backends
Over time, agent connection distribution can become skewed, if a backend experiences an outage, for instance. We can rebalance the cluster by forcing agents to reconnect periodically.

Add a configuration flag to the `sensu-agent` that specifies the amount of time, in minutes, that agents should remain connected before disconnecting and reconnecting to another backend. The configuration flag should be named `--max-session-length` and should be specified as a time duration. (i.e. 3s, 10m, 1h)

The default should be 15m.

Add jitter to the time when calculating when to reconnect agents, so that a thundering herd scenario does not occur. The jitter algorithm should be
```
wait = wait + time.Duration(rand.Float64()*float64(wait))
```

ccressent:
## What is this change?
```
This makes the agent disconnect and reconnect after some configurable
amount of time (+ built-in jitter). The aim of this change is to have
agents organically rebalance their connections to the backends over
time.

Signed-off-by: Cyril Cressent <cyril@sensu.io>
```

Close #3329.

Yes, added.

Yes: I am suggesting a different way to calculate the final session length (including jitter) such that `--max-session-length` feels more "natural". Let me know if it makes sense to go ahead with that.

```
maxSessionLength = x/2 + randFloat() * x
where x is the value given to the flag
```

Given that `randFloat()` gives a value in the range `[0, 1(`, it makes it so the maximum amount of time before a disconnect/reconnect is really bounded by the value given to the flag.

Yes. The connection manager in the agent is quite crusty and I broke various things several times before arriving at my current solution.

I haven't. New documentation for the flag will be required.

## Is this change a patch?

No.

### DIFF
--- a/agent/cmd/start.go
+++ b/agent/cmd/start.go
@@ -77,6 +77,7 @@ const (
 	flagRetryMin                 = "retry-min"
 	flagRetryMax                 = "retry-max"
 	flagRetryMultiplier          = "retry-multiplier"
+	flagMaxSessionLength         = "max-session-length"
 
 	// TLS flags
 	flagTrustedCAFile         = "trusted-ca-file"
@@ -138,6 +139,7 @@ func NewAgentConfig(cmd *cobra.Command) (*agent.Config, error) {
 	cfg.RetryMin = viper.GetDuration(flagRetryMin)
 	cfg.RetryMax = viper.GetDuration(flagRetryMax)
 	cfg.RetryMultiplier = viper.GetFloat64(flagRetryMultiplier)
+	cfg.MaxSessionLength = viper.GetDuration(flagMaxSessionLength)
 
 	// Set the labels & annotations using values defined configuration files
 	// and/or environment variables for now
@@ -320,6 +322,7 @@ func handleConfig(cmd *cobra.Command, arguments []string) error {
 	viper.SetDefault(flagRetryMin, time.Second)
 	viper.SetDefault(flagRetryMax, 120*time.Second)
 	viper.SetDefault(flagRetryMultiplier, 2.0)
+	viper.SetDefault(flagMaxSessionLength, 0*time.Second)
 
 	// Merge in flag set so that it appears in command usage
 	flags := flagSet()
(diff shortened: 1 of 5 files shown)


agent/cmd/start.go
```diff
@@ -441,6 +444,7 @@ func flagSet() *pflag.FlagSet {
 	flagSet.Duration(flagRetryMin, viper.GetDuration(flagRetryMin), "minimum amount of time to wait before retrying an agent connection to the backend")
 	flagSet.Duration(flagRetryMax, viper.GetDuration(flagRetryMax), "maximum amount of time to wait before retrying an agent connection to the backend")
 	flagSet.Float64(flagRetryMultiplier, viper.GetFloat64(flagRetryMultiplier), "value multiplied with the current retry delay to produce a longer retry delay (bounded by --retry-max)")
+	flagSet.Duration(flagMaxSessionLength, viper.GetDuration(flagMaxSessionLength), "maximum amount of time after which the agent will reconnect to one of the configured backends")
```
echlebek, line 447: Might be worthwhile to mention that the default is to stay connected forever?

echlebek (approved):
Looks good to me!

ccressent:
@echlebek
Any thoughts on the new timeout calculation I'm suggesting?

echlebek:
I think your suggestion is good! :+1:

cmd/loadit/main.go
```diff
@@ -34,7 +34,8 @@ var (
 	flagHugeEvents        = flag.Bool("huge-events", false, "send 1 MB events to the backend")
 	flagUser              = flag.String("user", agent.DefaultUser, "user to authenticate with server")
 	flagPassword          = flag.String("password", agent.DefaultPassword, "password to authenticate with server")
-	flagBaseEntityName    = flag.String("base-entity-name", "test-host", "base entity name to prepend with count number.")
+	flagBaseEntityName    = flag.String("base-entity-name", "test-host", "base entity name to prepend with count number")
+	flagMaxSessionLength  = flag.Duration("max-session-length", 0*time.Second, "maximum amount of time after which the agent will reconnect to one of the configure backends")
```
echlebek, line 38: There's a typo, s/configure/configured/

fguimond (approved):

PR #4594, opened 2022-01-25
Allow specifying more precise API versions

### LINKED ISSUE #4591: Support referring to precise versions in resource specifications
In Sensu resource specifications, there is a field that indicates the API version. It looks like
```
"api_version": "core/v2"
```

Support the full API version for api_version values, so that users can specify versions like
```
"api_version": "core/v2.2.0"
```

When Sensu's deployed version is smaller than this, it should refuse to attempt to parse the resource.

echlebek:
## What is this change?

This commit makes changes that allows users to specify richer version in
the api_version field of a wrapped resource. Users can now specify
versions like core/v2.3.0, to require that Sensu has at least version
2.3.0 available.

Since the first version of Sensu that implements this functionality has
core/v2.6.0 and core/v3.3.0, requiring versions smaller than this has no
effect.

There are no published modules for enterprise APIs yet, so only core
APIs can be gated this way as of this writing.

Closes #4591 

Yes

Modules are not yet available for enterprise APIs.

New documentation will eventually be required, but I don't think it needs to be altered just yet.

Testin'

## Is this change a patch?

No.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Changed
 - Agent now persists prometheus HELP messages as a metric tag.
+- Resource specifications can now use precise core/v2 and core/v3 module
+versions. For instance, api_version: core/v2.5.1 or core/v2.3. The resource
+validation system will reject the resource if Sensu does not have at least the
+specified version.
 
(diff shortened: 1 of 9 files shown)


echlebek:
@amdprophet after some meddling, now ready for review.

amdprophet (approved):

PR #4596, opened 2022-01-28
[GraphQL] Cap max value of Entity.status to math.MaxInt32

jamesdphillips:
Web browsers handle values greater than `math.MaxInt32` inconsistently and as a result the `graphql-go` nullifies said values. (Find relevant code [here](https://github.com/graphql-go/graphql/blob/master/scalars.go#L30-L32).) This patch caps the max value returned by the status resolver to `math.MaxInt32`.

---

### DIFF
--- a/backend/apid/graphql/entity.go
+++ b/backend/apid/graphql/entity.go
@@ -1,6 +1,7 @@
 package graphql
 
 import (
+	"math"
 	"sort"
 	"time"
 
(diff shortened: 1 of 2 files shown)


echlebek (approved):
Yikes. Still having problems with that uint field after all this time.

asachs01:
@jamesdphillips did this make it into 6.6.5?

PR #4598, opened 2022-01-29
[GraphQL] Partial count field

jamesdphillips:
* Adds a `partialCount` flag to pagination that communicates whether or not the `totalCount` in fact represents the total count of entities or if the count was abandoned.
* When queries entities, in the case where no filter was given, the total count of entities is retrieved from the metrics daemon to avoid having to scan the entire keyspace for determine the count.

---


jamesdphillips:
@c-kruse correct! @dabria is working on that right now.

PR #4599, opened 2022-01-29
[GraphQL] Use a sliding chunk size

jamesdphillips:
Switches to using a chunk size of between 250 and 500 records. In cases where the page size is small (eg. 25 items) this will lead to a bit more memory usage but it greatly reduce the amount to round trips. As I/O is likely a bigger bottleneck than CPU or memory, this should greatly reduce resolver execution time.

---


echlebek (approved):

jamesdphillips:
| Before  | After |
| ------------- | ------------- |
| Screen Shot 2022-02-01 at 3 00 32 PM | Screen Shot 2022-02-01 at 2 45 03 PM  |

Saw nearly a 3x improvement for p99 requests and surprisingly about 10X for p50.

PR #4604, opened 2022-02-01
Ringv2 Create Watcher per Subscription

c-kruse:
## What is this change?

Adds the context.Context to ringv2 watcher key

BSM and potentially other usages of ringv2 occasionally "abandon" subscriptions (cancel the context used to create the subscription and stop reading from the subscription channel) and then create a fresh but otherwise identical subscription with a new context from another goroutine. By returning the channel of an existing watcher form another context, there is a potential for subsequent subscriptions to unexpectedly be cancelled.

✅ 

@echlebek  I'm iffy about this one. It was a quick change so I made it and it seems to work, but it feels like the change could have a pretty large surface area.

No doc updates required.

Manual/local validation so far.

## Is this change a patch?

?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Fixed
+- Resolves unpredictable ringv2 behavior when identical subscriptions are
+created from different contexts.
+
 ## [6.6.4] - 2022-01-13
 
 ### Added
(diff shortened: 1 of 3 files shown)


c-kruse:
Actually sounds like @portertech  prefers BSM be updated to always use the postgres implementation. So the priority/value here is questionable.

echlebek (approved):
I think this is the right thing to do. It resolves a race condition and also makes it possible to create identical subscriptions, which is something the ring should support.

It's not something I conceived of as being a requirement when I was designing the package, but I believe the facility needs to be general enough to support this. Plus, the race condition around cancellation and re-establishment definitely needs to be fixed. Great work!

Issue #4610, opened 2022-02-07
[6.6.5]: Deadlock in types caused by rwlock in APIModuleVersions


vooon:
Introduced here:

https://github.com/sensu/sensu-go/commit/1b94c22e9bd250f7dc003375201f38752c1ada4f#diff-7bd1b6df952fed43510ebd5c251eb0d9686e4a385757760326e5f05987f807e0R300

APIModuleVersions() wants to write-lock packageMapMu in already read-locked env.
So ResolveType() and ResolveRaw() stuck in deadlock.

## Expected Behavior
- node starts

## Current Behavior
- node stuck in pipelined creation

## Possible Solution
Call APIMiduleVersions outside of read-locked context.

## Steps to Reproduce (for bugs)
1. Start node, it won't listen to API because Init stuck.

## Context
- OSS Build

## Your Environment
- Build from source
- Go 1.17.6.

fguimond:
Note that this issue was not present in 6.6.5 but rather in the (unreleased) main branch.

PR #4611, opened 2022-02-07
types: Fix ResolveType/ResolveRaw deadlock

vooon:
## What is this change?

Fix bug #4610 by calling `APIModuleVersions()` before read-locking `packageMapMu`.

Required to be able to start the backend node.

Perhaps yes?

No.

No.

No.

Yes, after that backend works again.

## Is this change a patch?

Yes.


fguimond (approved):
Validated the fix, thanks for your contribution @vooon  🥇

Issue #4261, opened 2021-04-09
Check stop execution when secret provider is not reachable


guipe:
## Expected Behavior
When secret provider (Vault) is not reachable, all checks which use this provider must continue to be executed, and these check must pass to critical status.

## Current Behavior
When secret provider (Vault) is not reachable, all checks which use this provider stop execution, and these checks left in OK state.

## Steps to Reproduce (for bugs)

1. Stop communication between Sensu-go backend and Vault
2. Verify checks which use secret from Vault (checks are not more executed and their status is always OK)

## Context

When Vault is not reachable, all checks which use secret stay in OK status. 

These checks must fail as secret are not more reachable.

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent):sensuctl: 6.0.0 sensu-agent:6.1.2 sensu-backend: 6.2.5
* Installation method (packages, binaries, docker etc.): docker
* Operating System and version (e.g. Ubuntu 14.04): CentOS 7.8.2003

portertech:
This needs further discussion. How should Sensu really surface backend errors/issues?

calebhailey:
Let's schedule a Tuesday engineering discussion for this one to review outstanding proposals.

calebhailey:
Updates: 

* As a first step, we think we can improve the UX here by implementing #4442 and #4443, and generating a normal Sensu Event when a backend is unable to retrieve a secret (this also opens the door for exposing other scheduler & event processing events as Sensu Events)
* We are still hesitant to modify the scheduler behavior to allow checks to be scheduled when a secret it depends on is unavailable, so we will not be modifying this behavior at this time 
* We will continue to explore a solution for updating Events to indicate various configuration issues (e.g. unavailable secret and/or unretrievable asset) instead of the current behavior which may result in potentially misleading "stale" OK events

calebhailey:
We'll probably need to create the following issues in https://github.com/sensu/sensu-go and https://github.com/sensu/sensu-enterprise-go: 

- [ ] [sensu/sensu-go] Add a GenerateSensuEvent() or GenerateSensuBackendEvent() function to sensu-backend
- [ ] [sensu/sensu-enterprise-go] Generate a Sensu Event when a backend is unable to retrieve a secret from a SecretProvider (w/ `namespace:<system-namespace>` & `entity:<sensu-backend-id>`)

[...]

PR #4624, opened 2022-02-09
backend events

### LINKED ISSUE #4443: Backends create sensu-system namespace and its entity
## Feature Suggestion

Sensu backend entities should be visible in the Sensu API and web app. 

## Possible Implementation

On backend start:

- Check for `sensu-system` namespace
** Create if missing
- Check for existing entity in the `sensu-system` named after the backend's local hostname
** If missing, create a new entity with `entity_class` of `backend` and populate the entity system information
** If one exists, don't touch it

Backend uses its own entity when reporting cluster state errors.

## Context

See: #4442

fguimond:
* create system namespace
* create backend entity
* create new error for vault access errors
* generate events when vault not available
* unit tests

## What is this change?

On backend startup create the `sensu-system` namespace as well as a backend entity (if necessary). When the secrets provider returns an access error log a critical event against the backend entity. When it's accessible again log an ok event against the entity. There is a mechanism to prevent "spamming" the event bus with backend events, so backend events are generated no more than every 30 seconds, unless the status changes.

Closes #4443 and #4261 

Yes

No

No

@hillaryfraley I believe this will need to be documented.

Manual testing:
* Create a Vault container: 
```
docker run --cap-add=IPC_LOCK -e 'VAULT_DEV_ROOT_TOKEN_ID=myroot' \                                                                                                                     
    -e 'VAULT_DEV_LISTEN_ADDRESS=0.0.0.0:8200' -p 8200:8200 vault
```
* Navigate to `http://localhost:8200`, login using the `myroot` secret and click on `secrets`
* Create a new secret with the following information
** Path: `sensu/prom`; key: `http_url`; value: `http://localhost:9100/metrics`; hit Save
* Create a check using this secret (path has to be `secrets/data/sensu/prom#http_url`

In the web ui or using sensuctl you will see an OK event under the `sensu-system/backendname` entity.

## Is this change a patch?

No

### DIFF
--- a/backend/resource/backend.go
+++ b/backend/resource/backend.go
@@ -3,7 +3,6 @@ package resource
 import (
 	"context"
 	"errors"
-	"fmt"
 	"time"
 
 	"github.com/google/uuid"
@@ -79,8 +78,7 @@ func (br *BackendResource) GenerateBackendEvent(component string, status uint32,
 	}
 
 	now := time.Now().Unix()
-	checkName := fmt.Sprintf("%s-%s", corev2.BackendCheckName, component)
-	if lastEvent, ok := br.lastEvents[checkName]; ok {
+	if lastEvent, ok := br.lastEvents[component]; ok {
 		if lastEvent.status == status && now-lastEvent.timestampSec < br.repeatIntervalSec {
 			return nil
 		}
@@ -91,21 +89,18 @@ func (br *BackendResource) GenerateBackendEvent(component string, status uint32,
 		Timestamp: now,
 		Entity:    br.backendEntity,
 		Check: &corev2.Check{
-			ObjectMeta: corev2.ObjectMeta{
-				Name:      checkName,
-				Namespace: systemNamespaceName,
-			},
-			Issued:   now,
-			Executed: now,
-			Output:   output,
-			Status:   status,
+			ObjectMeta: corev2.NewObjectMeta(component, systemNamespaceName),
+			Issued:     now,
+			Executed:   now,
+			Output:     output,
+			Status:     status,
 		},
-		ObjectMeta: corev2.ObjectMeta{},
+		ObjectMeta: corev2.NewObjectMeta("", systemNamespaceName),
 		ID:         id[:],
 	}
 	err := br.bus.Publish(messaging.TopicEventRaw, event)
 	if err == nil {
-		br.lastEvents[checkName] = &eventInfo{
+		br.lastEvents[component] = &eventInfo{
 			status:       status,
 			timestampSec: now,
 		}
--- a/backend/secrets/error.go
+++ b/backend/secrets/error.go
@@ -1,20 +1,22 @@
 package secrets
 
-import "fmt"
+import (
+	"errors"
+	"fmt"
+)
+
+var (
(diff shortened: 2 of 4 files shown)


backend/resource/backend.go
```diff
@@ -0,0 +1,128 @@
...
+	return nil
+}
+
+func (br *BackendResource) GenerateBackendEvent(component string, status uint32, output string) error {
+	if br.backendEntity == nil {
+		return errors.New("backend entity doesn't exist")
+	}
+
+	now := time.Now().Unix()
+	checkName := fmt.Sprintf("%s-%s", corev2.BackendCheckName, component)
```
echlebek, line 82: I don't think prepending all the backend event check names with "backend" is necessary. We already have "backend" as the entity name, which is part of the key to retrieve an event.

backend/resource/backend.go
```diff
@@ -0,0 +1,128 @@
...
+			ObjectMeta: corev2.ObjectMeta{
+				Name:      checkName,
+				Namespace: systemNamespaceName,
+			},
+			Issued:   now,
+			Executed: now,
+			Output:   output,
+			Status:   status,
+		},
+		ObjectMeta: corev2.ObjectMeta{},
```
echlebek, line 103: The namespace should be defined, and labels/annotations should be set to empty slices. This gives a nicer user experience around filters and selectors.

backend/secrets/error.go
```diff
@@ -0,0 +1,34 @@
+package secrets
+
+import "fmt"
+
+type ErrInvalidSecretInfo string
+type ErrNoProviderDefined struct{}
```
echlebek, line 6: For errors with an invariant message, it's a little better and a little more idiomatic to use

```
var ErrNoProviderDefined = errors.New("no provider defined")
```

These can be easily compared by value, for example, `if err == ErrNoProviderDefined { // handle error }`.

In your use of ErrNoProviderDefined, you've returned a pointer value, which would not compare as easily. Similar for ErrSecretsNotSupported.

go.mod
```diff
@@ -75,6 +75,7 @@ require (
 	go.etcd.io/etcd/client/v3 v3.5.0
 	go.etcd.io/etcd/server/v3 v3.5.0
 	go.etcd.io/etcd/tests/v3 v3.5.0
+	go.uber.org/atomic v1.7.0
```
echlebek, line 78: :thinking: not sure where this dependency comes from, does it persist if you run `go mod tidy`?

echlebek (approved):
Great work Francis! I look forward to making use of these new concepts. Just a few small comments.

backend/resource/backend.go
fguimond: Agreed, fixed.

backend/resource/backend.go
fguimond: Fixed 👍

backend/secrets/error.go
fguimond: Fixed.

fguimond:
Created https://github.com/sensu/sensu-enterprise-go/pull/2040 to return proper error when the Vault is not accessible.

Issue #4141, opened 2020-12-18
sensuctl cluster-role create flags should be plural


calebhailey:
## Feature Suggestion

The `sensuctl cluster-role create <name>` flags should be `--resources` and `--verbs` (plural) instead of `--resource` and `--verb` (singular). 

## Possible Implementation

Sensu `ClusterRole` resources have roles containing `resources` and `verbs` fields (plural), but the corresponding `sensuctl cluster-role create <name>`  flags are the singular form `--resource` and `--verb`, which is quite confusing. We might not want to deprecate the singular form flags to prevent breaking any existing user workflows – is it possible to alias them? Can both the singular and plural forms work?

PR #4635, opened 2022-02-19
sensuctl cluster-role create flags should be plural

agm650:
## What is this change?

adding two new flags, "verbs" and "resources" to replace "verb" and "resource".
In order not to break things, old flags are markes as deprecated and will "map" to the new flags transparently for the user.

Solving #4141 

Changelog has been updated.

nope

none for now :)

Some examples are to be updated. A PR is going to be opened for this

Running integration test plus manual creation of cluster-resource and mixing old & new flags.
## Is this change a patch?

No

### DIFF
--- a/CHANGELOG-6.md
+++ b/CHANGELOG-6.md
@@ -27,6 +27,7 @@ mid-write.
 - The default embedded etcd heartbeat interval has been increased from 100 to 300.
 - The default embedded etcd election timeout has been increased from 1000 to 3000.
 - Upgraded etcd version from 3.5.0 to 3.5.2.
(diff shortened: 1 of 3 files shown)


portertech:
As it's a breaking change, I've added this PR to the upcoming 7.0 milestone.

agm650:
> As it's a breaking change, I've added this PR to the upcoming 7.0 milestone.

It's not "yet" a breaking change. Old commands are still working you just get a warning telling you that the non plural verbs are deprecated. 
The idea was that at some point they will be drop but for a while people using them will be notified

echlebek:
Changes applied in #4712

PR #4641, opened 2022-03-02
Add CheckConfig pipelines to Check on NewCheck

echlebek:
## What is this change?

This commit fixes a bug where check pipelines aren't represented in event.check objects.

The full check configuration should be represented in events.

Yes

## Is this change a patch?

Yes


amdprophet (approved):
I think we'll need to add pipelines to the Metrics type as well but it can be implemented in a separate PR.

PR #4658, opened 2022-03-11
Metric threshold annotations

### LINKED ISSUE #4652: Matching output_metric_thresholds should generate annotation(s)
The new [Metric Threshold Service Checks](https://github.com/sensu/sensu-go/pull/4585) are working great in 6.7.0 pre-releases, but the resulting events don't provide any indication of the reason for the non-zero status. 

The agent should add one or more annotations indicating when the event status is the result of a metric threshold. 

### Annotation Content 

```
The value of <name> (<[comma-separated list of tag key=value]>) exceeded the configured threshold (<threshold_type>: <threshold>, actual: <value>).
```

Example

```yaml
  output_metric_thresholds:
    - name: system_cpu_used
      tags:
        - name: cpu
          value: cpu-total
      thresholds:
        - max: '{{ .annotations.system_cpu_used_warning_threshold | default "80.0" }}'
          status: 1
        - max: '{{ .annotations.system_cpu_used_critical_threshold | default "90.0" }}'
          status: 2
    - name: system_mem_used
      thresholds:
        - max: '{{ .annotations.system_mem_used_warning_threshold | default "80.0" }}'
          status: 1
        - max: '{{ .annotations.system_mem_used_critical_threshold | default "90.0" }}'
          status: 2
```

In the scenario where `system_cpu_used{cpu="cpu-total"}` was 95%, the resulting annotation should read: 

> The value of "system_cpu_used" (cpu="cpu-total") exceeded the configured threshold (max: 90.0, actual: 95.0).

### Annotations

* `sensu.io/output_metric_thresholds/<name>[.<tag_value>|.<tag_value>]/<threshold_type>/<status>`

  Set one annotation per matching metric threshold. 

  Status values should be strings (`0` = "ok", `1` = "warning", `2` = "critical"; all other values greater than 2 = "unknown")

  Using the same example from above, the resulting annotation names should be: 

  `sensu.io/output_metric_thresholds/system_cpu_used.cpu-total/max/warning`
  `sensu.io/output_metric_thresholds/system_cpu_used.cpu-total/max/critical`

* `sensu.io/event/notification` 

  Set this annotation for the most severe metric threshold, by status. 

  Using the same example from above, the resulting annotation value would be the threshold with the highest status (`2`): 

  > The value of "system_cpu_used" (cpu="cpu-total") exceeded the configured threshold (max: 90.0, actual: 95.0).

fguimond:
## What is this change?

Closes #4652 

Add event annotations to indicate the reason of a non-0 metric status.

There is one annotation per threshold rule that is matched, plus another one indicating the global status for the check.

To provide context about the metric check status.

No, this is an extension to the original metric threshold feature which is not released yet.

@calebhailey 
There were no mentions about what annotation to use for a `null_status` match. I ended up adding the following annotation:

`sensu.io/output_metric_thresholds/non-matching/null-status/unknown` - `The non-matching rule did not match any metric point.`

@hillaryfraley - I'll let you determine whether or not additional content is necessary to highlight the additional annotation.

* Manual testing
* Unit tests

## Is this change a patch?

No.


agent/check_handler.go
```diff
@@ -401,11 +410,101 @@ func evaluateOutputMetricThresholds(event *corev2.Event) uint32 {
...
+}
+
+func getAnnotationKey(metricThreshold *corev2.MetricThreshold, measure string, status uint32) string {
+	var key strings.Builder
+
+	key.WriteString("sensu.io/output_metric_thresholds/")
+	key.WriteString(metricThreshold.Name)
+	for _, tag := range metricThreshold.Tags {
+		key.WriteString(".")
+		key.WriteString(tag.Value)
```
c-kruse, line 446: I can't think of any serious gotchas here so I'll defer to your judgement, but are we concerned at all about stability of this annotation key?

So far I've seen sensu do a good job of not reordering lists, but through operator intervention or unspecified behavior we could end up going from `sensu.io/output_metric_thresholds/my_threshold.foo.bar/null-status/critical` for tags `source=foo,dest=bar` to `sensu.io/output_metric_thresholds/my_threshold.bar.foo/null-status/critical` pretty easily.

fguimond: Very good point Christian, I did overlook that aspect. I'll make the change to sort the tags by key.

fguimond: @c-kruse - now that I was about to fix this I believe this is actually not an issue. The tags are stored in a slice, not a map. This makes sure the order is maintained. The user would have to explicitly change the order using sensuctl for your scenario to happen. But for any given evaluation of an event the order would remain the same across annotation.

@calebhailey do you have any input on this issue?

amdprophet (approved):

agent/check_handler.go
calebhailey: @fguimond nope.

calebhailey:
@fguimond I think we need to update the copy for "matching" `null_status` thresholds (i.e. thresholds for metrics that weren't collected). In practice if I have a metric thresholds like the following, I would set `null_status` if the `example_metric` measurement wasn't collected (or to be more specific, there were no `example_metric` measurements with the tag `foo: bar`). 

```
spec:
  ...: ...
  output_metric_thresholds:
  - name: example_metric
    tags:
      - name: foo
        value: bar
    thresholds:
      - max: "2.0"
        min: "1.0"
        null_status: 1
        status: 1
```

In this case, I think the annotation key should be: `sensu.io/output_metric_thresholds/example_metric/null` with the value of `WARNING: No metric matching "metric2" (foo="bar") was found; expected max: 2.0, min: 1.0.`

So the annotation key syntax for "matching" null_status thresholds should be: 

```
sensu.io/output_metric_thresholds/<name>[.<tag_value>|.<tag_value>]/null_status
```

And the annotation value for "matching" null_status thresholds should be: 

```
<STATUS>: No metric matching "metric2" (foo="bar") was found[; expected <threshold_type>: <threshold_value>].
```

Two further thoughts on this matter: 

1. We should only add annotations for "matching" null_status thresholds if the null_status != 0. 
2. We should move the null_status field so you can only set it once per metric. 

   As implemented, you could select a single metric and set two thresholds (a max warning and max critical threshold), and both could have null_status: 

   ```yaml
   spec:
     ...: ...
     output_metric_thresholds:
     - name: example_metric
       tags:
         - name: foo
           value: bar
       thresholds:
         - max: "80.0"
           null_status: 1
           status: 1
         - max: "90.0"
           null_status: 2
           status: 2   
   ```

   In practice, null_status should be set once for the measurement: 

   ```yaml
   spec:
     ...: ...
     output_metric_thresholds:
     - name: example_metric
       tags:
         - name: foo
           value: bar
       null_status: 1
       thresholds:
         - max: "80.0"
           status: 1
         - max: "90.0"
           status: 2   
   ```

   LMK if you'd rather make this latter change in a separate PR and I'll open up a separate issue for it. 


Let me know if this doesn't make sense.

fguimond:
@calebhailey I totally agree moving up the null_status to the metric threshold level. I have started implementing the change as part of this issue. In the case where there are more than one threshold what should be the annotation value? Example:

```
spec:
  ...: ...
  output_metric_thresholds:
  - name: example_metric
    tags:
      - name: foo
        value: bar
    null_status: 1
    thresholds:
      - min: "20.0"
        max: "80.0"
        status: 1
      - min: "10.0"
        max: "90.0"
        status: 2   
```

Do we want multiple lines, one per threshold like this?
```
WARNING: No metric matching "example_metric" (foo="bar") was found; expected min: 20.0, max: 80.0.
CRITICAL: No metric matching "example_metric" (foo="bar") was found; expected min: 10.0, max: 90.0.
```

agent/check_handler.go
fguimond: @calebhailey Sorry my question wasn't clear in this case. Should we sort the annotation before adding them to annotations or let them in the same order they were created?

In other words if we have the following scenario
```
spec:
  ...: ...
  output_metric_thresholds:
  - name: example_metric
    tags:
      - name: zoo
        value: none
      - name: foo
        value: bar
      - name: too
        value: dive
    thresholds:
      - max: "2.0"
        min: "1.0"
        null_status: 1
        status: 1
```
Do we want to output in sequential order
```
sensu.io/output_metric_thresholds/example_metric.none.bar.dive/...
```
or sorted by key
```
sensu.io/output_metric_thresholds/example_metric.bar.dive.none/...
```

I would personally lean toward sequential order but wanted to get your opinion.

Thanks!

calebhailey:
@fguimond good catch; I think the annotation could list all of the configured thresholds. Your example is good because it shows a complex set of thresholds (setting both min+max and a status), so perhaps we should describe each threshold by listing the min/max values _and the status_?

Based on your example, the null_status output could look like: 

```
WARNING: no metric matching "example_metric" (foo="bar") was found; expected min: 20.0 - max: 80.0 (status: warning) min:10.0 - max: 90.0 (status: critical). 
```

The null_status annotation syntax would then be: 

```
<null_status string value>: No metric matching "<name>" ([comma-separated list of tag key=value]) was found[; expected <thresholds>]
```

Thresholds can be a comma-separated list of descriptions in this format: 

* Min+max thresholds: `min: <min_value> - max: <max_value> (status: <status>)`
* Min-only thresholds: `min: <min_value> (status: <status>)`
* Max-only thresholds: `max: <max_value> (status: <status>)`

amdprophet (approved):

Issue #4636, opened 2022-02-22
Inconsistent and wrong arm_version on Rasberry Pi 3/4 devices


espenaf:
## Expected Behavior
sensu-agent property **arm_version** = 7. So when downloading assets, the correct ones are filtered and installed.

## Current Behavior
sensu-agent property arm_version might not be 6/7 on a freshly installed device, or changes over time. The following values have been observed:

* debian 10.11 / debian (Raspberry Pi 3): arm_version="1476395014"
* raspbian 10.11 / debian (Raspberry Pi 3): arm_version="1476395014"
* raspbian 11.2 / debian (Raspberry Pi 3): arm_version="-1476395001"
* raspbian 10.11 / debian (Raspberry Pi 4): arm_version="-1476395002"

On the raspbian 10.11 (Raspberry Pi 4) device, the arm_version changed after upgrading to sensu-agent 6.6.6, from "-1476395001" to "-1476395002".

This breaks download of assets using arm_version to filter.

## Possible Solution
Currently duplicating asset config with the following change to force a workaround:

`        - entity.system.arm_version == -1476395002`

## Steps to Reproduce (for bugs)
1. Download and install image from https://www.raspberrypi.com/software/operating-systems/ on a Raspberry Pi 3 or 4 
2. Install sensu-agent via either apt repo or via ansible with https://sensu.github.io/sensu-go-ansible/
3. Add agent to backend
4. View properties in Entities -> entity -> properties 

## Context
Have a consistent and working arm_version property that can be used for deploying assets on a broad set of Raspberry Pi devices and OS versions. Currently this needs several duplications builds in the asset config of all plugins used by the agent.

## Your Environment
Sensu Backend 6.6.6 with Sensu Agent 6.6.3 or 6.6.6. Tested on Raspberry Pi 3 and 4, running pure Debian 10 Buster, Raspbian 10 Buster and  Raspbian 11 Bullseye. All running 32 bit version of Debian/Raspbian.

sensu-discourse:
This issue has been mentioned on **Sensu Community**. There might be relevant details there:

https://discourse.sensu.io/t/sensu-agent-arm-version-wrong-on-debian-buster-raspberry-pi/2729/3

agm650:
Could it be that the resolution of the goarm variable is incorrect?
on a raspberry pi3 (11.2) I have 0x40000007 
So I have the version 7 defined somewhere 

In some other repository I found out that the arm version was stored in an int8.

In the original issue,
[...]

PR #4661, opened 2022-03-14
Correcting arm version

agm650:
## What is this change?

Correcting the detection of the arm version

At the present time the value retrieved for the arm version is invalid. Looking at some other code on internet ([here](https://gist.github.com/lucab/f7162ca2d95191c692edc12ea8ccaaef) or [here](https://github.com/creativeprojects/resticprofile/blob/master/arm.go) ) doing the same manipulation, it appears that the goarm needs to be define as a uint8.
Since int8/uint8 is not a valid type for protobuf exchange, the getARMVersion still return an int32. I'm just casting goarm to int32.

This is solving #4636

Yes and I updated the changelog

Nothing I can think off

Not really. I just realised that protobuf doesn't support int8/uint8 as a type so I had to refrain myself from updating the .proto file.

I don't think this change has any impact on the documentation.

I build the sensu-agent with GOARM 6 and 7. Then I executed the program on a raspberry pi. Then I checked that the version displayed in the webUI and also with sensuctl.
As I do not have any other ARM device I was unable to perform some more complete testing.

## Is this change a patch?

Not sure. the change is very small, so it might bee seen as a patch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 - Log agent IP address for connections with faulty TLS configurations.
+- ARM Version detection
 
 ### Changed
 - The default embedded etcd heartbeat interval has been increased from 100 to 300.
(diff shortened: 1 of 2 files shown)


echlebek (approved):
Thanks, I've verified that runtime.goarm is in fact a uint8 in the Go source code!

PR #4666, opened 2022-03-17
[GraphQL] Test if current user can perform a given action

jamesdphillips:
<p align="center"></p>

Adds a new interface that allows one to test whether they're able to perform an action. Inspired by Kubernetes' SelfSubjectAccessReview API[^1].

Closes https://github.com/sensu/sensu-enterprise-go/issues/873

### Schema

```graphql
#
# Extends root query object with an field resolver that allows one to test
# whether they're able to perform an action. Inspired by K8s
# SelfSubjectAccessReview API.
#
# @link https://github.com/kubernetes/kubernetes/blob/c1e69551be1a72f0f8db6778f20658199d3a686d/staging/src/k8s.io/api/authorization/v1/types.go#L45-L67
#
extend type Query @named(suffix: "SelfSubjectAccessReview") {
  """
  Checks whether or not the current user can perform an action; provides handy
  interface to preemptively test if an action will succeed. Returns null if the
  action is allowed and otherwise a FetchErr.
  """
  cani(type: TypeMetaInput!, meta: ObjectMetaInput!, verb: String!): FetchErr
}
```

### Usage

```graphql
query ICanHas {
  cani(
    type: { type: "CheckConfig", apiVersion: "core/v2" },
    meta: { namespace: "default" },
    verb: "create",
  )
}
```

[^1]: https://github.com/kubernetes/kubernetes/blob/c1e69551be1a72f0f8db6778f20658199d3a686d/staging/src/k8s.io/api/authorization/v1/types.go#L45-L67

PR #4674, opened 2022-03-24
[backend/api] Allow consumer to test if authorized to perform action

jamesdphillips:
Adds method to api package allowing the consumer to test if they are able to perform an action.

Required by https://github.com/sensu/sensu-enterprise-go/pull/2150.
Replaces https://github.com/sensu/sensu-go/pull/4666

---


echlebek (changes requested):
I think it's worth creating a specific type for verbs and defining constants for them. As in,
```
type RBACVerb string

const Get RBACVerb = "get"
```

Should make the API easier to use once we forget which order the params go in :)

echlebek (approved):

PR #4678, opened 2022-03-25
Fix a panic in socket handlers

### LINKED ISSUE #4675: Nil pointer dereference when TCP handler write fails
## Expected Behavior
TCP handler write failures are propagated as a regular error.

## Current Behavior
Panic, sensu-backend crashes.

## Possible Solution
Fix the condition that leads to the panic (nil assigned to conn, conn.Close() called)

echlebek:
## What is this change?

This changelog fixes a panic in sensu-backend when a socket handler gets an error mid-write. It also fixes another bug where socket handlers will stop respecting their timeout after the initial connection is established.

Closes #4675 

Yes

It was tricky to reproduce the bug with a test but I eventually succeeded.

Docs changes not required.

Fully verified with automated testing.

## Is this change a patch?

Yes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 ### Fixed
 - Fixed a bug where pipelines in check configuration were not represented in
 the check object of events that were produced with the check configuration.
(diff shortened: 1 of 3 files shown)


c-kruse (approved):
Clever way to get a failing test to reproduce the issue!

LGTM!

PR #4680, opened 2022-03-30
Subdue logic

### LINKED ISSUE #1955: Fix check subdue (broken and pulled prior to GA)
## The fix

https://github.com/sensu/sensu-go/issues/1955#issuecomment-1074144897

## Context

After spending some time debugging a test that only failed during certain times of day, I was finally able to isolate a test that I think behaves reliably.

The test that fails is `TestIsSubdued/valid_subdue`.

The reproduction I've created, which should hopefully fail any time of day, is here:

```
func TestTimeWindowWhenEarlyMorning(t *testing.T) {
	now, err := time.Parse(time.RFC1123Z, "Mon, 13 Aug 2018 5:30:00 -0900")
	if err != nil {
		t.Fatal(err)
	}
	subdue := &TimeWindowWhen{
		Days: TimeWindowDays{
			All: []*TimeWindowTimeRange{
				{
					Begin: now.Add(-time.Hour).Format(time.Kitchen),
					End:   now.Add(time.Hour).Format(time.Kitchen),
				},
			},
		},
	}
	subdued, err := subdue.InWindows(now)
	if err != nil {
		t.Fatal(err)
	}
	require.True(t, subdued)
}
```

Because `Begin` is an hour before `now`, and `After` is an hour after, the subdue should be true.

I believe the problem is that we are not consistently handling times in UTC. For instance, Check.IsSubdued does not use UTC, whereas the unit test TestIsSubdued does (to set up the test). In the test that I pasted to this issue, I have not used UTC, and observed a failure.

The InWindow method assumes that the time it receives is in UTC, so I think that something is not quite right.

What I can't figure out is why the TestIsSubdued test does not fail every time - it should have a constant offset from UTC. Because it only fails at certain times of day, I am concerned that may be a deeper problem with the InWindow method. However, at this time, that is just conjecture.

Note that the TestIsSubdued test will never fail on CI systems, because they are on UTC. This really is a nasty bug!

fguimond:

## What is this change?

Implement a new subdue specification.

Closes #1955 

Better handling of time offsets by using RFC-3339 date/time/offset format.

Yes of course!

* I was not able to set the check's subdue information. However if a customer somehow has the data should we clean it up or convert it to the new format?

* The timezone (or actually UTC offset) defined in the begin time as an offset is used to calculate the weekday, day of month, month and year instead of using UTC.
* Alternatively a proper timezone could be more appropriate and support daylight savings time, but it's not implemented yet.
* I took the liberty to remove the old subdue logic (not the schema) as it didn't appear we could set the check's `sudbue` information. If somehow a customer does have subdue data it will be ignored.

@hillaryfraley new documentation will be required for this.

Unit tests
Manual testing with a check similar to this:

```
type: CheckConfig
api_version: core/v2
metadata:
  created_by: fguimond
  labels:
    sensu.io/managed_by: sensuctl
  name: echo-subdued
  namespace: default
spec:
  check_hooks: null
  command: echo 111
  env_vars: null
  handlers: []
  high_flap_threshold: 0
  interval: 15
  low_flap_threshold: 0
  output_metric_format: ""
  output_metric_handlers: null
  pipelines: []
  proxy_entity_name: ""
  publish: true
  round_robin: false
  runtime_assets: null
  secrets: null
  stdin: false
  subdue: null
  subdues:
  - begin: "2022-03-30T11:15:00-04:00"
    end: "2022-03-30T11:17:00-04:00"
    repeat:
    - wednesdays
    - thursdays
  - begin: "2022-03-16T11:19:00-04:00"
    end: "2022-03-16T11:21:00-04:00"
    repeat:
    - weekdays
  - begin: "2022-03-09T11:23:00-04:00"
    end: "2022-03-09T11:25:00-04:00"
    repeat:
    - daily
  - begin: "2022-03-09T12:41:00-04:00"
    end: "2022-03-09T12:43:00-04:00"
    repeat:
    - weekly
  - begin: "2022-01-30T12:38:00-04:00"
    end: "2022-01-30T12:40:00-04:00"
    repeat:
    - monthly
```

## Is this change a patch?

No.


api/core/v2/time_window.go
```diff
@@ -129,3 +131,212 @@ func (t *TimeWindowWhen) InWindows(current time.Time) (bool, error) {
...
+	if err != nil {
+		return fmt.Errorf("invalid begin time format: %v", err)
+	}
+	end, err := t.GetEndTime()
+	if err != nil {
+		return fmt.Errorf("invalid end time format: %v", err)
+	}
+	if end.Unix()-begin.Unix() < 0 {
+		return errors.New("end time must be after begin time")
+	}
```
echlebek, line 178: if end.Before(begin) {}

api/core/v2/time_window.go
```diff
@@ -129,3 +131,212 @@ func (t *TimeWindowWhen) InWindows(current time.Time) (bool, error) {
...
+
+	if len(t.Repeat) == 0 {
+		return fmt.Errorf("missing repeat interval")
+	}
```
echlebek, line 182: Isn't never repeating a valid use case here? Could schedule a one-off subdue?

api/core/v2/time_window.go
```diff
@@ -129,3 +131,212 @@ func (t *TimeWindowWhen) InWindows(current time.Time) (bool, error) {
...
+	if err != nil {
+		return false
+	}
+
+	actualTime = actualTime.In(beginTime.Location())
+	duration := endTime.Unix() - beginTime.Unix()
+	beginHour, beginMin, beginSec := beginTime.Clock()
+
+	todayBegin := time.Date(actualTime.Year(), actualTime.Month(), actualTime.Day(), beginHour, beginMin, beginSec, 0, beginTime.Location())
+	todayEnd := time.Unix(todayBegin.Unix()+duration, 0)
```
echlebek, line 288: todayEnd := todayBegin.Add(duration)

api/core/v2/time_window.go
```diff
@@ -129,3 +131,212 @@ func (t *TimeWindowWhen) InWindows(current time.Time) (bool, error) {
...
+	if err != nil {
+		return false
+	}
+	endTime, err := t.GetEndTime()
+	if err != nil {
+		return false
+	}
+
+	actualTime = actualTime.In(beginTime.Location())
+	duration := endTime.Unix() - beginTime.Unix()
```
echlebek, line 284: duration := endTime.Sub(beginTime)

api/core/v2/time_window.go
```diff
@@ -129,3 +131,212 @@ func (t *TimeWindowWhen) InWindows(current time.Time) (bool, error) {
...
+	beginHour, beginMin, beginSec := beginTime.Clock()
+
+	todayBegin := time.Date(actualTime.Year(), actualTime.Month(), actualTime.Day(), beginHour, beginMin, beginSec, 0, beginTime.Location())
+	todayEnd := time.Unix(todayBegin.Unix()+duration, 0)
+	todayEnd = todayEnd.In(beginTime.Location())
+
+	return actualTime.Unix() > todayBegin.Unix() && actualTime.Unix() < todayEnd.Unix()
```
echlebek, line 291: return actualTime.After(todayBegin) && actualTime.Before(todayEnd)

api/core/v2/time_window.go
```diff
@@ -129,3 +131,212 @@ func (t *TimeWindowWhen) InWindows(current time.Time) (bool, error) {
...
+	if err != nil {
+		return false
+	}
+	endTime, err := t.GetEndTime()
+	if err != nil {
+		return false
+	}
+
+	actualTime = actualTime.In(beginTime.Location())
+	duration := endTime.Unix() - beginTime.Unix()
```
echlebek, line 261: duration := endTime.Sub(beginTime)

api/core/v2/time_window.go
```diff
@@ -129,3 +131,212 @@ func (t *TimeWindowWhen) InWindows(current time.Time) (bool, error) {
...
+	beginHour, beginMin, beginSec := beginTime.Clock()
+
+	thisWeekBegin := time.Date(actualTime.Year(), actualTime.Month(), actualTime.Day(), beginHour, beginMin, beginSec, 0, beginTime.Location())
+	dayOffset := int(weekday) - int(actualTime.Weekday())
+	thisWeekBegin = thisWeekBegin.AddDate(0, 0, dayOffset)
+	thisWeekEnd := time.Unix(thisWeekBegin.Unix()+duration, 0)
```
echlebek, line 267: thisWeekEnd := thisWeekBegin.Add(duration)

api/core/v2/time_window.go
```diff
@@ -129,3 +131,212 @@ func (t *TimeWindowWhen) InWindows(current time.Time) (bool, error) {
...
+	thisWeekEnd = thisWeekEnd.In(beginTime.Location())
+
+	return actualTime.Unix() > thisWeekBegin.Unix() && actualTime.Unix() < thisWeekEnd.Unix()
```
echlebek, line 270: return actualTime.After(thisWeekBegin) && actualTime.Before(thisWeekEnd)

echlebek (commented):
Looks good, but I think we should eliminate the conversions to unix time, as they are lossy and could be a source of bugs. Luckily the Time object provides functions that should make these conversions unnecessary.

c-kruse (commented):
It was difficult to tell so I had to add a test case to know for sure, but looks like we are extending the subdues into the past as well. I don't know that I would have necessarily made that assumption as an operator. e.g. "Begin: 2023-01-01, End: 2023-01-30, Repeat: annually" would apply to 2022-01-10.

Perhaps that's something that we'd want in the docs?

Test case if you'd want to include it. Otherwise maybe I missed the same assertion somewhere else?
```
diff --git a/api/core/v2/time_window_test.go b/api/core/v2/time_window_test.go
index b94a2b8a..04207255 100644
--- a/api/core/v2/time_window_test.go
+++ b/api/core/v2/time_window_test.go
@@ -382,6 +382,13 @@ func TestTimeWindowRepeated_InDayTimeRange(t *testing.T) {
                        "2022-03-30T06:00:00-04:00", // wednesday
                        time.Tuesday,
                        true,
+               }, {
+                       "positive week before",
+                       "2022-03-22T09:00:00+00:00",
+                       "2022-03-24T11:00:00+00:00",
+                       "2022-03-16T06:00:00-04:00", // wednesday before
+                       time.Tuesday,
+                       true,
                },
        }
```

fguimond:
> It was difficult to tell so I had to add a test case to know for sure, but looks like we are extending the subdues into the past as well. I don't know that I would have necessarily made that assumption as an operator. e.g. "Begin: 2023-01-01, End: 2023-01-30, Repeat: annually" would apply to 2022-01-10.

@c-kruse - This a very good observation that I overlooked. From the requirements at https://github.com/sensu/sensu-go/issues/1955#issuecomment-1074144897 it's very clear that we should not dwell on the past but rather look in the future. I think a simple check to make sure the current time is not before the begin time should do the trick.

> The following example — with the same begin timestamp as the previous example – is applied every week from Monday through Thursday after Sunday March 20st (PT), but does not apply to March 20th.
> ```
> subdue:
>   - begin: 2022-03-20T16:04:00-0700
>     end: 2022-03-20T18:58:00-0700
>     repeat:
>       - mondays
>       - tuesdays
>       - wednesdays
>       - thursdays
> ```

echlebek (approved):
Looks good! I think there are some commented-out lines of code, they should probably be removed before merge?

c-kruse (changes requested):
One more thing! Mind rebasing on top of develop/6?

c-kruse (commented):
And I'm actually suspicious about the time.In(time.Location()) type logic.

TIL I really do not understand time zone data. When I run the tests they fail for me. I think you should probably be able to reproduce with `TZ="PST8PDT" go test .`

fguimond:
> And I'm actually suspicious about the time.In(time.Location()) type logic.
> 
> TIL I really do not understand time zone data. When I run the tests they fail for me. I think you should probably be able to reproduce with `TZ="PST8PDT" go test .`

Weird, the only errors I got were the json-iter panic errors when running the tests as is. However when I disable the panicky tests the suite runs fine, either with the TZ env var set or with my machine actually set to a European time zone.

c-kruse (commented):
I'm out today, so I won't hold this up. Here's the full context of what I was seeing. My gut tells me that this may actually be a welcomed coincidence for people running servers with local time.

```
christian@davis core/v2 (subdue-logic) » go test .
current 2022-03-28 15:09:08 +0600 +0600 Monday
now 2022-04-07 22:31:51.284779687 +0600 +0600 Thursday
location
--- FAIL: TestTimeWindowRepeated_InTimeRange (0.00s)
    --- FAIL: TestTimeWindowRepeated_InTimeRange/positive_after_different_day (0.00s)
        time_window_test.go:473:
                Error Trace:    time_window_test.go:473
                Error:          Not equal:
                                expected: true
                                actual  : false
                Test:           TestTimeWindowRepeated_InTimeRange/positive_after_different_day
FAIL
FAIL    github.com/sensu/sensu-go/api/core/v2   0.074s
FAIL
```

It is that specific test, and it's obviously system dependent. My best guess is that time.Parse saw the 2022-03-22 date with the -7 offset and decided that it was my local tz, then somehow translates the target date (in DST that next winter) to a -8 offset? idk man. time is hard. good luck.

c-kruse (approved):
Looks good! Thanks!

fguimond:
@c-kruse thanks for the detailed explanations, I was able to reproduce the test error and find a solution. I ended up using time.ParseInLocation() with UTC time instead of time.Parse(). The latter seemed to base the offset information on top of the system timezone and apply the DST settings accordingly. Using UTC takes care of that problem and removes any unwanted/unexpected DST logic.

PR #4681, opened 2022-03-31
Remove json-iterator from the project

echlebek:
## What is this change?

Recently we found that json-iterator can panic due to incorrect use of
package unsafe. Since json performance is less of a concern than it used
to be, and since encoding/json has progressed in its performance since,
then, I believe it's wise to remove it from the project altogether.

I've also created new tags for core/v2, corev/3, and types modules, and
have updated the go.mods in the repository to reflect that.

It avoids a panic caused by unsafe code in json-iterator version 1.11 by simply removing the dependency on the library altogether.

I don't think so. The panic was never triggered in anything other than a development environment, and this is not really a use-facing change.

Automated testing is sufficient.

## Is this change a patch?

Technically yes from a semver perspective, but I think there is a low but non-zero chance of performance regression, so I'm more comfortable with applying it to the main branch.


amdprophet (approved):
Looks good. Have you compared before/after performance with this change?

PR #4682, opened 2022-04-04
Run tests for all modules

c-kruse:
## What is this change?

Add CircleCI job to test core/v2 and types modules

No use in tests that aren't ran!

Probs not.

Would y'all prefer build.sh always run types and core/v2 unit tests? I'm on the fence. It'd probably mean no go module caching in CI for those modules. Also go 1.18 seems to have problems with the types tests (making running the types test suite locally if you've upgraded to 1.18 a hassle - have so far successfully avoided a goenv type tool)

CircleCI is a mystery.

N/A

CircleCI

## Is this change a patch?

N

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,6 +36,18 @@ jobs:
           path: ~/sensu-go
       - go/mod-download-cached
       - run: go test ./...
+  test-types:
+    <<: *sensu_go_build_env
+    environment:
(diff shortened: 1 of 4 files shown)


echlebek (commented):
Thanks for doing this. Can you add one for core/v3 as well?

echlebek (approved):

Issue #4697, opened 2022-04-20
Proposal - Global core/v3 resources


c-kruse:
The core/v3.Resource interface does not currently have a way to indicate that a Resource is global.

Sensu currently tests whether or not core/v2 resources are global by checking if the SetNamespace method is a no-op. This behavior is also relied on when defaulting namespaces (for example to the context of sensuctl.) Since the corev3.Resource interface doesn't have this method, core/v3 resources have so far been assumed to be namespaced.

## Feature Suggestion

If we're going to be moving more over to the corev3 resource style in the future I think we will need a way to indicate which resource types are global.

## Possible Implementation
I think we should add an interface to the sensu-go/core/v3 package for this purpose.
```
type GlobalResource interface {
	// IsGlobalResource returns true when the resource
	// is not namespaced.
	IsGlobalResource() bool
}
```

Resources can then choose to optionally implement the IsGlobalResource function to explicitly indicate whether they are global or namespaced.

We would then update all of the compatibility layers to test for this. I've ran into these so far:
[corev3.V2ResourceProxy](https://github.com/sensu/sensu-go/blob/3a31c256f42c47adcd918079c7980ff201d6487e/api/core/v3/resource.go#L66)
[the types.compat package](https://github.com/sensu/sensu-go/blob/main/types/compat/compat.go#L37)

We could probably also update our codegen library to test for this interface for validation (asserting namespace is not set) and uri pathing.
## Context

Trying to get the enterprise secrets/v2 implementation up and running. The proposed SensuProvider is a global resource.

PR #4700, opened 2022-04-20
Add core/v3 support for global resources

c-kruse:
## What is this change?

Closes https://github.com/sensu/sensu-go/issues/4697

* Adds a GlobalResource interface that corev3 Resoruces
can optionally implement in order to indicate they are not namespaced.
* Also updates the V2ResourceProxy to honor the core/v2 behavior
of no-op SetNamespace implementations to indicate a resource
is global.
* Fixes some unit tests that were making poor assertions.
* Updates the types/compat package to check for GlobalResources.

This will enable global resources to be implemented using the core/v3.Resource interface.

Yes

I am a bit iffy on the effects of adding api/core/v3 to the replace block of types/go.mod.

Not really. May need a followup pr to bump the version number on the dependency from types on api/core/v3?

N/A

Using the enterprise SensuProvider api WIP.

Without this change sensuctl assumes all corev3 resources are namesapced so it defaults namespace to "default":
```
$ cat << EOF | sensuctl create
type: SensuProvider
api_version: secrets/v2
metadata:
  name: foo
spec:
...

Error: error putting resource #0 with name "foo" and namespace "default" (/api/enterprise/secrets/v2/sensu-providers/base64-encoded): namespace must be empty. sensu-providers are global resources
```

Including this change:
```
$ cat << EOF | sensuctl create
type: SensuProvider
api_version: secrets/v2
metadata:
  name: foo
spec:
...

$ sensuctl dump secrets/v2.SensuProvider
type: SensuProvider
api_version: secrets/v2
metadata:
  created_by: admin
  labels:
    sensu.io/managed_by: sensuctl
  name: foo
...
```

## Is this change a patch?

N

### DIFF
--- a/CHANGELOG-6.md
+++ b/CHANGELOG-6.md
@@ -16,6 +16,9 @@ stopping.
 ### Fixed
 - Fixed several data races in schedulerd.
 - Mitigate a data race in agentd sessions.
+### Added
+- GlobalResource interface in core/v3 allows core/v3 resources to
+be marked as global resources.
 
 ## [6.7.0] - 2022-04-21
 
--- a/api/core/v3/resource.go
+++ b/api/core/v3/resource.go
@@ -37,6 +37,14 @@ type GeneratedType interface {
 	Validate() error
 }
 
+// GlobalResource  is an interface for indicating
+// a resource's namespace strategy
+type GlobalResource interface {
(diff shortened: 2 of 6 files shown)


echlebek (approved):
LGTM! You'll need to tag core/v3, use that tag in types, and release a tag for types as well.

Issue #4616, opened 2022-02-07
Add postgres configuration to CLI flags and configuration file


echlebek:
Remove the PostgresConfig configuration method, which we have had ongoing reliability issues with. Add CLI flags and configuration file support for configuring postgresql.

We should support separate methods of configuring postgresql for configuration and state stores respectively. Users can choose between using the same database for state and configuration, or choose to use different databases.

Specification
-----------------
Configuration structure for backend config:
```
type StoreConfig struct {
  // ConfigurationStore specifies the selected configuration store to use (either "postgres", "etcd", or "dev")
  ConfigurationStore string

  // StateStore specifies the selected state store to use (either "postgres" or "dev")
  StateStore string

  // PostgresConfigurationStore contains postgres configuration store details. It's only valid to set this when
  // ConfigurationStore is set to be "postgres".
  PostgresConfigurationStore *PostgresConfig

  // EtcdConfigurationStore contains etcd configuration store details. It's only valid to set this when
  // ConfigurationStore is set to be "etcd".
  EtcdConfigurationStore *EtcdConfig

  // PostgresStateStore contains postgres state store details. It's only valid to set this when 
  // StateStore is set to be "postgres".
  PostgresStateStore *PostgresConfig
}

type EtcdConfig struct {
  TLS *corev2.TLSClientConfig
  URLs []string
}

type PostgresConfig struct {
  DSN string
}
```

Command-line flags, mapping to the above configuration structs:
```
--config-store
--state-store
--postgres-config-store-dsn
--postgres-state-store-dsn
--etcd-config-store-urls
--etcd-config-store-cert
--etcd-config-store-key
--etcd-config-store-ca-cert
```

There should also be associated environment variables, similar to how other configuration options work.

Postgresql is required for production deployments, but etcd is not.

When dev mode is specified, sensu-backend will start up an embedded etcd instance that cannot be clustered. The same instance will be used for state and configuration storage. The embedded etcd database directory will be a subdirectory of --state-dir, similarly to how embedded etcd works right now.

The existing etcd configuration flags are to be removed.

PR #4704, opened 2022-04-24
Remove support for embedded etcd

echlebek:
## What is this change?

    Remove embedded etcd clustering
    
    This is a substantial breaking commit. It removes several etcd
    configuration options for configuring embedded etcd clustering. At one
    time, this was the only way that sensu-go worked.
    
    We've learned that embedded etcd introduces a lot of operational
    fragility, and are now mandating that etcd be an external dependency.
    
    To improve hackability and developer velocity, there is now a
    zero-dependency developer mode for sensu-go. The developer mode makes
    use of a single-node etcd instance, but we reserve the right to change
    it to something else in the future of the 7 series releases.
    
    This is the first in a series of commits that will add full support for
    PostgreSQL to sensu-go. We have chosen to de-emphasize etcd in favour of
    PostgresQL, as we have found that it is far easier for users to deploy
    PostgresQL on cloud infrastructure (and possibly also on-prem
    infrastructure).
    
    In addition to these breaking changes, some routine gardening has been
    done. The concept of integration build tags has been removed.
    Integration test build tags are troublesome for static analysis tools
    and editors that compile on save. We introduced them at a time that Go
    didn't have test caching, and now it is time to de-introduce them. The
    whole senus-go test suite can run with the race detector enabled in just
    a few minutes.

Part of #4616

Yes

I decided to pay down some technical debt and remove the `+built integration` tags from our integration tests.

Documentation updates will be needed, as embedded etcd is no longer supported for the 7.x series. cc @hillaryfraley 

I manually tested a few things under an external etcd including backend init. All is well, but we'll need to update our staging environment to not use embedded etcd.

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -192,7 +192,7 @@ func devModeClient(ctx context.Context, config *Config, backend *Backend) (*clie
 	cfg.InitialAdvertisePeerURLs = cfg.ListenPeerURLs
 	cfg.AdvertiseClientURLs = config.EtcdClientURLs
 	cfg.Name = "dev"
-	cfg.LogLevel = "warn"
+	cfg.LogLevel = config.LogLevel
 	cfg.ClientLogLevel = config.EtcdClientLogLevel
 
 	// Start etcd
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -524,6 +524,13 @@ func Initialize(ctx context.Context, config *Config) (*Backend, error) {
 			clusterVersion = status.Version
 			break
 		}
+	} else {
+		status, err := b.Client.Status(ctx, config.EtcdClientURLs[0])
+		if err != nil {
+			logger.WithError(err).Error("error getting etcd cluster info")
+		} else {
+			clusterVersion = status.Version
+		}
 	}
 
 	// Load the JWT key pair
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -180,100 +180,20 @@ type StoreV2Updater interface {
 	UpdateStore(to storev2.Interface)
 }
 
-func newClient(ctx context.Context, config *Config, backend *Backend) (*clientv3.Client, error) {
-	if config.NoEmbedEtcd {
-		logger.Info("dialing etcd server")
-		tlsInfo := (transport.TLSInfo)(config.EtcdClientTLSInfo)
-		tlsConfig, err := tlsInfo.ClientConfig()
-		if err != nil {
-			return nil, err
-		}
-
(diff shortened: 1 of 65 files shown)


backend/backend.go
```diff
@@ -559,8 +524,6 @@ func Initialize(ctx context.Context, config *Config) (*Backend, error) {
 			clusterVersion = status.Version
 			break
 		}
-	} else {
```
amdprophet, line 562: For consistency, should we continue to report the cluster version when in dev mode? I believe there's some value in keeping it (e.g. developing checks against a dev sever).

backend/backend.go
```diff
@@ -180,101 +180,21 @@ type StoreV2Updater interface {
...
-	cfg.Name = config.EtcdName
-	cfg.LogLevel = config.EtcdLogLevel
+	cfg.ListenClientURLs = config.EtcdClientURLs
+	cfg.ListenPeerURLs = []string{"http://127.0.0.1:2380"}
+	cfg.InitialCluster = "dev=http://127.0.0.1:2380"
+	cfg.InitialClusterState = "new"
+	cfg.InitialAdvertisePeerURLs = cfg.ListenPeerURLs
+	cfg.AdvertiseClientURLs = config.EtcdClientURLs
+	cfg.Name = "dev"
+	cfg.LogLevel = "warn"
```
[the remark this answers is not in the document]
echlebek: OK, I think that makes sense.

backend/backend.go
echlebek: I had an issue with this, because dev mode uses the embedded client and not a real client, the call panics. But I can look into making it work somehow.

echlebek: Figured it out :muscle:

backend/schedulerd/cron_scheduler.go
```diff
@@ -101,9 +105,10 @@ func (s *CronScheduler) Interrupt(check *corev2.CheckConfig) {
 
 // Stop stops the cron scheduler.
 func (s *CronScheduler) Stop() error {
+	s.cancel()
+	s.stopWg.Wait()
 	cronCounter.WithLabelValues(s.check.Namespace).Dec()
 	logger.Info("stopping cron scheduler")
```
amdprophet, line 111: Does it make more sense to move this log line above the call to `s.stopWg.Wait()` similar to the `Stop()` method in `IntervalScheduler`?

amdprophet:
Just the one change to consider; everything else looks great!

amdprophet (approved):

PR #4717, opened 2022-04-29
Bump sensu-go and types api/core/v3 dependency to v3.6.2

c-kruse:

@echlebek looks like you might not usually PR these, but wanted to sanity check myself here.

Already pushed a api/core/v3.6.2 tag on the current head of develop/6.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -58,7 +58,7 @@ require (
 	github.com/robfig/cron/v3 v3.0.1
 	github.com/sensu/lasr v1.2.1
 	github.com/sensu/sensu-go/api/core/v2 v2.14.0
-	github.com/sensu/sensu-go/api/core/v3 v3.6.1
+	github.com/sensu/sensu-go/api/core/v3 v3.6.2
 	github.com/sensu/sensu-go/types v0.10.0
 	github.com/shirou/gopsutil/v3 v3.21.12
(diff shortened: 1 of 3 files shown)


echlebek (commented):
go.sum changes seem incomplete :thinking:

c-kruse:
@echlebek I think this is really more of a consequence of adding the replace directive in the types module https://github.com/sensu/sensu-go/commit/ac3387abab706a6ab49e94a9ab5d05e8b2aa2c2f#diff-2309f8be2bbb14fc4f3ced43b726b6b7c91cfb847a559aba04eed5de2a4b47d2R8

It is all `go mod tidy`d up.

We may want to consider removing the replace directives and just dealing with the labor of doing the "push, tag, go get, push" shuffle. If I'm reading [the docs](https://go.dev/ref/mod#go-mod-file-replace) correctly, the difference between how the sensu-go modules are built and tested with the replace directives vs how the enterprise version may resolve the types and api module versions is a little bit concerning.

For folks developing with go 1.18 it looks like multi-module workspaces work well even though we have go 1.13 indicated in our go.mods.

c-kruse:
@echlebek  let's get this merged with the incomplete go.sum artifact and follow up with removing those replace directives. Pulling those out and documenting it was a bigger change than I had hoped.

Issue #4609, opened 2022-02-05
Epic: Sensu 7.x: Rework storage and add OSS Postgresql support


echlebek:
## Description

For Sensu 7.x, we'd like to build upon what we've learned by building, maintaining, and supporting Sensu over the last five years. One of the most common points of difficulty in operating Sensu is having to build and maintain a performant etcd cluster. For many IT organizations, this requirement makes using Sensu too difficult. None of the major cloud providers offer hosted etcd, and running etcd on-prem can be tricky to get right for a lot of organizations.

When we initially hit scaling problems with etcd, we made the postgres event store available to help users deal with higher volumes of event writes than etcd could handle. While this was effective, the postgres event store also had an awkward configuration API, and effectively doubled the number of external dependencies of Sensu. It was also only available in the commercial edition of Sensu, complicating the development process between OSS and commercial versions of the software.

We want to double-down on our postgres support. We want to offer postgres support in both OSS Sensu and in the commercial product, and eventually we want Sensu to be able to work with only postgres. Additionally, we want postgres configuration to be done at application start up time, not as a special API.

- [x] https://github.com/sensu/sensu-go/issues/4614
- [x] https://github.com/sensu/sensu-go/issues/4615
- [x] https://github.com/sensu/sensu-go/issues/4616
- [x] https://github.com/sensu/sensu-go/issues/4628
- [x] https://github.com/sensu/sensu-go/issues/4899
- [x] https://github.com/sensu/sensu-go/issues/4900
- [x] https://github.com/sensu/sensu-go/issues/4901
- [x] https://github.com/sensu/sensu-go/issues/4902
- [x] https://github.com/sensu/sensu-enterprise-go/issues/2542

calebhailey:
Related: https://github.com/sensu/sensu-go/issues/4446

echlebek:
Related: https://github.com/sensu/sensu-go/issues/4625

ccressent:
While we are reworking storage, and since it's also suggested in other related issues that we could even, ultimately, not have etcd at all, I was wondering if it would be useful for us to think about relaxing the "Postgres" dependency to just an "SQL store" dependency. Maybe not as loose as "SQL store", but something that's Postgres, MySQL, SQLite compatible.

[...]

PR #4733, opened 2022-05-04
Re-work sensu-backend store configuration

### LINKED ISSUE #4616: Add postgres configuration to CLI flags and configuration file

echlebek:
## What is this change?

* Add postgres configuration for state store and config store
* Change etcd configuration so that it pertains only to config store
* Remove upgrade command (but not migration library bits)
* Modify init command

Closes #4616 

Yes

I verified that dev mode works manually.

## Is this change a patch?

No.


c-kruse (approved):
👍  Looks good to me. Esp like the help message template organization.

Only nitpick is that I think that checking that `etcd-config-store-urls` isn't empty when config-store is "etcd" might be nice. The startup error isn't particularly cryptic, but it doesn't point you right at the answer either.

```
$ ./sensu-backend start --cache-dir /tmp/sensu --state-dir /tmp/sensu
{"component":"backend","error":"etcdclient: no available endpoints","level":"fatal","msg":"error executing sensu-backend","time":"2022-05-05T13:13:25-07:00"}
```

calebhailey:
Related: #4609

Issue #4747, opened 2022-05-16
Add VMware as Cloud Provider for agent discovery


mkennedy85:
Add VMware as a Cloud Provider for agent discovery to populate the entity metadata in order to assist targeting entities by platform in hybrid environments.

## Feature Suggestion
The agent discovery to support VMware as a cloud provider that is returned and populates entity metadata.

## Possible Implementation
The change has been implemented and a PR will be opened as the discovery is identical to the other providers such as AWS and GCP.

## Context
In a hybrid environment, this allows for easier targeting of entities that are in a VMware environment as opposed to AWS or GCP. This allows for easier multi-platform entity management.

PR #4748, opened 2022-05-16
Author: Michael Kennedy <michael_l_kennedy@me.com>

mkennedy85:
Committer: Michael Kennedy <michael_l_kennedy@me.com>

Signed-off-by: Michael Kennedy <michael_l_kennedy@me.com>

  Adding support for VMware as a Cloud Provider returned by GetCloudProvider(ctx).

## What is this change?

Add VMware as a Cloud Provider applied to entity metadata through agent discovery.

The agent is capable of platform discovery and supports AWS and GCP. This same approach is capable and should detect VMware as a platform cloud provider.

Yes and the Changelog entry has been added as specified in the contribution guidelines.

The change is rather trivial and does not introduce error conditions, but rather simply detects a possible output returned from an existing exec command.

There did not appear to be documentation pertaining to this agent behavior, so no updates have been made.

This change was verified on a Linux host running on a VMware vSphere 6.7 hypervisor, and the output was verified from the exec command and then validated using the existing function.

## Is this change a patch?

No it is not.

### DIFF
--- a/CHANGELOG-7.md
+++ b/CHANGELOG-7.md
@@ -25,6 +25,7 @@ software is upgraded when there are active keepalive failures.
 entity subscriptions and/or a check named `deregistration`.
 - Upgraded Go version from 1.17.1 to 1.18.1.
 - Changed sensu-backend etcd configuration options.
+- Added VMware as a value return in `GetCloudProvider` function.
 
 ### Removed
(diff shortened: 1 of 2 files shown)


mkennedy85:
This PR correlates with https://github.com/sensu/sensu-go/issues/4747.

asachs01:
cc @portertech to take a gander.

PR #4750, opened 2022-05-19
Get rid of docker and go-ansiterm deps

ccressent:
## What is this change?

This removes our dependency on "github.com/docker/docker" and
"github.com/Azure/go-ansiterm".

This has the effect of:
- removing the Github security warnings, since we don't import docker/docker at all
- removing low level code paths that were dead anyway
- removing the dependencies those code paths were using

Also see commit message:
```
This branch never worked: calling GetTermSize() on fd 0 always results
in an error: "inappropriate ioctl for device". We have been using
width=80 all along!

In that light, I'm just deleting the branch, which automatically gets
rid of (very) low level dependencies issuing ioctl() system calls on
Unix and calls to GetConsoleScreenBufferInfo() in kernel32.dll on
Windows.
```

No.

No.

Yes, I should have been suspicious of those `GetTermSize(0)` calls and checked
that they weren't always erroring out _before_ doing a bunch of work to
reimplement `GetTermSize()` without the docker and azure dependencies...

(See 436c3c9d930d18b5e2edb65e1c5a8b2178d3e9fe)

No documentation changes.

- Build
- Tests
- Played with `sensuctl` for 3 minutes

## Is this change a patch?

No. But we should cherry-pick it onto `develop/6` and `release/6.7` so that we
get rid of those dependencies and the security warnings there too.


echlebek (approved):
Much better :+1:

Issue #4625, opened 2022-02-09
Epic: Sensu 7.x: Deprecate etcd


echlebek:
## Description

We want to diminish the role etcd plays in Sensu and ultimately remove support for it. That means removing the option to run an embedded etcd, which we have found to be quite troublesome in practice, and eventually developing postgres-only solutions to features previously supported by etcd.

## Tasks
- [ ] #4446
- [ ] #4617

PR #4751, opened 2022-05-19
Fix an issue with non-dev event store

echlebek:
The store was never assigned to the backend, and so non-dev mode is broken.


amdprophet (approved):

calebhailey:
Related: #4609 and/or #4625

PR #4770, opened 2022-06-08
Add store/v2 watcher

### LINKED ISSUE #4739: Add Etcd store/v2 Watcher Implementation
Design Spike doc: https://docs.google.com/document/d/1OfT6fLIqKggGt0pKcFiBGa6Jb5KgygX7DpSy9Pt0XSo/edit?usp=sharing

When --config-store is etcd we will need to switch to a store/v2 Watcher interface common with the postgres implementation for the store/v2 backed resources.

## Interface
Reasoning for the interface detailed in https://github.com/sensu/sensu-go/issues/4721

// store/v2 interface similar to GetResourceV3Watcher
type Watcher interface {
    Watch(context.Context, ResourceRequest) <-chan WatchEvent
}

type WatchEvent struct {
    Action store.WatchActionType // or equiv storev2 replacement CREATE|UPDATE|DELETE|WatchError?
    Resource corev3.Resource
}

## Implementation
This should be a fairly straight-forward adaption of our existing watchers. Examples: https://github.com/sensu/sensu-go/blob/main/backend/store/etcd/watcher.go, https://github.com/sensu/sensu-go/blob/4e17edcbe360e48ee88e9af028ee00a88484c197/backend/store/etcd/watchers.go#L131

echlebek:
## What is this change?

This commit adds an etcd-based watcher for store/v2. It ports the
original store's watcher tests to ensure that the same use cases are
addressed.

Closes #4739 

No, this is an implementation detail only.

It was tricky to get right.


backend/store/v2/etcdstore/watcher.go
```diff
@@ -0,0 +1,149 @@
...
+	WatchEventsLabelStatus       = "status"
+	WatchEventsLabelResourceType = "resource"
+	WatchEventsLabelNamespace    = "namespace"
+
+	WatchEventsStatusHandled = "handled"
+	WatchEventsStatusDropped = "dropped"
+)
+
+var (
+	WatchEventsProcessed = prometheus.NewCounterVec(
```
c-kruse, line 28: It's probably in our best interest to expose this metric from the postgres store implementation too. Thoughts on moving this to the backend/store/v2 package?

echlebek: Agreed!

amdprophet (approved):

calebhailey:
Related: #4609

PR #4772, opened 2022-06-15
Produce mapping of resource fieldsets to typenames

jamesdphillips:
## Summary

Adds new code generator that produces a mapping of typename to the resource's fieldset function (used by search.)

* Since the fieldset function is not a method on the type, it was previously impossible to call the function unless you knew the type ahead of time.
* The code generator works because we've thankfully followed a convention of naming the functions `{typename}Fields`.
* I went with code generation because I didn't want to have to manually create a list and ensure others keep it up to date over time.
* Required by https://github.com/sensu/sensu-enterprise-go/pull/2280

**Alternative Approach**

I could have refactored each of the fieldset functions so that they were a method on each resource, however the scope of such a change is far greater.

```go
func (c *CheckConfig) Fieldset() map[string]string {
  // ...
```

## Usage

```go
wrapper := types.WrapResource(corev2.FixtureCheckConfig("disk-check"))
fset, ok := types.LookupFieldsetFn(wrapper.ApiVersion, wrapper.Type)

selector, ok := selector.ParseFieldSelector(q)
selector.Matches(fset.Fields())
```


jamesdphillips:
Closing as we've decided to head in the direction laid out in the "alternate approach" section listed above

PR #4773, opened 2022-06-16
Prevent executing old or duplicate check requests

amdprophet:
## What is this change?

Adds a new map of check keys -> issued timestamps that is used to determine if received check requests are older than or duplicates of the last executed check request.

Sensu backends currently send a check request to all matching subscriptions. If an agent and check have multiple matching subscriptions then the agent will receive the check request for every matching subscription. This can result in one of two undesirable behaviours:
1. If one of the check requests are still being executed when another is received it will result in the `check execution still in progress` error being logged.
2. If the check requests are executed quickly enough to avoid the `check execution still in progress` error then check history & features that rely on it (e.g. flap detection) may behave strangely.

Yes.

No new documentation is required unless we want to document that agents and checks should only share one subscription.

I've added new test coverage.

## Is this change a patch?

No.

### DIFF
--- a/agent/check_handler.go
+++ b/agent/check_handler.go
@@ -136,12 +136,6 @@ func (a *Agent) getLastIssued(request *corev2.CheckRequest) int64 {
 	return issued
 }
 
-func (a *Agent) removeLastIssued(request *corev2.CheckRequest) {
-	a.lastIssuedMu.Lock()
-	defer a.lastIssuedMu.Unlock()
-	delete(a.lastIssued, checkKey(request))
-}
-
 func (a *Agent) setLastIssued(request *corev2.CheckRequest) {
 	a.lastIssuedMu.Lock()
 	defer a.lastIssuedMu.Unlock()
--- a/agent/check_handler.go
+++ b/agent/check_handler.go
@@ -32,7 +32,7 @@ const (
 )
 
 var (
-	dupCheckRequestErr = errors.New("check request has already been received - agent & check may have multiple matching subscriptions")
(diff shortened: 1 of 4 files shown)


agent/check_handler.go
```diff
@@ -31,6 +31,11 @@ const (
 	measureNullStatus = "null-status"
 )
 
+var (
+	dupCheckRequestErr = errors.New("check request has already been received - agent & check may have multiple matching subscriptions")
```
c-kruse, line 35: At least for me `&` doesn't get decoded and ends up looking like this:
```
check request has already been received - agent \u0026 check may...
```

Mind using "and" instead?

amdprophet: Nice catch, fixed!

agent/check_handler.go
```diff
@@ -31,6 +31,11 @@ const (
 	measureNullStatus = "null-status"
 )
 
+var (
+	dupCheckRequestErr = errors.New("check request has already been received - agent and check may have multiple matching subscriptions")
```
calebhailey, line 35: Nice 👌

calebhailey: @asachs01 FYI ☝️ this is the new error message. 

Should we also change the log level? This is an agent log message, so not quite as problematic as backend logging. Let us know if we should also adjust the log level.

amdprophet: The duplicate error is logged as a warning; old check request error is logged as an error. Let me know what y'all think.

asachs01: I think warn is appropriate @amdprophet. As for the error message, that looks great. 👍

Issue #4765, opened 2022-05-31
Add a namespaces table to Postgres


amdprophet:
Add a new table in Postgres to store Sensu namespaces. Namespaced resources in other tables should use a FK constraint to refer to the correct namespace. FK constraints should refer to an auto-incrementing ID for the namespace.

Depends on #4764.

PR #4777, opened 2022-06-21
Add support for namespaces in the Postgres store

amdprophet:
## What is this change?

Adds a new `namespaces` table to the Postgres store and updates the `entity_configs` & `entity_states` table to use a foreign key to namespaces.

Closes https://github.com/sensu/sensu-go/issues/4765.

Yes.

Debugging SQL queries when tests fail can be quite difficult. We may want to look into improving error output when SQL queries fail.

I've added a bunch of new test coverage to test the new resources.

## Is this change a patch?

No.

### DIFF
--- a/backend/store/postgres/namespaces_schema.go
+++ b/backend/store/postgres/namespaces_schema.go
@@ -1,0 +1,188 @@
+package postgres
+
+const namespaceSchema = `
+--
+CREATE TABLE IF NOT EXISTS namespaces (
+	id                 bigserial PRIMARY KEY,
+	name               text NOT NULL,
+	selectors          jsonb,
+	annotations        jsonb,
+	created_at         timestamptz NOT NULL DEFAULT NOW(),
+	updated_at         timestamptz NOT NULL DEFAULT NOW(),
+	deleted_at         timestamptz,
+	CONSTRAINT namespace_unique UNIQUE (name)
+);
+
+CREATE TRIGGER refresh_namespaces_updated_at BEFORE UPDATE
+	ON namespaces FOR EACH ROW EXECUTE PROCEDURE
+	refresh_updated_at_column();
+`
+
+const addNamespaceForeignKeys = `
+-- Add unique namespaces across existing tables to the new namespaces table
+INSERT INTO namespaces (name)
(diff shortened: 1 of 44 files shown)


backend/store/postgres/namespaces_schema.go
```diff
@@ -0,0 +1,188 @@
...
+SELECT
+	name,
+	selectors,
+	annotations
+FROM namespaces WHERE name IN (SELECT unnest($1::text[]))
+`
+
+const deleteNamespaceQuery = `
+-- This query deletes a namespace. Any related data (e.g. entity configs,
+-- entity state, etc.) will also be deleted via ON DELETE CASCADE triggers.
```
c-kruse, line 150: In the etcd implementation I ended up taking special care to make sure the DeleteNamespace store method didn't orphan any resources [here](https://github.com/sensu/sensu-go/blob/6d913a110dea26ab9ede89948d8934db9a1d2c8f/backend/store/v2/etcdstore/store.go#L431-L442). Polling the team, it sounded like folks preferred to disallow deleting non-empty namespaces to the automatic cleanup approach.

amdprophet: I’d like to keep the cascade triggers to prevent any orphaned relational data and then add logic to the store to prevent deletion of non-empty namespaces. This way it’s less likely to have orphaned data in the case that a namespace gets deleted outside of the store. Thoughts?

c-kruse: 👍  works for me. This is more of a business logic requirement than a data modeling one.

My case for not cascading deletes was that as I understand it the foreign key constraint itself is preventing orphaned data. Not sure if we get a super useful error that we could handle differently when that constraint is violated.

echlebek: Some thoughts on this.

1. It's actually useful to be able to delete entire namespaces and our customers want that.
2. It's bad to accidentally delete entire namespaces, but it's not clear that's an actual risk.
3. It's a lot more likely to orphan data when etcd is in use, especially when used in concert with postgres.

This is a bit of a usability issue. We don't offer an easy way to clear out the contents of a namespace, users have to go around and delete every resource type which is a burden.

calebhailey:
Related: #4609

PR #4779, opened 2022-06-23
Use static strings for StoreName() in compat.go

echlebek:
StoreName is supposed to return an invariant string, not a computed one.


amdprophet (approved):

calebhailey:
Related: #4609

PR #4780, opened 2022-06-24
Use only storev2 for rbac package

echlebek:
You must use go.work in order for this to work, core/v2 isn't tagged yet


calebhailey:
Related: #4609

PR #4782, opened 2022-06-28
Refactor store/v2 interface

echlebek:
Pulling out context.Context from ResourceRequest, and putting it into the storev2.Interface methods


calebhailey:
Related: #4609

PR #4784, opened 2022-06-28
Move agentd to storev2

c-kruse:
Moves agentd to the store/v2 package.

Updates the backend/store/cache/v2 package to remove dependency on etcd.
Migrates agentd to store/v2 by:
  * Moving from cache to cache/v2 package
  * Adding Config for HealthRouter instead of constructing it on creation to match apid and remove etcd dep.
  * Using new storev2 WatchEvent
  * Adding Config/dependency for an Authenticator to matche apid and remove dependency on UserStore.
* Updates usage of the store/v2 interface in session.go to use new calling convention (ctx, req).
* Update some backend/api package compile errors for now store/v2 interface calling conventions.

## Dear Cyril,
Haha, don't want to @ you to disrupt your rest. I suspect my changes to cache/v2 may step on your toes if you've already made some progress on the schedulerd task. I hope we've taken similar steps if you have 🤞.

### DIFF
--- a/backend/agentd/agentd.go
+++ b/backend/agentd/agentd.go
@@ -19,7 +19,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sensu/sensu-go/agent"
 	corev2 "github.com/sensu/sensu-go/api/core/v2"
-	"github.com/sensu/sensu-go/backend/apid/actions"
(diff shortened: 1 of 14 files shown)


echlebek (approved):
:boom: :shipit:

calebhailey:
Related: #4609

PR #4767, opened 2022-06-06
Add support for EntityConfig in Postgres store

amdprophet:
**NOTE:** #4756 should be merged before this.

## What is this change?

Adds support for EntityConfig resources in the Postgres store/v2 methods. I also added additional migrations to rename the `entities` table to `entity_states` to prevent any confusion around what the table is used to store.

## Why is this change necessary?

Closes #4744
Closes #4719.

## Does your change need a Changelog entry?

Yes.

## Were there any complications while making this change?

Some of the Postgres store/v2 tests did not provide much insight into why tests were failing. I've updated a few of them to be more clear.

## Have you reviewed and updated the documentation for this change? Is new documentation required?

No new documentation is required.

## How did you verify this change?

I've added new test coverage & existing tests pass.

## Is this change a patch?

No.

PR #4785, opened 2022-06-30
Add soft delete to entities & namespaces in pgsql

amdprophet:

This adds support for soft deletion which I had missed when implementing #4767 & #4777. I've also refactored some of the tests to ensure databases are cleaned up after tests & better error output when a StoreV2 method returns an error.


c-kruse (approved):
Looking great! Thanks for revisiting.

Definitely enough to get started, but createIfNotExistsEntityConfigQuery looks like it's probably a little _off_ unless we assume we'll have some near instant cleanup/hard delete followup.

amdprophet:
@c-kruse thanks for the review. We should discuss how to tackle the create if not exists functionality this week.

calebhailey:
Related: #4609

PR #4807, opened 2022-07-21
Make entity config & state available through same RBAC rule

jamesdphillips:
TODO

### DIFF
--- a/api/core/v3/entity_config.go
+++ b/api/core/v3/entity_config.go
@@ -1,0 +1,11 @@
+package v3
+
+import (
+	corev2 "github.com/sensu/sensu-go/api/core/v2"
+)
+
+var entityConfigRBACName = (&corev2.Entity{}).RBACName()
+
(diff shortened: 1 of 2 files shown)


api/core/v3/resource_generated_test.tmpl
```diff
@@ -189,15 +189,15 @@ func TestResourceUniqueness(t *testing.T) {
...
+			t.Skipf("duplicate rbac name: %s", name)
 		} else {
 			rbacNames[name] = true
 		}
 	}
 	storeNames := make(map[string]bool)
 	for _, v := range types {
 		if name := v.StoreName(); storeNames[name] {
-			t.Errorf("duplicate store suffix: %s", name)
+			t.Skipf("duplicate store suffix: %s", name)
```
echlebek, line 200: Let's just kill this test, since it is no longer valid under this scheme

PR #4808, opened 2022-07-21
[GraphQL] List entity config & state resources

jamesdphillips:
* Expose `EntityConfig` & `EntityState` resources through GraphQL service
* **Complication:** The core/v3 package previous checked that each resource in the package had a unique `RBACName()`, however, since we don't want distinct names for the config & state resources this proved to be problematic. As such I've disabled the uniqueness tests for the time being. Same as: https://github.com/sensu/sensu-go/pull/4807
* Add `Fields()` method to `EntityConfig` & `EntityState` resources so that they implement the `Fielder` interface; allows users to filter with field selectors. 
* I apologize in the advance for the size of the PR but as is often the case >= 70% is generated code.


jamesdphillips:
DCO bot woke up and chose violence,
must mutilate history to reconcile

api/core/v3/resource_generated_test.go
```diff
@@ -487,15 +487,15 @@ func TestResourceUniqueness(t *testing.T) {
 	rbacNames := make(map[string]bool)
 	for _, v := range types {
 		if name := v.RBACName(); rbacNames[name] {
-			t.Errorf("duplicate rbac name: %s", name)
+			t.Skipf("duplicate rbac name: %s", name)
```
jamesdphillips, line 490: EntityConfig & EntityState share the same RBAC name and as such make this test problematic.

jamesdphillips:
Failing check is related to [this change](https://github.com/sensu/sensu-go/blob/develop/6/go.mod#L86), however I believe this is a valid use case for `replace`?

echlebek (approved):
Great work! Should this be ported forward to main as well?

PR #4810, opened 2022-07-22
Backport watcher

echlebek:
## What is this change?

We need the store/v2 watcher for some upcoming 6.8 features.


c-kruse (approved):
Looks to be mainly a copy of what was developed in 7.x (without Context broken out of ResourceRequest), yes?

Interested to see what you do with the enterprise postgres store. Maybe it's okay to just have a stub implementation that returns an error for now? 

👍

echlebek:
Yes, it's basically a copy of 7.x stuff. And yes, the postgres storev2 will not support the functionality I'm thinking.

Issue #4812, opened 2022-07-25
Backend ID Getter Retry "etcdserver: too many requests"


c-kruse:
Digging into an enterprise stability issue https://github.com/sensu/sensu-enterprise-go/issues/2373 I've noticed a disproportional number of backend restarts coming from the backend id getter (~40% of the time) when the etcd cluster is under unsustainable load. These are almost always due to an "etcdserver: too many requests" error. We should use the `kvc.RetryRequest` utility function to retry when this error pops up.

## Expected Behavior
The BackendIDGetter daemon should retry requests when etcd responds with the "etcdserver: too many requests" status.

## Actual Behavior
There is effectively no retry logic included in this section (all results either return a non-nil error indicating a retry-able error, or true indicating success)

PR #4815, opened 2022-07-26
Retry Etcd Requests in BackendIDGetter

c-kruse:
Closes https://github.com/sensu/sensu-go/issues/4812

## What is this change?

It looks like we intend to have retry logic in the BackendIDGetter's getLease function, but in practice it currently never retries. 

In a large environment we noticed that when the etcd cluster is overwhelmed, this tends to be a weak spot for sensu-go, triggering a restart due to an "etcdserver: too many requests" error returned from an etcd client. We intentionally include retry logic for this case in most of our store access logic, and should be consistent here.

Yes

N

N. Although I've found our retry util very difficult to reason about and easy to use incorrectly, as it appears was the case here.

N/A

Unit tests

## Is this change a patch?

N

### DIFF
--- a/CHANGELOG-6.md
+++ b/CHANGELOG-6.md
@@ -15,6 +15,10 @@ be marked as global resources.
 ### Changed
 - Upgraded CI Go version to 1.17.12
 
+### Fixed
+- Fixed a bug where sensu-backend could crash if the BackendIDGetter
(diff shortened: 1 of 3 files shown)


echlebek (approved):
:+1:

Issue #4813, opened 2022-07-26
Fix Keepalive Event Prioritization


c-kruse:
In order to reduce the chance of a keepalive event creation storms, we currently intend to prioritize keepalive events from agents over other events in eventd. Unfortunately this feature was never completed.

## Expected Behavior
Agent keepalives are handled ahead of regular check events.

## Current Behavior
Agent keepalives are mixed in to the same message topic with check events and are not prioritized.

c-kruse:
Related to: https://github.com/sensu/sensu-go/issues/4292 https://github.com/sensu/sensu-go/issues/4273

PR #4816, opened 2022-07-27
Route Keepalive Events to Eventd with High Priority

c-kruse:
Closes https://github.com/sensu/sensu-go/issues/4813

## What is this change?

Changes the flow of keepalive events from keepalived to eventd. Currently keepalive and other events are mixed together in the same topic going into eventd, so are handled at the same priority. This change uses the (until now) unused wizard bus topic "sensu:keepalive-raw" as a higher-priority topic between keepalived and eventd, allowing eventd to process time sensitive keepalive events ahead of regular events.

Keepalive events are time sensitive. Under high event load today, keepalives can fall behind and cause a keepalive storm where the events are being generated faster than they can be processed. In order to mitigate this, we should prioritize the handling of keepalive events in eventd.

Yes

I could use a second set of eyes on the `// TODO`. Near as I can tell these events have been dropped for the past few minor releases. I'm unsure of the consequences of beginning to handle them again.

Performance testing was... interesting. I'd like to revisit with someone more familiar with the perf test suite, but I observed a failure mode in the etcd only mode that appeared more severe than past releases.

N/A

Integration tests, QA Crucible tests.

## Is this change a patch?

N

### DIFF
--- a/backend/agentd/session.go
+++ b/backend/agentd/session.go
@@ -650,11 +650,6 @@ func (s *Session) handleEvent(_ context.Context, payload []byte) error {
 		} else {
 			eventBytesSummary.WithLabelValues(metrics.EventTypeLabelCheck).Observe(float64(len(payload)))
 		}
-		// TODO (ck) Should these be going through keepalived (TopicKeepalive)
-		// or straight to eventd (TopicKeepaliveRaw)?
-		if event.Check.Name == corev2.KeepaliveCheckName {
-			return s.bus.Publish(messaging.TopicKeepaliveRaw, event)
-		}
 	} else if event.HasMetrics() {
 		eventBytesSummary.WithLabelValues(metrics.EventTypeLabelMetrics).Observe(float64(len(payload)))
 	}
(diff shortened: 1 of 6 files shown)


backend/agentd/session.go
```diff
@@ -650,6 +650,8 @@ func (s *Session) handleEvent(_ context.Context, payload []byte) error {
 		} else {
 			eventBytesSummary.WithLabelValues(metrics.EventTypeLabelCheck).Observe(float64(len(payload)))
 		}
+		// TODO (ck) Should these be going through keepalived (TopicKeepalive)
+		// or straight to eventd (TopicKeepaliveRaw)?
```
echlebek, line 654: This should stay as it is. Let's amend the comment to mention that it's a compatibility fallback.

echlebek (approved):

backend/agentd/session.go
c-kruse: Spent some time looking at the history here and decided to treat these like regular events as they were prior to 6.4.

amdprophet (approved):

PR #4819, opened 2022-07-27
PG Config Store V2 Watches

fguimond:
Move the watcher out of the storev2 to make it reusable in the config store.

### DIFF
--- a/backend/store/postgres/watcher.go
+++ b/backend/store/postgres/watcher.go
@@ -2,75 +2,26 @@ package postgres
 
 import (
 	"context"
-	"database/sql"
 	"fmt"
-	"sync"
 	"time"
 
-	"github.com/jackc/pgx/v4/pgxpool"
 	v3 "github.com/sensu/sensu-go/api/core/v3"
 	"github.com/sensu/sensu-go/backend/poll"
 	storev2 "github.com/sensu/sensu-go/backend/store/v2"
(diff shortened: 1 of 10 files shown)


backend/store/postgres/watcher.go
```diff
@@ -49,3 +63,139 @@ func (rs recordStatus) Row(id string, resource storev2.Wrapper) poll.Row {
...
+	if !ok {
+		if isConfig {
+			tableFactory = watchConfigStoreOverride
+		} else {
+			tableFactory = func(storev2.ResourceRequest, *pgxpool.Pool) (poll.Table, error) {
+				return nil, fmt.Errorf("default watcher not yet implemented")
+			}
+		}
+	}
+	table, err := tableFactory(req, w.store.GetPgxPool())
```
c-kruse, line 104: Now that we have separated storev2.Interface implementations into ConfigStore and (roughly) a state store do we still need this plumbing?

Doesn't need to happen now, but perhaps the WatchedStore interface could be expanded to replace most of this?

c-kruse (approved):
👍  Thank you for cleaning up a bit of my mess there!

Looks good.

fguimond:
> 👍 Thank you for cleaning up a bit of my mess there!
> 
> Looks good.

@c-kruse thank you for reviewing, hopefully it aligns with what you had in mind 👍

PR #4839, opened 2022-08-10
Migrate authentication machinery to store/v2

echlebek:
Removes store authentication dependencies and instead offers some new APIs built on top of store/v2.

### DIFF
--- a/backend/authentication/providers/basic/basic_test.go
+++ b/backend/authentication/providers/basic/basic_test.go
@@ -1,0 +1,205 @@
+package basic
+
+import (
+	"context"
+	"testing"
+
+	corev2 "github.com/sensu/sensu-go/api/core/v2"
+	"github.com/sensu/sensu-go/backend/authentication/bcrypt"
+	"github.com/sensu/sensu-go/backend/store"
+	storev2 "github.com/sensu/sensu-go/backend/store/v2"
+	"github.com/sensu/sensu-go/testing/mockstore"
+	"github.com/stretchr/testify/mock"
+)
+
+func TestProviderAuthenticate(t *testing.T) {
+	tests := []struct {
+		Name     string
+		Store    storev2.Interface
+		Username string
+		Password string
+		Hash     string
+		Error    bool
+	}{
+		{
+			Name:     "username not supplied",
+			Username: "",
+			Password: "hello",
+			Error:    true,
+		},
+		{
+			Name:     "password not supplied",
+			Username: "Hello",
+			Password: "",
+			Error:    true,
+		},
+		{
+			Name: "user not found",
+			Store: func() storev2.Interface {
+				s := new(mockstore.V2MockStore)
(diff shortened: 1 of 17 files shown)


backend/authentication/providers/basic/basic_test.go
```diff
@@ -0,0 +1,205 @@
...
+					PasswordHash: pw,
+				}
+				s.On("Get", mock.Anything, mock.Anything).Return(mockstore.Wrapper[*corev2.User]{Value: user}, nil)
+				return s
+			}(),
+			Username: "eric",
+			Password: "password",
+			Error:    false,
+		},
+	}
```
ccressent, line 100: Should you add a test to cover that branch for the legacy case where, if `User.PasswordHash == ""` we use `User.Password` instead?

ccressent (commented):
Looks good to me overall, but I'm not sure I understand what `JWT struct` and its `GetSecret()` is for, or what the new SymmetricKey corev3 type is for?

backend/authentication/providers/basic/basic_test.go
echlebek: That's a good question. I thought about the legacy case while I was working on this stuff. I concluded that we should probably drop core/v2.User and add core/v3.User instead to drop support for the legacy case. But I couldn't do that in this PR... I got stuck elsewhere in the system while testing.

What are your thoughts on preserving the legacy case? IMO it is harmful.

echlebek:
SymmetricKey is a store-compatible way to store the symmetric encryption key that we generate for JWTs. Previously it was stored directly into etcd.

JWT is an API client that comprises the business logic that previously existed in the store for creating JWT secrets. I expect that it will grow in functionality in a subsequent effort... this is somewhat partial work due to having to switch gears to ConfigStore.

ccressent:
OK so it's not really used yet but will be soon?

backend/authentication/providers/basic/basic_test.go
ccressent: I think moving from `core/v2.User` to `core/v3.User` is indeed a separate task; let's keep it separate in the interest of not having the scope of this refactor explode.

My only thought right now about that legacy `Password` field in that removing it and support for it sounds good from a technical standpoint, but I would first check what that means for users and operators, especially those of have a db that has been migrated over and over.

It could be as simple as adding a migration, but I'm not 100% sure.

For now, I think the ideal course of action is to add a quick test case for it, for the sake of coverage, and it will fail to build and get removed when we do that v2 -> v3 work?

ccressent (approved):

backend/authentication/providers/basic/basic_test.go
echlebek: There won't be any migration from 6 to 7, it'll be greenfield databases only.

echlebek:
Yep! It's going to be used instead of the custom signing key stored in etcd for JWT issuance.

PR #4841, opened 2022-08-11
Pull in fixes from develop/6 branch

jamesdphillips:


amdprophet (approved):
I don't think replace directives will work when using `go install` but I'm not aware of any other reasonable options. Looks good.

PR #4854, opened 2022-08-31
Include Backend Name with Agent WS Connection

c-kruse:
Closes https://github.com/sensu/sensu-enterprise-go/issues/2270

## What is this change?

Adds a backend-id header to the agent ws upgrade response and log it on the agent side.

It can be difficult to debug issues that appear to only affect specific backends. Logging more backend connection info on the agent side may assist users in identifying problem backends.

Y

N

N

N/A

## Is this change a patch?

N


fguimond (approved):
LGTM 👍

Issue #4847, opened 2022-08-19
Check Subdues are unreliable while crossing repeat intervals.


c-kruse:
Check subdues are currently implemented as an open interval (i.e. begin < now < end). This is a problem when the operator specifies overlapping repeating intervals. For example: "Subdue between 4pm on Friday and 8am Monday" may be configured like this:
```
subdues:
- begin: "2022-08-19T16:00:00Z"
   end: "2022-08-19T23:59:59Z"
  repeat:
    - fridays
- begin: "2022-08-19T00:00:00Z"
  end: "2022-08-19T23:59:59Z"
  repeat:
    - saturdays
- begin: "2022-08-19T00:00:00Z"
  end: "2022-08-19T23:59:59Z"
  repeat:
    - sundays
- begin: "2022-08-19T00:00:00Z"
  end: "2022-08-19T08:00:00Z"
  repeat:
    - mondays
```

Since these are implemented as closed intervals (not to mention the extra precision that may make a date truly be between 23:59:59 and 00:00:00 the following day), this strategy leads to a small window where checks are still scheduled each night around midnight.

## Expected Behavior
TBD

asachs01:
Hi @c-kruse I wanted to followup and see if you'd made any progress on this and if we have an idea as to when a fix might be released.

c-kruse:
@asachs01  I hadn't made any progress, but it should be a fairly quick and contained fix. Adding to the next patch release.

c-kruse:
@asachs01  I'm still planning on making adjustments to the implementation as it is indeed currently misleading. Read the code even more carefully and I've got an immediate workaround.

Instead of ending ranges at 23:59:59 in that day, users can specify an end time at 00:00:01 the next day and their sporadic missed subdues should go away.

i.e. this works without misses for 4PM friday until noon saturday 
```
subdues:
- begin: "2022-08-19T16:00:00Z"
   end: "2022-08-20T00:00:01Z"
  repeat:
    - fridays
- begin: "2022-08-19T00:00:00Z"
  end: "2022-08-19T12:00:00Z"
  repeat:
    - saturdays
```

PR #4859, opened 2022-09-08
Update Check Subdues to allow for overlapping intervals

c-kruse:

## What is this change?

Closes https://github.com/sensu/sensu-go/issues/4847

Changes the evaluation of check subdues to be inclusive of the begin time and up to (but excluding) one second **past** the end time.

With this change subdues are a little more intuitive to use. The example straight from our docs to subdue a check all weekend currently has a second long gap (Sat night between 23:59:59 and Sun morning 00:00:00.0001.) By making the begin time inclusive and the end time + 1 second exclusive there is no longer a potential gap.
```
  - begin: "2022-04-23T00:00:00-07:00"
    end: "2022-04-23T23:59:59-07:00"
    repeat:
    - weekends
```

Alternately, users can bypass this issue today by setting the end time 2 seconds later (in this case overlapping into the next day) to avoid any potential missed subdues.

Y

The choice to subdue for 1 second past the end time may be a little contentious/unintuitive. We could drop this and make it a more strict [start, end) interval, and update our documentation to remove any ambiguity around subdues intervals. I think it might just get a little wordy.

N

I don't think so. I think this change brings us in line with the docs.

Unit tests

## Is this change a patch?

Y

### DIFF
--- a/CHANGELOG-6.md
+++ b/CHANGELOG-6.md
@@ -5,6 +5,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic
 Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Changed
+- Check subdues are now evaluated on a half open interval such that there
+should be no unintended gaps between overlapping subdues.
+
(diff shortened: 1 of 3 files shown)


fguimond (approved):
👍 Thanks Christian!

PR #4862, opened 2022-09-12
Redact entity configuration metadata

jamesdphillips:
## What is this change?

Redacts entity config resource's metadata using configured list.

Redacts entity config when fetching entity config reources.

* Unfortunate but I pulled `FoundInArray` into core/v3 to avoid cyclical dependency 
* I felt this was best handled in the API package, hopefully this avoids any unintentional regressions in the future.
* Because the base module's dependencies weren't updated, the API package tests broke after the recently entity config RBAC change (#4808)

Unit tests & sanity check.


jamesdphillips:
Addresses https://github.com/sensu/sensu-enterprise-go/issues/2484

amdprophet (approved):

echlebek (approved):
Looks good! Can you please also apply to main?

PR #4866, opened 2022-09-19
Change sensu-backend init to be postgres-only

echlebek:
... and lots of associated changes!

This is another commit towards removing support for etcd and requiring postgres for all functionality. Currently sensu-backend is still broken, but it's closer after this commit than before it to working.


c-kruse (approved):
👍  didn't read through everything, but init is init-y. migrations, ClusterConfig and SymmetricKey all ✅

PR #4869, opened 2022-09-21
Temporarily Fix the api/core/v3 dependency

c-kruse:
## What is this change?

We've landed ourselves in a state. develop/6 currently does not build since the api/core/v3 dependency is pointing at a squashed commit.

This updates the dep to a commit that is not likely to ever be squished (the current develop/6 HEAD), but I believe we should follow up with merging the core/v3 changes into main and publishing a new v3.x tag for it to point at before a release.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -48,7 +48,7 @@ require (
 	github.com/robfig/cron/v3 v3.0.1
 	github.com/sensu/lasr v1.2.1
 	github.com/sensu/sensu-go/api/core/v2 v2.14.0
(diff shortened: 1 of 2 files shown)


echlebek:
Should the fix be ported to the core/v3.7.x series?

jamesdphillips (approved):
Sorry, I know I said I'd do this but got busy with other things... Thank you for getting er' done.

jamesdphillips:
> Should the fix be ported to the core/v3.7.x series?

I created https://github.com/sensu/sensu-go/issues/4871 so we don't forgot.

PR #4884, opened 2022-10-03
Auto-restart agent service on Windows

amdprophet:
## What is this change?

Updates the Windows service for Sensu Agent to auto-restart every 5 seconds after a failure.

Closes https://github.com/sensu/sensu-enterprise-go/issues/1575.

Yes.

I couldn't find a way of triggering a failure state in Sensu Agent to fully test this.

We may want to describe this behaviour in the docs; what do you think @hillaryfraley?

I can see the failure cases set in the service after this change.

Screen Shot 2022-10-03 at 15 51 06

## Is this change a patch?

No.

### DIFF
--- a/CHANGELOG-6.md
+++ b/CHANGELOG-6.md
@@ -7,17 +7,18 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
-
 ### Added
 - The authentication module now logs successful (INFO) and unsuccessful (ERROR)
   login attempts.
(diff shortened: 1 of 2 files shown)


c-kruse (approved):

hillaryfraley:
I created a docs PR to add this info. Does the agent service automatically restart on other platforms? We don't have that documented so I'm trying to figure out if agent service auto-restart will be specific to Windows or if this behavior is already happening for other platforms.

PR #4887, opened 2022-10-07
Add Operator Presence Control system

### LINKED ISSUE #4867: Create the operator presence control (opc) state schema
Operator Presence Control is an alternative term for a dead man's switch.

An operator in an OPC system must regularly inform the system of its operating condition, or the OPC system will mark it Absent. Once the operator has been marked absent, the OPC system will regularly issue reminders that the operator is absent until the operator checks in.

An operator can be any entity or condition. For instance, an agent entity can use the OPC system to implement the keepalives feature. The check scheduler can use the OPC system to implement check TTL. The round robin scheduler can use the OPC to detect when an agent is no longer available for scheduling.

To implement the OPC system, we'll create an OPC state table in postgresql.

An OPC state table is a relationship between an operator, its current operating condition, and the time at which the current operating condition was last updated. The operator can be identified by a namespace, operator type, and operator name.

Sensu schedulers can poll the OPC state table, seeking entries that are absent, or have been absent long enough that issuing a reminder is necessary.

The OPC table has the following layout, with `namespace` being a foreign key to the `namespaces` table.

| id BIGSERIAL | namespace BIGINT | operator_type INTEGER | operator_name TEXT | present BOOLEAN | last_update TIMESTAMPTZ | check_in_timeout INTERVAL | metadata JSONB |
|--------|----|----|---|---|---|---|---|

Create a migration that adds the table, and write a Go API to manipulate it. The Go API should have the following definitions:

```
// OperatorType is used to distinguish operators of different types.
// For instance, it's allowed to have two operators with the same name within
// a namespace, provided they are of a different type. For example, a namespace
// could contain a keepalive operator called "ketchup" and also a check TTL
// operator called "ketchup".
type OperatorType int

// OperatorKey holds the key fields of an operator, for identifying a unique
// operator by namespace, type and name.
type OperatorKey struct {
	// Namespace is the namespace of the operator.
	Namespace string

	// Type is the type of the operator.
	Type OperatorType

	// Name is the name of the operator.
	Name string
}

// OperatorState holds state information about an operator, including a
// reference to its controller.
type OperatorState struct {
	// Namespace is the namespace of the operator.
	Namespace string

	// Type is the type of the operator.
	Type OperatorType

	// Name is the name of the operator.
	Name string

	// Controller is an operator that is responsible for this operator. It is
	// not guaranteed to be non-nil.
	Controller *OperatorKey

	// CheckInTimeout is the amount of time that can pass before the operator
	// would be considered absent.
	CheckInTimeout time.Duration

	// Present indicates whether or not the operator is currently present and
	// accounted for.
	Present bool

	// LastUpdate indicates when the operator last checked in, or when the OPC
	// system noted it was still absent.
	LastUpdate time.Time

	// Metadata is the operator's metadata. It can be any arbitrary JSON data
	// structure, so it is kept as an encoded json.RawMessage.
	Metadata *json.RawMessage
}

const (
	// AgentOperator is the operator type for agent entities.
	AgentOperator OperatorType = 0

	// BackendOperator is the operator type for backend entities.
	BackendOperator OperatorType = 1

	// CheckOperator is the operator type for check TTLs functionality.
	CheckOperator OperatorType = 2
)

// MonitorOperatorsRequest is a request to watch a specific operator space
// for updates, whether those updates are generated as repeated absence alerts
// or check ins. The OPC will send all error encountered for each poll to the
// supplied error handler, if it is non-nil.
//
// In a clustered execution setting, it's often useful to use a distinct
// controller for each node so that nodes do not conflict in their monitoring,
// which involves resetting the LastUpdate field whenever the current time
// minus the time last seen is greater than the CheckInTimeout.
type MonitorOperatorsRequest struct {
	// Type is the OperatorType.
	Type OperatorType

	// Namespace is the namespace to monitor. It is optional.
	Namespace string

	// Name is the name to monitor. It is optional.
	Name string

	// ControllerType is the operator type of the controller.
	ControllerType OperatorType

	// ControllerNamespace is the namespace of the controller.
	ControllerNamespace string

	// ControllerName is the name of the controller.
	ControllerName string

	// Every is the operator database polling interval.
	Every time.Duration

	// ErrorHandler is a function tha handles any errors generated in the
	// monitoring process.
	ErrorHandler func(error)
}

// OperatorMonitor monitors operators for updates or missed check-ins.
type OperatorMonitor interface {
	// MonitorOperators continuously watches all operators in the operator
	// space for updates.
	MonitorOperators(context.Context, MonitorOperatorsRequest) <-chan []OperatorState
}

// OperatorConcierge is responsible for checking operators in and out.
type OperatorConcierge interface {
	// CheckIn checks in an operator. It can be for the first time or a
	// scheduled check in.
	CheckIn(context.Context, OperatorState) error

	// CheckOut checks out an operator. The OperatorMonitor will no longer
	// notice absences for the operator, until it checks in again.
	CheckOut(context.Context, OperatorKey) error
}
```

In case of disagreement between the `present` boolean field and the `LastUpdate` and `CheckInTimeout` fields, the `present` boolean field will be taken to be correct, the assumption being that the process responsible for updating the `present` field is laggy. However, warnings should be logged when this condition is encountered.

This system can be used to implement the following features of Sensu:

* Entity keepalives
* Check TTL
* Round-robin scheduling liveness
* Silences

Each particular system should use its own operator type and can choose its own metadata format.

echlebek:
## What is this change?

This commit adds a new persistent state manager to sensu-go called Operator Presence Control. Operator Presence Control or OPC is an accounting system for operator presence. An operator can be any entity, system or condition that has a notion of presence. For instance, agents can use the OPC system to implement keepalives, and the check scheduler can use the OPC system to implement check TTL.

When an operated is checked in, a row in the OPC SQL table is inserted or updated with the operator's details, the current time, and the duration that can elapse without further check-ins before the operator is considered to be absent.

Operators can be assigned controllers. Controllers are also operators, and are useful for sharding monitoring work. For instance, agent operators can be assigned backend operators as controllers. Each backend can then concern itself only with the agent operators that are assigned to it.

Controllers can become absent too. When that happens, the monitoring system will reassign operators missing a controller to the other controllers of the same type in the system. The operators will be equally balanced among all of the controllers.

Operators can also assign themselves a different controller by checking in with a different controller in their check-in request.

The OPC system differs from the existing notification systems in sensu-go. Notably, the system depends on postgresql instead of etcd, and uses polling instead of asynchronous notification. This should permit a much broader use of these types of notifications, since all notification events are handled in a batch within the same goroutine. We expect keepalive storms and check TTL storms to be much less damaging to system stability with this design.

Closes #4867 

No, this is an implementation detail only.


c-kruse (approved):
LGTM!

PR #4904, opened 2022-10-25
7x Migrate to core api

c-kruse:
## What is this change?

Migrates sensu-go to the new https://github.com/sensu/core and https://github.com/sensu/sensu-api-tools modules.

Depends on the following new/unreleased modules:
https://github.com/sensu/sensu-api-tools
https://github.com/sensu/core/tree/main/v2
https://github.com/sensu/core/tree/main/v3

Also depends on #4903 for the types module without the harmful circular dependency.


ccressent (approved):
Looks good ; it seems to me that if there are any issues, they would have to be addressed in `sensu-api-tools`.

echlebek (approved):

PR #4915, opened 2022-11-23
Refactor sensu-go to not depend on etcd

echlebek:
## What is this change?

This is a very large change, in terms of code volume and in feature set, and it sets the stage for the upcoming sensu-go 7.x series of releases. Etcd has been completely removed and sensu-go no longer makes use of it. Instead, postgresql 9.6 or greater is required. (Newer postgresql recommended)

On the heels of @c-kruse's recent work, the api modules have also been removed.

The configuration parameters and commands that relate to etcd maintenance and querying are gone. The mechanism for configuring postgresql is now a configuration file instead of a Sensu resource.

Replacements have been built out for almost all facilities that previously depending on etcd, but there are some notable omissions, including the queue library, which enabled adhoc scheduling. Adhoc scheduling is currently not available but will be added at a later time.

After years of supporting a product that depends on etcd, we've come to realize that many potential users have avoided sensu because they don't want to operate etcd at scale. Operating etcd is challenging and requires expensive hardware, and there are very few hosted and cloud options available. By contrast, managed postgresql is available on all major cloud providers and most smaller hosting providers as well.

It's been a big undertaking to convert sensu-go into a product that uses postgresql instead of etcd, especially because etcd was directly embedded into the product itself. In 2017, we made an all-in technical decision and had to live with the outcomes, positive and negative. Fast-forward to 2022, etcd has not seen widespread adoption by cloud providers, and sensu-go's write heavy workload turned out to be a poor fit for etcd.

Yes, I have made a note in the changelog.

We changed our minds a few times along the way about how we'd support postgresql, and this slowed down the time to implement this work. Additionally, we wanted to refactor large portions of sensu-go so that newer and better APIs could be used. As a result this work touches every component in sensu-backend.

We'll need to reformulate some of our documentation now that etcd is gone and postgresql is required.

The software is still broken at this point, even though most of the tests are passing. Currently there are bugs that are preventing sensu-go from running properly. Still, this change gets us to the point where multiple people can work on these bugs and get sensu-go back to a point where it can do typical check scheduling and event handling.


echlebek:
It's ready :grin:

c-kruse (approved):
🚢

PR #4924, opened 2022-12-07
Rename func 'NewGenericStore' to func 'Of'

echlebek:
This change renames NewGenericStore to Of. This results in nicer looking function calls, since the generic parameter confers most of the meaning in the function call. That's it, that's the change.


c-kruse (approved):
good gopls.

PR #4928, opened 2022-12-14
Do not implement core/v3 wrapper behaviour for LicenseFile

echlebek:
We need LicenseFile to render in a very particular way, and if labels and annotations are set non-nil, rendering round tripping will break existing licenses.


c-kruse (approved):
😬

echlebek:
It turns out we can avoid doing this.

PR #4931, opened 2022-12-19
Generic Store Watch

c-kruse:
## What is this change?

Adds a Watch function to the generic store. Also fixes some watch poller bugs.

This turned out to be necessary for resource replicators in the enterprise product.


echlebek (approved):
:shipit:

PR #4936, opened 2023-01-11
Use core/v3.Resource in the CLI

echlebek:
## What is this change?

Replaces all uses of corev2.Resource with corev3.Resource in the CLI. This avoids compatibility problems with the updated backend.


ccressent (approved):
Awesome, thanks for that!

PR #4939, opened 2023-01-13
Disable selectors when listing configuration

echlebek:
## What is this change?

The configuration store attempts to query selectors, but the table doesn't have a selectors column yet, so it doesn't work. I've disabled this for now by commenting out the relevant parts of code. We'll need an effort to add the selectors column shortly.


c-kruse (changes requested):
Nah. It was pretty close.
https://github.com/sensu/sensu-go/pull/4940

Issue #4787, opened 2022-07-04
Add --strip-networks flag to sensu-agent


echlebek:
We'd like to add a way to disable sending network information from sensu-agent to sensu-backend. The network information sent by sensu-agent can be quite lengthy, and in high volume monitoring environments, where each agent has many virtual network addresses, this can become a performance bottleneck.

Add a new boolean configuration option, `--strip-networks`, to sensu-agent, which works via the CLI, environment variables, and configuration file. When the option is enabled, sensu-agent should avoid collecting information for the `system.networks` property in the agent entity state.

ManishaKumari295:
The ticket is already worked on - https://github.com/sensu/sensu-go/pull/4941 . Hence this ticket can be closed i guess.

PR #4941, opened 2023-01-17
Strip network agent entity state

agm650:
## What is this change?

Disable sending network information from sensu-agent to sensu-backend when a flag is passed to the agent

Solving #4787

Yes

No

No, I think it was easy to implement.
By default it will still add the information, the only behaviour change will be when the flag will be set. 

Documentation has been updated and [a pull request](https://github.com/sensu/sensu-docs/pull/4162) was created

I run the automated test.
In addition to this I build the sensu-agent binary for x86-64 linux & arm7 linux. I tested them against and up to date sensu-go-backend (6.9.1).
I performed 3 different tests (between each tests agents were removed from the backend):

- Start the agent without the flag to observe that network information were still retrieved (and visible in the sensu-go webui)
- Start the agent with the flag on the command line to observe that network information were removed from the webUI
- Start the agent with the environment variable  to observe that network information were removed from the webUI

## Is this change a patch?

I think it can considered to be a patch since its small. If not I will change the target branch. let me know.

### DIFF
--- a/CHANGELOG-6.md
+++ b/CHANGELOG-6.md
@@ -5,6 +5,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic
 Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unrealased]
+
+### Added
+- Adding a flag at agent level to avoid collecting system.networks property in the agent entity state
+
 ## [6.9.1] - 2022-12-01
 
 ### Changed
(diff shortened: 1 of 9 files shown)


echlebek (changes requested):
Thanks for the pull request, I think this is great work! I would like to request some modifications, so that it is more testable.

1. system.Info should be modified to take a boolean parameter that governs whether or not network information is collected. For example: `func Info(includeNetworks bool) (types.System, error)`
2. The logic that was added in agent/agent.go should be represented in system.Info instead.
3. The test for system.Info should be written for both conditions to show that the parameter is working as intended.

agm650:
@echlebek : I made the requested modifications. Let me know if its ok.

echlebek (approved):
Thank you! :+1: 

LGTM.

PR #4944, opened 2023-01-23
Update CONTRIBUTING.md for Core Migration

c-kruse:
Now that sensu has migrated away from the github.com/sensu/sensu-go/core/* modules, update the CONTRIBUTING.md to reflect that.

I don't think that we really need to say anything about developing multiple modules at once. Now that there are no longer multiple modules in development in the sensu-go repo, it should work just like any other go project.


gd8 (approved):
This makes sense — thanks!

PR #4945, opened 2023-01-23
Add a Postgres-backed Queue implementation

c-kruse:
Replaces the types.Queue interface in the backend/queue package with minor cosmetic changes. Adds a queue implementation based on row-level locks on a new postgres table, queue_items. Needed to re-enable ad hoc scheduling.

### DIFF
--- a/backend/queue/queue.go
+++ b/backend/queue/queue.go
@@ -2,7 +2,8 @@ package queue
 
 import (
 	"context"
-	"errors"
+	"fmt"
+	"path"
 	"time"
 
 	"github.com/sensu/sensu-go/backend/store"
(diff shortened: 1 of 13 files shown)


backend/queue/queue.go
```diff
@@ -20,34 +21,68 @@ var (
...
-	return i.value
+// ClusteredQueue is a Client wrapper meant to function as a simple pub/sub
+// across all active sensu backends in a cluster.
+// Given a queueName, ClusteredQueue Enqueues to queueName/{{backendID}} for
+// each backend, and Reserves only from queueName/{{this backendID}}
+type ClusteredQueue struct {
+	// Client underlying queue implementation
+	Client Client
+	// GetBackendID gets the current backend process's id.
+	// TODO(ck): just hostname per backend/opc.go?
```
echlebek, line 61: I think we should move to backend IDs being hostnames, or optionally named by config file. If backends share the same hostname, that's on them and sensu execution will be weird. This is what agents do, so no reason for backends not to.

backend/store/postgres/queue.go
```diff
@@ -0,0 +1,92 @@
...
+	return err
+}
+
+// Reserve reserves a queue item.
+// When the queue is empty Reserve will block and poll for new items until either
+// an item is found, or the context is cancelled.
+//
+// When Reserve returns a Reservation, the caller MUST Ack or Nack that Reservation.
+// Otherwise a transaction + connection could be leaked depending on the session
+// settings (i.e. idle_in_transaction_session_timeout.)
```
echlebek, line 36: This seems a bit dangerous if it's possible to reserve items faster than it is to deal with them. Thoughts?

c-kruse: It is indeed a bit dangerous @echlebek.

For ad hoc scheduling it's a problem we wouldn't likely run into, as the plan is to have a single goroutine reserving and processing these items.

I could see us mitigating it by either adding a Capacity field to the Queue with some synchronized logic to limit open reservations, or by replacing the Reserve interface with a Handle method that accepts a callback and can be backed by a worker pool with a fixed size.

echlebek (approved):

PR #4949, opened 2023-01-30
Update tessend to not use etcd

ccressent:
## What is this change?

This updates `tessend` to make it not use etcd at all.

7.x work to get rid of etcd

I think it's already covered under the entry that says that etcd is not used/supported anymore in 7.x.
So not adding an entry.

There is still a TODO related to figuring out the number of backends in the cluster. I'm not sure what is the current best way to go about retrieving or figuring out this information. I thought I might be able to leverage the new OPC work, but it's not clear I can after all.

Many pieces of `tessend` were hardcoded to read/write keys in etcd, and many other moving pieces were necessary to complete this work (new store counting and watching methods, ...).

No documentation update needed: the user facing interface that lets one enable/disable `tessend` is not changing. This change is purely about how `tessend` works _internally_.

Manual testing, the old tests need to be reintroduced/reworked based on their relevance.

## Is this change a patch?

No.

### DIFF
--- a/backend/tessend/data.go
+++ b/backend/tessend/data.go
@@ -1,0 +1,59 @@
+package tessend
+
+import (
+	corev2 "github.com/sensu/core/v2"
+)
+
+var (
+	// Distribution indicates that Sensu Go was distributed from OSS but can be
+	// overridden
+	Distribution = "oss"
+)
+
+// Data is the payload sent to tessen
+type Data struct {
+	// General information about the Sensu installation.
+	Cluster Cluster `json:"cluster"`
+
+	// Metric data.
+	Metrics corev2.Metrics `json:"metrics"`
+}
+
+// Cluster is the cluster information sent to tessen
+type Cluster struct {
+	// ID is the ID of the sensu-enterprise-go cluster.
+	ID string `json:"id"`
+
+	// Distribution indicates how Sensu Go was distributed
+	Distribution string `json:"distribution"`
+
+	// Version is the Sensu release version (e.g. "1.4.1").
+	Version string `json:"version"`
+
+	// License is the Cluster's license.
+	License interface{} `json:"license"`
+}
+
+// LicenseFile represents the content of a license file
+type LicenseFile struct {
+	// License contains the actual license
+	License interface{} `json:"license"`
+}
+
+// Wrapper wraps the LicenseFile for unmarshalling
+type Wrapper struct {
+	Value LicenseFile `json:"spec"`
+}
+
+// StoreConfig contains information about what services Sensu is using to store
+// information.
+type StoreConfig struct {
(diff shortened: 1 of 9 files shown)


backend/tessend/data.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// StoreConfig contains information about what services Sensu is using to store
+// information.
+type StoreConfig struct {
+	// ConfigStore indicates the storage backend used to store Sensu configuration.
+	ConfigStore string `json:"config_store"`
+
+	// StateStore indicates the storage backend used to store Sensu state.
+	StateStore string `json:"state_store"`
```
echlebek, line 55: We're no longer creating a distinction between these stores. Rather there is the config store, and then several specific stores for special types. These include events, entities, entity configs, entity states, and namespaces.

backend/tessend/tessend.go
```diff
@@ -0,0 +1,829 @@
...
+
+// getPerResourceMetrics populates the data payload with the total number of each resource.
+func (t *Tessend) getPerResourceMetrics(now int64, data *Data) error {
+	var backendCount float64
+
+	// collect backend count
+
+	// TODO: This needs to be replaced with something else, that maybe we don't
+	// even have yet. Eric mentioned using the Operator Presence Control data,
+	// for which we don't really have an abstraction yet.
```
echlebek, line 680: Christian added support for listing OPC items recently, so we can incorporate that now

backend/tessend/tessend.go
```diff
@@ -0,0 +1,829 @@
...
+	defer resourceMetricsMu.RUnlock()
+	ticker := time.NewTicker(t.duration)
+	defer ticker.Stop()
+	for metricName, resource := range resourceMetrics {
+		select {
+		case <-t.ctx.Done():
+			return t.ctx.Err()
+		case <-ticker.C:
+		}
+		count, err := t.store.GetConfigStore().Count(t.ctx, storev2.NewResourceRequestFromV2Resource(resource))
```
echlebek, line 718: This will be a bit of an issue for the "special" data types unfortunately, since they need to come from their respective store.

echlebek (changes requested):
It's looking good to me, the only issue is that some of the types are special now and don't come from the config store, but rather from other stores. So they need to be explicitly handled unfortunately.

c-kruse (commented):
Cluster Member count looks right to me!

The note Eric pointed out regarding the special types still looks outstanding, though. I think functioning counters Events and Entities probably two of most essential bits of tessen. Other than that, ✅ .

backend/tessend/data.go
echlebek: I see, OK

echlebek (approved):

PR #4953, opened 2023-02-09
Add store based synchronization.

c-kruse:
Adds a store-based Mutex and synchronized executor.

### Background Info on Advisory Locks

https://www.postgresql.org/docs/15/explicit-locking.html#ADVISORY-LOCKS

### DIFF
--- a/backend/store/postgres/sync.go
+++ b/backend/store/postgres/sync.go
@@ -1,0 +1,113 @@
+package postgres
+
+import (
+	"context"
+	"errors"
+	"sync"
+	"time"
+
+	"github.com/jackc/pgx/v5"
+	"github.com/sensu/sensu-go/backend/store"
+)
+
+// SynchronizedExecutor functions as a postgresql advisory lock based mutex.
+type SynchronizedExecutor struct {
+	DB DBI
+
+	CheckinInterval time.Duration
+}
+
+// Execute blocks until the Mutex can be locked, executes the handler function,
+// then unlocks the Mutex. Implemented with postgresql advisory locks.
+func (se *SynchronizedExecutor) Execute(ctx context.Context, mux store.Mutex, handler store.MutexHandler) error {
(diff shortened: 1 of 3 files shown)


backend/store/postgres/sync.go
```diff
@@ -0,0 +1,110 @@
...
+			select {
+			case <-ctx.Done():
+				return nil
+			case <-time.After(next.Sub(now)):
+			}
+		}
+		tx, err := se.tryLock(ctx, mux)
+		if err != nil {
+			logger.WithError(err).Errorf("error trying to lock mutex: %d", mux)
+			continue
```
echlebek, line 36: It seems like in this circumstance we're going to be pretty mean to the database potentially. `next` isn't getting bumped yet, and there is no backoff on creating database transactions that could fail over and over. Perhaps the code should bump next before trying the lock, or use backoff in the retry loop.

backend/store/postgres/sync.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+		if tx != nil {
+			defer func() {
+				if err := tx.Rollback(ctx); err != nil {
```
echlebek, line 40: I think this is valid, but I have experienced customer support issues with unconditional rollbacks, as I have also used them in the past. The reason is that rollbacks are things that postgresql operators look for, as signs of problems. If the system is malfunctioning for other reasons, it can be a false signal. Is it necessary to rollback here, or would commit have the same effect?

echlebek (commented):
I'm optimistic, just a couple comments/questions!

backend/store/postgres/sync.go
c-kruse: It's my understanding that the difference between rollback would be negligible.  Will change these to Commits.

echlebek (approved):
Let's gooooooo :dark_sunglasses:

PR #4954, opened 2023-02-16
Add in-memory event store

echlebek:
* Add in-memory event store
* Modify postgresql event store to not do in-place merging
* Modify postgresql queries to stop relying on last_ok and timestamp columns
* Add configuration for constraining postgresql max tps

This is still largely un-tested but should basically work.


backend/store/memory/event_store.go
```diff
@@ -0,0 +1,348 @@
...
+}
+
+func NewEventStore(config EventStoreConfig) *EventStore {
+	return &EventStore{
+		backingStore:      config.BackingStore,
+		flushInterval:     config.FlushInterval,
+		db:                newMemoryDB(),
+		silenceStore:      config.SilenceStore,
+		flushC:            make(chan struct{}),
+		eventWriteLimiter: rate.NewLimiter(config.EventWriteLimit, int(config.EventWriteLimit/10)),
```
c-kruse, line 129: Was curious to see golang.org/x/time/rate in action. Looks like it didn't get fully hooked up.

ccressent:
Wasn't there a design document you wrote about an in-memory event store? Or am I remembering incorrectly?

backend/store/memory/event_store.go
```diff
@@ -0,0 +1,366 @@
...
+		entry := value.(*eventEntry)
+		entry.Mu.Lock()
+		defer entry.Mu.Unlock()
+		if !entry.Dirty {
+			return true
+		}
+		eventBytes, err := snappy.Decode(nil, entry.EventBytes)
+		if err != nil {
+			// developer error, fatal
+			panic(err)
```
ccressent, line 92: Is it really a developer error?

Can't `snappy.Decode()` above fail even if your code were 100% correct, say because the memory holding `entry.EventBytes` got corrupted somehow (malicious or accidental memory manipulation by something else on the system, cosmic ray flipping a bit, ...)?

backend/store/memory/event_store.go
```diff
@@ -0,0 +1,366 @@
...
+		}
+		var event corev2.Event
+		if err := proto.Unmarshal(eventBytes, &event); err != nil {
+			// developer error, fatal
```
ccressent, line 96: ditto

backend/store/memory/event_store.go
```diff
@@ -0,0 +1,366 @@
...
+			panic(err)
+		}
+		if err := m.limiter.Wait(ctx); err != nil {
+			// context cancellation
+			return false
+		}
+		_, _, storeErr = s.UpdateEvent(ctx, &event)
+		if storeErr != nil {
+			return false
```
ccressent, line 105: I'm wondering if it's not worth having a way to still try to store the rest of the events even if one of them fails to be stored as we range over the map.

I know it makes tracking and returning errors more complex, but my main concern is what happens if we stop going over the map early as we do here: a single problematic event blocks all the other events "behind it in the map" from being persisted.

Would it not be better if that single problematic event didn't have such an impact on the other values to be persisted? At least the system would still mostly work rather than having to somehow do something with the problematic event so that the rest of the system keeps working.

backend/store/memory/event_store.go
```diff
@@ -0,0 +1,366 @@
...
+		if err != nil {
+			if _, ok := err.(*store.ErrNotFound); !ok {
+				return nil, nil, err
+			}
+		}
+	} else {
+		decompressed, err := snappy.Decode(nil, entry.EventBytes)
+		if err != nil {
+			// fatal developer error
+			panic(err)
```
ccressent, line 235: ditto

backend/store/memory/event_store.go
```diff
@@ -0,0 +1,366 @@
...
+		}
+		var ev corev2.Event
+		if err := proto.Unmarshal(decompressed, &ev); err != nil {
+			// fatal developer error
+			panic(err)
```
ccressent, line 240: ditto

backend/store/memory/event_store.go
```diff
@@ -0,0 +1,366 @@
...
+
+	// Mu is necessary in the rare instance when multiple goroutines are competing
+	// to store different versions of the same event. This is only necessary due to
+	// a design flaw in eventd, and could possibly be mitigated there.
+	Mu sync.Mutex
+}
+
+const eventOutputTruncatedBytesLabel = "sensu.io/output_truncated_bytes"
+
+type SilenceStoreI interface {
```
ccressent, line 44: I don't think this should to be exported? And I'm not clear why we can't use the interface defined in the `store` package already?

ccressent (commented):
Pardon the oversimplification, but to make sure I understand the idea correctly: this is in essence a write cache for events?

backend/cmd/start.go
```diff
@@ -72,7 +72,8 @@ const (
 	flagName                  = "name"
 
 	// Postgres store
-	flagPGDSN = "pg-dsn"
+	flagPGDSN    = "pg-dsn"     // postgresql connection string
+	flagPGMaxTPS = "pg-max-tps" // postgresql maximum transactions per second cap
```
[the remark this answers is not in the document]
echlebek: I'm not sure how that would be done on postgresql, for potentially ephemeral clients, let me look into it.

backend/store/store.go
```diff
@@ -659,3 +659,13 @@ type Initializer interface {
...
+
+type noMergeEventKey struct{}
+
+func NoMergeEventContext(ctx context.Context) context.Context {
+	return context.WithValue(ctx, noMergeEventKey{}, struct{}{})
+}
+
+func IsNoMergeEventContext(ctx context.Context) bool {
+	return ctx.Value(noMergeEventKey{}) != nil
+}
```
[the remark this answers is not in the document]
echlebek: I don't think it is tbh, this is a pretty common way to do it.

backend/store/postgres/dynamic_query.go
```diff
@@ -34,8 +41,15 @@ type getEventTmplData struct {
...
+var countEventsTmpl = template.Must(template.New("CountEvents").Parse(`
+WITH ns AS (
+	SELECT id FROM namespaces
+	WHERE name = $1
+)
+SELECT count(*)
 FROM events
+FULL OUTER JOIN ns
+ON events.namespace = ns.id
 WHERE
```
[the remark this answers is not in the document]
echlebek: Yeah, all this complexity is related to the various different ways to list events. With a namespace, without one, with paging, without paging, selectors...

backend/store/memory/event_store.go
echlebek: I suppose so, but that would necessitate a panic too.

backend/store/memory/event_store.go
echlebek: At this point the event is validated, so it's not likely to be a problematic individual event, but rather a systemic problem with the event store, like perhaps the connection has failed or something. In that case it's best not to keep trying and wait for the next write-out window.

backend/cmd/start.go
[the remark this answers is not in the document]
ccressent: Now that I've read the rest of the PR more carefully, I think this should be an option on your new store rather than something that mentions Postgres?

My understanding is that your new store could potentially be independent of Postgres and that this TPS configuration is just to rate limit its writing to its backing store, whatever that is (Postgres or something else).

I suppose what I'm trying to say is that I found it confusing that the new option seems to configure some type of overall TPS cap with Postgres while that's not what it's doing.

backend/store/memory/event_store.go
echlebek: Yeah this is a bit crufty, I can look at refactoring things.

echlebek:
@ccressent this is essentially a write cache for events, but also has event store semantics, namely the event merging algo. I still need to implement cache eviction for when agents disconnect, as well. Stay tuned!

backend/store/memory/event_store.go
ccressent: Why? Wouldn't logging the issue and skipping over the entry be a way to go about it?

backend/store/memory/event_store.go
ccressent: That's fair, thank you.

backend/store/memory/event_store.go
echlebek: If the system has corrupted memory then it is working in an undefined state and should probably crash.

backend/cmd/start.go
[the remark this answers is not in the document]
echlebek: I'm up for selecting a better name to prevent confusion! Any suggestions?

echlebek:
> all the events from the backing store could end up in the cache all at once

Once eviction is added, only events that are related to connected agents will be in the cache.

> I think a table that holds pointers to dirty entries to be written in the next flush would be a better design.

I considered doing this, but I think it would potentially use even more memory in the worst case, when all entries are dirty, which would be the common case in a heavily loaded system.

c-kruse (approved):

PR #4957, opened 2023-02-23
Fix known issues in flapping integration tests

c-kruse:
* Queue testspec: Expect queue.Reserve to return error wrapping context error on timeout instead of exact match.
* Queue testspec: don't panic in test on reserve error.
* TestSynchronizedExecutor: Adapt test to not rely on predictable goroutine scheduling.


echlebek (approved):
:crossed_fingers:

PR #4961, opened 2023-02-27
Add log-millisecond-timestamps backend configuration flag

c-kruse:
The log-millisecond-timestamps configuration allows users to configure a sensu backend to use RFC3339 formatted timestamp strings with millisecond precision in its backend logs.

## What is this change?

Closes https://github.com/sensu/sensu-enterprise-go/issues/2518

Y

N

N

Yes! New configuration flag.

`./sensu-backend start -c <base config path> --log-millisecond-timestamps --log-level debug |& jq .time`

## Is this change a patch?

N

### DIFF
--- a/CHANGELOG-6.md
+++ b/CHANGELOG-6.md
@@ -11,6 +11,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 ### Added
 - Adding a flag at agent level to avoid collecting system.networks property in the agent entity state
 - Added silences sorting by expiration to GraphQL service
+- Added log-precision-time backend configuration flag
 
 ## [6.9.1] - 2022-12-01
 
--- a/CHANGELOG-6.md
+++ b/CHANGELOG-6.md
@@ -11,7 +11,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 ### Added
 - Adding a flag at agent level to avoid collecting system.networks property in the agent entity state
 - Added silences sorting by expiration to GraphQL service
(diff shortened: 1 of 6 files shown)


ccressent:
My main comment is about the name of the flag. I don't believe `log-precision-time` is very descriptive; one needs to go look at the documentation to figure out what that means. What about `log-millisecond-timestamps`?

More generally, maybe it's a good opportunity to future proof this: what about a flag that configures the timestamp format, say `--log-timestamp-format`, taking a timestamp format string and defaulting to whatever our current default is. Then we add additional convenience alias flags as needed, such as `log-millisecond-timestamps` here. Such a flag would just be an alias to the appropriate `--log-timestamp-format <format>`.

That way we don't come back to this every time someone, somewhere, wants a different timestamp. And if a given format is used enough to warrant it, it may get its own alias (quick and easy).

What do you think?

c-kruse:
@ccressent  thanks for the suggestion! I've renamed the flag to `log-millisecond-timestamps`.

While I like the idea of allowing a user specified time layout for myself, I think go's time layout is probably a little too novel to throw at operators. I probably spend 5 minutes rereading `go doc time.Layout` each time I use it myself. I wager we'd get a lot of folks using strftime style formatting i.e. `%Y-%m-%dT%H:%M:%S

ccressent (approved):
Thanks for the changes; I too end up reading the docs every time I need to come up with a time format in Go... thanks for finding a good middle ground!

PR #4962, opened 2023-02-27
GraphQL service max query depth

gd8:
## What is this change?

This adds a GraphQL validator to limit max query depth. The validator is run on queries from un-authed requests

Logic inspired by:
- https://github.com/cungminh2710/graphql-depth-limiter (JS library for `graphql-js`, which is the basis for our `graphql-go`)
- https://github.com/graph-gophers/graphql-go/blob/master/internal/validation/validation.go (another GraphQL library in Go that has a max-depth validation built-in)

The query depth limit doesn't break un-authorized GraphQL queries we use which are listed here https://github.com/sensu/sensu-enterprise-go/issues/1952#issuecomment-1209485803

Relates to https://github.com/sensu/sensu-enterprise-go/issues/1952

Updated

Added tests. Also manually testing with a Web-UI and running queries with curl.

### DIFF
--- a/CHANGELOG-6.md
+++ b/CHANGELOG-6.md
@@ -5,6 +5,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic
 Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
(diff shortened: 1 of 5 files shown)


dabria (approved):
Looks good to me!

echlebek (approved):

PR #4968, opened 2023-03-09
Publish consistent data types to EntityConfig topic

c-kruse:


echlebek (approved):
Looks good, can we delete the store.WatchEvent types?

PR #4969, opened 2023-03-10
Make keepalive monitor less CPU intensive

echlebek:
This change adds indices to the opc table and also relaxes the keepalive monitoring loop so that it is only every 10 seconds.


c-kruse (approved):

echlebek:
An unrelated test case failed. I think it's just a flaky test.

PR #4976, opened 2023-03-22
Improve general cache efficiency

echlebek:
This change adds new store listing features, and leverages those features to make caching less expensive.

Specifically, two new SelectionPredicate fields have been added, UpdatedSince and IncludeDeletes. This tells the store to only select resources that have been updated since a particular time stamp, and include soft-deleted entries. The ability to do this allows cachev2 to work efficiently with any store that supports these list options.

The SelectionPredicate fields have been supported in the ConfigStore and the EntityConfigStore. We can hold off on implementing these features in other stores for now.

The cache/v2 itself has been updated to leverage these features and now has a simpler update strategy as a result.

Integration tests have been added for both the ConfigStore and EntityConfigStore types in the cache/v2 package.


backend/apid/handlers/get_v3_test.go
```diff
@@ -80,6 +80,14 @@ func TestHandlers_GetV3Resource(t *testing.T) {
 				return
 			}
 
+			if got != nil {
+				meta := got.GetMetadata()
+				// delete these to facilitate comparison
```
c-kruse, line 85: lol

c-kruse (approved):

PR #4977, opened 2023-03-23
fix entity config poller and include test

c-kruse:
Fixes bug in EntityConfigPoller where we were iteratively appending to a slice of updates, but instead of appending new updates we were modifying and appending the same update (a pointer scoped to outside of the loop.)

Also does some housekeeping of the pollers/watcher code: Cleans up some nasty tests and obsolete tests, replaces them with tests using the real implementation of the entity config poller.

@echlebek  I'm pretty sure this is the "scheduling" bug.


c-kruse:
aaand I've made it very hard to see what I've changed by moving things between files.

See the first commit for the actual bug fix: https://github.com/sensu/sensu-go/pull/4977/commits/199b263c5ce1330be56c8c1f0a53423d685e26c0#diff-1b90938f51ff73ab53f5cac51355ce22115e89fd2563fa9a5e56b996786730bb

PR #4983, opened 2023-03-30
Implement secure API keys

echlebek:
This commit adds secure API keys to Sensu, breaking compatibility with Sensu 6.x API keys. API keys are no longer stored in plaintext. Instead, their salted hash is stored, and after the initial key grant, the key cannot be retrieved from Sensu.

The sensuctl api-key grant command works the same way as before, but now responds slightly differently, and also includes a uniquely chosen name for the API key.

API keys can also now be created with sensuctl create with a supplied hash value, when migrating keys between clusters.


backend/apid/middlewares/authentication.go
```diff
@@ -75,43 +77,32 @@ func (a Authentication) Then(next http.Handler) http.Handler {
...
-	user := &corev2.User{
-		Username: apiKey.Username,
-	}
-
-	req = storev2.NewResourceRequestFromResource(user)
-	wrapper, err = store.GetConfigStore().Get(ctx, req)
-	if err != nil {
-		return nil, err
-	}
+	for _, apiKey := range apiKeys {
```
c-kruse, line 86: Suppose the way bcrpyt works we can't actually look up a particular hash, huh?

You worried at all about this becoming a bourdon in highly automated environments?

c-kruse (approved):
What a relief to have this sorted!

backend/apid/middlewares/authentication.go
echlebek: That's correct. Since salt selection is internal to how bcrypt works, we can't hash the password and get the same result. I had forgotten about that fact going into this.

If we could select our own salts with another algorithm like scrypt then we would have the ability to get the same hash. But that in itself would weaken our security posture, because you are supposed to use a completely random salt.

We could improve lookup efficiency by having the client include the user the API key is associated with, which is an additional client burden, but could reduce server-side costs a lot. For now, I'm tempted to wait and see if it becomes a problem.

echlebek: I suppose yet another trick we could pull is to include both the username and the API key in an encoded payload and hand _those_ out instead of UUIDs. Then we'd be able to perfectly query the set of API keys for a given user, and the client experience would be the same.

echlebek: We can improve the efficiency of lookups in a future effort.

Issue #4455, opened 2021-10-26
Make asset build matching order deterministic


jspaleta:
Currently when asset build filter rules overlap in an asset definition, allowing for a multiple builds to be used on a target agent platform, its undetermined which matching asset will be used. 

The nondeterministic nature of which matching asset will be downloaded makes overlapping assets situations  difficult to test,confirm and resolve. 

If there was a way to enforce ordering of asset builds, so that each build, this would allow for deterministic behavior in the case of overlapping filter rules.

## Possible Implementation 
explicit match order using a new priority attribute.

### Use case
Existing asset has a build for debian 7 with filter "debian platform release >= 7"
Introduce new build for debian 8 with filter "debian platform release >= 8"
Set priority higher for debian 8 build relative to debian 7 build
Use asset on agent running on debian 9, result is deterministic use of debian 8 build.

amdprophet:
Dupe of / related to https://github.com/sensu/sensu-go/issues/3655.

Issue #3655, opened 2020-03-30
Asset builds with highest counted elements of build filter takes precedence 


tarcinil:
## Feature Suggestion
When multiple builds have passing filters use the build with the highest (counted elements) matched filters.

## Possible Implementation
N/A - I haven't reviewed the current code.

## Context
> For assets with multiple builds, only the first build which matches will be downloaded and installed.

Based on this from [Asset Documentation](https://docs.sensu.io/sensu-go/latest/reference/assets/#asset-build-evaluation), it means that if you have:

```
filters: ["entity.system.os == 'linux'", "entity.system.arch == 'amd64'"]
```
vs.
```
filters: ["entity.system.os == 'linux'", "entity.system.arch == 'amd64'", "some.label.zone == 'foo'"]
```

Present Implementation would use the build information from the first `build` slice.

Feature: Because the 2nd slice has 3 total elements, it takes precedence.  
Edge Case: same present condition, if 2 slices of the same size exists, use first.

echlebek:
Interesting idea, thanks!

nikkixdev:
We should audit how filtering behaves currently, @amdprophet believes it's the first match. If that's the case, implementing this could be a breaking change for some use cases. Potentially a good candidate for 6.0.0, but needs more investigation first.

calebhailey:
I removed the `wontfix` label, but this might have to sit in the backlog until the next major release (2021) as it would be a breaking change, as commented above.

tarcinil:
That is fine.  I have worked around the issue in Ruby.

PR #4990, opened 2023-04-17
Use asset build with most specific filters

amdprophet:
## What is this change?

Updates `asset.FilteredManager` to evaluate all asset build filters and returns the asset build with the most specific (highest number of filters) matching the entity. This is a breaking change as it can alter which asset build is installed on existing clusters.

Closes https://github.com/sensu/sensu-go/issues/3655
Closes https://github.com/sensu/sensu-go/issues/4455

Yes.

No.

No.

The docs should be updated to describe this behaviour.
https://github.com/sensu/sensu-docs/issues/4185

I used TDD to define the behaviour before I made the change. The tests should be sufficient.

## Is this change a patch?

No.


echlebek (approved):
Looks good to me!

PR #5004, opened 2023-05-05
Use Configuration Store Selectors

c-kruse:
Changes the configuration store to be compatible with sensu-go selectors, and use them in the rest api.


echlebek (approved):
Looks about like what I'd expect! :+1:

PR #5005, opened 2023-05-05
Add support for etag and preconditions to store

echlebek:
## What is this change?

This commit adds support for store preconditions and etags. Store users can request IfMatch and IfNoneMatch by way of context value. Resources returned by the store now encode a base64-encoded ETag in their annotations.

It's also possible to detect the side effects a given store method had, by providing a TxInfo object to the method, also via context value. The store will write into TxInfo whether the object has been created, updated, or deleted.

### DIFF
--- a/backend/store/postgres/config_schema.go
+++ b/backend/store/postgres/config_schema.go
@@ -43,22 +43,41 @@ INSERT INTO configuration ( api_version
 						  , annotations
 						  , fields
 						  , resource
+						  , etag
 )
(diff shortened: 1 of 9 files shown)


echlebek:
Fixing up some missed problems

echlebek:
Please ignore build failure from appveyor

echlebek:
Not ready for review, I'll need to base this on Christian's pending work.

echlebek:
@c-kruse @ccressent ready for review. I'm going to start working on support in APID now.

c-kruse (approved):

PR #5006, opened 2023-05-05
Migrate AppVeyor job to CircleCI

amdprophet:
Migrates the AppVeyor unit test job to CircleCI


echlebek:
Looks like we need to somehow deregister appveyor from the repo :thinking:

amdprophet:
@echlebek it's most likely still running because the target branch (main) has an appveyor config.

echlebek:
OK! You may fire when ready. :ringed_planet:

Issue #4987, opened 2023-04-07
Proposal to also add OK status to metric threshold annotations & Improvement threshold annotations format


jadiunr:
## Feature Suggestion

In the current Sensu Go (v6.9.2) , Annotations are added to Events when the metric thresholds set by output_metric_thresholds are exceeded.
However, they are only added when the status is Warning or Critical, and no annotations are added when the status is **OK**.

In some use cases, such as using the  [sensu-slack-handler](https://github.com/sensu/sensu-slack-handler) to notify metric threshold annotations, it is somewhat difficult to understand why the metric threshold alert is resolved because there are no annotations on the Resolved event when the metric threshold returns to normal.
(It can see what happened by looking at Critical events, though...)

![Screenshot_2023-04-08_03-37-06](https://user-images.githubusercontent.com/19828725/230661413-552d99c2-1e41-4f76-9134-a0deab28b628.png)

Based on the above, I suggest that annotations be added even when the threshold is within the normal range.

## Possible Implementation

PR: #5013 #5014

```json
  "metadata": {
    "namespace": "default",
    "annotations": {
      "sensu.io/notifications/ok": "The value of test_metric.value is within the configured threshold (max: 200, actual: 100).",
      "sensu.io/output_metric_thresholds/test_metric.value/max/ok": "The value of test_metric.value is within the configured threshold (max: 200, actual: 100)."
    }
  },
```

## Context

None in particular.
See above for details.

c-kruse:
@jadiunr thank you for the proposal and the PRs! Everything looks sound to me.

Pinging @fguimond who is working on pulling things together for a 6.10.0 release early next week. Perhaps we can move forward with this for a 6.10.1 release?

echlebek:
One of the biggest changes between sensu classic and sensu-go is that successful events now move all the way through the pipeline. (Previously only failing events did this.) I think this is a great enhancement because it reflects that same spirit of design for metrics thresholds. Thanks for the great issue and PRs.

jadiunr:
Thank you for replying!

I am considering having multiple thresholds for annotations as a better improvement idea.

```json
"metadata": {
  "namespace": "default",
  "annotations": {
    "sensu.io/notifications/ok": "The value of test_metric.value is within the configured threshold (max: 200 (status: warning) 300 (status: critical), actual: 100).",
[...]

PR #5014, opened 2023-05-17
Improvement threshold annotations format (for 6.x)

jadiunr:
## What is this change?

This PR adds annotations to the event even if the threshold check is normal.
In addition, the format is changed so that the min-max range can be output in a single annotation.

See #4987

Adding the threshold annotation when normal will allow to ensure that the thresholds are normal when notifying some communication platform of the Event.
This change also makes it easier to ensure that the threshold checks are working correctly.

Yes, required as **Added** and **Changed**.

This PR is for Sensu Go 6.x (develop/6 branch).

This section [Add event annotations based on metric threshold evaluation](https://docs.sensu.io/sensu-go/6.9/observability-pipeline/observe-schedule/metrics/#add-event-annotations-based-on-metric-threshold-evaluation) may need to be rewritten.

Setup a test environment using `output_metric_thresholds` and verified that the annotations have been added.

## Is this change a patch?

No.

### DIFF
--- a/CHANGELOG-6.md
+++ b/CHANGELOG-6.md
@@ -13,6 +13,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 - Added silences sorting by expiration to GraphQL service
 - Added log-millisecond-timestamps backend configuration flag
 - Added a session store, used to detect and prevent refresh token reuse
+- Added threshold annotation even when OK status
 
 ### Changed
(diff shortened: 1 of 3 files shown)


echlebek:
Sorry, but we can't accept this PR as it could potentially break workflows. However, the sibling PR for main is accepted.

PR #5016, opened 2023-05-25
Add support for conditional requests

### LINKED ISSUE #4341: Update resource PUT APIs to support 200 Created and 304 Not Modified
## Feature Suggestion

The POST APIs (e.g. CheckConfig POST) already return HTTP response code 201 Created and 409 Conflict for when the resource already exists. Extending the PUT APIs to indicate that a resource was either 201 Created, 200 OK (updated) or 304 Not Modified would enable better user experiences to be built when working with resources. Example: #3079 

## Context

Most core/v2 resource PUT APIs currently always return HTTP 201 when successful as the store interface does not facilitate communicating which action was taken.

Update these handlers to return
HTTP 201 Created when the resource did not previously exist
HTTP 200 OK when the resource was updated
HTTP 304 Not Modified when the resource was unchanged
## Tech Notes

**Router:**

The apid router will likely need rearranged to allow for multiple non-error status codes per route. The existing interface `type actionHandlerFunc func(r *http.Request) (interface{}, error)` only allows for variable error codes. Successful codes are hard coded in the router per-method.

**Etcd Store:**

The likely most performant solution in the etcd store to know whether or not a resource was created, updated or unchanged is to keep it in a single transaction and request prev_kv on the put operation. This could end up being flakey due to etcd shenanigans like compaction.

**Store Interface Options:**

The resources using the standard PUT handler all use the generic CreateOrUpdateResource store method today. The signature of this interface does not provide a way to communicate back whether or not a resource has been created, replaced or unchanged. Our options here are to update the store interface to return the previous value of the updated resource or to return a set of flags indicating what action was taken.

echlebek:
This commit adds support for the conditional HTTP headers If-Match and If-None-Match. It also changes the way ETag is produced for all resources.

Clients can use If-Match and If-None-Match in the way described by RFC 7232 when making requests.

Closes #4341


echlebek:
@c-kruse this isn't fully complete yet. I need to add test coverage in a few places. But I wanted to get this onto your reviewing docket as it's quite a lot of changes, more than I was hoping for tbh.

c-kruse (commented):
On a little bit of my own journey here, but i'm like 90% sure patch is busted anyways looking at the config store code.

echlebek:
@c-kruse Probably good enough to start reviewing, though I suspect there might be some missing pieces. I'm reviewing myself as well.

echlebek:
I've done quite a bit of manual end to end testing on this now, I think it's good to go.

c-kruse (approved):
Very exciting to see these working!

PR #5040, opened 2023-11-20
Use native Semgrep in CI

amdprophet:
The GitHub Action that we were using for Semgrep was deprecated and has now been removed. Use native Semgrep instead.


ccressent (commented):
:+1:

PR #5050, opened 2024-02-01
Git issue #5009 Asset cache can lead to denial of service if asset database is deleted -Fix

### LINKED ISSUE #5009: Asset cache can lead to denial of service if asset database is deleted
## Expected Behavior
When the asset database is deleted, but the asset cache is not, the system should still work as expected.

## Current Behavior
When the asset database is deleted, but the asset cache is not deleted, then the asset downloader encounters errors about files already existing when trying to populate the cache.

## Possible Solution
Replace existing files when the database is missing an entry.

SudhanshuBawane:
Closed #5009 

## Description
The change is regarding the denial of service from agent when asset.db gets deleted due to some external scenario. 
Which creates a DOS as SHA contains the some reference to the previous asset.db and hence forth the new one does not get created properly.

## Change in behavior
To prevent the DOS of the agent in above scenario and keep the agent working properly. For new asset.db creation. 

## Added 

- The new cleanUp function for cleaning up the SHA dir from cache to prevent a DOS in case of asset.db deletion.
- Added the test to check all scenarios.

## Changed

- The expandWithDuration function in boltdb_manager has been changed a little for the cleanUp function support.

## Fixed

- This fixes the DOS in agent in case of asset database is deleted.

## Change verification
The changes can be verified not only by the test cases but also by checking the same behavior. That is delete the asset.db and make note of it's size. Then re-run the agent and it will run without populating error file exits and asset.db get recreated with same size.

### DIFF
--- a/CHANGELOG-6.md
+++ b/CHANGELOG-6.md
@@ -8,6 +8,14 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### 2024-02-01
+
+### Changed
+- The expandWithDuration in boltdb_manager to have a assetSHA dir cleanup.
+
+### Added
+- Added the cleanup in expander to clean the assetSHA in case of asset cache getting deleted
+
 ### Changed
 - Upgraded CI Go version to 1.21.3
 - Upgraded jwt version to 4.4.3
--- a/CHANGELOG-6.md
+++ b/CHANGELOG-6.md
@@ -10,11 +10,8 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### 2024-02-01
 
-### Changed
-- The expandWithDuration in boltdb_manager to have a assetSHA dir cleanup.
-
-### Added
-- Added the cleanup in expander to clean the assetSHA in case of asset cache getting deleted
+### Fixed
+- Assets can now be re-installed when the asset.db file has been deleted.
 
 ### Changed
 - Upgraded CI Go version to 1.21.3
--- a/asset/boltdb_manager.go
+++ b/asset/boltdb_manager.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/spf13/viper"
 	"os"
 	"path/filepath"
 
@@ -23,6 +24,7 @@ const (
 	// ExpandDuration is the name of the prometheus summary vec used to track
 	// average latencies of asset expansion.
 	ExpandDuration = "sensu_go_asset_expand_duration"
+	FlagCacheDir   = "cache-dir"
 )
 
 var (
@@ -240,6 +242,14 @@ func (b *boltDBAssetManager) expandWithDuration(tmpFile *os.File, asset *corev2.
 			Observe(v * float64(1000))
 	}))
 	defer timer.ObserveDuration()
+
+	assetSHA := asset.Sha512
+	CacheDir := viper.GetString(FlagCacheDir)
+	fullPath := filepath.Join(CacheDir, assetSHA)
+
+	if err := CleanUp(fullPath); err != nil { //fix for git issue 5009
+		logger.Printf("error cleaning up the SHA dir: %s", err)
+	}
 
 	assetPath = filepath.Join(b.localStorage, asset.Sha512)
 	return assetPath, b.expander.Expand(tmpFile, assetPath)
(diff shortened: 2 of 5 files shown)


asset/boltdb_manager.go
```diff
@@ -241,6 +243,14 @@ func (b *boltDBAssetManager) expandWithDuration(tmpFile *os.File, asset *corev2.
 	}))
 	defer timer.ObserveDuration()
 
+	assetSHA := asset.Sha512
+	CacheDir := viper.GetString(FlagCacheDir)
+	fullPath := filepath.Join(CacheDir, assetSHA)
+
+	if err := CleanUp(fullPath); err != nil { //fix for git issue 5009
+		logger.Printf("error cleaning up the SHA dir: %s", err)
```
fguimond, line 251: It would be nice if more context information such as the SHA  and full path could be provided with the log message. You can use `WithFields` and `WithError` .

Additionally please use an explicit log level such as `Errorf` or `Warnf` instead of using the generic `Printf`. In this case I believe warning would be an appropriate log level.

You can see an example at https://github.com/sensu/sensu-go/blob/76c7a167e572bb370438e38b678fa19c3d973de1/backend/pipeline/filter/legacy.go#L126

asset/expander.go
```diff
@@ -90,3 +91,14 @@ func sniffType(f io.ReadSeeker) (filetype_types.Type, error) {
 
 	return ft, nil
 }
+
+// Sudhanshu - CleanUp the SHA for the git issue 5009 fix. Making sure that in case of DOS when asset.db gets deleted it gets cleanUp so that asset can be re-downloded
```
fguimond, line 95: Nitpick: no need to add your name and issue number :)

fguimond:
A couple of minor things but looks good otherwise!

fguimond:
Don't forget to add a changelog entry.

asset/expander.go
```diff
@@ -90,3 +91,13 @@ func sniffType(f io.ReadSeeker) (filetype_types.Type, error) {
...
+
+// cleanup of the assetSHA when cache dir gets force deleted
+func CleanUp(fullPath string) error {
+	errorSHA := os.RemoveAll(fullPath)
+	if errorSHA != nil {
+		return errorSHA
+	}
+	return nil
+
+}
```
echlebek, line 103: This is equivalent to os.RemoveAll and its use can be replaced with os.RemoveAll.

CHANGELOG-6.md
```diff
@@ -8,6 +8,14 @@ Versioning](http://semver.org/spec/v2.0.0.html).
...
 ## Unreleased
 
+### 2024-02-01
+
+### Changed
+- The expandWithDuration in boltdb_manager to have a assetSHA dir cleanup.
+
+### Added
+- Added the cleanup in expander to clean the assetSHA in case of asset cache getting deleted
+
```
echlebek, line 18: The changelog entry should say 

```
### Fixed
- Assets can now be re-installed when the asset.db file has been deleted.
```

echlebek (changes requested):

CHANGELOG-6.md
SudhanshuBawane: This has been modified

asset/expander.go
SudhanshuBawane: This change has been made

asset/expander.go
```diff
@@ -90,3 +91,8 @@ func sniffType(f io.ReadSeeker) (filetype_types.Type, error) {
 
 	return ft, nil
 }
+
+// cleanup of the assetSHA when cache dir gets force deleted
+func CleanUp(fullPath string) error {
+	return os.RemoveAll(fullPath)
+}
```
echlebek, line 98: Please remove this and use os.RemoveAll directly in boltdb_manager.go.

asset/expander_test.go
```diff
@@ -113,3 +113,33 @@ func TestExpandInvalidArchive(t *testing.T) {
...
+	err = CleanUp(SHAFilePath)
+	if err != nil {
+		t.Errorf("CleanUp returned an error: %v", err)
+	}
+
+	_, err = os.Stat(SHAFilePath)
+	if !os.IsNotExist(err) {
+		t.Errorf("CleanUp did not remove the dummy file as expected")
+	}
+}
```
echlebek, line 145: This only tests os.RemoveAll and is therefore unnecessary

echlebek (changes requested):
Please add a test that shows that the boltdb manager now exhibits correct behaviour when the asset.db has been deleted.

SudhanshuBawane:
> Please add a test that shows that the boltdb manager now exhibits correct behaviour when the asset.db has been deleted

SudhanshuBawane:
> > Please add a test that shows that the boltdb manager now exhibits correct behaviour when the asset.db has been deleted

May I know what kind of behavior is expected like what I need to check ?

PR #5055, opened 2024-03-12
Disabling agent account doesn’t terminate active connections - Fix

SudhanshuBawane:
closed [2608](https://github.com/sensu/sensu-enterprise-go/issues/2608)

## Description

To fix the behavior of the agent not getting disconnected while user is disabled.

## Fixed

The agent is now getting disconnected as soon as the user is disabled unlike before.

## Added

- The new watcher configuration for user has been added.
- The session code has been modified to consume the new user configuration into the bus.

## Change verification

To verify the change as normal run backend and agent, as soon as both are running and connected, Disable the user via sensuctl command. In backend you will see a shutting down a agent session prompt and in agent will fail to do a handshake with bad credential prompt.


sarvodaya-kumar-26 (approved):

rakibhossainctr:
ready to merge

backend/agentd/agentd.go
```diff
@@ -308,6 +320,24 @@ func (a *Agentd) handleEvent(event store.WatchEventEntityConfig) error {
 	return nil
 }
 
+// adding the UserConfig updates to the etcd bus for the watcher to consume
+func (a *Agentd) handleUserEvent(event *store.WatchEventUserConfig) error {
+	if event.User == nil {
```
chavakula, line 328: Check if event is null here, if event is null it will panic

backend/agentd/session.go
```diff
@@ -321,6 +347,45 @@ func (s *Session) sender() {
...
+				}
+				logger.Println("The update operation has been performed on user")
+			default:
+				err := fmt.Errorf("session received unexpected user watch event action: %s", watchEvent.Action.String())
+				logger.WithError(err).Error("user update handling error")
+				continue
+			}
+
+			if watchEvent.User == nil {
+				logger.Error("session received nil user in watch event")
```
chavakula, line 373: You should continue or return after detecting nil user

backend/agentd/session.go
```diff
@@ -321,6 +347,45 @@ func (s *Session) sender() {
...
+			}
+			lagger := logger.WithFields(logrus.Fields{
```
chavakula: lagger => logger

backend/agentd/session.go
```diff
@@ -321,6 +347,45 @@ func (s *Session) sender() {
...
+				"action":    watchEvent.Action.String(),
+				"user":      watchEvent.User.Username,
+				"namespace": watchEvent.User.GetMetadata().GetNamespace(),
+			})
+			lagger.Debug("user update received")
+
+			bytes, err := s.marshal(watchEvent.User)
```
chavakula: bytes is a standard library package name.  avoid using them

backend/agentd/watcher.go
```diff
@@ -70,3 +69,44 @@ func GetEntityConfigWatcher(ctx context.Context, client *clientv3.Client) <-chan
...
+				ch <- &store.WatchEventUserConfig{
+					Action: response.Type,
+				}
+				continue
+			}
+
+			// unmarshal the user config
+			var userConfig corev2.User
+			if err := proto.Unmarshal(response.Object, &userConfig); err != nil {
+				continue
```
chavakula, line 104: Log error message here

chavakula (approved):

Issue #5063, opened 2024-08-21
etcd slowly increase in size, sessions cleanup missing


elfranne:
Sensu store sessions inside etcd but does not clean up. This will slowly increase the size of etcd, potentially stopping etcd as there is a hard limit at 2GB (default).

## Expected Behavior
Clean up old/expired sessions

## Current Behavior
Old sessions are kept under `/sensu.io/user_sessions/admin/[uuid]` inside etcd.

You can see the number of sessions kept in etcd:
`etcdctl [insert authentication and endpoint info] --command-timeout=200s --dial-timeout=200s get "/sensu.io/user_sessions/admin" --prefix=true --keys-only | grep -c -v -e '^\s*$' `. After a few month we were getting close to 1.7GB, with nearly 5.000.000 keys in the etcd, and over 4.9m just being sessions.

## Possible Solution
Quick fix is to delete all sessions:
`etcdctl [insert authentication and endpoint info] --command-timeout=200s --dial-timeout=200s del "/sensu.io/user_sessions/admin" --prefix=true`
This reduced etcd from 1.7GB to 392MB.

Proper fix would be to clean expired sessions.

## Steps to Reproduce (for bugs)
The scale of the issue would probably hard to reproduce but we can see that the sessions are kept in etcd and keeps increasing.

## Your Environment
Sensu 6.10
Ubuntu 22.04 LTS

jhenderson-pro:
Hey @elfranne. 👋 

Thank you for the well-written issue and testing. I'll have our Engineering team take a look at this.

valsharess:
This is a major bug, is there any update from Engineering? We had 10.5 million user session keys that we had to cleanup, this has brought memory usage down by 9GB and reduced the database from 1.8GB to 20MB.

I don't want to have to check on this every few months as we use the api quite a lot so the sessions start piling up.

ManishaKumari295:
> This is a major bug, is there any update from Engineering? We had 10.5 million user session keys that we had to cleanup, this has brought memory usage down by 9GB and reduced the database from 1.8GB to 20MB.
> 
> I don't want to have to check on this every few months as we use the api quite a lot so the sessions start piling up.

Hi , the issue is being worked upon with code fixes . We will update you about the fix once it is integrated and available. Trying our best to make it soon.

sensu-discourse:
This issue has been mentioned on **Sensu Community**. There might be relevant details there:

[...]

PR #5064, opened 2024-09-12
Etcd cleanup

ManishaKumari295:
Closed https://github.com/sensu/sensu-go/issues/5063

### Description
The user sessions gets piled up in etcd occupying upto 2bg space .

### Change in behavior
Deleting the etcd user_sesssions them after 6 minutes of creation ( Access token of session gets expired in 5 mnts) This change will make sure the memory occupancy uptil 1.8 GB  will be eradicated 

### Added
Added TTL time of 6 minutes in session.go file for each entry regarding user-sessions. 

### Fixed
This fixes the etcd space occupancy issue .

### Change verification
- Do sensuctl configure.
- Check respective entry in etcd ar /sensu.io/user-sessions/admin/xxyyzz[session_id]
- this entry will be deleted automatically after 11 minutes if the backend keep .
- I have the video of default behaviour ( which caused the issue) and the video of fixed behaviour with TTL=2mins , unfortunately cannot be uploaded due to big size.

### DIFF
--- a/CHANGELOG-6.md
+++ b/CHANGELOG-6.md
@@ -9,8 +9,8 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 ## [6.11.1] - 2024-09-12
 
 ### Changed
-- Added TTl to each entry in user-session within etcd
-- TTl value is 6 minutes
+- ADD TTL to each user session in the etcd data store to prevent leak.
+- TTl value is DefaultAccessTokenLifeSpan + 1 minute
 
 ## [6.11.0] - 2024-01-31
 
--- a/CHANGELOG-6.md
+++ b/CHANGELOG-6.md
@@ -10,7 +10,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Changed
 - Added TTl to each entry in user-session within etcd
-- TTl value is 11 minutes
+- TTl value is 6 minutes
 
 ## [6.11.0] - 2024-01-31
 
--- a/CHANGELOG-6.md
+++ b/CHANGELOG-6.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic
 Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [6.11.1] - 2024-09-12
+
+### Changed
+- Added TTl to each entry in user-session within etcd
+- TTl value is 11 minutes
+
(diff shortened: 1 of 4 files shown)


ManishaKumari295:
> Thanks for the fix @ManishaKumari295 but it seems there is a issue with the branch...

Hi @elfranne , the branch seems to be fine now. Please verify the commited changes. 
Thanks

CHANGELOG-6.md
```diff
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic
 Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [6.11.1] - 2024-09-12
+
+### Changed
+- Added TTl to each entry in user-session within etcd
```
fguimond, line 12: I would rephrase this to something like this and have just one entry:
```
### Fixed
- ADD TTL to each user session in the etcd data store to prevent leak.
```

backend/store/etcd/session.go
```diff
@@ -30,12 +30,15 @@ func (s *Store) GetSession(ctx context.Context, username, sessionID string) (str
 }
 
 // UpdateSession applies the supplied state to the session uniquely identified
-// by the given username and session ID.
+// by the given username and session ID and TTL of 6 minutes added considering access token expires in 5 minutes
```
fguimond, line 33: Instead of hard-coding the value to 6 minutes I would calculate it based on the token expiration. I think something like (token expiration * 1.5) would be ok.

fguimond:
Great approach @ManishaKumari295. Just a couple of minor things and it should be good to go!

elfranne:
I have limited knowledge in the inner workings of the Sensu, please take my findings with a grain of salt.
After some research I can see that Etcd has a feature to create a [lease grant](https://etcd.io/docs/v3.5/dev-guide/api_reference_v3/):
> LeaseGrant creates a lease which expires if the server does not receive a keepAlive within a given time to live period. All keys attached to the lease will be expired and deleted if the lease expires. Each expired key generates a delete event in the event history.

But it seems we are now creating grants based on username and sessionID and usuing new grants at every UpdateSession. Will this not result of a similar situation where we now have grants filling up etcd?

Another approach would be to create a lease grant per user at the token creation and do a LeaseKeepAlive. There might also be necessary to create a cleanup of lease grants for deleted users.

@fguimond, out of curiosity, why would you keep expired token?
 >(token expiration * 1.5)

PR #5065, opened 2024-09-26
Etcd cleanup

### LINKED ISSUE #5063: etcd slowly increase in size, sessions cleanup missing

ManishaKumari295:
closed #5063

**Description**
The user sessions gets piled up in etcd occupying upto 2bg space .

**Change in behavior**
Deleting the etcd user_sesssions them after 6 minutes of creation ( Access token of session gets expired in 5 mnts) This change will make sure the memory occupancy uptil 1.8 GB will be eradicated

**Added**
Added TTL time of 6 minutes in session.go file for each entry regarding user-sessions.

**Fixed**
This fixes the etcd space occupancy issue .

**Change verification**
Do sensuctl configure.
Check respective entry in etcd ar /sensu.io/user-sessions/admin/xxyyzz[session_id]
this entry will be deleted automatically after 11 minutes if the backend keep .
I have the video of default behaviour ( which caused the issue) and the video of fixed behaviour with TTL=2mins , unfortunately cannot be uploaded due to big size

### DIFF
--- a/CHANGELOG-6.md
+++ b/CHANGELOG-6.md
@@ -9,8 +9,8 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 ## [6.11.1] - 2024-09-12
 
 ### Changed
-- Added TTl to each entry in user-session within etcd
-- TTl value is 6 minutes
+- ADD TTL to each user session in the etcd data store to prevent leak.
+- TTl value is DefaultAccessTokenLifeSpan + 1 minute
 
 ## [6.11.0] - 2024-01-31
 
--- a/CHANGELOG-6.md
+++ b/CHANGELOG-6.md
@@ -10,7 +10,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Changed
 - Added TTl to each entry in user-session within etcd
-- TTl value is 11 minutes
+- TTl value is 6 minutes
 
 ## [6.11.0] - 2024-01-31
 
--- a/CHANGELOG-6.md
+++ b/CHANGELOG-6.md
@@ -10,7 +10,6 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Changed
 - ADD TTL to each user session in the etcd data store to prevent leak.
-- TTl value is DefaultAccessTokenLifeSpan + 1 minute
 
 ## [6.11.0] - 2024-01-31
 
--- a/CHANGELOG-6.md
+++ b/CHANGELOG-6.md
@@ -6,10 +6,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic
 Versioning](http://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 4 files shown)


backend/store/etcd/session.go
```diff
@@ -30,12 +30,18 @@ func (s *Store) GetSession(ctx context.Context, username, sessionID string) (str
 }
 
 // UpdateSession applies the supplied state to the session uniquely identified
-// by the given username and session ID.
+// by the given username and session ID with attached lease and TTl for each key
 func (s *Store) UpdateSession(ctx context.Context, username, sessionID, state string) error {
-	if _, err := s.client.Put(ctx, userSessionPath(username, sessionID), state); err != nil {
+
+	leaseDuration := jwt.DefaultAccessTokenLifespan
+	ttl := int64(leaseDuration.Minutes()+1) * 60
```
elfranne, line 37: Any reason to adding a minute to the lease?

CHANGELOG-6.md
```diff
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic
 Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [6.11.1] - 2024-09-12
+
+### Changed
+- ADD TTL to each user session in the etcd data store to prevent leak.
+- TTl value is DefaultAccessTokenLifeSpan + 1 minute
```
fguimond, line 13: Just one comment per change is required. The second entry can be removed.

backend/store/etcd/session.go
ManishaKumari295: This is to keep a buffer in deletion, no other reason.

fguimond (commented):
LGTM

ManishaKumari295:
Merging to develop/6

PR #5073, opened 2024-12-05
.initialized key issue

### LINKED ISSUE #5047: Required ".initialized" key in external etcd gives sensu user access to all key spaces
## Expected Behavior

We want to use an external etcd for our sensu backen. We expected that after configuring it as described in the [documentation](https://docs.sensu.io/sensu-go/latest/operations/deploy-sensu/cluster-sensu/#configure-key-space-access) is enough.

## Current Behavior

While initializing the sensu backend, it fails with "permission denied" error. Because it requires the access to ".initialized" key too. After giving access to that key, initialization goes through without problem, but then the problem is that the sensu user has access to `/` key space too.

## Possible Solution

Move the required ".initialized" key to "/sensu.io/.initialized". 

1. Deploy an etcd cluster
2. Create sensu user and its roles as described in [documentation](https://docs.sensu.io/sensu-go/latest/operations/deploy-sensu/cluster-sensu/#configure-key-space-access)
3. Init the sensu backend as described in [documentation](https://docs.sensu.io/sensu-go/latest/operations/deploy-sensu/cluster-sensu/#start-etcd)

## Context

We want to limit the acces of sensu user to `/sensu.io/` key space only, because we plan to have other applications using other key spaces.

## Your Environment

* Sensu version used (sensuctl, sensu-backend, and/or sensu-agent): 6.10.0
* etcd version: 3.5.11
* Installation method (packages, binaries, docker etc.): docker

SudhanshuBawane:
Closed #5047 

## Description

We wanted to implement a simple add-on to use an external etcd for our sensu backend. After configuring it as described in the(https://docs.sensu.io/sensu-go/latest/operations/deploy-sensu/cluster-sensu/#configure-key-space-access) will be enough.

## Changed

Just a change of the key replacing ".initialized" key to "/sensu.io/.initialized".

### DIFF
--- a/backend/store/etcd/initialization.go
+++ b/backend/store/etcd/initialization.go
@@ -11,9 +11,14 @@ import (
 	"go.etcd.io/etcd/client/v3/concurrency"
 )
 
+//const (
+//	initializationLockKey = ".initialized.lock"
+//	initializationKey     = ".initialized"
+//)
+
 const (
-	initializationLockKey = ".initialized.lock"
-	initializationKey     = ".initialized"
+	initializationLockKey = "/sensu.io/.initialized"
+	initializationKey     = "/sensu.io/.initialized"
 )
 
 // StoreInitializer ...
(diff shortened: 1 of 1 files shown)


fguimond:
I believe we need to migrate the old `.initialized` key to `sensu.io/.initiazlied` otherwise customers will need to re-initialize their environment.

PR #5118, opened 2025-07-16
Fallback Pipelines

### LINKED ISSUE #4407: Stop execution of asset consumers if asset.GetAssets() or asset.GetAll() errors
## Expected Behavior
If an asset definition is failed to be fetched from the store or if an asset fails to download/extract, the consumer of the resource (check, handler, etc.) should not proceed with execution.

if you have multiple assets that provide a binary with the same name, an asset failing to be retrieved from the store could result in a different binary being executed

## Current Behavior
We currently log errors when calling `asset.GetAssets()` but continue on with execution of the resource, regardless of what kind of error occurs.

We attempt to stop execution if `store.ErrInternal` is returned from `asset.GetAll()` but it can never return that error as it doesn't interact with the store; `asset.GetAssets()` does.

https://github.com/sensu/sensu-go/blob/main/backend/pipeline/handler.go#L243-L255
https://github.com/sensu/sensu-go/blob/3732d6c2eb7a7f2584fa766871f2d61a42d0c1fe/asset/store.go#L11

If we allow for execution to continue after asset retrieval failures, it could result in other binaries with the same name (from the system or other runtime assets) being executed instead of the expected binary.

## Possible Solution
Update `asset.GetAssets()` to return errors. Update error checks to return if an error is returned from `asset.GetAssets()` or `asset.GetAll()`.

rakibhossainctr:
## What is this change?

1. Updating agent events to carry FallbackPipeline to backend
2. Updating GetAssets to raise errors on missing runtime assets

When the execution of a workflow pipe handler results in an error (e.g. fork/exec failure), the error is logged and any other workflows in the pipeline are skipped. This does not include pipe handlers that return a non-zero exit status. The AdapterV1 Pipeline Adapter does not currently distinguish between zero and non-zero exit statuses. The remaining workflows are also skipped if the execution of a filter or mutator results in an error.

NA

NA

- refactoring was required

NA

1. Created handler with non-zero status (wrong command) and verified

## Is this change a patch?

NA


chavakula:
@rakibhossainctr test cases are failing, please check

chavakula:
@rakibhossainctr need to add latest core/V2 version in go.mod

go.sum
```diff
@@ -1575,10 +1575,10 @@ github.com/ruudk/golang-pdf417 v0.0.0-20201230142125-a7e3863a1245/go.mod h1:pQAZ
 github.com/ryanuber/columnize v0.0.0-20160712163229-9b3edd62028f/go.mod h1:sm1tb6uqfes/u+d4ooFouqFdy9/2g9QGwK3SQygK0Ts=
 github.com/schollz/progressbar/v2 v2.13.2/go.mod h1:6YZjqdthH6SCZKv2rqGryrxPtfmRB/DWZxSMfCXPyD8=
 github.com/sean-/seed v0.0.0-20170313163322-e2103e2c3529/go.mod h1:DxrIzT+xaE7yg65j358z/aeFdxmN0P9QXhEzd20vsDc=
-github.com/sensu/core/v2 v2.20.0 h1:QJQbqZiXSQ2dUE8LUvwUcKaqxHjYYCQwJ1Mil0cap4U=
-github.com/sensu/core/v2 v2.20.0/go.mod h1:fzr3qioLAHyLsBfq9I6ELVjL01NUNjqmQx4kj2tbDLo=
-github.com/sensu/core/v3 v3.9.0 h1:7wz4ILGwPbHHTT/8sKTtflBAgOuJgTaKhUlN4lyUdSY=
-github.com/sensu/core/v3 v3.9.0/go.mod h1:AuupbIzmjYdzEkaEdmpzP/ZgrqQgHTovXVejzyTN6u4=
+github.com/sensu/core/v2 v2.0.0-20250805061914-c7adcd108861 h1:uaYGNaj7BJ4blwAeJMWlv29k35nPSO2RKkR1OfrKeSs=
```
fguimond: Just a reminder that this will need to be updated once the `core` PR is merged.

amdprophet (approved):
Looks good! 👍

PR #5121, opened 2025-08-11
Validate runtime assets

### LINKED ISSUE #5120: Add validation for missing runtime_assets in CheckConfig resources
## Feature Suggestion
Currently, when creating a CheckConfig with a runtime_assets field, there is no validation or warning if the asset name(s) specified are missing or invalid. This can lead to runtime errors that are not immediately obvious at the time of creation.

This feature request is to add a validation step or warning mechanism during check creation to identify and notify the user if any of the referenced runtime_assets do not exist in the namespace.

## Possible Implementation
- Add backend-side validation in the check creation/update handler.

- When processing CheckConfig.spec.runtime_assets, ensure each asset exists in the given namespace.

- If any asset is missing:
** Option 1: Return a validation error and reject creation (like other invalid fields).
** Optionally: Add a sensuctl check lint or check validate command for dry-run validations.

## Context
While building checks that depend on Bonsai assets, I realized that Sensu allows check creation with missing assets, but then fails silently (or logs a failure only at execution time). This led to extra debugging effort to trace why checks weren't executing as expected.

It would be a major usability improvement to alert users early — ideally during sensuctl create — that the asset doesn't exist.

rakibhossainctr:
## What is this change?
Adds backend-side validation for runtime_assets in CheckConfig creation and update.
If any referenced asset does not exist in the namespace, the request will be rejected with a validation error.

Previously, Sensu allowed creating a CheckConfig with non-existent assets, leading to runtime execution failures that were only visible in logs.
This change improves usability by detecting missing or invalid assets during check creation/update, preventing avoidable runtime errors and debugging overhead.

Yes — this is a user-facing change in validation behavior for check creation/update.

No.

No major complications

Yes

Added unit tests for check creation/update with:

Verified via sensuctl create that creation fails with a clear error message when missing assets are referenced.

## Is this change a patch?
No


backend/apid/routers/checks.go
```diff
@@ -63,6 +68,95 @@ func (r *ChecksRouter) Mount(parent *mux.Router) {
...
+		return nil, actions.NewError(actions.InvalidArgument, err)
+	}
+
+	if len(config.RuntimeAssets) > 0 {
+		if missing, names := r.hasMissingAssets(req, config.RuntimeAssets); missing {
+			return nil, actions.NewError(actions.InvalidArgument, fmt.Errorf("runtime assets missing: %v", names))
+		}
+	}
+
+	req.Body = io.NopCloser(bytes.NewBuffer(body))
```
fguimond: This code block is repeated 3 times (in Patch, Post, Put). It might be worth isolating it in its own function.

backend/apid/routers/checks.go
```diff
@@ -63,6 +68,95 @@ func (r *ChecksRouter) Mount(parent *mux.Router) {
...
+
+func (r *ChecksRouter) hasMissingAssets(req *http.Request, runtimeAssets []string) (bool, []string) {
+	var assets []*corev2.Asset
+	if err := r.handlers.Store.ListResources(req.Context(), r.assetResource.StorePrefix(), &assets, &store.SelectionPredicate{}); err != nil {
+		return false, nil
+	}
+
+	set := make(map[string]struct{}, len(assets))
+	for _, v := range assets {
+		set[v.Name] = struct{}{}
```
fguimond, line 128: Correct me if I'm wrong but I believe assets are namespaced. If so a name match will not be enough, you need to make sure the proper asset exists in the same namespace as the CheckConfig.

backend/apid/routers/checks.go
```diff
@@ -25,8 +28,9 @@ type checkController interface {
 
 // ChecksRouter handles requests for /checks
 type ChecksRouter struct {
-	controller checkController
-	handlers   handlers.Handlers
+	controller    checkController
+	handlers      handlers.Handlers
+	assetResource corev2.Asset
```
fguimond: This appears to be used only to access the `StorePrefix` function. I don't think it's worth having it as a member of this struct. It could be created inline.

backend/apid/routers/checks.go
```diff
@@ -63,6 +68,95 @@ func (r *ChecksRouter) Mount(parent *mux.Router) {
...
+		}
+	}
+
+	req.Body = io.NopCloser(bytes.NewBuffer(body))
+	return r.handlers.CreateOrUpdateResource(req)
+}
+
+func (r *ChecksRouter) hasMissingAssets(req *http.Request, runtimeAssets []string) (bool, []string) {
+	var assets []*corev2.Asset
+	if err := r.handlers.Store.ListResources(req.Context(), r.assetResource.StorePrefix(), &assets, &store.SelectionPredicate{}); err != nil {
```
fguimond: Related to the first comment, replace
```
	if err := r.handlers.Store.ListResources(req.Context(), r.assetResource.StorePrefix(), &assets, &store.SelectionPredicate{}); err != nil {```
with 
```
	if err := r.handlers.Store.ListResources(req.Context(), (corev2.Asset{}).StorePrefix(), &assets, &store.SelectionPredicate{}); err != nil {
```

backend/apid/routers/checks.go
rakibhossainctr: The ListResources method already retrieves assets exclusively from the namespace specified in the CheckConfig.

chavakula (approved):

PR #5128, opened 2025-11-06
fix: updating the changelog with correct details

### LINKED ISSUE #2883: Implement ttl_status attribute to maintain 1.x parity
## Expected Behavior
When creating a check with a ttl, I should be able to configure whether or not I would a warning, or a critical event.

## Current Behavior
Currently, this is not present, but is in Core: https://docs.sensu.io/sensu-core/1.7/reference/checks/#check-attributes

## Possible Solution
We would need to implement this in Go. See #782, as it appears this might have been missed. 

1. Create a check via the API w/ the ttl specified (preferably a low one)
2. Observe the check staying in warning status

rakibhossainctr:

### DIFF
--- a/CHANGELOG-6.md
+++ b/CHANGELOG-6.md
@@ -9,7 +9,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 ## [Unreleased]
 
 ### Added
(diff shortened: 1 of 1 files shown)


chavakula (approved):
approved

